PR #12, opened 2024-07-08
feat(test): Add unit tests for aicore-clients

shibeshduw:
Adds unit tests for the AI core clients generated by using Cloud SDK. The following 5 AI Core endpoints covered:

- Artifact
- Configuration
- Deployment
- Execution
- Scenario

Closes [AI/gen-ai-hub-sdk-js#47](https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/47)


packages/ai-core/src/artifact-api.ts
```diff
@@ -23,68 +23,86 @@ export const ArtifactApi = {
...
-    name?: Name_1;
-    kind?: 'model' | 'dataset' | 'resultset' | 'other';
-    artifactLabelSelector?: string[];
-    $top?: number;
-    $skip?: number;
-    $search?: string;
-    searchCaseInsensitive?: boolean;
-    $expand?: 'scenario';
-  }) =>
+  artifactQuery: (
```
KavithaSiva, line 29: [q] Why is `queryParameters` no longer optional? (Valid for all the API introduced in this PR.)

packages/ai-core/src/artifact-api.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        {
+          count: 1,
+          resources: [
+            {
+              createdAt: '2024-07-08T13:36:41Z',
+              description: 'dataset for training test',
+              id: '08b9e4a4-2cea-4a3c-a509-9308d92bda85',
+              kind: 'dataset',
+              modifiedAt: '2024-07-08T13:36:41Z',
+              name: 'i745181-test-data',
```
KavithaSiva, line 34: [suggestion] Please cleanup your personal id and also clean up actual ids from tests.
Valid for all tests.

packages/ai-core/src/application-api.ts
```diff
@@ -24,106 +24,135 @@ export const ApplicationApi = {
...
    * @returns The request builder, use the `execute()` method to trigger the request.
    */
-  kubesubmitV4ApplicationsGetAll: (queryParameters?: {
-    $top?: number;
-    $skip?: number;
-    $count?: boolean;
-  }) =>
+  kubesubmitV4ApplicationsGetAll: (
+    queryParameters?: { $top?: number; $skip?: number; $count?: boolean },
+    headerParameters?: { Authorization?: string }
```
KavithaSiva, line 32: [q] Why is the `Authorization` header only added to a few API, while all others lack it. I assume every request would need this header.
If we supply this header via the destination instead in some API for e.g Artifact, why can't the same be followed here?

packages/ai-core/src/artifact-api.ts
shibeshduw: Support for generating OpenAPI clients with header parameters was added in [v3.16.0](https://github.com/SAP/cloud-sdk-js/releases/tag/v3.16.0). Here's the related [PR](https://github.com/SAP/cloud-sdk-js/pull/4708) if you would like the context. The main reason is since headerParameters is required, an optional parameter (queryParameters) cannot come before it. So even though the properties in queryParams are optional, the object itself has to passed (could be empty {}).

packages/ai-core/src/application-api.ts
shibeshduw: Since we only added support for in header parameters, these were already in the spec but are only visible now. Its a good point. I assume we would just supply this via the service binding. So the `Authorization` headers can be removed @jjtang1985 ?

jjtang1985: @shibeshduw , as long as this is generated, it's ok, because the generated client should reflect what's defined in the spec.
I double check the Java client and the spec, where the `Authorization` header is shown in both cases, under the application api.

packages/ai-core/src/artifact-api.test.ts
```diff
@@ -0,0 +1,76 @@
...
+describe('artifact', () => {
+  const destination: HttpDestination = {
+    url: 'https://ai.example.com'
+  };
+
+  afterEach(() => {
+    nock.cleanAll();
+  });
+
+  it('get artifact parses a successful response', async () => {
```
deekshas8, line 19: [req] `It` should read like a sentence and describe how the test behaves.  E.g., `parses a successful response for get request`

packages/ai-core/src/artifact-api.test.ts
```diff
@@ -0,0 +1,76 @@
...
+          name: 'training-test-data',
+          scenarioId: 'foundation-models',
+          url: 'ai://default/spam/data'
+        }
+      ]
+    };
+
+    nock(destination.url).get('/lm/artifacts').reply(200, expectedResponse, {
+      'Content-Type': 'application/json',
+      'AI-Resource-Group': 'default'
```
deekshas8, line 38: [q] Aren't these part of the request header?

shibeshduw: You're right. The resource group is only a part of the request header. Only content type should be a part of response. Thanks for catching it.

deekshas8: I would also not bother so much with response headers for now.

deekshas8 (approved):

PR #18, opened 2024-07-11
chore: Automate reuse check

jjtang1985:


.github/workflows/build.yml
```diff
@@ -39,6 +39,8 @@ jobs:
         with:
           node-version: ${{ vars.DEFAULT_NODE_VERSION }}
           cache: 'pnpm'
+      - name: REUSE Compliance Check
+        uses: fsfe/reuse-action@v1.1
```
deekshas8, line 43: [q] Should we consider upgrading the version? I see they already release v4.

jjtang1985, line 43:
```suggestion
        uses: fsfe/reuse-action@v4.0
```

.github/workflows/build.yml
```diff
@@ -39,6 +39,8 @@ jobs:
         with:
           node-version: ${{ vars.DEFAULT_NODE_VERSION }}
           cache: 'pnpm'
+      - name: REUSE Compliance Check
+        uses: fsfe/reuse-action@v4.0
```
jjtang1985, line 43:
```suggestion
        uses: fsfe/reuse-action@v4
```

deekshas8 (approved):

PR #20, opened 2024-07-12
chore: Adjust compile script

deekshas8:
Adjust the script so `pnpm compile` when run separately also runs compile recursively

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,8 @@
     "directory": "type-tests"
   },
   "scripts": {
-    "postinstall": "pnpm -r -w=false run compile",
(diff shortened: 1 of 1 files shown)


KavithaSiva (approved):

KavithaSiva:
We were wondering how the build worked for the original multi-module split PR, but, looks like we run the `postinstall` step and this is why it worked without issues.

deekshas8:
> We were wondering how the build worked for the original multi-module split PR, but, looks like we run the postinstall step and this is why it worked without issues.

Correct, I completely missed that compile should also work independently :/

PR #24, opened 2024-07-15
chore: Update orchestration client

deekshas8:


packages/gen-ai-hub/src/orchestration/api/schema/completion-post-request.ts
```diff
@@ -13,8 +13,4 @@ export type CompletionPostRequest = {
   orchestration_config: OrchestrationConfig;
   input_params: Record<string, InputParamsEntry>;
   messages_history?: ChatMessages;
-  /**
-   * Whether to return module results, which returns detailed information about each module execution. Be careful with the content of the individual module results as they may contain unchecked user or model content.
-   */
-  return_module_results?: boolean;
```
KavithaSiva, line 19: ~[q] Should we not support users providing this option?~ Nvm, I see the option was removed with the spec update.

KavithaSiva (approved):

PR #27, opened 2024-07-18
feat: Add message history option

deekshas8:
This PR adds the messages_history option to the input parameters.


packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
```diff
@@ -28,92 +37,89 @@ describe('GenAiHubClient', () => {
     nock.cleanAll();
   });
 
-  it(' calls chatCompletion and parses response', async () => {
+  afterAll(() => {
```
KavithaSiva, line 40: [q] Why is this required? The test seems to pass, even if we skip this.

packages/gen-ai-hub/src/client/openai/openai-client.test.ts
```diff
@@ -58,60 +54,47 @@ describe('openai client', () => {
...
       };
-      const mockResponse = fs.readFileSync(
-        path.join(
-          'test-util',
-          'mock-data',
-          'openai',
-          'openai-chat-completion-success-response.json'
-        ),
-        'utf8'
+      const mockResponse = parseMockResponse(
```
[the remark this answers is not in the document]
deekshas8: No you're right, I forgot to pass the type here

packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
```diff
@@ -28,92 +37,89 @@ describe('GenAiHubClient', () => {
...
+    clearXsuaaServices();
+  });
```
KavithaSiva, line 42: [q] Why is this required? The test seems to pass, even if we skip this.

packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
```diff
@@ -28,92 +37,89 @@ describe('GenAiHubClient', () => {
...
+
+  it('calls chatCompletion with minimum configuration and parses response', async () => {
     const module_configurations: ModuleConfigs = {
       templating_module_config: {
         template: [{ role: 'user', content: 'Hello!' }]
       },
```
jjtang1985, line 48: For prompt templating, `templating_module_config`, I would not ask the user to provide the complete object like the current status below, as it should be built by the SDK:
```
templating_module_config: {
         template: [{ role: 'user', content: 'Hello!' }]
       }
```

Instead, I would use the same API as our OpenAI client to keep similar UX.
Variables (next task in the prompt templating) can also be treated in the same way, as the only difference from the API perspective, would be additional information (input_param) to tell the value of the variables.

packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
deekshas8: It's a left over from testing with the real endpoint. I'll remove this.

packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
```diff
@@ -28,92 +37,89 @@ describe('GenAiHubClient', () => {
...
         url: 'completion'
       }
     );
-    const result = await client.chatCompletion(request);
-    const expectedResponse: CompletionPostResponse = JSON.parse(mockResponse);
-    expect(result).toEqual(expectedResponse);
+    expect(client.chatCompletion(request)).resolves.toEqual(mockResponse);
   });
 
-  it('throws error for incorrect input parameters', async () => {
```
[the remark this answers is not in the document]
deekshas8: You're right, we need to test other error cases. This specific case wasn't valid anymore as we have literal types for role.

packages/gen-ai-hub/test-util/mock-http.ts
```diff
@@ -40,3 +41,12 @@ export function mockInference<D extends BaseLlmParameters>(
...
     .reply(response.status, response.data);
 }
+
+export function parseMockResponse<T>(client: string, fileName: string): T {
+  const fileContent = fs.readFileSync(
+    path.join('test-util', 'mock-data', client, fileName),
+    'utf-8'
+  );
+
+  return JSON.parse(fileContent);
```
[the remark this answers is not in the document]
deekshas8: I think `as` in general is not a preferred practice. I'll think about this a bit more

packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
deekshas8: As discussed, we will do this in a separate PR

packages/gen-ai-hub/test-util/mock-http.ts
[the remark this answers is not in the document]
KavithaSiva: We discuss with Marika in general about response parsing and fix this here later.

packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
KavithaSiva: More tests to be added in a follow-up PR.

PR #33, opened 2024-07-26
feat: Add content filtering

KavithaSiva:
Proposed Content filtering API:

**Option 1: (Current approach)**
This is based on the assumption that the orchestration service only respects the first configuration provided for a filter service provider.

```TypeScript
client.chatCompletion({
      llmConfig: {
      },
      prompt: {
        template: [{ role: 'user', content: 'Hello!' }],
        messages_history: [{ role: 'system', content: 'Test system message' }]
      },
      filterConfig: { 
        input: { azureContentSafety: { Hate: 0, SelfHarm: 2 }},
        output: { 
          azureContentSafety: { Sexual: 4, Violence: 6 },
          someOtherFilterServiceProvider: { Hate: 0, SelfHarm: 2 }
        },
      }
  })
);

```

Other options:
**Option 2:**
```TypeScript
client.chatCompletion({
      llmConfig: {
      },
      prompt: {
        template: [{ role: 'user', content: 'Hello!' }],
        messages_history: [{ role: 'system', content: 'Test system message' }]
      },
      filterConfig: { 
        input: { azureContentSafety: { Hate: 0, SelfHarm: 2 }},
        output: [
          { azureContentSafety: { Sexual: 4, Violence: 6 }},
          { someOtherFilterServiceProvider: { Hate: 0, SelfHarm: 2 }}
        ]
      }
  })
);
```

**Option 3:**
```TypeScript
client.chatCompletion({
      llmConfig: {
      },
      prompt: {
        template: [{ role: 'user', content: 'Hello!' }],
        messages_history: [{ role: 'system', content: 'Test system message' }]
      },
      filterConfig: { 
        input: { type: 'azureContentSafety', config: { Hate: 0, SelfHarm: 2}},
        output: [
          { type: 'azureContentSafety', config: { Sexual: 4, Violence: 6 }},
          { type: 'someOtherFilterServiceProvider', config: { Hate: 0, SelfHarm: 2 }}
        ]
      }
  })
);
```


KavithaSiva:
With formatting though the proposed API still looks verbose, WDY think? is this an issue:

client.chatCompletion({
      llmConfig: {
      },
      prompt: {
        template: [{ role: 'user', content: 'Hello!' }],
        messages_history: [{ role: 'system', content: 'Test system message' }]
      },
      filterConfig: {
        input: {
          AzureContentSafety: {
            Hate: 0,
            SelfHarm: 2
          }
        },
        output: [
          {
            AzureContentSafety: {
              Sexual: 4,
              Violence: 6
            }
          },
          {
            AzureContentSafety: {
              Hate: 0,
              SelfHarm: 2,
              Sexual: 4,
              Violence: 6
            }
          }
        ]
      }
  })
);

ChristophMatthies:
Suggestion:

```jsonc
filterConfig: { 
    input: { azureContentSafety: { Hate: 0, SelfHarm: 2 }},
    output: {
      azureContentSafety: { Sexual: 4, Violence: 6 },
      someOtherFilterServiceProvider: { Hate: 0, SelfHarm: 2 }
    }
}
```

marikaner:
To be honest I am not happy with either of the options. If we simplify whatever is passed in the config, I think it is confusing if it is this close to the original, but still different. That opens potential for errors. I don't know whether this was already discussed before, but I would suggest one of the following options:
- Keep the original spec from the orchestration service. This has the disadvantage, that it might be too verbose and difficult to use.
- Breakout from the object structure and pass it through a more high level API. I could imagine something in this direction:
```
createFilterConfig({
  input: createInputFilter({
    type: 'azureContentSafety',
    config: { Hate: 0, SelfHarm: 2}
  }),
 output: createOutputFilter({
    type: 'azureContentSafety',
    config: { Sexual: 4, Violence: 6 }
  })
})
```

or

```
createFilterConfig(
  createInputFilter({
    type: 'azureContentSafety',
    config: { Hate: 0, SelfHarm: 2}
  ),
 createOutputFilter({
    type: 'azureContentSafety',
    config: { Sexual: 4, Violence: 6 }
  })
)
```

jjtang1985:
I would have the following:
- Vanilla orchestration service payload structure ([ref](https://github.com/SAP/ai-sdk-js/discussions/44#discussion-6997811)), which should be used by SDK internally and this internal function might also be used by the user, if they like.
- High level API, so user can pass a `ContentFilterConfig` object. (we also need to offer one or multiple ways to build the object)

This looks similar to what Marika mentioned.

KavithaSiva:
> I would have the following:
> 
> * Vanilla orchestration service payload structure ([ref](https://github.com/SAP/ai-sdk-js/discussions/44#discussion-6997811)), which should be used by SDK internally and this internal function might also be used by the user, if they like.
> * High level API, so user can pass a `ContentFilterConfig` object. (we also need to offer one or multiple ways to build the object)
> 
> This looks similar to what Marika mentioned.

I am assuming then that we can add separate config objects for both `llmConfig` and `prompt` as well.

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
```diff
@@ -47,6 +50,10 @@ export function constructCompletionPostRequest(
         },
         llm_module_config: input.llmConfig
       },
+      ...(input.filterConfig &&
+        input.filterConfig?.input &&
+        input.filterConfig?.output &&
```
MatKuhr, line 55: I think this is an issue, because only setting one of the two, or even setting none [is allowed](https://github.tools.sap/AI/llm-orchestration/blob/main/src/spec/api.yaml#L288). But the current code leave the filter config empty, if one of the two is not defined, right?

packages/gen-ai-hub/src/orchestration/orchestration-types.ts
```diff
@@ -39,16 +41,48 @@ export interface PromptConfig {
...
 export type LlmConfig = LLMModuleConfig;
 
+/**
+ * Wrapper object to configure Filters.
+ */
+export interface FilterConfig {
+  /**
+   * Input configuration for filtering provider.
+   */
+  input?: FilterServiceProvider;
```
MatKuhr, line 51: We could consider:
```suggestion
  input?: FilterServiceProvider | [FilterServiceProvider];
```

That would also allow:
```
filterConfig: {
        input: [] as FilterServiceProvider,
```

Not sure how to best implement this internally, but using `if (Array.isArray(config.input))` was what copilot suggested me 😉 

Only downside I can think of is that the type signature becomes a bit more complex.

packages/gen-ai-hub/src/orchestration/api/schema/azure-content-safety.ts
```diff
@@ -0,0 +1,15 @@
+/*
```
MatKuhr, line 1: Is this really generated code? I am wondering why it is being added just now?

KavithaSiva:
The latest state as per Junjie's recommendations are https://github.com/SAP/ai-sdk-js/pull/58

PR #45, opened 2024-07-31
chore: Canary Release

marikaner:
Add canary release workflow


tsconfig.json
```diff
@@ -50,7 +50,7 @@
 
     /* Emit */
     "declaration": true /* Generate .d.ts files from TypeScript and JavaScript files in your project. */,
-    "declarationMap": true /* Create sourcemaps for d.ts files. */,
+    // "declarationMap": true                            /* Create sourcemaps for d.ts files. */,
```
tomfrenken, line 53: Since it's unclear to me why we would remove the files, I would put them back until Marika is back, other than that LGTM

tomfrenken (approved):

tsconfig.json
deekshas8: Added it back.

PR #49, opened 2024-08-01
chore: Add codeql config

marikaner:
I switched the CodeQL setup to the advanced setup, so that we can only scan the prodcutive code.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,94 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+  schedule:
+    - cron: '19 2 * * 4'
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    # Runner size impacts CodeQL analysis time. To learn more, please see:
+    #   - https://gh.io/recommended-hardware-resources-for-running-codeql
+    #   - https://gh.io/supported-runners-and-hardware-resources
+    #   - https://gh.io/using-larger-runners (GitHub.com only)
+    # Consider using larger runners or machines with greater resources for possible analysis time improvements.
+    runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
+    timeout-minutes: ${{ (matrix.language == 'swift' && 120) || 360 }}
+    permissions:
+      # required for all workflows
+      security-events: write
+
+      # required to fetch internal or private CodeQL packs
+      packages: read
+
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,94 @@
...
+    runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
+    timeout-minutes: ${{ (matrix.language == 'swift' && 120) || 360 }}
+    permissions:
+      # required for all workflows
+      security-events: write
+
+      # required to fetch internal or private CodeQL packs
+      packages: read
+
+      # only required for workflows in private repositories
```
tomfrenken, line 39: I guess we can remove this, given the comment?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,94 @@
...
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
```
tomfrenken, line 15: Do you think adding workflow_dispatch here could be useful? I usually like it when I can manually trigger an action, not sure if that's necessary in its case though.

.github/workflows/codeql.yml
marikaner: Should we, though? This is the config generated from GitHub, in case we have to regenerate it at some point I'd like to have the smallest diff possible.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,94 @@
...
+    # If the analyze step fails for one of the languages you are analyzing with
+    # "We were unable to automatically build your code", modify the matrix above
+    # to set the build mode to "manual" for that language. Then modify this step
+    # to build your code.
+    # ℹ️ Command-line programs to run using the OS shell.
+    # 📚 See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
+    - if: matrix.build-mode == 'manual'
+      shell: bash
+      run: |
+        echo 'If you are using a "manual" build mode for one or more of the' \
```
tomfrenken, line 84: It probably can be replaced/removed?

.github/workflows/codeql.yml
marikaner: Generally, I agree, but in this case I don't think we will ever do it. We have never needed it in SAP Cloud SDK. Also, same argument as above, I would like to touch as little as possible, because this file was generated.

.github/workflows/codeql.yml
marikaner: Same as above.

.github/workflows/codeql.yml
tomfrenken: Hmm if we ever would have to regenerate it, I agree, but I don't see a reason why we would need to in the future, what do you think?
Because otherwise, I would like to keep the workflow files with the fewest bloat possible.

.github/workflows/codeql.yml
tomfrenken: Yeah, I agree.

tomfrenken (approved):
All in all LGTM, just a few open discussion points.

.github/workflows/codeql.yml
marikaner: But then this is personal preference, right? I just don't see any worth in putting extra effort into touching this file.

PR #51, opened 2024-08-01
chore: Refactor E2E Test using Jest

MatKuhr:


deekshas8:
@MatKuhr I got it working by adding `"@sap-ai-sdk/sample-code": "workspace:^"` to the root package.json. Making `e2e-tests` a package and adding this there might be a better way then. I see you already do have a disabled.package.json ;)

tests/e2e-tests/package.json
```diff
@@ -0,0 +1,26 @@
...
+		"dist/**/*.js.map",
+		"dist/**/*.d.ts",
+		"dist/**/*.d.ts.map"
+	],
+	"dependencies": {
+		"@sap-ai-sdk/sample-code": "workspace:^"
+	},
+	"scripts": {
+		"compile": "tsc",
+		"e2e-test": "NODE_OPTIONS=--experimental-vm-modules jest",
```
MatKuhr, line 18: Among other required settings, this seems to be required, because [jest support for ESM is experimental](https://jestjs.io/docs/ecmascript-modules). See also [this SO question](https://stackoverflow.com/a/75001235).

However, this means tests don't work if triggered via an IDE. One has to configure their jest runner to pass this node option.

pnpm-workspace.yaml
```diff
@@ -5,6 +5,7 @@ packages:
   # sample code
   - 'sample-code'
   # exclude packages that are inside test directories
+  - '!tests/e2e-tests' # including the e2e tests makes causes pnpm to explode
```
MatKuhr, line 8: excluding this here has the disadvantage that the e2e tests are not compiled with the rest of the code and one always has to run `pnpm i` before running e2e tests.

But if I include it I end up with around 1500 new files that usually should end up in `.dist`, but are instead mixed into the source code, so it seems to mess with src vs. dist directories..

jest.config.mjs
```diff
@@ -166,6 +166,8 @@ const config = {
   // This option allows use of a custom test runner
   // testRunner: "jest-circus/runner",
 
+  // extensionsToTreatAsEsm: ['.ts'],
+
   // A map from regular expressions to paths to transformers
   transform: {
```
KavithaSiva, line 172: [suggestion] Looks like this is no longer required if we have changed the preset to `ts-jest/presets/default-esm`.

deekshas8 (approved):

pnpm-workspace.yaml
```diff
@@ -4,7 +4,9 @@ packages:
   - 'packages/gen-ai-hub'
   # sample code
   - 'sample-code'
+  # e2e tests
+  - 'tests/e2e-tests'
   # exclude packages that are inside test directories
   - '!poc/'
   - '!test-util'
```
KavithaSiva, line 11: [suggestion] `poc` no longer exists and `test-util` is currently under a different path.

PR #52, opened 2024-08-02
chore: Add Fosstars rating

marikaner:
As in the title. The report will be pushed to the default "fosstars" branch.


tomfrenken (approved):
LGTM

PR #58, opened 2024-08-07
feat: Add content filtering with convenience functions

KavithaSiva:
**Proposed state:**
```TypeScript
client.chatCompletion({
  llmConfig: {
  },
  prompt: {
    template: [{ role: 'user', content: 'What is the capital of Germany?' }]
  },
  filterConfig: {
    input: azureContentFilter({ Hate: 4, SelfHarm: 2 }),
    output: azureContentFilter({ Sexual: 0, Violence: 6 })
  }
});
```
or you could assign `filterConfig` property  with the content filtering JSON structure the orchestration service accepts:

```TypeScript
client.chatCompletion({
  llmConfig: {
  },
  prompt: {
    template: [{ role: 'user', content: 'What is the capital of Germany?' }]
  },
  filterConfig: {
    input: {
      filters: [
        {
          type: 'azure_content_safety',
          config: {
            Hate: 4,
            SelfHarm: 4,
            Sexual: 4,
            Violence: 4
          }
        }
      ]
    },
    output: {
      filters: [
        {
          type: 'azure_content_safety',
          config: {
            Hate: 0,
            SelfHarm: 2,
            Sexual: 2,
            Violence: 2
          }
        }
      ]
    }
  }
});
```
Note: I also re-generated the orchestration client.


packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
```diff
@@ -4,8 +4,12 @@ import { HttpDestination } from '@sap-cloud-sdk/connectivity';
 import { mockGetAiCoreDestination } from '../test-util/mock-context.js';
 import { mockInference, parseMockResponse } from '../test-util/mock-http.js';
 import { BaseLlmParametersWithDeploymentId } from '../core/index.js';
-import { CompletionPostResponse } from './client/api/index.js';
+import {
+  CompletionPostResponse,
+  FilteringModuleConfig
+} from './client/api/index.js';
 import { GenAiHubCompletionParameters } from './orchestration-types.js';
+import { createAzureFilter } from './orchestration-filter-utility.js';
```
MatKuhr, line 12: wdy think about:

```suggestion
import { azureContentFilter } from './orchestration-filter-utility.js';
```

The fact that we create a more complex object is somewhat irrelevant for the user, they might instead think that something on the server side is "created".. 


I don't have a strong preference, I am just wondering 😄

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
```diff
@@ -39,22 +39,24 @@ export class GenAiHubClient {
...
-    }
+        llm_module_config: input.llmConfig,
+        ...(input?.filterConfig && {
+          filtering_module_config: input.filterConfig
+        })
+      }
+    },
+    input_params: input.prompt.template_params
+      ? input.prompt.template_params
+      : {},
```
MatKuhr, line 56: why change this? `input_params` are still optional, right?

packages/gen-ai-hub/src/orchestration/orchestration-filter-utility.ts
```diff
@@ -0,0 +1,21 @@
+import { AzureContentSafety, FilteringConfig } from './client/api/index.js';
+
+/**
+ * Convenience function to create Azure filters.
+ * @param filter - Filtering configuration for Azure filter.
+ * @returns An object with the Azure filtering configuration.
+ */
+export function createAzureFilter(filter: AzureContentSafety): FilteringConfig {
```
MatKuhr, line 8: We could consider:

```suggestion
export function createAzureFilter(filter: AzureContentSafety = undefined): FilteringConfig {
```

because of [this default behavior](https://github.tools.sap/AI/llm-orchestration/issues/466#issuecomment-7346528).
So, effectively:
```ts
createAzureFilter();    // okay
createAzureFilter({});  // illegal, throws
```

packages/gen-ai-hub/src/test-util/mock-data/orchestration/genaihub-chat-completion-filter-config.json
```diff
@@ -0,0 +1,71 @@
+{
```
MatKuhr, line 1: is this file really needed for currently? Is this response data evaluated in any way?

MatKuhr (commented):
lgtm with some minor notes

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
KavithaSiva: In the generated client it is no longer optional, but in the spec it is [still optional](https://github.tools.sap/AI/llm-orchestration/blob/main/src/spec/api.yaml#L66).
cc @tomfrenken

packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
KavithaSiva: I like it, applied the changes.

packages/gen-ai-hub/src/test-util/mock-data/orchestration/genaihub-chat-completion-filter-config.json
KavithaSiva: It is only used as a mock response right now. I will still keep it for convenience for the future when we work on validating the parsed response.

packages/gen-ai-hub/src/orchestration/orchestration-filter-utility.ts
KavithaSiva: I like the way to enable default filters (`createAzureFilter();`). But, I couldn't use the default value of `undefined` for the filter parameter as `Type 'undefined' is not assignable to type 'AzureContentSafety'.`
I used the default value of `{ default: true }` instead.

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
KavithaSiva: I have regenerated the orchestration client.

MatKuhr (approved):
LGTM!
Nice that we have the test coverage and also the option for the default filter, where no config is given.
Maybe someone else wants to look more closely at the TS coding details, but semantically it makes sense to me 👍🏻

packages/gen-ai-hub/src/orchestration/orchestration-filter-utility.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    };
+    genaihubCompletionParameters.filterConfig = filterConfig;
+    const completionPostRequest: CompletionPostRequest =
+      constructCompletionPostRequest(genaihubCompletionParameters);
+    expect(
+      completionPostRequest.orchestration_config.module_configurations
+        .filtering_module_config
+    ).toEqual(expectedFilterConfig);
+  });
+  it('constructs filter configuration with both input and ouput', async () => {
```
marikaner, line 81:
```suggestion

  it('constructs filter configuration with both input and ouput', async () => {
```

packages/gen-ai-hub/src/orchestration/orchestration-filter-utility.test.ts
```diff
@@ -0,0 +1,175 @@
...
+        ]
+      }
+    };
+    expect(
+      completionPostRequest.orchestration_config.module_configurations
+        .filtering_module_config
+    ).toEqual(expectedFilterConfig);
+  });
+
+  it('skips filter configuration', async () => {
```
marikaner, line 156:
```suggestion
  it('omits filter configuration if not set', async () => {
```

packages/gen-ai-hub/src/orchestration/orchestration-filter-utility.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    ).toBeUndefined();
+  });
+
+  it('throw error when configuring empty filter', async () => {
+    const createFilterConfig = () => {
+      {
+        azureContentFilter({});
+      }
+    };
+    expect(createFilterConfig).toThrow;
```
marikaner, line 173: [req] 1. You need to call the `toThrow()` method otherwise it will not be executed correctly. I believe `expect(createFilterConfig).not.toThrow;` should have the same effect.
2. Consider also checking that the correct error message is thrown.
```suggestion
    expect(createFilterConfig).toThrow();
```

packages/gen-ai-hub/src/orchestration/orchestration-filter-utility.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Convenience function to create Azure filters.
+ * @param filter - Filtering configuration for Azure filter. If skipped, the default Azure filter configuration is used.
+ * @returns An object with the Azure filtering configuration.
+ */
+export function azureContentFilter(
+  filter?: AzureContentSafety
+): FilteringConfig {
+  if (filter && Object.keys(filter).length === 0) {
```
marikaner, line 11: [q] I think in the very beginning we agreed to use the Cloud SDK styleguide (maybe we should link it in this repo). Therefore: https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#use-truthyfalsy-checks-where-possible
```suggestion
  if (filter && !Object.keys(filter).length) {
```

packages/gen-ai-hub/src/orchestration/orchestration-filter-utility.ts
```diff
@@ -0,0 +1,23 @@
...
+    throw new Error('Filter property cannot be an empty object');
+  }
+  const inputFilterConfig: FilteringConfig = {
+    filters: [
+      {
+        type: 'azure_content_safety',
+        ...(filter ? { config: filter } : {})
```
marikaner, line 18: You can simplify this:

```suggestion
        ...(filter && { config: filter })
```

packages/gen-ai-hub/src/orchestration/orchestration-filter-utility.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @param filter - Filtering configuration for Azure filter. If skipped, the default Azure filter configuration is used.
+ * @returns An object with the Azure filtering configuration.
+ */
+export function azureContentFilter(
+  filter?: AzureContentSafety
+): FilteringConfig {
+  if (filter && Object.keys(filter).length === 0) {
+    throw new Error('Filter property cannot be an empty object');
+  }
+  const inputFilterConfig: FilteringConfig = {
```
marikaner, line 14: You could just return this directly instead of doing this below.
```suggestion
  return {
```

packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
```diff
@@ -70,6 +74,75 @@ describe('GenAiHubClient', () => {
     expect(client.chatCompletion(request)).resolves.toEqual(mockResponse);
   });
 
+  it('calls chatCompletion with filtering configuration and parses response', async () => {
```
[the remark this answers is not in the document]
KavithaSiva: Done.

PR #60, opened 2024-08-08
chore: Update orchestration client and spec

tomfrenken:
With this PR, we update the spec and client of the orchestration & ai core service.
We also use the latest version of the cloud SDK and add a generation script.

Additionally, we remove a redundant type test and add a temporary workaround input_params, they will be fixed in an upcoming PR.


packages/gen-ai-hub/package.json
```diff
@@ -23,12 +23,14 @@
     "test": "NODE_OPTIONS=--experimental-vm-modules jest",
     "lint": "eslint . && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -c",
     "lint:fix": "eslint . --fix && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -w --log-level error",
-    "generate": "openapi-generator -i ./src/orchestration/spec/ -o ./src/orchestration/client"
+    "generate:orchestration": "openapi-generator --options-per-service='./src/orchestration/spec/options-per-service.json' --overwrite --generateESM -i ./src/orchestration/spec/api.yaml -o ./src/orchestration && pnpm lint:fix",
+    "postgenerate": "ts-node ../../scripts/update-imports.ts src/orechestration/api"
```
deekshas8, line 27: [q] We dont need this command here and the script now right?

tomfrenken: Good catch, that was just a leftover from my previous approach 😅

packages/ai-core/yarn.lock
```diff
@@ -0,0 +1,1203 @@
...
+  dependencies:
+    "@jsdevtools/ono" "^7.1.3"
+    call-me-maybe "^1.0.1"
+    js-yaml "^3.13.1"
+
+"@apidevtools/openapi-schemas@^2.1.0":
+  version "2.1.0"
+  resolved "https://registry.yarnpkg.com/@apidevtools/openapi-schemas/-/openapi-schemas-2.1.0.tgz#9fa08017fb59d80538812f03fc7cac5992caaa17"
+  integrity sha512-Zc1AlqrJlX3SlpupFGpiLi2EbteyP7fXmUOGup6/DnkRgjP9bgMM/ag+n91rsv0U1Gpz0H3VILA/o3bW7Ua6BQ==
+
```
deekshas8, line 18: [req] delete this.

packages/ai-core/src/application-api.ts
```diff
@@ -5,19 +5,19 @@
...
+  BckndArgoCDApplicationStatus,
+  BckndArgoCDApplicationBaseData,
+  BckndArgoCDApplicationModificationResponse,
+  BckndArgoCDApplicationDeletionResponse,
+  BckndArgoCDApplicationRefreshResponse
 } from './schema/index.js';
 /**
  * Representation of the 'ApplicationApi'.
- * This API is part of the 'AI_CORE_API' service.
+ * This API is part of the 'src' service.
```
deekshas8, line 20: Strange comment. This might come from one of the properties in the options-per-service.json, maybe packageName?

tomfrenken: Indeed, I'll play around with the options and check if we can improve it

scripts/update-imports.ts
```diff
@@ -0,0 +1,68 @@
+/* eslint-disable no-console */
```
deekshas8, line 1: I think this is also not needed anymore?

deekshas8 (approved):
LGTM

PR #65, opened 2024-08-12
chore: Fix nightly canary release

marikaner:
This should fix the nightly canary release.
@jjtang1985 should we add the check, whether anything was changed since the last release now or later?

### DIFF
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -47,6 +47,6 @@ jobs:
           pnpm changeset pre enter ${date}
           pnpm changeset version
           pnpm changeset pre exit
-          pnpm changeset publish --tag ${{ inputs.tag }}
+          pnpm changeset publish --tag ${{ github.event_name == 'schedule' && 'canary' || inputs.tag }}
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPMJS_ACCESS_TOKEN }}


jjtang1985:
> should we add the check, whether anything was changed since the last release now or later?

@marikaner 
What would be the definition of the change?
I assume it's not only about our source code, but also dependency updates?
I would only skip, if none of the following is updated:
- source code
- dependency updates

marikaner:
> > should we add the check, whether anything was changed since the last release now or later?
> 
> @marikaner What would be the definition of the change? I assume it's not only about our source code, but also dependency updates? I would only skip, if none of the following is updated:
> 
> * source code
> * dependency updates



> > should we add the check, whether anything was changed since the last release now or later?
> 
> @marikaner What would be the definition of the change? I assume it's not only about our source code, but also dependency updates? I would only skip, if none of the following is updated:
> 
> * source code
> * dependency updates
> 
> If it's not straight forward, we can postpone the "skip" mechanism.

Generally, I agree, but this is more difficult to achieve, because you have to check what was changed and not only if there were commits (which is how we do it in the SAP Cloud SDK).
I was rather wondering if you think this has priority as of this week or whether it should be done after GA or whenever there is time.

jjtang1985:
Let's just postpone the "skip" thing for now, @marikaner

jjtang1985 (approved):
LGTM

.github/workflows/publish-canary.yml
```diff
@@ -47,6 +47,6 @@ jobs:
           pnpm changeset pre enter ${date}
           pnpm changeset version
           pnpm changeset pre exit
-          pnpm changeset publish --tag ${{ inputs.tag }}
+          pnpm changeset publish --tag ${{ github.event_name == 'schedule' && 'canary' || inputs.tag }}
```
tomfrenken, line 50: How about
```
pnpm changeset publish --tag ${{ github.event_name == 'schedule' ? 'canary' : inputs.tag }}
```

tomfrenken: Because I'm not sure if the conditional logic you applied works properly in github actions

tomfrenken: I guess my suggestion comes a little late 😬

marikaner: To my understanding the way I did it will work in GH workflows but the operators you suggested don't exist in GH workflows.

marikaner: => https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/expressions#example

tomfrenken: Hmm good point, I guess you followed the[ example for their ternary equivalent](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/expressions#example), makes sense.

PR #68, opened 2024-08-13
chore: Refactor client test structure and generation script

tomfrenken:
While re-generating our clients, we noticed that deprecated files are not automatically removed, existing ones are only overwritten.
As we want to get rid of any manual steps during and after generation, we need to clear the output directory in the future.

With this PR, we switch from overwriting to clearing the output directory.
To keep our tests, they are moved to a separate testing directory.


KavithaSiva (commented):
Lgtm, please also consider cleaning up the imports like @deekshas8 suggested.

KavithaSiva (approved):

PR #69, opened 2024-08-13
chore: Simplify canary release config

marikaner:
This removes some of the config that I added previously:
1. Keep only one package in the dummy changelog for canary - due to the fixed versioning this is enough.
2. Do not ignore packages - all packages that should not be released have the private flag and this prevents them from being published.

This means for as long as we keep fixed versioning there is not need to adjust the config when changing the package structure.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,5 +7,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["@sap-ai-sdk/sample-code", "@sap-ai-sdk/e2e-tests", "@sap-ai-sdk/type-tests"]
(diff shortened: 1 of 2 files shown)


.changeset/config.json
```diff
@@ -7,5 +7,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["@sap-ai-sdk/sample-code", "@sap-ai-sdk/e2e-tests", "@sap-ai-sdk/type-tests"]
+  "ignore": []
```
deekshas8, line 10: [q] Can we remove this empty property altogether then?

deekshas8 (approved):
LGTM

PR #71, opened 2024-08-14
chore: Extend Openapi request builder

deekshas8:
Closes https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/70

- override the `.execute` and `.executeRaw`  methods in generated openapi clients. 
- update post generation script to change imports `@sap-cloud-sdk/openapi` -> `@sap-ai-sdk/core`

Reviewer Notes: 
- The generated client has a `.skipCsrfToken` method that the user can still and invoke, but it doesn't really do anything since we internally always set the fetch option to false. I am still looking into ways how we can hide this from users altogether.
- The pipeline breaks because a new version of SDK is not released yet


packages/core/src/http-client.ts
```diff
@@ -60,6 +60,7 @@ export async function executeRequest<D extends BaseLlmParameters>(
   requestConfig?: CustomRequestConfig
```
tomfrenken, line 60: not directly related to your change, but what do you think about changing the type "D" to "Data" or something more speaking?

tomfrenken:
I can see that the default-api.ts in the orchestration client is still using the "old" request builder, any thoughts on that?

scripts/update-imports.ts
```diff
@@ -0,0 +1,53 @@
...
+async function traverseDirectory(dirPath: string): Promise<void> {
+  const files = await fs.readdir(dirPath);
+
+  for (const file of files) {
+    const filePath = path.join(dirPath, file);
+
+    try {
+        const stat = await fs.stat(filePath);
+  
+        if (stat.isDirectory() && path.basename(filePath) !== 'schema') {
```
tomfrenken, line 32: Given the remaining purpose of the script, do you think it might be a little overkill to check if it's schema & index, or do you think the improved efficiency is worth the additional complexity? I don't really have a preference.

deekshas8:
> I can see that the default-api.ts in the orchestration client is still using the "old" request builder, any thoughts on that?

Actually this is not needed (and perhaps should be deleted pos generation?). I removed it manually for now.

tomfrenken (dismissed):
LGTM, only thing missing are the automation for removing the default-api & the export in the index file.

tomfrenken (approved):

PR #72, opened 2024-08-15
chore: Add filter utility to improve discoverability

KavithaSiva:
This PR introduces `FilterUtility` to make new functionality being added to filter configuration like for e.g a new service provider discoverable.
Still choosing to retain the direct function usage (`azureContentFilter()`) as well because currently we only have one option and it shortens code.

Proposed state:
```Typescript
client.chatCompletion({
  llmConfig: {
  },
  prompt: {
    template: [{ role: 'user', content: 'What is the capital of Germany?' }]
  },
  filterConfig: {
     //Both options are allowed
    input: FilterUtility.azureContentFilter({ Hate: 4, SelfHarm: 2 }), 
    output: azureContentFilter({ Sexual: 0, Violence: 6 })
  }
});
```

### DIFF
--- a/packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
+++ b/packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
@@ -12,7 +12,10 @@ import {
   GenAiHubClient,
   constructCompletionPostRequest
 } from './orchestration-client.js';
-import { azureContentFilter } from './orchestration-filter-utility.js';
+import {
+  azureContentFilter,
+  FilterUtility
+} from './orchestration-filter-utility.js';
 describe('GenAiHubClient', () => {
   const client = new GenAiHubClient();
   const deploymentConfiguration: BaseLlmParametersWithDeploymentId = {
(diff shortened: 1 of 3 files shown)


tomfrenken:
I like the idea of improved discoverability, however, I really dislike long call chains (e.g. `MyCoolClass.someFunctionForSomething()`) therefore I would be in favor of adding a dedicated section for our filter function(s) as they grow.

tomfrenken:
On the other hand, depending on how the different filters will be chained together, we might even consider a similar implementation to the cloud sdk's resilience package: https://sap.github.io/cloud-sdk/docs/js/guides/resilience

KavithaSiva:
> On the other hand, depending on how the different filters will be chained together, we might even consider a similar implementation to the cloud sdk's resilience package: https://sap.github.io/cloud-sdk/docs/js/guides/resilience

Could you sketch out an example on how this would look?

packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
```diff
@@ -78,7 +81,7 @@ describe('GenAiHubClient', () => {
       },
       filterConfig: {
         input: azureContentFilter({ Hate: 4, SelfHarm: 2 }),
-        output: azureContentFilter({ Sexual: 0, Violence: 4 })
+        output: FilterUtility.azureContentFilter({ Sexual: 0, Violence: 4 })
```
jjtang1985, line 84: I prefer the original API.

jjtang1985 (changes requested):
I prefer the original API, which is short and JS native.

PR #73, opened 2024-08-15
chore: Add a PR template

KavithaSiva:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
+## Context
+
+AI/gen-ai-hub-sdk-js-backlog#ISSUENUMBER.
```
deekshas8, line 3: [q] Does this work? The backlog is in the internal repo, I think you need the full URL

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
...
+
+Please provide a short description of what your change does and why it is needed.
```
deekshas8, line 5: [req] I would put this line under a section `## Context` and remove `Feature Scope`.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Feature scope:
+ 
+- [ ] Task1 
+- [ ] Task2 
+  - [ ] SubTask1 
+
+## Definition of Done
+
+- [ ] Functionality scope stated & covered
```
deekshas8, line 15:
```suggestion
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
...
+- [ ] Tests cover the scope above
```
deekshas8, line 16:
```suggestion
- [ ] Code is tested (Unit, E2E)
```

[pp]

.github/PULL_REQUEST_TEMPLATE.md
KavithaSiva: Seems to work for the Java SDK, where they have a similar setup.

.github/PULL_REQUEST_TEMPLATE.md
KavithaSiva: I am sorry, I did not understand. Why do you want to remove `Feature Scope` ?

.github/PULL_REQUEST_TEMPLATE.md
KavithaSiva: Why do you want to remove this? Is this because you think `Feature scope` and this check box is redundant?

.github/PULL_REQUEST_TEMPLATE.md
deekshas8: I tried this in another PR and it didnt resolve the URL

.github/PULL_REQUEST_TEMPLATE.md
deekshas8: [pp] I honestly am not a fan of this. The scope of the ticket is already defined in the linked BLI and not everything is a feature. Unless I misunderstood what is meant here.
The context/description should be enough to describe overall changes or serve as a note to reviewers about what they should look out for. I would not restrict how one describes this.

KavithaSiva: But why do you think there is a restriction here? It is just a reminder to add in the description right?
For smaller BLI's I guess people can just link the BLIs but for e.g API proposals or PRs that cover a BLI partially, I think this could be super helpful. 
Is it okay if we just make this optional depending on the BLI?

.github/PULL_REQUEST_TEMPLATE.md
KavithaSiva: Strange, I don't understand how this works for Java then. Will check with them.

.github/PULL_REQUEST_TEMPLATE.md
KavithaSiva: Ah, I see that you made it optional :D.

.github/PULL_REQUEST_TEMPLATE.md
KavithaSiva: I would still like to keep this one as I removed the `Feature scope` checkboxes now.

.github/PULL_REQUEST_TEMPLATE.md
KavithaSiva: I added an autolink now, so now the shorthand works.

.github/PULL_REQUEST_TEMPLATE.md
tomfrenken: I don't get what this is supposed to be?
I assume the functionality scope is defined in the block above where you put
```
<!-- Please provide a description of what your change does and why it is needed. -->
```

KavithaSiva: Removed this check-box now.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
...
+AI/gen-ai-hub-sdk-js-backlog#ISSUENUMBER.
+
+<!-- Please provide a description of what your change does and why it is needed. -->
+
+## Definition of Done
+
+- [ ] Functionality scope stated & covered
+- [ ] Code is tested (Unit, E2E)
+- [ ] Error handling created / updated & covered by the tests above
+- [ ] ~Aligned changes with the Java SDK~
```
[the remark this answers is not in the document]
KavithaSiva: They are optional, made optional checkboxes more readable now.

marikaner (commented):
I am not a big fan of lists to check off on the PR. This should be either covered by the inherent DoD that we all agree to (which includes implementation and tests at minimum) or through the ticket. Documentation is also part of the DoD, but often forgotten therefore in specific cases I think it is a good idea to have a reminder in the ticket (which we already do), but why again on the PR?
Release notes are covered by the PR check: for `chore:` you don't need to add anything, for `fix:` and `feat:` the PR check will fail if you don't add a changeset.
Alignment with Java will also not be needed on every PR. So I would vote for something simpler and potentially peak into our old template, that still gives you a commented out checklist if you need it, but doesn't force you to check off boxes: https://raw.githubusercontent.com/SAP/cloud-sdk-js/main/.github/PULL_REQUEST_TEMPLATE.md

tomfrenken:
I agree with Marika that the checkboxes become somewhat useless when we already have the same in the related tickets.
I do see a benefit for OSS contributors though, as they do not make use of the tickets.

At the end of the day, I don't have a strong preference of either approach, therefore I leave it to you 🤷

.github/PULL_REQUEST_TEMPLATE.md
MatKuhr: FYI When we originally added this in Java this was also sort of meant as a reminder to create / maintain a meaningful description for your PR 😉

MatKuhr:
my 2 cents: personally I like a checklist on the PR.

> same in the related tickets.

Personally, I prefer to have those on the PR rather than a ticket. On a PR I can judge much better if it requires docs, e2e tests, type tests etc. or not, and a checklist makes it explicit. In Java we do this by crossing out anything, i. e. crossing out documentation is an explicit statement that docs are not needed for this PR (compared to leaving it unchecked, which might as well mean one forgot about it).

IMO that is also a great way to indicate to reviewers which parts are done. E.g. reviewers can read the checklist as "productive code done, unit tests done, type tests and e2e tests still missing".

 Of course, no strong opinion. I'll try to remember to add docs, even without the checklist 😉

tomfrenken:
How about we keep the checkboxes, but make sure we do not duplicate them in the tickets, this way we get rid of possible duplicates and @KavithaSiva and @MatKuhr are happy 👀

tomfrenken:
If we notice we keep ignoring the checkboxes or think they add no value we can still remove them.
(I remember this happened a few years ago with the js template, but perhaps it's a culture thing)

tomfrenken (approved):
LGTM

PR #74, opened 2024-08-18
fix: Custom request config type

jjtang1985:
If I understand correctly, user is not allowed to overwrite the request body.
If so, this is a fix, as `HttpRequestConfig` (from cloud sdk) uses `data` as the key for request body.

@deekshas8, below is the list of allowed keys, in the custom request config.
Is it intended to customise so many things especially I see `http(s)Agent` and `proxy`?
Screenshot 2024-08-18 at 17 55 16


deekshas8:
@jjtang1985 I agree the current type has many properties that we probably don't need. That's why I opened this [ticket](https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/69) to have a discussion about this. We didn't have this before.
For the initial implementation I just kept it as close to the generic http-client from the SDK.

deekshas8 (approved):
Good catch. Thanks for taking care of this. LGTM.

PR #75, opened 2024-08-18
feat: Model Discovery

MatKuhr:
## Context

Closes AI/gen-ai-hub-sdk-js-backlog#57

## API Design

Variant 1:
```ts
client.chatCompletion(OpenAiModels.GPT_4o, prompt);
client.chatCompletion({ name: 'gpt-4o', type: 'chat', version: '0613'}, prompt);
```

Autocompletion is available for `name` and `type`.

Variant 2:
```ts
client.chatCompletion(OpenAiModels.GPT_4o(), prompt);
client.chatCompletion(OpenAiModels.GPT_4o('0613'), prompt);

Autocompletion is available for the version number.

```

Variant 3:
```ts
client.chatCompletion('gpt-4o', prompt);
client.chatCompletion({ name: 'gpt-4o', version: '0613' }, prompt);
```

Autocompletion is available for the name.

**Decision:** We choose alternative 3, as it is the most JS-like style. We can still choose to offer `1` or `2` in addition the future, if we see need for it.


packages/gen-ai-hub/src/core.ts
```diff
@@ -0,0 +1,71 @@
+import { DeploymentApi, AiDeployment } from '@sap-ai-sdk/ai-core';
+
+// TODO: docs
+/* eslint-disable */
+
+export type DeploymentResolver = DeploymentId | (() => Promise<AiDeployment>);
+export type DeploymentId = string;
+
```
MatKuhr, line 8: relevant context also [here](https://github.com/SAP/ai-sdk-js/pull/50#discussion_r1702004299) and [here](https://github.com/SAP/ai-sdk-js/pull/50#discussion_r1701879493)

packages/gen-ai-hub/src/client/openai/openai-client.ts
```diff
@@ -1,60 +1,93 @@
...
+  OpenAiChatModel,
+  OpenAiEmbeddingModel
 } from './openai-types.js';
 
 const apiVersion = '2024-02-01';
 
 /**
  * OpenAI GPT Client.
  */
-export class OpenAiClient implements BaseClient<BaseLlmParameters> {
```
MatKuhr, line 19: see https://github.com/SAP/ai-sdk-js/pull/50#discussion_r1702010223

packages/core/src/http-client.ts
```diff
@@ -54,19 +54,17 @@ export interface EndpointOptions {
...
  */
-export async function executeRequest<Data extends BaseLlmParameters>(
+export async function executeRequest(
   endpointOptions: EndpointOptions,
-  data: Data,
+  data: any,
   requestConfig?: CustomRequestConfig
 ): Promise<HttpResponse> {
   const aiCoreDestination = await getAiCoreDestination();
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
MatKuhr, line 63: driveby fix

tomfrenken:
When we talked about the model_params in the refinement, I had an idea in regards to your static variant.

What if instead of
```
client.chatCompletion(OpenAiModels.GPT_4o(), prompt);
client.chatCompletion(OpenAiModels.GPT_4o('0613'), prompt);
```

You instantiate the models first, and then use them to create a config with proper autocomplete, e.g.:
```
const gptModel = OpenAiModels.GPT_4o('0613')
client.chatCompletion(gptModel.createConfig(<model_params_with_autocomplete>), prompt);
```

or depending on the implementation:
```
client.chatCompletion(
   OpenAiModels
   .GPT_4o('0613')
   .createConfig(<<model_params_with_autocomplete>>), prompt);
```

packages/gen-ai-hub/src/core.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @param opts.modelVersion - The version of the model of the deployment.
+ * @returns An AiDeployment, if a deployment was found, fails otherwise.
+ */
+export async function resolveDeployment(opts: {
+  scenarioId: string;
+  executableId?: string;
+  modelName?: string;
+  modelVersion?: string;
+}): Promise<AiDeployment> {
+  // TODO: is there a more elegant way to write this in TS?
```
tomfrenken, line 24: You could use a ternary here:
```
const query =  { scenarioId: opts.scenarioId, status: 'RUNNING' };
if(opts.executableId) query.executableIds = [opts.excutableId]
```
Or a little more complex but as one object:
```
const query: any = {
  scenarioId: opts.scenarioId,
  status: 'RUNNING',
  ...(opts.executableId && { executableIds: [opts.executableId] })
};
```

packages/gen-ai-hub/src/core.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  // TODO: add a cache
+  let deploymentList: AiDeployment[];
+  try {
+    deploymentList = await DeploymentApi.deploymentQuery(query, {
+      'AI-Resource-Group': 'default'
+    })
+      .execute()
+      .then(res => res.resources);
```
tomfrenken, line 43: why mix await and .then?

How about
```ts
const { deploymentQuery } = DeploymentApi; <--- keeps the call a little shorter, should be below the imports
const defaultResourcesGroup = { 'AI-Resource-Group': 'default' }
deploymentList = (await deploymentQuery(query, defaultResourcesGroup).execute()).resources
```

packages/gen-ai-hub/src/core.ts
```diff
@@ -0,0 +1,71 @@
...
+  } catch (error) {
+    throw new Error('Failed to fetch the list of deployments: ' + error);
+  }
+
+  if (opts.modelName) {
+    deploymentList = deploymentList.filter(
+      (deployment: any) =>
```
tomfrenken, line 50: Could a more narrow type be used for deployment?
Worst case example:
```
interface deployment {
  modelName: string;
  [key: string]: any; 
}

Comment (tomfrenken) [reply packages/gen-ai-hub/src/core.ts:50]: E.g. the modelExtractor function seems to assume a pretty nested object:
```
deployment.details?.resources?.backend_details?.model;
```

File: packages/gen-ai-hub/src/core.ts [hunk 7]
```diff
@@ -0,0 +1,71 @@
...
+  }
+  if (opts.modelVersion) {
+    // feature idea: smart handling of 'latest' version
+    deploymentList = deploymentList.filter(
+      (deployment: any) =>
+        modelExtractor(deployment)?.modelVersion === opts.modelVersion
+    );
+  }
+  if (deploymentList.length === 0) {
+    // TODO: return undefined instead?
```
tomfrenken, line 62: Error makes sense to me, what would be the benefit of undefined?

packages/gen-ai-hub/src/core.ts
```diff
@@ -0,0 +1,71 @@
...
+    query = {
+      scenarioId: opts.scenarioId,
+      status: 'RUNNING',
+      executableIds: [opts.executableId]
+    };
+  } else {
+    query = { scenarioId: opts.scenarioId, status: 'RUNNING' };
+  }
+
+  // TODO: add a cache
```
tomfrenken, line 36: Perhaps link the ticket already if there is one 👀

sample-code/src/aiservice.ts
```diff
@@ -1,22 +1,14 @@
...
-const deployments: { [model: string]: string } = {
-  'gpt-4-32k': 'd577d927380c98ea',
-  'gpt-35-turbo': 'd66d1927bf590375',
-  ada: 'd0084a63ebd7bcd3'
-};
-
 /**
  * Ask GPT about the capital of France.
  * @returns The answer from GPT.
  */
```
tomfrenken, line 8: As a certified `.then` hater, and for consistency, I would transform all these function to async / await functions
```ts
export async function chatCompletion(): Promise<any> {
return (await openAiClient
    .chatCompletion(OpenAiModels.GPT_35_TURBO, {
      messages: [{ role: 'user', content: 'What is the capital of France?' }]
    })).choices[0].message.content
}
```

tomfrenken: Also the any type seems a bit wonky

tomfrenken (commented):
Overall looks like a great improvement, generally, I would like to stick to `async / await` instead of `.then().catch()`, and keep function calls shorter by destructuring the APIs, instead of:
```
WayTooLongApiCallFromSomeApi.theOneFunctionIActuallyNeed() <--- I get Java PTSD flashbacks
```

Additionally, if there is a way to avoid the many `any` types, we should try to narrow them as much as we can.

Other than that just some general suggestions for the open TODOs 👍

packages/gen-ai-hub/src/core.ts
```diff
@@ -0,0 +1,71 @@
+import { DeploymentApi, AiDeployment } from '@sap-ai-sdk/ai-core';
```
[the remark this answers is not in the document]
tomfrenken: I guess putting it into this package can't be prevented 😬 
Perhaps we can call it something else than "core" though, to avoid the naming conflict.
Perhaps a `utils` subdirectory, and this is `resolver.ts`?

packages/gen-ai-hub/src/client/openai/openai-client.ts
tomfrenken: I believe the purpose of the `BaseClient` was rather for future use, so that other clients of e.g. Gemini / Claude etc. all align with the same base client, do you think this would make sense to keep / help with the implementation of additional clients?

packages/gen-ai-hub/src/client/openai/openai-types.ts
```diff
@@ -1,4 +1,114 @@
-import { BaseLlmParameters } from '@sap-ai-sdk/core';
+// TODO: docs
+/* eslint-disable */
+export const OpenAiModels = {
```
marikaner, line 3: [q] Why do we need chat + embedding models combined?

packages/gen-ai-hub/src/client/openai/openai-types.ts
```diff
@@ -1,4 +1,114 @@
...
+      type: 'chat',
+      version: version
+    };
+  }
+  private constructor() {}
+}
+
+/**
+ *
+ * Alternative proposal for referencing models.
```
marikaner: I added below alternative for discussion.

packages/gen-ai-hub/src/utils/deployment-resolver.test.ts
```diff
@@ -0,0 +1,133 @@
...
+import {
+  mockClientCredentialsGrantCall,
+  mockAiCoreEnvVariable,
+  aiCoreDestination
+} from '../../../../test-util/mock-http.js';
+import { resolveDeployment } from './deployment-resolver.js';
+
+describe('Deployment resolver', () => {
+  beforeEach(() => {
+    mockAiCoreEnvVariable();
```
deekshas8, line 11: [req] mocking AI core env variable is done in the global setup. SO you don't need to do it anywhere else.

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
```diff
@@ -1,34 +1,42 @@
...
-    const dataWithInputParams = {
-      deploymentConfiguration: data.deploymentConfiguration,
-      ...constructCompletionPostRequest(data)
-    };
+  ): Promise<CompletionPostResponse> {
+    const body = constructCompletionPostRequest(data);
+    const deployment =
+      typeof deploymentResolver === 'function'
+        ? (await deploymentResolver()).id
+        : deploymentResolver;
```
marikaner, line 33: Regarding my comment before (OpenAIClient): Maybe this can even be extracted to core, so that it is not repeated even more. I think it is fine to duplicate it here due to the package boundaries (I mean it's 4 loc), however once we add more model-specific clients we will need to solve it somehow anyways.

marikaner (changes requested):
I haven't finished my review yet, some of the questions could probably answered by me looking into it more thoroughly.

packages/gen-ai-hub/src/utils/deployment-resolver.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+  if (deploymentList.length === 0) {
+    throw new Error(
+      'No deployment matched the given criteria: ' + JSON.stringify(opts)
+    );
+  }
+  return deploymentList[0];
+}
+
+const modelExtractor = (deployment: AiDeployment) =>
```
marikaner, line 64: style guide
```suggestion
const extractModel = (deployment: AiDeployment) =>
```

packages/gen-ai-hub/src/utils/deployment-resolver.ts
```diff
@@ -0,0 +1,65 @@
...
+  deployment.details?.resources?.backend_details?.model;
```
marikaner, line 65: style guide
```suggestion
function extractModel (deployment: AiDeployment) {
  deployment.details?.resources?.backend_details?.model;
}
```

packages/gen-ai-hub/package.json
```diff
@@ -28,6 +28,7 @@
   },
   "dependencies": {
     "@sap-ai-sdk/core": "workspace:^",
+    "@sap-ai-sdk/ai-core": "workspace:^",
```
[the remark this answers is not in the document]
MatKuhr: looking up the deployment requires 
```ts
{ DeploymentApi,  AiDeployment,  AiDeploymentStatus } from '@sap-ai-sdk/ai-core';
```

So we can't put this in `core`. I agree, somewhat awkward, but didn't want to add yet another package 😄

sample-code/src/aiservice.ts
MatKuhr: Fixed ✅

packages/gen-ai-hub/src/orchestration/orchestration-client.test.ts
```diff
@@ -28,8 +24,7 @@ describe('GenAiHubClient', () => {
   });
 
   it('calls chatCompletion with minimum configuration', async () => {
-    const request: GenAiHubCompletionParameters = {
-      deploymentConfiguration,
+    const request = {
```
MatKuhr, line 27: Follow up for making the constants usable with orchestration: https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/91
Follow up for adding convenience for `model_params`: https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/89#issuecomment-7440288

packages/gen-ai-hub/src/client/openai/openai-client.ts
MatKuhr: (just for completeness) In addition to YAGNI, [this change](https://github.com/SAP/ai-sdk-js/pull/75/files#diff-3b1b4fce21b3de60726e49eea5f4559cf76699d393e75495f681ffdee58d62deL57) is related. The `executeRequest` is also used for non-inference calls. There deployment ID doesn't make any sense (which is why I am removing it), and consequently I am removing the `<Data extends BaseLlmParameters>` on that method, which was the only usage.

packages/gen-ai-hub/src/client/openai/openai-client.ts
```diff
@@ -1,60 +1,107 @@
...
+   * @param model - The model to use for the chat completion.
    * @param data - The input parameters for the chat completion.
+   * @param deploymentResolver - A deployment id or a function to retrieve it.
    * @param requestConfig - The request configuration.
    * @returns The completion result.
    */
   async chatCompletion(
+    model: OpenAiChatModel | { name: OpenAiChatModel; version: string },
     data: OpenAiChatCompletionParameters,
+    deploymentResolver?: DeploymentResolver,
```
MatKuhr, line 33: Follow-up BLI to improve the `deploymentResolver`: https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/92

packages/gen-ai-hub/src/client/openai/openai-client.ts
```diff
@@ -1,60 +1,107 @@
...
+async function resolveOpenAiDeployment(
+  model: string | { name: string; version: string },
+  resolver?: DeploymentResolver
+) {
+  if (typeof resolver === 'string') {
+    return resolver;
+  }
+  const llm =
+    typeof model === 'string' ? { name: model, version: 'latest' } : model;
+  const deployment = await resolveDeployment({
```
MatKuhr, line 88: if `resolver` is a function it is currently ignored. See https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/92

packages/gen-ai-hub/src/index.ts
```diff
@@ -1,17 +1,10 @@
+export * from './client/index.js';
```
MatKuhr, line 1: Is this fine?

packages/gen-ai-hub/src/utils/deployment-resolver.ts
```diff
@@ -0,0 +1,80 @@
...
+  try {
+    deploymentList = (await deploymentQuery(query, resourceGroup).execute())
+      .resources;
+  } catch (error) {
+    throw new Error('Failed to fetch the list of deployments: ' + error);
+  }
+
+  if (opts.model) {
+    deploymentList = deploymentList.filter(
+      deployment => extractModel(deployment)?.name === opts.model!.name
```
MatKuhr, line 61: I tried using:

```suggestion
      deployment => extractModel(deployment)?.name === opts.model.name
```

but `tsc` complains that `opts.model` might be undefined. Which I don't get, how can that be undefined after the `if` check? So I decided `tsc` is stupid and the tests covering the case where `model` is undefined seem to agree with me 😄

packages/gen-ai-hub/src/index.ts
marikaner: As long as no checks fail, yes. I think the API check is not in place yet, but once it is this might cause problems, not sure. I'd suggest to fix it then if needed.

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
```diff
@@ -1,34 +1,42 @@
...
 
 /**
  * Get the orchestration client.
  */
-export class GenAiHubClient {
+export class OrchestrationClient {
   /**
    * Creates a completion for the chat messages.
    * @param data - The input parameters for the chat completion.
+   * @param deploymentResolver - A deployment id or a function to retrieve it.
```
marikaner, line 19:
```suggestion
   * @param deploymentResolver - A deployment ID or a function to retrieve it.
```

packages/gen-ai-hub/src/utils/deployment-resolver.ts
marikaner, line 61: Yes, `tsc` is stupid like that sometimes. This case seems particularly straight forward, therefore I was about to create an issue in TS, but I found that this is another design limitation: `tsc` does not know what else you are doing to that object in your callback (e.g. you could set it to `undefined`), therefore it resets the narrowing you have done before. (https://github.com/microsoft/TypeScript/issues/9998, https://github.com/microsoft/TypeScript/issues/11498)

You can mitigate this by creating a local variable `const model = opts.model;` and work with that. I think it is not worth it.

Generally I dislike the use of `!` as this can be dangerous. Maybe we change code in another place and assert this cannot be `undefined` while it is. In most cases it is better to resort to `?`, because this will resolve to `undefined` and not result in an error at runtime.

```suggestion
      deployment => extractModel(deployment)?.name === opts.model?.name
```

packages/gen-ai-hub/src/utils/deployment-resolver.ts
```diff
@@ -0,0 +1,80 @@
...
+    );
+    if (opts.model.version) {
+      // feature idea: smart handling of 'latest' version: treat 'latest' and the highest version number as the same
+      deploymentList = deploymentList.filter(
+        deployment => extractModel(deployment)?.version === opts.model!.version
+      );
+    }
+  }
+
+  if (deploymentList.length === 0) {
```
marikaner, line 71:
```suggestion
  if (!deploymentList.length) {
```

marikaner (dismissed):
LGTM, I left a few minor comments and answers to questions you asked.

packages/gen-ai-hub/src/utils/deployment-resolver.ts
tomfrenken: I read Matthias comment -> fixed it (after finding out what you already found out 🤦) -> now read your comment 😬, I went with the const model = opts.model approach

tomfrenken (dismissed):
Unintentionally dismissed Marika's LGTM when I resolved one of her comments 😬.

LGTM

packages/gen-ai-hub/src/client/openai/openai-client.ts
```diff
@@ -1,60 +1,107 @@
...
+    );
     const response = await executeRequest(
       {
-        url: `/inference/deployments/${data.deploymentConfiguration.deploymentId}/chat/completions`,
+        url: `/inference/deployments/${deploymentId}/chat/completions`,
         apiVersion
       },
       data,
-      requestConfig
+      mergeRequestConfig(requestConfig)
```
deekshas8, line 46: [q] Aren't we calling merge twice? Once here and then again in `executeRequest`?

packages/gen-ai-hub/src/client/openai/openai-client.ts
```diff
@@ -1,60 +1,107 @@
...
+function mergeRequestConfig(
+  requestConfig?: CustomRequestConfig
+): HttpRequestConfig {
+  return {
+    method: 'POST',
+    headers: {
+      'content-type': 'application/json'
+    },
+    params: { 'api-version': apiVersion },
+    ...requestConfig
```
deekshas8, line 105: here the requestConfig will entirely replace the headers and params keys.

MatKuhr: yes, this is intended. My thinking was that the open AI client should only mix in whatever is specific to the open ai client.

tomfrenken (approved):
LGTM

PR #78, opened 2024-08-19
fix: Missing visibility of azureContentFilter

jjtang1985:
This function is supposed to be used publicly, which needs to be added to the index file


KavithaSiva (approved):
LGTM

PR #79, opened 2024-08-19
chore: Add cjs check

tomfrenken:
With this PR, we add the option to transpile to commonJS and add it to our build action to ensure we can compile to CJS on every PR.


tomfrenken:
Important note: With the introduction of exports as-is, imports are now a lot more restrictive, only allowing for usage of functions in the index.js file.

If we want to make use of internal functionality, we have to extend the imports file like for example, this:
```
"exports": {
  ".": {
    "import": "./dist/index.js",
    "require": "./dist/cjs/index.js",
    "types": "./dist/index.d.ts"
  },
  "./internal": {
    "import": "./dist/internal.js",
    "require": "./dist/cjs/internal.js"
  }
}
```
after creating an internal.ts file, just in the cloud sdk.

.github/workflows/e2e-test.yaml
```diff
@@ -40,7 +40,7 @@ jobs:
       - name: "Check Sample App Starts Up"
         working-directory: ./sample-code
         run: |
-          pnpm run local &
+          pnpm local &
```
marikaner, line 43: I am pretty sure I read somwhere in the pnpm docs, that it is recommended to use `run` in CI. I honestly don't care much as I don't see the risk of omitting `run`. So go for it, just keep this in the back of your mind if issues do occur.

package.json
```diff
@@ -9,7 +9,8 @@
   "type": "module",
   "scripts": {
     "postinstall": "pnpm compile",
-    "compile": "pnpm -r -w=false run compile",
+    "compile": "pnpm -r -w=false compile",
```
marikaner, line 12: I know this was here before, but what is `-w` supposed to be? According to the docs this does not exist and I would assume for a boolean option to be `false` by default.

packages/ai-core/package.json
```diff
@@ -8,16 +8,25 @@
...
     "dist/**/*.d.ts",
     "dist/**/*.d.ts.map"
   ],
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/cjs/index.js",
+      "types": "./dist/index.d.ts"
+    }
+  },
```
marikaner, line 26: [req] Why do we expose this here? This was basically make this hybrid, which did not want to do until further notice.

packages/ai-core/tsconfig.cjs.json
```diff
@@ -0,0 +1,7 @@
+{
```
marikaner, line 1: [req] With this config the CJS config will be shipped when releasing. This should not be the case.

packages/core/src/index.ts
```diff
@@ -2,7 +2,8 @@ export {
   executeRequest,
   BaseLlmParametersWithDeploymentId,
   BaseLlmParameters,
-  CustomRequestConfig
+  CustomRequestConfig,
+  EndpointOptions
```
marikaner, line 6: [q] Is it intended to make this part of the public API? Looks somewhat "internal" to me.

marikaner (changes requested):
I got a few questions + we should not ship and expose the CJS code.

.github/workflows/e2e-test.yaml
tomfrenken: Hmm I see, I wanted to get rid of the inconsistent use of `run`, if we see issues occurring, I would vote for adding it to every pnpm command.

package.json
tomfrenken: It's the very first option in the docs?
https://pnpm.io/pnpm-cli#-w---workspace-root

tomfrenken: Hmm after removing it I do not see any difference in the output, so you might be onto something

packages/core/src/index.ts
tomfrenken: Yeah I had a brief discussion about it and it is intended to be public.

marikaner (approved):
LGTM

PR #84, opened 2024-08-22
chore: Adjust orchestration tests

KavithaSiva:
Adjusts tests as part of AI/gen-ai-hub-sdk-js-backlog#44

The way we define placeholders for parameters in the template changed, see [here](https://github.tools.sap/AI/llm-orchestration/blob/main/src/spec/api.yaml#L278), this PR adjusts tests accordingly.
I also enhanced the Orchestration E2E test, just in case, to be aware if something changes, but this can be discussed.

Before:
```TypeScript
template: [ { role: 'user', content: 'Create {number} paraphrases of {phrase}' } ]
```
After:
 ```TypeScript
template: [ { role: 'user', content: 'Create {{?number}} paraphrases of {{?phrase}}' } ]
```

### DIFF
--- a/packages/gen-ai-hub/src/orchestration/orchestration-completion-post-request.test.ts
+++ b/packages/gen-ai-hub/src/orchestration/orchestration-completion-post-request.test.ts
@@ -46,6 +46,7 @@ describe('constructCompletionPostRequest()', () => {
     expect(completionPostRequest).toEqual(expectedCompletionPostRequest);
   });
 
+  // Todo: Adapt the test after Cloud SDK fix for: https://github.com/SAP/cloud-sdk-backlog/issues/1234
   it('with model configuration and empty template', async () => {
     genaihubCompletionParameters.prompt.template = [];
     const expectedCompletionPostRequest: CompletionPostRequest = {
@@ -262,6 +263,7 @@ describe('constructCompletionPostRequest()', () => {
     expect(completionPostRequest).toEqual(expectedCompletionPostRequest);
   });
 
+  // Todo: Adapt the test after Cloud SDK fix for: https://github.com/SAP/cloud-sdk-backlog/issues/1234
   it('with model configuration, prompt template empty filter configuration', async () => {
     genaihubCompletionParameters.filterConfig = {};
     const expectedCompletionPostRequest: CompletionPostRequest = {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/gen-ai-hub/src/orchestration/orchestration-completion-post-request.test.ts [hunk 1]
```diff
@@ -46,10 +46,33 @@ describe('constructCompletionPostRequest()', () => {
     expect(completionPostRequest).toEqual(expectedCompletionPostRequest);
   });
 
+  it('with model configuration and empty template', async () => {
+    genaihubCompletionParameters.prompt.template = [];
```
tomfrenken, line 50: This is changing state for all tests, as it is globally scoped, either delete it or assign this value already when the variable is initialized.

tomfrenken: Never mind, since we reassign it beforeEach it should be fine.

packages/gen-ai-hub/src/orchestration/orchestration-completion-post-request.test.ts
```diff
@@ -46,10 +46,33 @@ describe('constructCompletionPostRequest()', () => {
...
+    const expectedCompletionPostRequest: CompletionPostRequest = {
+      orchestration_config: {
+        module_configurations: {
+          templating_module_config: {
+            template: []
+          },
```
KavithaSiva, line 56: Note: This results in a bad request in the orchestration service, I am not sure, if we should restrict this use case somehow in the client.

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -20,7 +20,18 @@ describe('orchestration', () => {
         model_params: { max_tokens: 50, temperature: 0.1 }
       },
       prompt: {
-        template: [{ role: 'user', content: 'Hello!' }]
+        template: [
```
tomfrenken, line 23: Can this test be extended to check if the templating works as we expect it to?

tomfrenken (commented):
Overall, LGTM, just one more open question regarding the E2E test.

tests/e2e-tests/src/orchestration.test.ts
KavithaSiva: That's a good idea, done.

packages/gen-ai-hub/src/orchestration/orchestration-completion-post-request.test.ts
KavithaSiva: I have asked Franzi if they could add a `minItems` in the spec for `template` as well.

tomfrenken (approved):
LGTM

PR #90, opened 2024-08-27
feat: Combine deployment resolution and model

marikaner:
## Context

AI/gen-ai-hub-sdk-js-backlog#92.

This combines the deployment ID and model into one parameter. I also refactored some of the code to be more concise and fixed some issues that I found along the way.

Unfortunately the API does not get better for orchestration. Regardless, I would like to merge this one before adjusting it in here and getting inspired from python.

## Definition of Done

- [x] Documentation updated
- [x] (Optional) ~Aligned changes with the Java SDK~
- [x] (Optional) ~Release notes updated~

### DIFF
--- a/packages/gen-ai-hub/src/client/openai/openai-client.test.ts
+++ b/packages/gen-ai-hub/src/client/openai/openai-client.test.ts
@@ -59,11 +59,9 @@ describe('openai client', () => {
         chatCompletionEndpoint
       );
 
-      const response = await client.chatCompletion(
-        'gpt-35-turbo',
-        prompt,
-        '1234'
-      );
+      const response = await client.chatCompletion(prompt, {
+        deploymentId: '1234'
+      });
       expect(response).toEqual(mockResponse);
     });
 
@@ -86,7 +84,7 @@ describe('openai client', () => {
       );
 
       await expect(
-        client.chatCompletion('gpt-4', prompt, '1234')
+        client.chatCompletion(prompt, { deploymentId: '1234' })
       ).rejects.toThrow('status code 400');
     });
   });
@@ -111,11 +109,9 @@ describe('openai client', () => {
         },
         embeddingsEndpoint
       );
-      const response = await client.embeddings(
-        'text-embedding-ada-002',
-        prompt,
-        '1234'
-      );
+      const response = await client.embeddings(prompt, {
+        deploymentId: '1234'
+      });
       expect(response).toEqual(mockResponse);
     });
 
@@ -138,7 +134,7 @@ describe('openai client', () => {
       );
 
       await expect(
-        client.embeddings('text-embedding-3-large', prompt, '1234')
+        client.embeddings(prompt, { deploymentId: '1234' })
       ).rejects.toThrow('status code 400');
     });
   });
--- a/packages/gen-ai-hub/src/client/openai/openai-client.ts
+++ b/packages/gen-ai-hub/src/client/openai/openai-client.ts
@@ -105,6 +105,7 @@ function mergeRequestConfig(
 ): HttpRequestConfig {
   return {
     method: 'POST',
+    ...requestConfig,
     headers: mergeIgnoreCase(
       {
         'content-type': 'application/json'
@@ -114,7 +115,6 @@ function mergeRequestConfig(
     params: mergeIgnoreCase(
       { 'api-version': apiVersion },
       requestConfig?.params
-    ),
-    ...requestConfig
+    )
   };
 }
--- a/packages/gen-ai-hub/src/client/openai/openai-client.ts
+++ b/packages/gen-ai-hub/src/client/openai/openai-client.ts
@@ -5,13 +5,13 @@ import {
   getDeploymentId,
   type ModelDeployment
 } from '../../utils/deployment-resolver.js';
-import {
-  type OpenAiChatCompletionParameters,
-  type OpenAiEmbeddingParameters,
-  type OpenAiEmbeddingOutput,
-  type OpenAiChatCompletionOutput,
-  type OpenAiChatModel,
-  type OpenAiEmbeddingModel
+import type {
+  OpenAiChatCompletionParameters,
+  OpenAiEmbeddingParameters,
+  OpenAiEmbeddingOutput,
+  OpenAiChatCompletionOutput,
+  OpenAiChatModel,
+  OpenAiEmbeddingModel
 } from './openai-types.js';
 
 const apiVersion = '2024-02-01';
--- a/packages/gen-ai-hub/src/client/openai/openai-client.ts
+++ b/packages/gen-ai-hub/src/client/openai/openai-client.ts
@@ -1,12 +1,9 @@
 import { type HttpRequestConfig } from '@sap-cloud-sdk/http-client';
 import { type CustomRequestConfig, executeRequest } from '@sap-ai-sdk/core';
-import { mergeIgnoreCase, pickValueIgnoreCase } from '@sap-cloud-sdk/util';
+import { mergeIgnoreCase } from '@sap-cloud-sdk/util';
 import {
-  type ModelDeployment,
-  type FoundationModel,
-  type Model,
-  isDeploymentIdConfiguration,
-  resolveDeployment
+  getDeploymentId,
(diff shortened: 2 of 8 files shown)


packages/gen-ai-hub/src/utils/deployment-resolver.test.ts
```diff
@@ -73,9 +75,47 @@ describe('Deployment resolver', () => {
...
     await expect(
-      resolveDeployment({ scenarioId: 'foundation-models' })
+      resolveDeployment({
+        scenarioId: 'foundation-models',
+        model: { name: 'gpt-4o', version: '0613' }
+      })
     ).rejects.toThrow('No deployment matched the given criteria');
   });
+
+  it('should consider group ID', async () => {
```
MatKuhr, line 85:
```suggestion
  it('should consider custom resource group', async () => {
```

packages/gen-ai-hub/src/client/openai/openai-client.test.ts
```diff
@@ -86,7 +84,7 @@ describe('openai client', () => {
       );
 
       await expect(
-        client.chatCompletion('gpt-4', prompt, '1234')
+        client.chatCompletion(prompt, { deploymentId: '1234' })
```
MatKuhr, line 87: [pp] I'd feel that
```suggestion
        client.chatCompletion({ deploymentId: '1234' }, prompt)
```
is a bit more intuitive, especially if we are adding `new OpenAiClient('gpt-4o').chatCompletion(prompt)`

packages/gen-ai-hub/src/client/openai/openai-client.ts
```diff
@@ -21,22 +25,17 @@ const apiVersion = '2024-02-01';
...
-   * @param deploymentResolver - A deployment id or a function to retrieve it.
+   * @param deploymentConfig - This configuration is used to retrieve a deployment. Depending on the configuration use either the given deployment ID or the model name to retrieve matching deployments. If model and deployment ID are given, the model is verified against the deployment.
    * @param requestConfig - The request configuration.
    * @returns The completion result.
    */
   async chatCompletion(
-    model: OpenAiChatModel | { name: OpenAiChatModel; version: string },
     data: OpenAiChatCompletionParameters,
-    deploymentResolver?: DeploymentResolver,
+    deploymentConfig: DeploymentConfiguration<OpenAiChatModel>,
```
MatKuhr, line 35: [pp] Declaring this `DeploymentConfiguration` for me puts the focus on the deployment, which is rather a detail we would want to hide from users for most use cases. Instead, I'd prefer the API to naturally hint towards the models.

packages/gen-ai-hub/src/utils/deployment-resolver.ts
```diff
@@ -1,19 +1,65 @@
...
+/**
+ * Type guard to check if the given deployment configuration is a deployment ID configuration.
+ * @param deploymentConfig - Configuration to check.
+ * @returns `true` if the configuration is a deployment ID configuration, `false` otherwise.
+ */
+export function isDeploymentIdConfiguration(
+  deploymentConfig: DeploymentConfiguration
+): deploymentConfig is DeploymentIdConfiguration {
+  return (
+    typeof deploymentConfig === 'object' && 'deploymentId' in deploymentConfig
```
MatKuhr, line 47: [q] How does this work, considering the payload actually contains `id`?

packages/gen-ai-hub/src/client/openai/openai-client.test.ts
marikaner: I think I agree, no strong opinion though. When I didn't it somehow felt wrong. The only thing that makes this better though, is that in orchestration it is an optional parameter and it couldn't be optional if it came first.

packages/gen-ai-hub/src/client/openai/openai-client.ts
marikaner: Makes sense. I don't have the greatest ideas for renaming though. `ModelConfiguration` is not exact enough as it excludes the deployment ID. Just putting some ideas out there: `ModelDeployment`, `DeployedModelConfig`, something with an "or" inbetween, but I don't like that.

packages/gen-ai-hub/src/utils/deployment-resolver.ts
marikaner: I think you are referring to a different type `DeploymentConfiguration` is the parameter type and this function is a type guard that checks whether it is the "subtype" `DeploymentIdConfiguration`.

packages/gen-ai-hub/src/client/openai/openai-client.test.ts
marikaner: @MatKuhr I think at this point in time, this was a useless change. I can revert it, but I would suggest to keep it and then fix it with the next PR where we introduce something of this kind: `new OpenAiClient('gpt-4o').chatCompletion(prompt)`.

packages/gen-ai-hub/src/client/openai/openai-client.ts
marikaner: @deekshas8 @tomfrenken Any suggestions?

packages/gen-ai-hub/src/client/openai/openai-client.test.ts
MatKuhr: Why use the type in orchestration at all? Because there the model is a mandatory part of the payload..

But agreed, let's move that to a potential PR for `new OpenAiClient('gpt-4o').chatCompletion(prompt)` 😉

packages/gen-ai-hub/src/utils/deployment-resolver.ts
MatKuhr: ah okay never mind, I see that in a different place the `id` is extracted from the full response.

packages/gen-ai-hub/src/client/openai/openai-client.test.ts
marikaner: In orchestration it is a bit different, because users could potentially pass the deployment ID in addition to the mandatory payload (as of now). I will also note this down in my proposals for the follow up.

packages/gen-ai-hub/src/client/openai/openai-client.ts
```diff
@@ -1,16 +1,17 @@
...
+  type ModelDeployment
 } from '../../utils/deployment-resolver.js';
 import {
-  OpenAiChatCompletionParameters,
-  OpenAiEmbeddingParameters,
-  OpenAiEmbeddingOutput,
-  OpenAiChatCompletionOutput,
-  OpenAiChatModel,
-  OpenAiEmbeddingModel
+  type OpenAiChatCompletionParameters,
```
deekshas8, line 9: [pp] import type { ... } so it's not repeated.

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
```diff
@@ -16,25 +13,31 @@ export class OrchestrationClient {
...
     const body = constructCompletionPostRequest(data);
-    const deployment =
-      typeof deploymentResolver === 'function'
-        ? (await deploymentResolver()).id
-        : deploymentResolver;
+    deploymentId =
+      deploymentId ??
+      (
+        await resolveDeployment({
+          scenarioId: 'orchestration',
```
deekshas8, line 30: [q] In this case, we don't need to consider the resource group passed in request config (if)?

packages/gen-ai-hub/src/utils/deployment-resolver.ts
```diff
@@ -27,56 +65,121 @@ export interface FoundationModel {
...
+  executableId: string,
+  requestConfig?: CustomRequestConfig
+): Promise<string> {
+  if (isDeploymentIdConfiguration(modelDeployment)) {
+    return modelDeployment.deploymentId;
+  }
+
+  return (
+    await resolveDeployment({
+      scenarioId: 'foundation-models',
```
deekshas8, line 166: [pp] This function is for a specific scenario-id, maybe add this to the description?

deekshas8 (approved):
LGTM.

PR #93, opened 2024-08-29
feat: Carve out orchestration config and model deployment resolution

marikaner:
## Context

AI/gen-ai-hub-sdk-js-backlog#92.

Move out the orchestration and model deployment config into the constructor. (API design decided)
This is a proposal to ease discussion on the open discussion points:
- should it be possible to set the deployment ID explicitly through configuration?
- should it be possible to set the resource group explicitly through configuration?
- should the request config also be dragged up into the constructor?
  - should it then be possible to overwrite the general request config?

**update**
- New issue: We now have combined embedding + chat model types. To avoid this we could have separate clients for embedding and chat:
```ts
new OpenAIChatClient('chat-model').chatCompletion({...});
new OpenAIEmbeddingClient('embedding-model').embeddings({...});
```
This is not ideal because there is only one function on the class then, making it a bit weird to have a class in the first place.
An alternative could be to do it similar to the python lib:
```ts
new OpenAIClient().chatCompletion('chat-model').run({...});
new OpenAIClient().embeddings('embedding-model').run({...});
```
Then again, we are chaining functions, which is not the worst but something we tried to avoid.

(I will update and add tests once we have a final decision)

## Definition of Done

- [ ] Documentation updated


packages/gen-ai-hub/src/client/openai/openai-client.ts
```diff
@@ -18,20 +17,22 @@ const apiVersion = '2024-02-01';
  * OpenAI Client.
  */
 export class OpenAiClient {
+  // TODO: document constructor
+  //* @param modelDeployment - This configuration is used to retrieve a deployment. Depending on the configuration use either the given deployment ID or the model name to retrieve matching deployments. If model and deployment ID are given, the model is verified against the deployment.
+  constructor(private modelDeployment: ModelDeployment<OpenAiChatModel>) {}
```
marikaner, line 22: I left the deployment ID config in here as it was already part of the model and required less changes. If we decide to not allow setting it explicitly I will make a proposal without.

packages/gen-ai-hub/src/index.ts
```diff
@@ -1,7 +1,7 @@
 export * from './client/index.js';
 export {
   OrchestrationClient,
-  OrchestrationCompletionParameters,
+  OrchestrationModuleConfig as OrchestrationCompletionParameters,
```
marikaner, line 4:
```suggestion
  OrchestrationModuleConfig,
```

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
```diff
@@ -1,37 +1,45 @@
...
 
 /**
  * Get the orchestration client.
  */
 export class OrchestrationClient {
+  // TODO: either use the deploymentId from here or only allow setting it through the request config path
+  // TODO: document constructor
+  constructor(
+    private config: OrchestrationModuleConfig,
+    private deploymentIdConfig?: DeploymentIdConfiguration
```
marikaner, line 21: I left the deployment ID config in here as it was already implemented below and required less changes, merely changed it to an object instead of string. If we decide to not allow setting it explicitly I will make a proposal without.

packages/gen-ai-hub/src/orchestration/orchestration-types.ts
```diff
@@ -2,40 +2,40 @@ import {
...
+
   /**
-   * Template Parameters.
+   * Template parameters.
    */
-  template_params?: Record<string, InputParamsEntry>;
+  inputParams?: Record<string, InputParamsEntry>;
 }
 
+// TODO: why do we have this extra type? and if there is a reason, why does it not apply to the filtering module?
```
marikaner, line 24: Does anyone know the answer to this question?

deekshas8: For naming consistency (because the other Type and fieldname were called PromptConfig). 
The content filtering was done later and there it was missed. 

[pp] I would like if they were shorter, so without moduleConfig in the name (or at least without module, it feels unnecessary). But no strong opinion.

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
MatKuhr: For orchestration, in the future users will be able to store some parts of the config on server side and reference it by an ID. Accepting the deployment ID here would likely cause confusion once we add another opaque ID.

Personally, I'd remove the deployment ID it for now.

MatKuhr (commented):
just because I stumbled across [this](https://github.com/SAP/ai-sdk-js/blame/main/sample-code/src/orchestration.ts#L23), maybe you can also remove that with this PR.

packages/gen-ai-hub/src/orchestration/orchestration-types.ts
marikaner: But this only refers to the type, right? I would improve the names, but potentially leave the types. But also pp ;)

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
marikaner: Yes, I thought so, but I expect, that these references are part of the orchestration configuration. I don't really understand why you think it is *likely* to cause confusion, can you elaborate?

marikaner:
> just because I stumbled across [this](https://github.com/SAP/ai-sdk-js/blame/main/sample-code/src/orchestration.ts#L23), maybe you can also remove that with this PR.

Yes, once we have a final alignment on the API desgin, I will fix all other code as well, but you also mean to explicit reference, right?

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
tomfrenken: Without knowing the details, I would assume you can set the deployment id also in the server side, which probably takes precedence, meaning you can select a deployment_id client side, but it will be ignored, which would be confusing.

marikaner:
My opinion:
> should it be possible to set the deployment ID explicitly through configuration?

We supported this feature from the beginning and in the current API design it does not hurt. I think setting it through the request config is too inconvenient. I could imagine leaving it out in the orchestration API as I believe this is a less likely case, but then again, when needed, it is very inconvenient to overwrite and you have to do it per request.

> should it be possible to set the resource group explicitly through configuration?

I prefer that over reading it from the headers. The headers are set per request, which implies that they apply to the request only.
The other alternative, using `default` and not allowing other resource groups is too restrictive in my opinion. If users need it, they will try setting the header, which will not work together with deployment resolution. The API allows to set this, we should have a convenient way to set this as well - maybe discourage, but not disallow it.

> should the request config also be dragged up into the constructor?

No. I am not even sure we should allow the request config on the request in the first place. When the settings allowed by the API are configurable, I don't see a point in manipulating the request.
My impression on this is stronger for orchestration than foundation models, however, even there we could consider only adding this, when really needed and there are no other solutions in sight. This is different than the general OpenAPI and OData API handling, because we know the APIs explicitly and it's not like everyone can interpret the standard their own way.

> should it then be possible to overwrite the general request config?

Nope.

I believe we will have to solve these challenges at some point and we are already quite far in the discussion, so I feel leaving those things out would be a step back.

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
marikaner: @MatKuhr is that what you meant? I thought by config we mean the orchestration config and it does not have any references to the deployment ID.

tomfrenken:
> should it be possible to set the deployment ID explicitly through configuration?

Yes?
Perhaps my memory is already fading, but assuming that you can deploy multiple versions of a model, and you want to 
select a specific one, setting the deployment id would be a requirement?
Therefore if I deploy 2 `gpt 4.o` models, one with an old one with a new modelVersion, I must be able to set the deployment id? (Or is my understanding here wrong?)
>should it be possible to set the resource group explicitly through configuration?

No.
I don't think so, the only reason for setting a resource group that deviates from the default is a multi-tenant scenario, and there I would assume to set this on a per-request level.
>should the request config also be dragged up into the constructor?

Yes.
I would assume for all non-multi-tenant scenarios the request config should not change after client initialization.
If I want to consume a different model, I would usually initialize a different client as well.
>should it then be possible to overwrite the general request config?

The only use case for this in my head might be a multi-tenant scenario, where you want to adjust certain values specific to tenant selection, but if this were the case we might as well narrow the type accordingly. (e.g., only resource group & deployment id or so)

tomfrenken:
Hmm thinking further about points 3. & 4.

The only part of the request config I might be adjusting is the middleware and the only things I would adjust on a request level are resource groups and deployment ids (e.g., tenant-specific information).

Instead of exposing the request config anywhere, we could also consider exposing only a tenant config / selector and a .middleware() function, and a tenantConfig, thoughts?

marikaner:
> Hmm thinking further about points 3. & 4.
> 
> The only part of the request config I might be adjusting is the middleware and the only things I would adjust on a request level are resource groups and deployment ids (e.g., tenant-specific information).
> 
> Instead of exposing the request config anywhere, we could also consider exposing only a tenant config / selector and a .middleware() function, thoughts?

Middleware is less straight forward and the fancy version of a request config, isn't it. I wonder if this is a realistic use case.

tomfrenken:
> > Hmm thinking further about points 3. & 4.
> > The only part of the request config I might be adjusting is the middleware and the only things I would adjust on a request level are resource groups and deployment ids (e.g., tenant-specific information).
> > Instead of exposing the request config anywhere, we could also consider exposing only a tenant config / selector and a .middleware() function, thoughts?
> 
> Middleware is less straight forward and the fancy version of a request config, isn't it. I wonder if this is a realistic use case.

Good point, I was only thinking about adding resilience middlewares, for example langchain clients have built-in retry methods. Instead of exposing generic middlewares, we might expose the option to set a specific middleware from a curated set of resilience options?

tomfrenken:
Regarding the general multi-tenant setup that seemed to confuse me, in my head the setup could look like this:
1. A provider app might initialize a set of clients and offers a set of endpoints to subscribers
2. On these endpoints, only tenant-specific information would be set at execution time, which would serve for tenant-isolation. Is that understanding wrong?

marikaner:
> Regarding the general multi-tenant setup that seemed to confuse me, in my head the setup could look like this:
> 
> 1. A provider app might initialize a set of clients and offers a set of endpoints to subscribers
> 2. On these endpoints, only tenant-specific information would be set at execution time, which would serve for tenant-isolation.
> 
> Is that understanding wrong? From your description it sounded rather like I might initialize a client per tenant instead.
> 
> I would assume both is conceptually possible, if we want to support both, you need to be able to change tenant information at construction and execution.

Yes, but this is a very specific edge case and I think the user should take care of managing it. You cannot really provide "clients" through endpoints, so this doesn't even really apply here, does it?

packages/gen-ai-hub/src/orchestration/orchestration-client.ts
MatKuhr: > references are part of the orchestration configuration

That would be my expectation as well, but there is no decision on the details yet. I'd say it's not really problematic since the IDs will be in different objects, but still might increase the potential for user errors a little bit.

MatKuhr:
My opinion:

> should it be possible to set the deployment ID explicitly through configuration?

No, because I wouldn't know of a use case. We only need this ourselves for ease of testing.
Users would only need this when having more than one deployment per model and version, which would behave identical currently.
I am not aware of any parameters the user could set additionally to configure the two deployments differently.

If this is possible at a later point in time, we still should probably rather expand the filtering options for the user, e.g. by asking for a configuration name.

Hard coding the deployment ID guids should (hopefully) never be required for productive code.

> should it be possible to set the resource group explicitly through configuration?

Probably yes. I'd just give this as an option in the constructor and apply it to any requests the client does.

> should the request config also be dragged up into the constructor?

As you pointed out, dragging this up is awkward. But:

> I am not even sure we should allow the request config on the request in the first place.

I do think we need it. Custom headers and query parameters will be needed for some use cases, potentially even different URL prefixes. Exposing everything through dedicated API is probably not manageable.

> should it then be possible to overwrite the general request config?

Not sure what this is about 😄

marikaner:
Thanks 😊 

> > should it then be possible to overwrite the general request config?
> 
> Not sure what this is about 😄

When we would drag up request config, should it also be possible to set it in the request individually.

deekshas8:
> should it be possible to set the deployment ID explicitly through configuration?
I agree with Matthias's explanation here. This reminded me that the customer (beyondtech) also pointed this out in the demo (in the demo version, we had deploymentId parameter), so I am basing my assumption on: this is not a common use case for customers to pass it.

> should it be possible to set the resource group explicitly through configuration?
Yes. Unless it's something that can change per request (which atm I understand is not the case), setting it on client init makes sense to me. Also having a dedicated property than setting { header: { 'ai-resource-group': ...} } seems better API wise (but pp).

> should the request config also be dragged up into the constructor?
In general I would prefer this being a per "request" basis than when instantiating a client. Having this twice (in ctor and again for overriding in each call) is a bit awkward.

.changeset/silent-tools-beg.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ai-sdk/gen-ai-hub': minor
+'@sap-ai-sdk/core': minor
+---
+
+[New Functionality] Ensure that all external APIs are fully exported and any non-exported APIs are marked with `@internal`.
```
marikaner, line 6: Where did this come from?!

marikaner:
Thanks for all the comments, I am closing this PR in favor of this one: https://github.com/SAP/ai-sdk-js/pull/105

PR #95, opened 2024-08-30
feat: Add script to check public API visibility

shibeshduw:
## Context

Closes: AI/gen-ai-hub-sdk-js-backlog#94

This PR adds a script to check for all public APIs (similar to [cloud-sdk-js](https://github.com/SAP/cloud-sdk-js/blob/main/scripts/check-public-api.ts) with small adjustments) to ensure that the external API is exported completely and visible to the user. It also checks that everything not exported to the public API is marked with `@internal`.

## Definition of Done

- [ ] Documentation updated


scripts/check-public-api-cli.ts
```diff
@@ -0,0 +1,14 @@
...
+const logger = createLogger('check-public-api');
+
+/*
+For a detailed explanation what is happening here have a look at `0007-public-api-check.md` in the implementation documentation.
+ */
+checkApiOfPackage(process.cwd()).catch(err => {
+  logger.error(err);
+  process.exit(1);
+});
+// const cwd = "C:\\Users\\I745181\\Documents\\AISDK\\ai-sdk-js\\packages\\gen-ai-hub\\dist";
```
tomfrenken, line 13: I suppose this was for testing purposes?

shibeshduw: Whoopsie
Yeah it was

packages/gen-ai-hub/src/index.ts
```diff
@@ -1,10 +1,80 @@
-export * from './client/index.js';
+export type {
+  OpenAiChatModel,
```
tomfrenken, line 2: Did we check what part of this API we actually need to expose? Or is this simply exporting everything, but named?

shibeshduw: This is exporting everything that is not marked internal.

packages/gen-ai-hub/src/index.ts
```diff
@@ -1,10 +1,80 @@
...
+  ChatMessages,
+  UnmaskingConfig,
+  TokenUsage,
+  TemplatingModuleConfig,
+  TemplatingModuleResult,
+  PresidioEntities,
+  OrchestrationConfig,
+  ModuleResults,
+  ModuleConfigs,
+  Masking,
```
tomfrenken, line 55: Some of these do not seem to exist 🤔

shibeshduw: Ah i guess some of them got updated/deleted after the new orchestration spec was used to generate clients.

packages/gen-ai-hub/src/index.ts
tomfrenken: I'm not sure if we followed the @internal convention so far though :D for now this is fine, could you create a follow-up ticket to check if everything is properly tagged with @internal? And also make this a pre-requisite for the GA ticket.

scripts/check-public-api.ts
```diff
@@ -0,0 +1,359 @@
...
+  cwd: string,
+  barrelFileName: string
+): Promise<void> {
+  const barrelFilePath = join(cwd, barrelFileName);
+  if (existsSync(barrelFilePath) && (await lstat(barrelFilePath)).isFile()) {
+    const dirContents = (
+      await glob('*', {
+        ignore: [
+          '**/*.spec.ts',
+          '**/*.test.ts',
```
tomfrenken, line 314: I see you added .test.ts, however, I think you can also remove .spec.ts (since we do not use it)

scripts/check-public-api.ts
```diff
@@ -0,0 +1,359 @@
+/* eslint-disable jsdoc/require-jsdoc */
+
+import { join, resolve, parse, basename, dirname, sep, posix } from 'path';
```
tomfrenken, line 3: Some of these imports seem to be unnecessary

scripts/check-public-api.ts
```diff
@@ -0,0 +1,359 @@
...
+import path from 'path';
+import { fileURLToPath } from 'url';
+import { promises, existsSync } from 'fs';
+import { glob } from 'glob';
+import { createLogger, flatten, unixEOL } from '@sap-cloud-sdk/util';
+import mock from 'mock-fs';
+import { CompilerOptions } from 'typescript';
+import {
```
tomfrenken, line 11: Eslint is complaining about internal imports here, since it can't be prevented probably just exclude this line

tomfrenken (changes requested):
Overall LGTM, however some minor changes are still necessary.

Also, please add the check to our pipeline, e.g. same as we do in the cloud sdk.

packages/gen-ai-hub/src/index.ts
shibeshduw: I think for the AI sdk, we will just be enforcing that everything that is not exported through the public api is marked with @internal. The added script already checks for this.

tomfrenken: Yes, after the script runs on every PR and forces everyone to reflect if an API should be internal or public this makes sense.
Currently, though, this convention was not consistently followed, and with your PR you are adding every untagged API to the public API, even though this might not be intentional.

tomfrenken: Therefore we should at least check every now "public" API to make sure they should genuinely be public before GA.
After that this should never be necessary again.

shibeshduw: Ah okay I got what your mean now. I'll create a follow up to properly check for @internal consistency.

tomfrenken (approved):
LGTM

PR #96, opened 2024-08-30
chore: Add cache for deployment ID

marikaner:
## Context

AI/gen-ai-hub-sdk-js-backlog#78.

Introduce a cache for deployment IDs.
I did not offer a possibility for users to change the default validity time, let me know if you see reasons to introduce this.

For langchain integration we will probably need to cache model information as well. I am wondering whether we should cache the ID + model or just the whole deployment directly. @MatKuhr @deekshas8 @tomfrenken 

## Definition of Done

- [x] Documentation updated


packages/gen-ai-hub/src/utils/deployment-cache.ts
```diff
@@ -0,0 +1,83 @@
+// TODO: should we move this to @sap-cloud-sdk/connectivity? or add some kind of exception to the rule if possible?
```
marikaner, line 1: @deekshas8 @tomfrenken Did we align on whether the SAP Cloud SDK for AI is "allowed" to reach to the internal API? I think it should be ok, but we should then fix the lint rule for this.

packages/gen-ai-hub/src/utils/model.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/**
+ * Get the model information from a deployment.
+ * @param deployment - AI core model deployment.
+ * @returns The model information.
+ * @internal
+ */
+export function extractModel(
+  // TODO: this type does not seem to be correct (backend_details => backendDetails), check why
```
marikaner, line 30: I checked and it seems to be an actual mistake in the API spec. Requested a fix.

packages/gen-ai-hub/src/utils/deployment-cache.ts
tomfrenken: I agree, however, only for internals of the sap cloud SDK, not internals of the AI SDK, which mich make the rule a little more complicated.

packages/gen-ai-hub/src/utils/deployment-cache.ts
```diff
@@ -0,0 +1,83 @@
...
+import { Cache } from '@sap-cloud-sdk/connectivity/internal.js';
+import { type AiDeployment } from '@sap-ai-sdk/ai-core';
+import { type DeploymentResolutionOptions } from './deployment-resolver.js';
+import { extractModel, type FoundationModel } from './model.js';
+
+function getCacheKey({
+  scenarioId,
+  executableId,
+  model,
+  resourceGroup = 'default'
```
tomfrenken, line 12: If we set the default for the resource group here, we might as well do the same for all the other parameters?
At least this would be consistent and makes the string interpolation less loaded.

packages/gen-ai-hub/src/utils/deployment-cache.ts
```diff
@@ -0,0 +1,83 @@
...
+    model: extractModel(deployment)
+  };
+}
+
+/**
+ * Cache for deployments.
+ * @internal
+ */
+export const deploymentIdCache = createDeploymentCache(
+  new Cache(5 * 60 * 1000) // 5 minutes
```
tomfrenken, line 82: What is our invalidation strategy here?

I remember for the Cloud SDK we used the auth token lifetime, e.g.:
```ts
  const expiresIn = first(destination.authTokens || [])?.expiresIn;
  const expirationTime = expiresIn
    ? Date.now() + parseInt(expiresIn) * 1000
    : undefined;
 ```
 If I understand the PR correctly, if a deployment is removed, we would send failing request for up to 5 minutes?

File: packages/gen-ai-hub/src/utils/deployment-resolver.ts [hunk 5]
```diff
@@ -1,6 +1,8 @@
-import { DeploymentApi, AiDeployment } from '@sap-ai-sdk/ai-core';
+import { DeploymentApi, type AiDeployment } from '@sap-ai-sdk/ai-core';
 import { CustomRequestConfig } from '@sap-ai-sdk/core';
 import { pickValueIgnoreCase } from '@sap-cloud-sdk/util';
+import { deploymentIdCache } from './deployment-cache.js';
```
deekshas8, line 4: [q] If this is shared code between the clients, I think it should not lie in the gen-ai-hub package? Maybe ai-core?

packages/gen-ai-hub/src/utils/deployment-cache.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      model: { name: 'gpt-4o', version: 'latest' }
+    };
+    deploymentIdCache.set(opts, {
+      id: 'deployment-id',
+      details: {
+        resources: {
+          backend_details: { model: { name: 'gpt-4o', version: 'latest' } }
+        }
+      }
+    } as unknown as AiDeployment);
```
tomfrenken, line 23: Probably some typescript hiccup, but why as unknown as AiDeployment?

packages/gen-ai-hub/src/utils/deployment-cache.ts
marikaner: I can only do this for `executableId`, that is a good idea.

packages/gen-ai-hub/src/utils/deployment-cache.ts
marikaner: yes

packages/gen-ai-hub/src/utils/deployment-resolver.ts
marikaner: good call, I just left it where the deployment resolver was. I would like to do this on a separate PR, as this would likely lead to issues with my other PR (deployment resolver).

packages/gen-ai-hub/src/utils/deployment-cache.test.ts
marikaner: exactly
Screenshot 2024-09-02 at 14 54 19

packages/gen-ai-hub/src/utils/deployment-cache.ts
tomfrenken: 5 minutes feels rather arbitrary

packages/gen-ai-hub/src/utils/deployment-cache.ts
deekshas8: Seems like Junjie did: https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/40

packages/gen-ai-hub/src/utils/deployment-cache.ts
deekshas8: I do see a `cache.clear(`), so for testing its still fine

marikaner: I will check how realistic it is to implement this. I think it generally makes sense, however the more flexibility the user has, the more difficult it will be to decide whether a 404 is based on an incorrect deployment id or something else.

marikaner: we agreed that we will do this as a follow if possible

tomfrenken (approved):
Pipeline seems to have caught some minor stuff, other than that LGTM

packages/gen-ai-hub/src/utils/deployment-cache.ts
MatKuhr: I think it's unlikely that deployments will change often.

* Any new models will still be picked up, regardless of cache duration.
* Changing the deployment of an existing model has no benefits and I wouldn't know a use case for doing that
* Deleting a deployment can be done, but requests would fail regardless of caching or not, because the desired model no longer exists. The failures would be slightly more explicit without the cache, but the 404 should also be somewhat clear, in particular with proper logging 😉 
* For local testing caching is usually not a problem

In short: I think we should cache quite long by default, e.g. 12hrs or even more

packages/gen-ai-hub/src/utils/deployment-cache.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    expect(
+      deploymentCache.get({
+        ...opts,
+        model: { name: 'gpt-4o' }
+      })?.id
+    ).toEqual('deployment-id1');
+  });
+
+  it('should ignore model versions without model name', () => {
```
MatKuhr, line 105: any particular reason why? Ideally, this never happens, right?

packages/gen-ai-hub/src/utils/deployment-cache.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      })
+    ]);
+
+    expect(deploymentCache.get(opts)).toBeUndefined();
+    expect(
+      deploymentCache.get({
+        ...opts,
+        model: {
+          name: 'gpt-35-turbo',
+          version: 'latest'
```
MatKuhr, line 54: I would expect the test to pass with the following change, but it fails:

```suggestion
          name: 'gpt-35-turbo'
```

packages/gen-ai-hub/src/utils/deployment-cache.ts
```diff
@@ -0,0 +1,81 @@
...
+     * @param opts - Deployment resolution options to set the deployments for. Model information in the deployment resolution options are ignored.
+     * @param deployments - Deployments to retrieve the IDs and models from.
+     */
+    setAll: (
+      opts: Omit<DeploymentResolutionOptions, 'model'>,
+      deployments: AiDeployment[]
+    ): void => {
+      // go backwards to cache the first deployment ID for each model
+      [...deployments].reverse().forEach(deployment => {
+        cache.set(getCacheKey({ ...opts, model: extractModel(deployment) }), {
```
MatKuhr, line 59: I think this strategy of building the cache-key is a bit awkward and for the main use case probably even broken?

Broken: The model version is part of the cache key, but when users query using only the model name, a cache miss apparently occurs. See [this comment](https://github.com/SAP/ai-sdk-js/pull/96/files#r1741103818). Or am I missing something?

Awkward: We are partially building the cache key from the actual result, and partially from the user input. I guess the issue here is that `executableId` is not part of the response payload. This again leads to unnecessary cache misses where the same deployment is queried, sometimes with and other times without the executable ID. But I guess since it's not part of the response we don't have much of a choice here. We could query configurations in addition (which have the executable ID), but that would probably be over-optimizing 😄

packages/gen-ai-hub/src/utils/deployment-cache.test.ts
marikaner: Yes, I wanted to make sure that we don't accidentally allow it though. The way we build the cache key this should not be possible anyways.

packages/gen-ai-hub/src/utils/deployment-cache.ts
marikaner: > I think this strategy of building the cache-key is a bit awkward and for the main use case probably even broken?
> 
> Broken: The model version is part of the cache key, but when users query using only the model name, a cache miss apparently occurs. See [this comment](https://github.com/SAP/ai-sdk-js/pull/96/files#r1741103818). Or am I missing something?

Good catch, I will add this in a follow up.
 
> Awkward: We are partially building the cache key from the actual result, and partially from the user input. I guess the issue here is that `executableId` is not part of the response payload. This again leads to unnecessary cache misses where the same deployment is queried, sometimes with and other times without the executable ID. But I guess since it's not part of the response we don't have much of a choice here. We could query configurations in addition (which have the executable ID), but that would probably be over-optimizing 😄

I thought about that as well, but at least when you use the same options twice you get a cached result, so might be ok. The other problem is worse ;)

PR #98, opened 2024-09-02
chore: Add response handling

deekshas8:
## Context

AI/gen-ai-hub-sdk-js-backlog#90.

Add a wrapper around completion output in both openai and orchestration clients. This wrapper class provides access to the normal types response, raw response as well as convenience methods

## Definition of Done

- [x] Documentation updated


packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+  /**
+   * Reason for stopping the model.
+   */
+  get finish_reason() {
+    return this.data.choices[0].finish_reason;
+  }
+
+  /**
+* Parses the Open AI response and returns the content.
```
marikaner, line 27: It seems that the linting didn't work here?

packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Usage of tokens in the response.
+   */
+  get usage() {
+    return this.data.usage;
+  }
+  /**
+   * Reason for stopping the model.
+   */
+  get finish_reason() {
```
marikaner, line 22: [req] Snake case does not follow our style guide. I would actually suggest to have a function instead of a getter. You could also consider allowing passing an index.

packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,47 @@
...
+      throw new Error('Invalid choice index.');
+    }
+    return choices[choiceIndex].message.content
+  }
+
+  /**
+   * Get the raw HttpResponse.
+   * @returns HttpResponse.
+   */
+  getRawResponse(): HttpResponse {
```
marikaner, line 44: [pp] Do we need a function for this? I would prefer to just expose this as a public (maybe readonly) property.

packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,47 @@
...
+    return this.data.choices[0].finish_reason;
+  }
+
+  /**
+* Parses the Open AI response and returns the content.
+* @param response - The  Open AI response.
+* @param choiceIndex - The index of the choice to parse.
+* @returns The message content.
+*/
+  getContent(choiceIndex: number = 0): string | null {
```
marikaner, line 32: [q] When would I get `null`? Maybe add this to the docs. It seems that we filter out `undefined`, so theoretically this can only happen when the message content itself really is there, but `null`. (see next comment)

marikaner: [req] I think we discussed that if there is a finish reason other than `success` or whatever the specific success message is, we would throw an error with the finish reason. Potentially this would also filter out the `null` messages (although, I thought those messages are usually empty string?)

packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,47 @@
...
+    const choices = this.data.choices;
+    if (choiceIndex < 0 || choiceIndex >= choices.length) {
+      throw new Error('Invalid choice index.');
```
marikaner, line 35: Just a few thoughts:
1. If we throw an error I would like to add the given index to the message and maybe use the phrasing "choice index out of bounds" or so.
2. Do we really want to throw an error in this case? I guess when we do it on finish reason "content_filter", we can probably do that in this case too. I think I generally prefer the more graceful exit without throwing errors (also for the finish reason), but I currently don't have a better reason than my gut feeling.

packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,47 @@
...
+export class OpenAiChatCompletionResponse {
+  public data: OpenAiChatCompletionOutput
+  constructor(private rawResponse: HttpResponse,) {
+    this.data = rawResponse.data;
+  }
+
+  /**
+   * Usage of tokens in the response.
+   */
+  get usage() {
```
marikaner, line 16: [pp] I would prefer to be a function `getUsage()`.

packages/gen-ai-hub/src/client/openai/openai-types.ts
```diff
@@ -345,7 +345,7 @@ export interface OpenAiEmbeddingParameters {
 /**
  * Usage statistics for the completion request.
  */
-interface OpenAiUsage {
+export interface OpenAiUsage {
```
marikaner, line 348: [q] Is this used somewhere or is it just because it is supposed to be public API?

packages/gen-ai-hub/src/orchestration/orchestration-response.ts
```diff
@@ -0,0 +1,47 @@
+import { HttpResponse } from "@sap-cloud-sdk/http-client";
```
marikaner, line 1: same as openai

marikaner (changes requested):

packages/gen-ai-hub/src/client/openai/openai-types.ts
deekshas8: This will be fixed in the other PR (generated client). Ideally all generated types should be exported.

packages/gen-ai-hub/src/orchestration/orchestration-response.ts
```diff
@@ -0,0 +1,66 @@
...
+  CompletionPostResponse,
+  TokenUsage
+} from './client/api/schema/index.js';
+
+const logger = createLogger({
+  package: 'gen-ai-hub',
+  messageContext: 'orchestration-response'
+});
+/**
+ * Respresentation of Orchestration response.
```
tomfrenken, line 13:
```suggestion
 * Representation of an Orchestration response.
```

packages/gen-ai-hub/src/orchestration/orchestration-response.ts
```diff
@@ -0,0 +1,66 @@
...
+   * Get the typed Completion Post Response.
+   */
+  public readonly data: CompletionPostResponse;
+  constructor(public readonly rawResponse: HttpResponse) {
+    this.data = rawResponse.data;
+  }
+
+  /**
+   * Usage of tokens in the response.
+   * @returns The usage tokens.
```
tomfrenken, line 26:
```suggestion
   * @returns The token usage.
```

packages/gen-ai-hub/src/orchestration/orchestration-response.ts
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * Get the typed Completion Post Response.
+   */
+  public readonly data: CompletionPostResponse;
+  constructor(public readonly rawResponse: HttpResponse) {
+    this.data = rawResponse.data;
+  }
+
+  /**
+   * Usage of tokens in the response.
```
tomfrenken, line 25:
```suggestion
   * Usage of tokens for the response.
```

tomfrenken: Not sure if for is the right suggestion, but in my head the tokens were used to create the response, so this might be more appropriate?

sample-code/src/aiservice.ts
```diff
@@ -16,9 +13,8 @@ export async function chatCompletion(): Promise<string> {
     },
     'gpt-35-turbo'
   );
-  const assistantMessage = response.choices[0]
-    .message as OpenAiChatAssistantMessage;
-  return assistantMessage.content!;
+  const assistantMessage = response.getContent();
```
tomfrenken, line 16: You could also:
```ts
return response.getContent()!
```

Unless you wanted to. make explicit that an assistant message is returned

sample-code/src/orchestration.ts
```diff
@@ -7,10 +7,10 @@ import {
  * Ask GPT about the capital of France.
  * @returns The answer from the orchestration service in Gen AI Hub.
  */
-export function orchestrationCompletion(): Promise<CompletionPostResponse> {
+export async function orchestrationCompletion(): Promise<CompletionPostResponse> {
   const orchestrationClient = new OrchestrationClient();
 
-  return orchestrationClient.chatCompletion(
+  const response = await orchestrationClient.chatCompletion(
```
tomfrenken, line 13: It's a little bit of a red flag the same type can be used even though you now only return the data property.

Should the type be different / more narrow, or do you think it's fine?

packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,47 @@
...
+  public readonly data: OpenAiChatCompletionOutput;
+  constructor(public readonly rawResponse: HttpResponse) {
+    this.data = rawResponse.data;
+  }
+
+  /**
+   * Usage of tokens in the response.
+   * @returns The usage tokens.
+   */
+  getUsageTokens(): OpenAiUsage {
```
tomfrenken, line 23:
```suggestion
  getTokenUsage(): OpenAiUsage {
```

packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export class OpenAiChatCompletionResponse {
+  public readonly data: OpenAiChatCompletionOutput;
+  constructor(public readonly rawResponse: HttpResponse) {
+    this.data = rawResponse.data;
+  }
+
+  /**
+   * Usage of tokens in the response.
+   * @returns The usage tokens.
```
tomfrenken, line 21:
```suggestion
   * @returns The used tokens.
```

packages/gen-ai-hub/src/client/openai/openai-types.ts
```diff
@@ -444,7 +444,7 @@ export interface OpenAiChatCompletionChoice extends OpenAiCompletionChoice {
   /**
    * Completion chat message.
    */
-  message: OpenAiChatMessage;
+  message: OpenAiChatAssistantMessage;
```
tomfrenken, line 447: How did this type change occur?

packages/gen-ai-hub/src/client/openai/openai-client.test.ts
```diff
@@ -62,7 +62,11 @@ describe('openai client', () => {
       const response = await client.chatCompletion(prompt, {
         deploymentId: '1234'
       });
-      expect(response).toEqual(mockResponse);
+
+      expect(response).toBeInstanceOf(OpenAiChatCompletionResponse);
+      expect(response.data).toEqual(mockResponse);
+      expect(response.getContent()).toEqual(expect.any(String));
+      expect(response.getFinishReason()).toEqual(expect.any(String));
```
tomfrenken, line 69: In the orchestration test you also test the .getUsageTokens() function, is this intentionally left out?

tomfrenken (commented):
All in all looks good, just a few questions & suggestions.

packages/gen-ai-hub/src/orchestration/orchestration-response.ts
deekshas8: I just copied this from the spec

packages/gen-ai-hub/src/client/openai/openai-types.ts
deekshas8: As discussed, this is avoid unnecessary check in the convenient method, the type will eventually be corrected in a followup PR

packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  /**
+   * Usage of tokens in the response.
+   * @returns Token usage.
+   */
+  getTokenUsage(): OpenAiUsage {
+    return this.data.usage;
+  }
+  /**
```
marikaner, line 26:
```suggestion

  /**
```

packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,48 @@
...
+   */
+  getTokenUsage(): OpenAiUsage {
+    return this.data.usage;
+  }
+
+  /**
+   * Reason for stopping the model.
+   * @returns The finish reason.
+   */
+  getFinishReason(): string | null {
```
marikaner, line 31: [req] I would add the optional index here as well for consistency.

tomfrenken (approved):
LGTM

marikaner (changes requested):
We could consider introducing some shared interface or similar to make sure we expose a common set of properties and functions in the responses throughout orchestration and foundation models. However, I think for now this would be overkill.

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -32,11 +32,15 @@ describe('orchestration', () => {
...
-    expect(response.module_results.templating).not.toHaveLength(0);
-    expect(response.module_results.templating?.[0].content).toBe(
-      'Create 3 paraphrases of I love coffee and I dislike cats'
-    );
-    expect(response.orchestration_result.choices).not.toHaveLength(0);
+    expect(response.data.module_results).toBeDefined();
+    expect(response.data.module_results.templating).not.toHaveLength(0);
+    expect(response.data.orchestration_result.choices).not.toHaveLength(0);
+    expect(response.getContent()).toBe(expect.any(String));
+    expect(response.getFinishReason()).toBe(expect.any(String));
```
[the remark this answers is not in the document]
deekshas8: There can be multiple finish reasons, and if we change the mock json response, then this test would have to be updated. So I tried to keep string fields generic.

deekshas8:
> We could consider introducing some shared interface or similar to make sure we expose a common set of properties and functions in the responses throughout orchestration and foundation models.

I will create a ticket so we can do/consider this as a follow up

packages/gen-ai-hub/src/client/openai/openai-response.test.ts
```diff
@@ -0,0 +1,47 @@
+import { parseMockResponse } from '../../../../../test-util/mock-http';
+import { OpenAiChatCompletionResponse } from './openai-response';
+import { OpenAiChatCompletionOutput } from './openai-types';
+
+describe('OpenAi Response', () => {
```
marikaner, line 5:
```suggestion
describe('OpenAI response', () => {
```

packages/gen-ai-hub/src/client/openai/openai-response.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  it('should return default choice index with convenience functions', () => {
+    expect(openAiChatClientResponse.getFinishReason()).toBe('stop');
+    expect(openAiChatClientResponse.getContent()).toBe(
+      "The deepest place on Earth is located in the western Pacific Ocean. It's called the Mariana Trench."
+    );
+  });
+
+  it('should return undefined when convenience function is called with incorrect index', () => {
+    expect(openAiChatClientResponse.getFinishReason(1)).toBeUndefined();
```
marikaner, line 44: [nth] you could consider also checking that the error is logged.

packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,51 @@
...
+import { createLogger } from '@sap-cloud-sdk/util';
+import { OpenAiChatCompletionOutput, OpenAiUsage } from './openai-types.js';
+
+const logger = createLogger({
+  package: 'gen-ai-hub',
+  messageContext: 'openai-response'
+});
+
+/**
+ * Open AI chat completion response.
```
marikaner, line 11:
```suggestion
 * OpenAI chat completion response.
```

packages/gen-ai-hub/src/client/openai/openai-types.ts
```diff
@@ -100,7 +100,7 @@ export interface OpenAiChatAssistantMessage {
   /**
    * Message content.
    */
-  content: string | null;
+  content?: string;
```
marikaner, line 103: I thought this was a generated type and could be null as well, e.g. if for some reason the content from the API is actually `null`?

packages/gen-ai-hub/src/client/openai/openai-types.ts
```diff
@@ -426,7 +426,7 @@ export interface OpenAiCompletionChoice {
   /**
    * Reason for finish.
    */
-  finish_reason: string | null;
+  finish_reason?: string;
```
marikaner, line 429: Same here.

packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Usage of tokens in the response.
+   * @returns Token usage.
+   */
+  getTokenUsage(): OpenAiUsage {
+    return this.data.usage;
+  }
+
+  /**
+   * Reason for stopping the model.
```
marikaner, line 28:
```suggestion
   * Reason for stopping the completion.
```

packages/gen-ai-hub/src/client/openai/openai-response.ts
```diff
@@ -0,0 +1,51 @@
...
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns The finish reason.
+   */
+  getFinishReason(choiceIndex = 0): string | undefined {
+    if (choiceIndex < 0 || choiceIndex >= this.data.choices.length) {
+      logger.error(`${choiceIndex} is not a valid choice index.`);
```
marikaner, line 34:
```suggestion
      logger.error(`Choice index ${choiceIndex} is out of bounds.`);
```

packages/gen-ai-hub/src/orchestration/orchestration-response.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  private validateChoiceIndex(choiceIndex: number): boolean {
+    const choices = this.data.orchestration_result.choices;
+    if (choiceIndex < 0 || choiceIndex >= choices.length) {
+      logger.error(`${choiceIndex} is not a valid choice index.`);
+      return false;
+    }
+    return true;
+  }
```
marikaner, line 72: There are multiple suggestions in the code below:

```suggestion
  getFinishReason(choiceIndex = 0): string | undefined {
    this.validateChoiceIndex(choiceIndex));
    return this.getChoices()[choiceIndex]?.finish_reason;
  }

  /**
   * Parses the orchestration response and returns the content.
   * @param choiceIndex - The index of the choice to parse.
   * @returns The message content.
   */
  getContent(choiceIndex = 0): string | undefined {
    this.validateChoiceIndex(choiceIndex));

    if (
      this.getChoices()[choiceIndex].message.content ===
        '' &&
      this.getChoices()[choiceIndex].finish_reason ===
        'content_filter'
    ) {
      throw new Error(
        'Content generated by the LLM was filtered by the output filter. Please try again with a different prompt or filter configuration.'
      );
    }
    return this.getChoices()[choiceIndex]?.message?.content;
  }

  private getChoices() {
    return this.data.orchestration_result.choices;
  }
  
  private validateChoiceIndex(choiceIndex: number): void {
    if (choiceIndex < 0 || choiceIndex >= this.getChoices().length) {
      logger.error(`Choice index ${choiceIndex} is out of bounds.`);
    }
  }
   
    
```

packages/gen-ai-hub/src/orchestration/orchestration-response.ts
```diff
@@ -0,0 +1,68 @@
...
+import {
+  CompletionPostResponse,
+  TokenUsage
+} from './client/api/schema/index.js';
+
+const logger = createLogger({
+  package: 'gen-ai-hub',
+  messageContext: 'orchestration-response'
+});
+/**
```
marikaner, line 12:
```suggestion

/**
```

PR #102, opened 2024-09-03
fix: Deployments should be cached for non-model-information

marikaner:
## Context

AI/gen-ai-hub-sdk-js-backlog#78.

Follow up from #96.

When retrieving a deployment for a model name without version and there are only deployments where each model has a version, the second time we use this deployment, the cache wasn't used.

## Definition of Done

- [ ] Documentation updated

### DIFF
--- a/packages/gen-ai-hub/src/utils/deployment-cache.ts
+++ b/packages/gen-ai-hub/src/utils/deployment-cache.ts
@@ -58,18 +58,13 @@ function createDeploymentCache(cache: Cache<Deployment>) {
       const cacheEntries = [...deployments]
         .reverse()
         .map(deployment => transformDeploymentForCache(deployment))
-        .flatMap(entry => {
-          const entries = [entry];
-          // if the entry has a model, also cache it without model and version
-          if (entry.model) {
-            entries.push({ id: entry.id });
-          }
-          // if the entry has a model version, also cache it for without model version
-          if (entry.model?.version) {
-            entries.push({ id: entry.id, model: { name: entry.model.name } });
-          }
-          return entries;
-        });
+        .flatMap(entry => [
+          entry,
+          { id: entry.id },
+          ...(entry.model
+            ? [{ id: entry.id, model: { name: entry.model.name } }]
+            : [])
+        ]);
 
       cacheEntries.forEach(entry => {
         cache.set(getCacheKey({ ...opts, model: entry.model }), {
--- a/packages/gen-ai-hub/src/utils/deployment-cache.ts
+++ b/packages/gen-ai-hub/src/utils/deployment-cache.ts
@@ -55,7 +55,7 @@ function createDeploymentCache(cache: Cache<Deployment>) {
       deployments: AiDeployment[]
     ): void => {
       // go backwards to cache the first deployment ID for each model
-      const cacheEntries = [...deployments]
+      [...deployments]
         .reverse()
         .map(deployment => transformDeploymentForCache(deployment))
         .flatMap(entry => [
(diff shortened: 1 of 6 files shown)


packages/gen-ai-hub/src/utils/deployment-cache.ts
```diff
@@ -51,20 +52,57 @@ function createDeploymentCache(cache: Cache<Deployment>) {
...
     setAll: (
-      opts: Omit<DeploymentResolutionOptions, 'model'>,
+      opts: DeploymentResolutionOptions,
       deployments: AiDeployment[]
     ): void => {
       // go backwards to cache the first deployment ID for each model
-      [...deployments].reverse().forEach(deployment => {
-        cache.set(getCacheKey({ ...opts, model: extractModel(deployment) }), {
-          entry: transformDeploymentForCache(deployment)
+      const cacheEntries = [...deployments].reverse().map(deployment => ({
```
marikaner, line 59: @MatKuhr what do you think about that?

My 2 cents: semantically much better, implementation-wise more error prone and complicated to read. I am not convinced ;)

packages/gen-ai-hub/src/utils/deployment-cache.ts
```diff
@@ -51,20 +52,50 @@ function createDeploymentCache(cache: Cache<Deployment>) {
...
     setAll: (
-      opts: Omit<DeploymentResolutionOptions, 'model'>,
+      opts: DeploymentResolutionOptions,
       deployments: AiDeployment[]
     ): void => {
       // go backwards to cache the first deployment ID for each model
-      [...deployments].reverse().forEach(deployment => {
-        cache.set(getCacheKey({ ...opts, model: extractModel(deployment) }), {
-          entry: transformDeploymentForCache(deployment)
+      const cacheEntries = [...deployments]
```
marikaner, line 59: @MatKuhr how do you like that?

packages/gen-ai-hub/src/utils/deployment-cache.ts
```diff
@@ -51,20 +52,75 @@ function createDeploymentCache(cache: Cache<Deployment>) {
...
     setAll: (
-      opts: Omit<DeploymentResolutionOptions, 'model'>,
+      opts: DeploymentResolutionOptions,
       deployments: AiDeployment[]
     ): void => {
-      // go backwards to cache the first deployment ID for each model
-      [...deployments].reverse().forEach(deployment => {
-        cache.set(getCacheKey({ ...opts, model: extractModel(deployment) }), {
-          entry: transformDeploymentForCache(deployment)
+      const cacheEntries = [...deployments].map(deployment =>
```
marikaner, line 58: @MatKuhr how do you like that?

my 2 cents: I think it is semantically better, however more error prone.

marikaner: Ok, after multiple tries and still failing tests, I am reverting this. Feel free to check the commit, if you are interested.

packages/gen-ai-hub/src/utils/deployment-cache.ts
```diff
@@ -51,13 +51,29 @@ function createDeploymentCache(cache: Cache<Deployment>) {
...
+          // if the entry has a model, also cache it without model and version
+          if (entry.model) {
+            entries.push({ id: entry.id });
+          }
+          // if the entry has a model version, also cache it for without model version
+          if (entry.model?.version) {
+            entries.push({ id: entry.id, model: { name: entry.model.name } });
+          }
+          return entries;
+        });
```
marikaner, line 72: This could still be shortened if we remove the checks:
```suggestion
        .flatMap(entry => [entry, { id: entry.id }, { id: entry.id, model: { name: entry.model.name } }]);
```

Even though this is a little less performant, I think I even prefer that, but no strong opinion.

MatKuhr (approved):

PR #111, opened 2024-09-05
fix: Internal files should also be ESM

marikaner:
## Context

I noticed the files were exporting CJS instead of ESM. Somehow this causes issues in ai-core where I wanted to add this, but not in the other packages ¯\\\_(ツ)\_/¯


shibeshduw (approved):
Looks straightforward.

PR #112, opened 2024-09-05
chore: generate CJS build info files in dist-cjs

marikaner:
## Context

Accidentally noticed that those files are written into the dist dir, which might not be expected (doesn't cause issues afaik though).


tomfrenken:
Weird

PR #107, opened 2024-09-04
chore: Move utils

marikaner:
## Context

AI/gen-ai-hub-sdk-js-backlog#100.

Moving it out as it will be shared between later separated packages.

PR #113, opened 2024-09-05
fix: Update public api script

shibeshduw:
## Context

Updating the script to ignore AI_CORE_CLIENTS during the checking process because by default, everything in there should be public. 
This is just a temporary workaround to merge https://github.com/SAP/ai-sdk-js/pull/107. We would ideally split the script to handle external imports and internal imports separately in the future.

## Definition of Done

- [ ] Documentation updated

### DIFF
--- a/scripts/check-public-api.ts
+++ b/scripts/check-public-api.ts
@@ -138,7 +138,7 @@ export async function checkApiOfPackage(pathToPackage: string): Promise<void> {
         usePrettier: false
       }
     }, { exclude: includeExclude?.exclude!, include: ['**/*.ts'] });
-    await checkBarrelRecursive(pathToSource);
+    // await checkBarrelRecursive(pathToSource);
 
     const indexFilePath = join(pathToSource, 'index.ts');
     checkIndexFileExists(indexFilePath);
--- a/scripts/check-public-api.ts
+++ b/scripts/check-public-api.ts
@@ -177,7 +177,9 @@ export function checkIndexFileExists(indexFilePath: string): void {
 export async function typeDescriptorPaths(cwd: string): Promise<string[]> {
   const files = await glob('**/*.d.ts', { cwd });
   return files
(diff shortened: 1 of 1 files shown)


scripts/check-public-api.ts
```diff
@@ -292,7 +294,7 @@ export async function checkBarrelRecursive(cwd: string): Promise<void> {
   (await readdir(cwd, { withFileTypes: true }))
     .filter(dirent => dirent.isDirectory())
     .forEach(async subDir => {
-      if (['__snapshot__', 'spec'].includes(subDir.name)) {
+      if (!['__snapshot__', 'spec', 'tests'].includes(subDir.name)) {
```
marikaner, line 297: Is this needed when it is on the ignore list?

scripts/check-public-api.ts
```diff
@@ -272,7 +274,7 @@ export async function parseIndexFile(filePath: string): Promise<string[]> {
   const starFiles = captureGroupsFromGlobalRegex(
     /export \* from '([\w\/.-]+)'/g,
     fileContent
-  );
+  ).filter(file => file !== './client/AI_CORE_API/index.js');
```
marikaner, line 277: Why do we need to skip those, if the \* check works?

scripts/check-public-api.ts
```diff
@@ -177,7 +177,9 @@ export function checkIndexFileExists(indexFilePath: string): void {
 export async function typeDescriptorPaths(cwd: string): Promise<string[]> {
   const files = await glob('**/*.d.ts', { cwd });
   return files
-    .filter(file => !file.endsWith('index.d.ts'))
+    .filter(file => 
+      !file.endsWith('index.d.ts') &&
+      !file.startsWith('client\\AI_CORE_API\\'))
```
marikaner, line 182: Why do we need to skip those?

marikaner (approved):

PR #120, opened 2024-09-09
feat: Add Langchain integration

tomfrenken:
## Context

AI/gen-ai-hub-sdk-js-backlog#24.

With this PR, we add a langchain module, currently only containing an OpenAI client.

## Definition of Done

- [x] Documentation updated
~- [ ] (Optional) Aligned changes with the Java SDK~
- [x] (Optional) Release notes updated (covered by Junjie's PR)


sample-code/package.json
```diff
@@ -26,6 +26,8 @@
     "@sap-ai-sdk/ai-api": "workspace:^",
     "@sap-ai-sdk/foundation-models": "workspace:^",
     "@sap-ai-sdk/orchestration": "workspace:^",
+    "@sap-ai-sdk/langchain": "workspace:^",
+    "@langchain/core": "^0.2.30",
```
deekshas8, line 30: [q] So users would need to install this separately as well to use types?

deekshas8 (commented):
Review still WIP. Left some minor questions in the meanwhile

packages/langchain/src/openai/util.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+/**
+ * Chunk an array into smaller arrays of specified chunk size.
+ * @param arr - Input array to be chunked.
+ * @param chunkSize - Size of each chunk.
+ * @returns Array of chunks.
+ * @internal
+ */
+export function chunkArray(arr: string[], chunkSize: number): string[][] {
```
deekshas8, line 168: [q] Why not use the `splitInChunks` from cloud sdk util?

packages/langchain/src/openai/util.ts
```diff
@@ -0,0 +1,212 @@
...
+    case 'human':
+      return 'user';
+    case 'system':
+      return 'system';
+    case 'function':
+      return 'function';
+    case 'tool':
+      return 'tool';
+    case 'generic':
+      return (message as ChatMessage).role as OpenAiChatMessage['role'];
```
deekshas8, line 77: [q] If a message type is generic, the role field still aligns with the recognized roles in OpenAiChatMessage?

packages/langchain/src/openai/types.ts
```diff
@@ -0,0 +1,64 @@
...
+  OpenAiChatCompletionParameters,
+  OpenAiChatModel,
+  OpenAiEmbeddingModel
+} from '@sap-ai-sdk/foundation-models';
+import type { ConfigurationOptions } from '@sap-ai-sdk/ai-api';
+
+/**
+ * Input for Text generation for OpenAI GPT.
+ */
+export type OpenAIChatModelInput = Omit<
```
deekshas8, line 18: [nth] A comment why we omit certain properties here would help. If I remember correctly it was because in langchain they are required but not in our azure openai client (or maybe the other way around), but it's not obvious right away.

packages/langchain/src/openai/types.ts
```diff
@@ -0,0 +1,64 @@
...
+      | 'tools'
+      | 'response_format'
+    >,
+    Pick<
+      OpenAiChatCompletionParameters,
+      'tool_choice' | 'functions' | 'tools' | 'response_format'
+    > {}
+
+/**
+ * Input for Text generation for OpenAI GPT.
```
deekshas8, line 57: [req] Description is incorrect

packages/langchain/src/openai/util.test.ts
```diff
@@ -0,0 +1,106 @@
+// mapResponseToChatResult
```
deekshas8, line 1: [req] leftover comments

packages/ai-api/src/utils/deployment-resolver.ts
```diff
@@ -46,8 +46,17 @@ export interface ResourceGroupConfiguration {
...
   | ModelNameT
-  | ((ModelConfiguration<ModelNameT> | DeploymentIdConfiguration) &
-      ResourceGroupConfiguration);
+  | ConfigurationOptions<ModelNameT>;
+
+/**
+ * The configuration options for a model deployment.
+ * @typeParam ModelNameT - String literal type representing the name of the model.
+ */
+export type ConfigurationOptions<ModelNameT> = Xor<
```
marikaner, line 55: [req] This name is not speaking at all, could also be `Option` or `Configuration` or `OptionsConfiguration` - aka. quite random ;)
Consider something like `ModelDeploymentConfig` or `ModelDeploymentDetails` or similar.

packages/langchain/README.md
```diff
@@ -0,0 +1,122 @@
+# @sap-ai-sdk/langchain
+
+This package contains Langchain compliant models, based on the @sap-ai-sdk clients.
```
marikaner, line 3: [req] It is called "LangChain" (pascal case). Please make sure to always refer to it by its correct name.

packages/langchain/README.md
```diff
@@ -0,0 +1,122 @@
...
+```
+$ npm install @sap-ai-sdk/langchain
+
+$ npm install @langchain/openai // if you want to use OpenAI models
+```
+
+## Pre-requisites
+
+- [Enable the AI Core service in BTP](https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/initial-setup).
+- Project configured with Node.js v20 or higher and native ESM support enabled.
```
marikaner, line 16: [req]All other points are in imperative. I don't care whether you use the imperative or past tense for this, but it should be consistent.

packages/langchain/README.md
```diff
@@ -0,0 +1,122 @@
...
+- For testing your application locally:
+  - Download a service key for your AI Core service instance.
+  - Create a `.env` file in the sample-code directory.
```
marikaner, line 19: [q] How is that related to "testing your application locally"? Please note, that those are the instruction on npm. Seems a bit confusing to me and rather related to our sample code / E2E tests.

packages/langchain/README.md
```diff
@@ -0,0 +1,122 @@
...
+  - Add an entry `AICORE_SERVICE_KEY='<content-of-service-key>'`.
+
+## Usage
+
+All client's comply with [Langchain's interface](https://python.langchain.com/v0.2/api_reference/openai/chat_models/langchain_openai.chat_models.azure.AzureChatOpenAI.html#langchain_openai.chat_models.azure.AzureChatOpenAI), therefore you should be able to use them as per usual.
```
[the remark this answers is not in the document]
marikaner: [pp] "you should be able to use them as usual" seems like:
1. we are unsure whether it works, indicating low quality on our side.
2. doppelt-gemoppelt. If it complies with langchain it means you can use it as expected, so why mention it explicitly. I would omit the part after the comma.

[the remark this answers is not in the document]
marikaner: [q/req] Why are we linking to the python API foe AzureOpenAI for Chat here. Seems oddly specific for the whole LangChain package and irrelevant as it is python. I think it would be more useful to link to JS (=>https://js.langchain.com/docs/introduction) and maybe not link to a specifc API version, but a general explanation of the concept.

packages/langchain/src/openai/util.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  'foundation-models',
+  'openai-chat-completion-success-response.json'
+);
+
+const chatCompletionEndpoint = {
+  url: 'inference/deployments/1234/chat/completions',
+  apiVersion: '2024-02-01'
+};
+
+const prompt = {
```
deekshas8, line 29: [pp] Can you specify the types for each variable, it's easier to understand which type is expected to transform into which.

packages/langchain/src/openai/chat.ts
```diff
@@ -0,0 +1,74 @@
...
+import type { ChatResult } from '@langchain/core/outputs';
+import { AzureChatOpenAI, AzureOpenAI } from '@langchain/openai';
+import { OpenAiChatClient as OpenAiChatClientBase } from '@sap-ai-sdk/foundation-models';
+import { mapLangchainToAiClient, mapResponseToChatResult } from './util.js';
+import type { OpenAiChatModelInput, OpenAiChatCallOptions } from './types.js';
+
+/**
+ * OpenAI Language Model Wrapper to generate texts.
+ */
+export class OpenAiChatClient extends AzureChatOpenAI {
```
deekshas8, line 12: [q] I think Zhongpin changed the type names to include Azure as a prefix. We should probably do it for langchain package as well?

packages/langchain/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+The only difference is in the initialization of the client, where you have th option to pass either:
```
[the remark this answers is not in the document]
marikaner: [pp] => "the only difference" implies that we are only almost compliant with the langchain interface and now we explain a workaround, because we do something special. To my understanding this is not the case.

Consider something more neutral in this direction:
```suggestion
To initialize the client pass either:
```

Still, I would even prefer it to lead with the main use case of passing the model name, so that people can find this  as a TLDR and then follow up with additional information like the fact that you can pass a model version and resource group optionally or instead pass the deployment ID, if you want to discover it on your own.

packages/langchain/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+or
+
+```ts
+    deploymentId: string,
+    resourceGroup?: string
+    ...others
+```
+
+If you pass API keys they are ignored, since you're not inteded to call the vendor's endpoints directly.
```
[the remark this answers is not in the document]
marikaner: [req] I think you cannot say "you are not intended" => "du bist nicht beabsichtigt". "it is not intended" or "you are not supposed to" is correct.
Also I would keep "are" ignored as I remember from the technical writing course to avoid future tense ;)

packages/langchain/README.md
```diff
@@ -0,0 +1,122 @@
...
+Instead, the credentials in the binding are used to call SAP's LLM Proxy.
```
marikaner, line 44: [q] is this the official name?

packages/langchain/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+### OpenAI
+
+We offer two types of clients for OpenAI models.
+Currenty these are chat and embedding models.
```
marikaner, line 49: [pp] agreed with Deeksha +:
```suggestion
```

marikaner (changes requested):
Just took a first glance, will continue review later

packages/langchain/src/openai/chat.ts
```diff
@@ -0,0 +1,74 @@
...
+  declare CallOptions: OpenAiChatCallOptions;
```
deekshas8, line 13: [q] why not  `callOptions`, also is it meant to be public?

sample-code/src/langchain-openai.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Embed 'Hello, world!' using the OpenAI ADA model.
+ * @returns An embedding vector.
+ */
+export async function embedQuery(): Promise<number[]> {
+  const client = new OpenAiEmbeddingClient({
+    modelName: 'text-embedding-ada-002'
+  });
+  const response = await client.embedQuery('Hello, world!');
+
+  return response;
```
marikaner, line 26:
```suggestion
  return client.embedQuery('Hello, world!');
```

sample-code/src/server.ts
```diff
@@ -62,6 +63,32 @@ app.get('/ai-api/get-deployments', async (req, res) => {
...
+    res
+      .status(500)
+      .send('Yikes, vibes are off apparently 😬 -> ' + error.message);
+  }
+});
+
+app.get('/langchain/embedding', async (req, res) => {
+  try {
+    const result = await embedQuery();
+    if (result.length === 0) {
```
marikaner, line 80: [req] style guide!
```suggestion
    if (!result.length) {
```

marikaner (changes requested):
- please get a bit more familiar with the style guide and try to apply it to your code
- please make sure to reference "LangChain" correctly everywhere
- lots of questions ;)

looks quite good overall though, but big prs get big reviews ;)

packages/langchain/src/openai/types.ts
```diff
@@ -0,0 +1,64 @@
...
+import type {
+  AzureOpenAiChatModel,
+  AzureOpenAiEmbeddingModel
+} from '@sap-ai-sdk/core';
+import type { ConfigurationOptions } from '@sap-ai-sdk/ai-api';
+
+/**
+ * Input for Text generation for OpenAi GPT.
+ */
+export type OpenAiChatModelInput = Omit<
```
MatKuhr, line 18: [q] Why omit these things? Why shouldn't users be allowed to set these hyperparameters when instantiating the client?

sample-code/src/server.ts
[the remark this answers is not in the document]
tomfrenken: Mainly because I tried to use the same approach as we do with our existing embedding endpoint.
Do you think I should change this (then probably in both places)?

packages/langchain/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "@sap-ai-sdk/foundation-models": "workspace:^",
+    "@langchain/core": "^0.2.30",
+    "zod-to-json-schema": "^3.23.2"
+  },
+  "peerDependencies": {
+    "@langchain/openai": "^0.2.8"
+  },
+  "peerDependenciesMeta": {
+    "@langchain/openai": {
+      "optional": true
```
MatKuhr, line 42: [q] while I'm generally in favor of the optional approach, I'm wondering if it's worth it. Currently, the [langchain openai package](https://www.npmjs.com/package/@langchain/openai?activeTab=dependencies) brings only 4 dependencies. In addtion, our `foundation-models` package brings _all_ foundation models anyway. How much stuff would `@langchain/openai` add?

sample-code/package.json
tomfrenken: Users can also create these objects on their own or alternatively only pass strings.

packages/langchain/src/openai/util.ts
```diff
@@ -0,0 +1,212 @@
...
+  };
+}
+
+/**
+ * Maps a {@link BaseMessage} to OpenAI's Message Role.
+ * @param message - The message to map.
+ * @returns The OpenAI Message Role.
+ * @internal
+ */
+export function mapBaseMessageToRole(
```
MatKuhr, line 62: [q] isn't there some function inside `@langchain/openai` that we can use for this, and only re-export it with the correct type?

I.e. doing
```suggestion
export function mapToolToOpenAiTool( msg: BaseMessage): OpenAiChatMessage['role'] { 
  return langchainFunction(msg) as OpenAiChatMessage['role']; 
}
```

sample-code/src/langchain-openai.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Ask GPT about the capital of France.
+ * @returns The answer from GPT.
+ */
+export async function generate(): Promise<string> {
+  const client = new OpenAiChatClient({ modelName: 'gpt-35-turbo' });
+  const response = await client.generate([
+    [new HumanMessage('What is the capital of France?')]
+  ]);
+  return response.generations[0][0].text;
```
MatKuhr, line 13: Instead of `response.generations[0][0].text;` I'd rather demo some actual langchain usage:

```suggestion
import { StringOutputParser } from "@langchain/core/output_parsers";

const parser = new StringOutputParser();
client.pipe(parser);
```

otherwise, the above code arguably has not much if any benefit or other distinction compared to the standalone OpenAI client

packages/langchain/src/openai/util.ts
```diff
@@ -0,0 +1,212 @@
...
+ * Maps the langchain's input interface to our own client's input interface
+ * @param client The Langchain OpenAI client
+ * @param options The Langchain call options
+ * @param messages The messages to be send
+ * @returns A AI SDK compatibile request
+ * @internal
+ */
+export function mapLangchainToAiClient(
+  client: OpenAiChatClient,
+  options: OpenAiChatCallOptions,
```
ZhongpinWang, line 188: [q] Is this type really correct? I see `OpenAiChatCallOptions` is defined as:

```ts
export interface OpenAiChatCallOptions
  extends Omit<
      ChatOpenAICallOptions,
      | 'tool_choice'
      | 'promptIndex'
      | 'functions'
      | 'function_call'
      | 'tools'
      | 'response_format'
    >,
    Pick<
      OpenAiChatCompletionParameters,
      'tool_choice' | 'functions' | 'tools' | 'response_format'
    > {}
```

It omits `functions` and `tools` (which has `schema` property supporting zod schema) from langchain and pick them from our sdk, which uses json schema. But I assume, options here should still be in the original langchain format or?

packages/langchain/README.md
tomfrenken: I figured you have to follow the same steps when using this library and then testing your app.
I can see how this might be redundant though, as this could then be said for any library ever. 🤷

packages/langchain/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "check:public-api": "node --loader ts-node/esm ../../scripts/check-public-api-cli.ts"
+  },
+  "dependencies": {
+    "@sap-ai-sdk/ai-api": "workspace:^",
+    "@sap-ai-sdk/foundation-models": "workspace:^",
+    "@langchain/core": "^0.2.30",
+    "zod-to-json-schema": "^3.23.2"
+  },
+  "peerDependencies": {
+    "@langchain/openai": "^0.2.8"
```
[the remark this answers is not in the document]
tomfrenken: Well version v0.3 exist since Friday 🤷

packages/langchain/src/openai/chat.ts
```diff
@@ -0,0 +1,74 @@
...
+  private openAiChatClient: OpenAiChatClientBase;
+
+  constructor(fields: OpenAiChatModelInput) {
```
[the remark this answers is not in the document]
tomfrenken: This is what we do! However, JavaScript users might unintentionally provide these fields still, in which case we overwrite them as they could break the intended behavior.

packages/langchain/README.md
```diff
@@ -0,0 +1,122 @@
+# @sap-ai-sdk/langchain
+
+This package contains Langchain compliant models, based on the @sap-ai-sdk clients.
+
+### Installation
```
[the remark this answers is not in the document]
tomfrenken: I believe the license etc. has to be in here still?
The packages are released as (somewhat) independent parts on npm, or am I missing something?

packages/langchain/src/openai/types.ts
tomfrenken: They still are able to pass them, we omit the false type (langchains type is outdated) and add the correct type (the one we've generated), together, they form a type where these parameters can still be set, but are now (correctly) optional.

sample-code/src/server.ts
tomfrenken: While I agree I simply aligned my "new" endpoints with the existing implementation, if I change it here, I assume I should also change the current /embedding endpoint?

packages/langchain/README.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+A chat completion example might be:
+
+```ts
+const response = await chatClient.generate([
+  [new SystemMessage('You are an IT support agent answering questions.')],
+  [new HumanMessage('Why is my internet not working?')]
+]);
+```
```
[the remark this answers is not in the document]
tomfrenken: Yeah that's a good point, I tried to align the documentation to the LLM-Common documentation to some degree, but I suppose the only difference between the different clients will be the name, not their interface.

packages/langchain/src/openai/chat.ts
tomfrenken: CallOptions is a property from the superclass, by declaring it I only give it its proper type.
I guess I could overwrite the visibility from the superclass, but I'm not sure if that's a good idea, what do you think?

🤔

packages/langchain/README.md
tomfrenken: I wasn't sure what the name of the component we call with the credentials is, do you have a better suggestion? 😬

packages/langchain/src/openai/chat.ts
```diff
@@ -0,0 +1,74 @@
...
+    const defaultValues = new AzureOpenAI({ apiKey: 'dummy' });
+    const stop = fields.stop
+      ? Array.isArray(fields.stop)
+        ? fields.stop
+        : [fields.stop]
+      : defaultValues.stop;
```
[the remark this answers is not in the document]
tomfrenken: Stop is handled differently because the types from langchain and us are crashing on a different level so to say.
All other types are matching the "real" type, but are actually optional.
Stop is not matching the "real" type (string | string[] | undefined) the superclass is inferring this type: (string[] | undefined), to match the superclass' type, we cast it to an array if it isn't one.

I agree I could use a util function for this though.

packages/ai-api/src/utils/deployment-resolver.ts
tomfrenken: Yes agreed, I was also not satisfied with the name, but there were so many renamings that I wanted to wait until somethings sticks before I change it (again) 😬

packages/langchain/src/openai/embedding.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    this.btpOpenAIClient = new OpenAiEmbeddingClientBase({ ...fields });
+  }
+
+  override async embedDocuments(documents: string[]): Promise<number[][]> {
+    const chunkedPrompts = chunkArray(
+      this.stripNewLines
+        ? documents.map(t => t.replace(/\n/g, ' '))
+        : documents,
+      this.batchSize
```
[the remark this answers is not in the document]
tomfrenken: There are no other formating options, the chat client doesn't have any of these.

packages/langchain/src/openai/types.ts
MatKuhr: hmm okay, but then why are we not using our type directly e.g. in the constructor?

packages/langchain/src/openai/chat.ts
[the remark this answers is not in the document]
tomfrenken: I agree that it would be fine to switch to ChatOpenAi, the question is what do we prefer to support, users switch from OpenAI or AzureOpenAi, on the implementation side its easy to go either way.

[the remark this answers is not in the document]
tomfrenken: Do you have a preference?

packages/langchain/package.json
tomfrenken: Well `langchain/openai` would always be on top of our types, as there is no overlap.
As the number of vendors grows this package would also grow, therefore I would go for this approach regardless.
I suppose it would not be that easy for foundation models, since an alternative approach would require us to create separate packages for all vendors.

If many support issues start popping up because users do not properly install the required packages, we can reconsider of course.

MatKuhr: > Well langchain/openai would always be on top of our types, as there is no overlap.

Hmm not sure I understand that correctly. My understanding is that, if the user doesn't import it, running `new OpenAiChatClient({ modelName: 'gpt-35-turbo' });` will fail. And further, that `@sap-ai-sdk/langchain` already depends on `foundation-models`, which will bring all vendors anyway. 

Not sure what you mean with overlap, but essentially not including `langchain/openai` is worth it if and only if it would bring a bunch of unnecessary packages, right?

packages/langchain/src/openai/chat.ts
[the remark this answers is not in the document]
MatKuhr: I would prefer not to offer non-functional API wherever we can 😄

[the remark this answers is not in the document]
tomfrenken: Well the don't offer the non-functional part, it's omitted from the type, but if users make the switch, it still works as we overwrite the additional parts.

packages/langchain/src/openai/types.ts
tomfrenken: Because the langchain type also includes properties which are not part of the API

packages/langchain/package.json
tomfrenken: Yes or rather if the package size increases by (arguably) too much by including every single vendor.
From a quick glimpse on npm the average package size seems to be around 200-400kb * vendors (assuming no overlap, although there might be some), I would consider especially for users who are wary of big package sizes (e.g. cap) smaller packages are certainly the favorable outcome.

packages/langchain/src/openai/chat.ts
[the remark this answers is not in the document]
MatKuhr: Urgh I see, things like `azureOpenAIBasePath` are in addition also defined on `ChatOpenAI`, not just `AzureChatOpenAI` 🙄

But eitheray, `AzureChatOpenAI` is pretty useless, right? We could even consider to directly use `BaseChatModel` with our own types. I don't see much that `ChatOpenAI` is adding that we can't easily add ourselves based on the already existing client...

packages/langchain/package.json
MatKuhr: AFAI can see, [langchain/openai](https://www.npmjs.com/package/@langchain/openai?activeTab=dependencies) only brings one additional dependency: [openai](https://www.npmjs.com/package/openai). The other ones overlap with [langchain/core](https://www.npmjs.com/package/@langchain/core?activeTab=dependencies). 

However, `openai` itself is pretty large with 2.5MB, so probably still better to have it optional then.

packages/langchain/src/openai/types.ts
MatKuhr: For example? Are those things we actually need? Would there maybe be a benefit in using a union type of our type + additional langchain fields instead?

packages/langchain/src/openai/chat.ts
marikaner: No, go with the standard. But is this the standard way to give it a type? If not, why do we need a different type?

packages/langchain/src/openai/chat.ts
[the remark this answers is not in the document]
tomfrenken: It's 100% useless and arguably less clean than using BaseChatModel, however, I did it to make the switch as easy as possible.
Do you think going with BaseChatModel is the preferred approach over easier migration?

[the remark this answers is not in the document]
MatKuhr: I guess so? I think users just rename `BTPOpenAIChat` to `ChatOpenAI` or whatever we decide on, and all the passed parameters would remain valid, regardless of which approach we take, wright?

packages/langchain/src/openai/util.ts
tomfrenken: I don't think so, I saw llm-commons doing this and was wondering the same thing.
I was searching through langchains API docs but didn't find any class which has the type "generic".

packages/langchain/src/openai/types.ts
MatKuhr: Understood, but the things that don't overlap are either:
* missing from our OpenAI client (e.g. logprobs)
   * ->  and we should add them there, right?
*  Or they are missing on purpose and we don't want them (e.g. __includeRawResponse)..

packages/langchain/src/openai/util.ts
tomfrenken: I'll check if LLM Access has some special message type for their API, if they do then I should probably just change the type ..., otherwise I will remove this special case handling.

packages/langchain/src/openai/types.ts
tomfrenken: Yeah that's a good point 🤔 , I'll adjust it!

packages/langchain/src/openai/chat.ts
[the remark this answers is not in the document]
tomfrenken: That is true, however, the default behavior would change.
LangChain's OpenAI client sets lots of defaults (e.g. temperature to 0.7, ...) whereas we would not do that if we just use the base model.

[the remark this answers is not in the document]
MatKuhr: Okay, I'd think changing the defaults would be an acceptable change. Maybe we need some research on what reasonable defaults should be anyway and also apply them in the standalone client. For me it would feel unexpected if e.g. the temperature default changes between the standalone client and langchain usage..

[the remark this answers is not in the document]
tomfrenken: I would expect the server to have reasonable defaults, as it already allows for these values to be undefined, therefore these defaults seem to be redundant IMO.

packages/langchain/src/openai/util.test.ts
deekshas8: No, I imagined these are existing types right? For example, request would be type `OpenAiChatCompletionParameters` ?

packages/langchain/src/openai/util.ts
tomfrenken: Realizing such type issues had to take a pretty in-depth review, so thanks for that already :D 
I think using our types here was correct since the same type was used to invoke any method of our langchain client.

I did change almost all types & classes by now though, so it should be (hopefully) clear which type we are intended to use.

packages/ai-api/src/index.ts
```diff
@@ -4,5 +4,5 @@ export type {
   ModelDeployment,
   DeploymentIdConfiguration,
   ModelConfiguration,
-  ResourceGroupConfiguration
+  ResourceGroupConfiguration,
```
marikaner, line 7: ?

packages/langchain/README.md
```diff
@@ -0,0 +1,117 @@
+# @sap-ai-sdk/LangChain
```
marikaner, line 1: [req] nope!

packages/langchain/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@sap-ai-sdk/foundation-models": "workspace:^",
+    "@langchain/core": "0.3.1",
+    "zod-to-json-schema": "^3.23.2",
+    "@sap-cloud-sdk/util": "^3.20.0"
+  },
+  "peerDependencies": {
+    "@langchain/openai": "^0.3.0"
+  },
+  "peerDependenciesMeta": {
+    "@langchain/openai": {
```
marikaner, line 40: [q] is this still needed?

packages/langchain/package.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "dependencies": {
+    "@sap-ai-sdk/ai-api": "workspace:^",
+    "@sap-ai-sdk/foundation-models": "workspace:^",
+    "@langchain/core": "0.3.1",
+    "zod-to-json-schema": "^3.23.2",
+    "@sap-cloud-sdk/util": "^3.20.0"
+  },
+  "peerDependencies": {
+    "@langchain/openai": "^0.3.0"
```
marikaner, line 37: [q] is this still needed?

packages/langchain/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@sap-ai-sdk/langchain",
+  "version": "0.0.0",
+  "description": "",
```
marikaner, line 4: [req] Please add a description

packages/langchain/src/openai/chat.ts
```diff
@@ -0,0 +1,77 @@
...
+import { BaseChatModel } from '@langchain/core/language_models/chat_models';
+import { AzureOpenAiChatModel } from '@sap-ai-sdk/core';
+import {
+  mapLangchainToAiClient,
+  mapResponseToChatResult
+} from './util.js';
+import type { OpenAiChatCallOptions, OpenAiChatModelParams } from './types.js';
+
+/**
+ * OpenAI Language Model Wrapper to generate texts.
```
marikaner, line 14: [pp] As this is not a title but a description, I wouldn't use title case.
```suggestion
 * OpenAI language model wrapper to generate texts.
```

packages/langchain/src/openai/chat.ts
```diff
@@ -0,0 +1,77 @@
+import { CallbackManagerForLLMRun } from '@langchain/core/callbacks/manager';
+import { BaseMessage } from '@langchain/core/messages';
+import type { ChatResult } from '@langchain/core/outputs';
+import { OpenAiChatClient as OpenAiChatClientBase } from '@sap-ai-sdk/foundation-models';
```
marikaner, line 4: Make sure to merge main!

packages/langchain/src/openai/types.ts
```diff
@@ -0,0 +1,55 @@
...
+import type {
+  AzureOpenAiChatModel,
+  CustomRequestConfig
+} from '@sap-ai-sdk/core';
+import type { ModelConfiguration, ResourceGroupConfiguration } from '@sap-ai-sdk/ai-api';
+
+/**
+ * Input type for OpenAI chat models.
+ */
+export type OpenAiChatModelParams = Omit<OpenAiChatCompletionParameters, 'messages'> &
```
marikaner, line 13: [req] For consistency all of those should be prefixed with "Azure"

packages/langchain/src/openai/chat.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.resourceGroup = fields.resourceGroup;
+    this.temperature = fields.temperature;
+    this.top_p = fields.top_p;
+    this.logit_bias = fields.logit_bias;
+    this.user = fields.user;
+    this.n = fields.n;
+    this.stop = fields.stop;
+    this.presence_penalty = fields.presence_penalty;
+    this.frequency_penalty = fields.frequency_penalty;
+    this.max_tokens = fields.max_tokens;
```
marikaner, line 45: idea: as those are all really the same names you could consider this:

```suggestion
    ['modelName', 'modelVersion', 'resourceGroup', 'temperature', 'top_p', 'logit_bias', 'user', 'n', 'stop', 'presence_penalty', 'frequency_penalty', 'max_tokens' ].forEach(field => {
    this[field] = fields[field];
  });
```

not sure this is better though.

packages/langchain/src/openai/chat.ts
```diff
@@ -0,0 +1,77 @@
...
+        this.openAiChatClient.run(
+          mapLangchainToAiClient(this, options, messages), options
+        )
+    );
+
+    // we currently do not support streaming
+    await runManager?.handleLLMNewToken(
+      typeof res.data.choices[0].message.content === 'string'
+        ? res.data.choices[0].message.content
+        : ''
```
marikaner, line 71: [q] can this be anything else but string or undefined? otherwise consider this:
```suggestion
      res.data.choices[0].message.content || ''
```

marikaner: Also, isn't `res` our response? Then you could access is much easier with `res.getMessageContent()` or whatever the name of the function was.

packages/langchain/src/openai/embedding.ts
```diff
@@ -0,0 +1,47 @@
...
+  override async embedDocuments(documents: string[]): Promise<number[][]> {
+    const documentEmbeddings = await Promise.all(
+      documents.map(document => this.createEmbedding({ input: document }))
+    );
+    return documentEmbeddings.map(embedding => embedding.data.map(entry => entry.embedding)).flat();
+  }
+
+  override async embedQuery(input: string): Promise<number[]> {
+    const embeddingResponse = await this.createEmbedding({ input });
+    return embeddingResponse.data[0].embedding;
```
marikaner, line 37: [req] Same here - I assume we get the response from our client. Please use the convenience around that if possible.

packages/langchain/src/openai/util.ts
```diff
@@ -0,0 +1,196 @@
...
+  return !!tools?.every(tool =>
+    Array.isArray((tool as StructuredTool).lc_namespace)
+  );
+}
+
+function mapToolChoice(toolChoice?: LangChainToolChoice): ToolChoice | undefined {
+  if (!toolChoice) {
+    return undefined;
+  }
+
```
marikaner, line 152: [pp] This is not needed.
```suggestion
```

packages/langchain/src/openai/types.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Input type for OpenAI embedding models.
+ */
+export type OpenAiEmbeddingModelParams = ModelConfiguration<AzureOpenAiChatModel> &
+  ResourceGroupConfiguration &
+  BaseLLMParams;
+
+/**
+ * OpenAI toolchoice type.
+ */
+export type ToolChoice =
```
marikaner, line 33: [q] does this need to be public? Seems to me that it should be internal. Or maybe even moved to where is is used and never exported at all. Also applies to `LangChainToolChoice`. Consider whether other types should be hidden as well.

packages/langchain/src/openai/util.ts
```diff
@@ -0,0 +1,196 @@
...
+): OpenAiChatMessage {
+  return {
+    content: message.content,
+    name: message.name,
+    role: mapBaseMessageToRole(message),
+    function_call: message.additional_kwargs.function_call,
+    tool_calls: message.additional_kwargs.tool_calls,
+    tool_call_id:
+      message._getType() === 'tool' ? (message as ToolMessage).tool_call_id : ''
+  } as OpenAiChatMessage;
```
marikaner, line 132: [q] Why is this needed? If it is compatible I would not expect this.

packages/langchain/tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "./src",
+    "outDir": "./dist",
+    "tsBuildInfoFile": "./dist/.tsbuildinfo",
+    "composite": true
+  },
+  "include": ["src/**/*.ts"],
+  "exclude": ["dist/**/*", "test/**/*", "**/*.test.ts", "node_modules/**/*"],
+  "references": [{ "path": "../foundation-models" }, { "path": "../ai-api" }]
```
marikaner, line 11: [q] Is this needed?

packages/langchain/src/openai/util.test.ts
```diff
@@ -0,0 +1,103 @@
...
+describe('Mapping Functions', () => {
+  beforeEach(() => {
+    mockClientCredentialsGrantCall();
+  });
+
+  afterEach(() => {
+    nock.cleanAll();
+  });
+
+  it('should parse an OpenAi response to a (Langchain) chat response', async () => {
```
marikaner, line 81:
```suggestion
  it('should parse an OpenAI response to a (LangChain) chat response', async () => {
```

packages/langchain/src/openai/util.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    const result = mapResponseToChatResult(openAiMockResponse);
+    expect(result).toMatchSnapshot();
+  });
+
+  it('should parse a Langchain input to an ai sdk input', async () => {
```
marikaner, line 86:
```suggestion
  it('should parse a LangChain input to an AI SDK input', async () => {
```

packages/langchain/src/openai/util.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    mockInference(
+      {
+        data: prompt
+      },
+      {
+        data: openAiMockResponse,
+        status: 200
+      },
+      chatCompletionEndpoint
+    );
```
marikaner, line 96: [q] Why do we need to mock the inference, when run is spied on? It should never come to this.

packages/langchain/src/openai/util.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    const client = new AzureOpenAiChatClient({ deploymentId: '1234' });
```
marikaner, line 98: [pp] This whole test seems more like an integration test or a test for the client than testing the util. The test above uses the function directly, why don't we use `mapBaseMessageToOpenAiChatMessage()` directly as well?

tests/e2e-tests/src/open-ai-langchain.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+// Pick .env file from root directory
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+dotenv.config({ path: path.resolve(__dirname, '../.env') });
+
+describe('Langchain OpenAI Access', () => {
+  it('should complete a chat', async () => {
+    const result = await generate();
+    expect(result).toBeDefined();
```
marikaner, line 14: [req] unnecessary
```suggestion
```

tests/e2e-tests/src/open-ai-langchain.test.ts
```diff
@@ -0,0 +1,23 @@
...
+    expect(result).toContain('Paris');
+  });
+
+  it('should compute an embedding vector', async () => {
+    const result = await embedQuery();
+    expect(result).toBeDefined();
```
marikaner, line 20: [req] unnecessary
```suggestion
```

marikaner (changes requested):
Looks much better already! I am still missing some tests for chat and embedding. 
I left a comment on the utils test, that in my opinion tests one level higher, maybe this is a bit double if you create separate tests, however I would expect that in the utils test we check that the mapping works with specific parameters and in chat + embedding we test that this is used correctly (simple sanity check would suffice).

packages/langchain/tsconfig.json
tomfrenken: I would argue yes, because this package depends on the referenced packages, this ensures they are always aligned.

packages/langchain/src/openai/util.ts
```diff
@@ -0,0 +1,205 @@
...
+ * @param tool - Base class for tools that accept input of any shape defined by a Zod schema.
+ * @returns The OpenAI chat completion function.
+ */
+function mapToolToOpenAiFunction(
+  tool: StructuredTool
+): AzureOpenAiChatCompletionFunction {
+  return {
+    name: tool.name,
+    description: tool.description,
+    parameters: zodToJsonSchema(tool.schema)
```
ZhongpinWang, line 31: Step 1: Here `tool` has the property `schema`. This function is called in line 40 `mapToolToOpenAiTool`

packages/langchain/src/openai/util.ts
```diff
@@ -0,0 +1,205 @@
...
+  };
+}
+
+/**
+ * Maps a LangChain {@link StructuredTool} to {@link AzureOpenAiChatCompletionTool}.
+ * @param tool - Base class for tools that accept input of any shape defined by a Zod schema.
+ * @returns The OpenAI chat completion tool.
+ */
+function mapToolToOpenAiTool(
```
ZhongpinWang, line 40: Step 2: `tool` here should of course also have the `schema` property. This function is called in `mapLangchainToAiClient`. Check step 3.

packages/langchain/src/openai/util.ts
```diff
@@ -0,0 +1,205 @@
...
+    temperature: client.temperature,
+    top_p: client.top_p,
+    logit_bias: client.logit_bias,
+    n: client.n,
+    stop: options?.stop ?? client.stop,
+    functions: isStructuredToolArray(options?.functions)
+      ? options?.functions.map(mapToolToOpenAiFunction)
+      : options?.functions,
+    tools: isStructuredToolArray(options?.tools)
+      ? options?.tools.map(mapToolToOpenAiTool)
```
ZhongpinWang, line 189: Step 3: So each entry in `options.tools` should have `schema` property. Check step 4.

packages/langchain/README.md
jjtang1985:
```suggestion
Instead, credentials in the binding are required to access SAP AI Core capabilities, specifically for consuming LLMs via the SAP Generative AI Hub.
```

packages/ai-api/src/utils/deployment-resolver.ts
```diff
@@ -4,7 +4,6 @@ import {
 } from '../client/AI_CORE_API/index.js';
 import { deploymentCache } from './deployment-cache.js';
 import { extractModel, type FoundationModel } from './model.js';
-
 /**
```
marikaner, line 7:
```suggestion

/**
```

packages/langchain/README.md
```diff
@@ -0,0 +1,102 @@
...
+  - Download a service key for your AI Core service instance.
+  - Create a `.env` file in the sample-code directory.
+  - Add an entry `AICORE_SERVICE_KEY='<content-of-service-key>'`.
+
+## Usage
+This package provides both chat and embedding clients, currently supporting Azure OpenAI.
+All clients comply with [LangChain's interface](https://js.langchain.com/docs/introduction).
+
+### Client Initialization
+To initialize a client, you only need to provide the model name:
```
marikaner, line 33:
```suggestion
To initialize a client, provide the model name:
```

packages/langchain/src/openai/chat.ts
```diff
@@ -0,0 +1,78 @@
...
+        this.openAiChatClient.run(
+          mapLangchainToAiClient(this, options, messages),
+          options.requestConfig
+        )
+    );
+
+    // we currently do not support streaming
+    await runManager?.handleLLMNewToken(
+      typeof res.getContent() === 'string' ? (res.getContent() as string) : ''
+    );
```
marikaner, line 74: consider this to avoid explicit type assertions

```suggestion
    const content = res.getContent();
    
    // we currently do not support streaming
    await runManager?.handleLLMNewToken(
      typeof content === 'string' ? content : ''
    );
```

eslint.config.js
```diff
@@ -33,6 +33,20 @@ export default [
       'jsdoc/require-jsdoc': 'off'
     }
   },
+  {
+    files: [
+      '**/packages/orchestration/src/client/api/default-api.ts'
```
deekshas8, line 38: [info] This file is deleted post generation, so this can be removed.

packages/langchain/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Table of Contents
+1. [Installation](#installation)
+2. [Pre-requisites](#pre-requisites)
+3. [Usage](#usage)
+    - [Client Initialization](#client-initialization)
+    - [Chat Clients](#chat-clients)
+    - [Embedding Clients](#embedding-clients)
+4. [Support, Feedback, Contribution](#support-feedback-contribution)
+5. [License](#license)
```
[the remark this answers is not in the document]
deekshas8: @KavithaSiva

packages/langchain/src/openai/util.ts
```diff
@@ -0,0 +1,233 @@
...
+      return 'user';
+    case 'system':
+      return 'system';
+    case 'function':
+      return 'function';
+    case 'tool':
+      return 'tool';
+    default:
+      throw new Error(`Unknown message type: ${message._getType()}`);
+  }
```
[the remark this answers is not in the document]
tomfrenken: If I write it like this typescript gets a stroke, I've implemented a similar solution now using a map.

packages/langchain/README.md
```diff
@@ -0,0 +1,102 @@
...
+  modelVersion: '24-07-2021',
+  resourceGroup: 'my-resource-group'
+});
+```
+
+### Chat Clients
+The chat clients allow you to interact with Azure OpenAI chat models, accessible via SAP Gen AI Hub. 
+To invoke the client, you only have a to pass a prompt.
+
+#### Simple Example
```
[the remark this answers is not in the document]
deekshas8: [pp] I would leave out this heading here, put a colon before the last sentence 
`To invoke the client, you only have to pass a prompt: `

packages/langchain/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+```ts
+const chatClient = new AzureOpenAiChatClient({
+  modelName: 'gpt-4o',
+  modelVersion: '24-07-2021',
+  resourceGroup: 'my-resource-group'
+});
+```
+
+### Chat Clients
```
deekshas8, line 50: [q] Should it be singular?

eslint.config.js
```diff
@@ -33,6 +33,20 @@ export default [
...
+    ],
+    rules: {
+      '@typescript-eslint/explicit-module-boundary-types': 'off'
+    }
+  },
```
marikaner, line 43: then let's go ;) 
```suggestion
```

packages/langchain/src/openai/chat.ts
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * LangChain chat client for Azure OpenAI consumption on SAP BTP.
+ */
+export class AzureOpenAiChatClient
+  extends BaseChatModel<AzureOpenAiChatCallOptions>
+  implements AzureOpenAiChatModelParams
+{
+  modelName: AzureOpenAiChatModel;
+  modelVersion?: string;
+  resourceGroup?: string;
```
MatKuhr, line 22: [minor] it's slightly awkward to have this mixed with the hyperparameters. Hypothetical, though somewhat unlikely case: OpenAI adds a hyperparameter called model version.

We could consider moving the hyperparameters out of the constructor to a separate method, e.g. 
```TS
new AzureOpenAiChatClient({ modelName: 'gpt-35-turbo' })
             .withConfig({ temperature: 0.5 });
```

However, probably not worth the refactoring effort for now, but maybe an idea for later

packages/langchain/README.md
```diff
@@ -0,0 +1,102 @@
...
+- [Enable the AI Core service in SAP BTP](https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/initial-setup).
+- Bind the service to your application.
+- Ensure the project is configured with Node.js v20 or higher, along with native ESM support.
+- For testing your application locally:
+  - Download a service key for your AI Core service instance.
+  - Create a `.env` file in the sample-code directory.
+  - Add an entry `AICORE_SERVICE_KEY='<content-of-service-key>'`.
+
+## Usage
+This package provides both chat and embedding clients, currently supporting Azure OpenAI.
```
deekshas8, line 29:
```suggestion
This package provides chat and embedding clients, currently supporting Azure OpenAI.
```

packages/langchain/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+```ts
+const chatClient = new AzureOpenAiChatClient({
+  modelName: 'gpt-4o',
+  modelVersion: '24-07-2021',
+  resourceGroup: 'my-resource-group'
+});
+```
+
+### Chat Clients
```
deekshas8:
```suggestion
### Chat Client
```

packages/langchain/README.md
```diff
@@ -0,0 +1,101 @@
...
+]);
+```
+
+## Support, Feedback, Contribution
+This project is open to feature requests/suggestions, bug reports etc. via [GitHub issues](https://github.com/SAP/ai-sdk-js/issues).
+
+Contribution and feedback are encouraged and always welcome. For more information about how to contribute, the project structure, as well as additional contribution information, see our [Contribution Guidelines](https://github.com/SAP/ai-sdk-js/blob/main/CONTRIBUTING.md).
+
+## License
+The SAP Cloud SDK for AI is released under the [Apache License Version 2.0.](http://www.apache.org/licenses/)
```
deekshas8, line 101:
```suggestion
The SAP Cloud SDK for AI is released under the [Apache License Version 2.0.](http://www.apache.org/licenses/).
```

packages/langchain/README.md
```diff
@@ -0,0 +1,102 @@
...
+const parser = new StringOutputParser();
+const llmChain = promptTemplate.pipe(client).pipe(parser);
+const response = await llmChain.invoke({
+  language: 'german',
+  text: 'What is the capital of France?'
+});
+```
+
+### Embedding Clients
+Embedding clients allow embedding either text or documents (represented as arrays of strings).
```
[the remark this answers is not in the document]
tomfrenken: The only difference is the name of the client, the initialization is exactly the same, therefore I wouldn't add an additional section.

packages/langchain/src/openai/chat.ts
tomfrenken: This would not be aligned with all other langchain clients, where this is done in the constructor though

packages/langchain/src/openai/util.ts
```diff
@@ -0,0 +1,232 @@
...
+    name: message.name,
+    content: message.content,
+    role: mapBaseMessageToRole(message),
+    function_call: message.additional_kwargs.function_call,
+    tool_calls: message.additional_kwargs.tool_calls,
+    tool_call_id: mapToolCallId(message)
+  });
+}
+
+function mapBaseMessageToContent(baseMessage: BaseMessage): AzureOpenAiChatMessage['content'] {
```
MatKuhr, line 154: function is unused ATM..

tests/type-tests/package.json
```diff
@@ -16,6 +16,8 @@
     "@sap-ai-sdk/foundation-models": "workspace:^",
     "@sap-ai-sdk/orchestration": "workspace:^",
     "@sap-ai-sdk/core": "workspace:^",
+    "@sap-ai-sdk/langchain": "workspace:^",
+    "@langchain/core": "^0.2.30",
```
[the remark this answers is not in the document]
ZhongpinWang: [q] should we fix the version or rather use `^`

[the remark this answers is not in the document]
marikaner: as this is below 1, you cannot make any semver assumptions: https://semver.org/#spec-item-4
so doesn't really matter ;)

packages/langchain/README.md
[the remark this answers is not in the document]
MatKuhr: while that is true, users don't know that and in the below code example can only guess what `embeddingClient` refers to. We also can't assume that all consumers of embedding functionality are also consumers of chat functionality. Thus we need to have something IMO, either a small piece of code or text

jjtang1985:
🥳

deekshas8 (approved):
LGTM.

PR #125, opened 2024-09-10
feat: Use literal string for orchestration models

ZhongpinWang:
## Context

AI/gen-ai-hub-sdk-js-backlog#102.

Model name autocompletion is supported now:

image

Custom strings are also allowed in case some model has not been included in the list yet. 

Model types moved to `@sap-ai-sdk/core` for reuse by both `foundation-models` and `orchestration` packages.

## Definition of Done

- [ ] Documentation updated


packages/core/src/model-types.ts
```diff
@@ -0,0 +1,52 @@
+type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
+
+/**
+ * Azure OpenAI models for chat completion.
+ */
+export type AzureOpenAiChatModel = LiteralUnion<
```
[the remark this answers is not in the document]
marikaner: I think this is a good idea. The names will be even longer through that, but theoretically we could also provide a second client for OpenAI only, right? @deekshas8 also interested in your opinion.

packages/core/src/model-types.ts
```diff
@@ -0,0 +1,52 @@
...
+ * All available models for chat completion.
+ */
+export type ChatModel = LiteralUnion<
+  AzureOpenAiChatModel | GcpVertexAiChatModel | AwsBedrockChatModel
+>;
+
+/**
+ * All available models for embeddings.
+ */
+export type EmbeddingModel = LiteralUnion<AzureOpenAiEmbeddingModel>;
```
marikaner, line 52: [q] Is this used anywhere?

tests/type-tests/test/model.test-d.ts
```diff
@@ -0,0 +1,10 @@
+import {
+  ChatModel,
+  AzureOpenAiChatModel,
+  EmbeddingModel
+} from '@sap-ai-sdk/core';
+
+expect<ChatModel>('custom-model');
```
marikaner, line 7: I assume you also checked autocompletion works manually? 😊

tests/type-tests/test/openai.test-d.ts
```diff
@@ -18,9 +18,9 @@ expectType<Promise<OpenAiChatCompletionResponse>>(
 );
 
 /**
- * Chat Completion with invalid model.
+ * Chat Completion with custom model.
  */
-expectError(
+expectType(
```
marikaner, line 23: [q] Is this correct for OpenAI? If we have all models, it should not be possible to pass anything else, right? The string variant might only be needed for orchestration?

packages/core/src/model-types.ts
ZhongpinWang: Oh true, we can remove it. Probably we will never need one. I will remove it.

tests/type-tests/test/model.test-d.ts
ZhongpinWang: Sure see the screenshot above. Works like a charm.

marikaner: Oops, totally ignored the description. 🙊

marikaner (approved):
I left a few comments, but I have no concerns.

tests/type-tests/test/openai.test-d.ts
ZhongpinWang: During the implementation, I found out that for Azure openai, the model `gpt-4o-mini` was missing. That is why I thought, it could be the case that we could also miss some models when ai core added something there. Because no auto sync, I can imagine that sometime in the future, user will request us to support some new model from openai, and it could only be fixed via a new release. That's also why I used `LiteralUnion` for openai as well to support custom string.

marikaner: I think that is fair. If the model really is incorrect, you will receive a runtime error anyways. 👍

packages/core/src/model-types.ts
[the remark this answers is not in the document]
KavithaSiva: I also like the idea of adding `Azure`, and if we do it, would also want to be adapted in the Java SDK as well.
But, do you know if there are any differences between the `AzureOpenAI` and  plain `OpenAI` clients?

I wasn't aware that Anthropic Claude is served by AWS. Where did you read about this?

packages/core/src/model-types.ts
KavithaSiva: Why do you think more Embedding models won't be added?

marikaner: This is more about the fact that there are all models for all vendors and we don't need this aggregation for embeddings. Only for chat (for orchestration).

packages/core/src/model-types.ts
```diff
@@ -0,0 +1,52 @@
+type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
```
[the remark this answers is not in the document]
ZhongpinWang: It was a pretty interesting finding for me as well.😄 Check this issue for more technical details -> https://github.com/microsoft/TypeScript/issues/29729

A simple explanation would be:

If union literal strings with `string`, the type system will collapse them all together into `string`. But the beauty of this trick is that it makes the type system think `string & Record<never, never>` is not `string`, and thus, not collapsible with literal string, so that literal strings get to survive.

packages/core/src/model-types.ts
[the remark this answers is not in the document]
deekshas8: @KavithaSiva https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/models-and-scenarios-in-generative-ai-hub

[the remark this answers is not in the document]
ZhongpinWang: I not sure if having a client for pure OpenAI would be possible because that would require AI Core to support it first. But Anyway, `AzureOpenAI` is how the model is called in AI Core deployment as well as in the [SAP Help documentation](https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/create-deployment-for-generative-ai-model-in-sap-ai-core#using-the-api).

Regarding Anthropic Claude, the executable id of all `anthropic--claude-*` models is `aws-bedrock`. So it is provided via AWS. You can also find model details in the yaml file I mentioned in the backlog item.

PR #128, opened 2024-09-11
feat: Add AI Core e2e tests

shibeshduw:
## Context
Added e2e test flow for `DeploymentApi` that covers starting a deployment(POST), stopping a deployment(PATCH) and deleting a deployment(DELETE), along with a precondition and postcondition check(GET).

Test run: https://github.com/SAP/ai-sdk-js/actions/runs/10811128561

Closes: AI/gen-ai-hub-sdk-js-backlog#59.

## Definition of Done

- [ ] Documentation updated


tests/e2e-tests/src/ai-core.test.ts
```diff
@@ -1,20 +1,141 @@
...
+        createdDeploymentId!,
+        { 'AI-Resource-Group': 'i745181' }
+      ).execute();
+
+      expect(deleteResponse).toBeDefined();
+      expect(deleteResponse.message).toBe('Deletion scheduled');
+    });
+
+    afterAll(async () => {
+      expect(createdDeploymentId).toBeDefined();
```
MatKuhr, line 113: [q] why is this assertion here? It is already asserted in the "modify" test..

tests/e2e-tests/src/ai-core.test.ts
```diff
@@ -1,20 +1,141 @@
...
+        }
+        await new Promise(resolve => setTimeout(resolve, RETRY_INTERVAL));
+      }
+      throw new Error(
+        `Deployment ${deploymentId} did not reach ${targetStatus} status within the expected time. ` +
+          'Please manually stop and delete the deployment.'
+      );
+    }
+
+    const sanitizedState = (state: AiDeploymentList | undefined) => ({
```
MatKuhr, line 45: [q] Why do we need this?

shibeshduw: For the pre and post condition check, I am currently checking if deploymentQuery (GET all deployments) is equal before and after the flow is executed, but it turns out that the 'modifiedAt' property is updated for all deployments on every get request. So I use this helper to remove this property while comparing the state.

tests/e2e-tests/src/ai-core.test.ts
shibeshduw: It is not technically required but If creating a deployment fails, then `createdDeploymentId` would be undefined. This assertion is present in every test after creating deployment to stop execution for these tests.

tests/e2e-tests/src/ai-core.test.ts
```diff
@@ -1,20 +1,141 @@
...
+      ...state,
+      resources: state?.resources.map(resource =>
+        _.omit(resource, ['modifiedAt'])
```
deekshas8, line 48: [q] Why not just use `delete` keyword here?

shibeshduw: Updated it 👍

tests/e2e-tests/package.json
```diff
@@ -25,6 +25,7 @@
     "lint:fix": "eslint . --fix && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -w --log-level error"
   },
   "devDependencies": {
+    "@types/lodash": "^4.17.7",
```
deekshas8, line 28: [q] Is this still needed?

tests/e2e-tests/src/ai-core.test.ts
deekshas8: With delete you won't get a warning. But this is also okay I guess

tests/e2e-tests/src/ai-api.test.ts
```diff
@@ -1,20 +1,141 @@
...
 dotenv.config({ path: path.resolve(__dirname, '../.env') });
 
-describe('ai-api', () => {
-  it('should get deployments', async () => {
-    const deployments = await DeploymentApi.deploymentQuery(
-      {},
-      { 'AI-Resource-Group': 'default' }
-    ).execute();
-    expect(deployments).toBeDefined();
+describe('AI Core APIs', () => {
```
marikaner, line 16:
```suggestion
describe('AI APIs', () => {
```

tests/e2e-tests/src/ai-api.test.ts
```diff
@@ -1,20 +1,141 @@
...
+  describe('DeploymentAPI', () => {
```
marikaner, line 17:
```suggestion
  describe('DeploymentApi', () => {
```

marikaner: (also check the others)

tests/e2e-tests/src/ai-api.test.ts
```diff
@@ -1,20 +1,141 @@
...
+
+    async function waitForDeploymentToReachStatus(
+      deploymentId: string,
+      targetStatus: 'RUNNING' | 'STOPPED'
+    ): Promise<AiDeployment> {
+      for (let i = 0; i < MAX_RETRIES; i++) {
+        const deploymentDetail = await DeploymentApi.deploymentGet(
+          deploymentId,
+          {},
+          { 'AI-Resource-Group': 'i745181' }
```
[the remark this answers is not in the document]
marikaner: yes please 😊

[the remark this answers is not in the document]
MatKuhr: +1 for ai-sdk-js-e2e. its better to not use default here to not interfere with other e2e tests

tests/e2e-tests/src/ai-api.test.ts
```diff
@@ -1,20 +1,141 @@
...
+      ).execute();
+
+      expect(deleteResponse).toBeDefined();
+      expect(deleteResponse.message).toBe('Deletion scheduled');
+    });
+
+    afterAll(async () => {
+      expect(createdDeploymentId).toBeDefined();
+
+      await new Promise(r => setTimeout(r, 20000));
```
marikaner, line 115: [q] Why?

marikaner (changes requested):

tests/e2e-tests/src/ai-api.test.ts
shibeshduw: Takes around 5-10 seconds, sometimes more, for the deployment to be deleted.

tests/e2e-tests/src/ai-api.test.ts
```diff
@@ -1,20 +1,141 @@
...
+          { 'AI-Resource-Group': 'i745181' }
+        ).execute();
+        if (deploymentDetail.status === targetStatus) {
+          return deploymentDetail;
+        }
+        await new Promise(resolve => setTimeout(resolve, RETRY_INTERVAL));
+      }
+      throw new Error(
+        `Deployment ${deploymentId} did not reach ${targetStatus} status within the expected time. ` +
+          'Please manually stop and delete the deployment.'
```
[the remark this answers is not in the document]
shibeshduw: When we create/modify the deployment, we are waiting for the status to change. If the status doesn't change in the time that we set, then the rest of the flow doesn't work. So we would need to clear all the deployments in this case. 

I guess your comment below already covers this, if we add a cleanup step then this is no longer required.

tests/e2e-tests/src/ai-api.test.ts
```diff
@@ -1,20 +1,141 @@
...
+        createResponse.id,
+        'RUNNING'
+      );
+      expect(runningDeployment.status).toBe('RUNNING');
+      expect(runningDeployment.deploymentUrl).toBeTruthy();
+      createdDeploymentId = runningDeployment.id;
+    }, 180000);
+
+    it('should modify the deployment to stop it', async () => {
+      expect(createdDeploymentId).toBeDefined();
```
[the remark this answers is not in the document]
shibeshduw: Maybe we could use jest's pattern matching to run individual tests but also keep this flow? Plus we also add a cleanup step as you mentioned.

tests/e2e-tests/src/deployment-api.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+const sanitizedState = (state: AiDeploymentList | undefined) => ({
+  ...state,
+  resources: state?.resources.map(
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    ({ modifiedAt, ...rest }) => rest
+  )
+});
+
+function getDeploymentId(id: string | undefined): string {
```
marikaner, line 117: I assume the idea here is that later on this can be used to fetch a deployment ID in case it does not exist, right?

PR #130, opened 2024-09-12
feat: Apply schema check for Azure OpenAI client response

ZhongpinWang:
## Context

AI/gen-ai-hub-sdk-js-backlog#54.

Use Zod schema to type check OpenAI chat and embedding reponses at runtime.

## Definition of Done

- [ ] Documentation updated


ZhongpinWang:
@marikaner @deekshas8 I would appreciate your opinions for this PR, especially if you like the approach with Zod. I can see some potential refactorings and automation, but feel free to mention them as well.

packages/foundation-models/package.json
```diff
@@ -20,24 +20,27 @@
...
   "scripts": {
-    "compile": "tsc",
+    "compile": "pnpm run generate-zod && tsc",
     "compile:cjs": "tsc -p tsconfig.cjs.json",
     "test": "NODE_OPTIONS=--experimental-vm-modules jest",
     "lint": "eslint \"**/*.ts\" && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -c",
     "lint:fix": "eslint \"**/*.ts\" --fix && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -w --log-level error",
-    "check:public-api": "node --loader ts-node/esm ../../scripts/check-public-api-cli.ts"
+    "check:public-api": "node --loader ts-node/esm ../../scripts/check-public-api-cli.ts",
+    "generate-zod": "ts-to-zod src/openai/openai-types.ts src/openai/openai-types-schema.ts && sed -i 's|export const|\\/**\\n * @internal\\n *\\/\\nexport const|' src/openai/openai-types-schema.ts"
```
marikaner, line 29: [pp] This is a bit too much bash magic for my taste, is there a more explicit way without me reading the sed man?

packages/foundation-models/src/openai/openai-embedding-response.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * OpenAI embedding response.
+ */
+export class OpenAiEmbeddingResponse {
+  /**
+   * The embedding response.
+   */
+  public readonly data: OpenAiEmbeddingOutput;
+  constructor(public readonly rawResponse: HttpResponse) {
```
marikaner, line 19: [q] Oh, so now we actually do not check this in our tests, but on every user request?
[pp] I expected this to just be a test, running this every time seems too much to me. At least it should be optional...

marikaner (commented):
Generally I am not opposed to the "response" classes itself, might make access to data more convenient. However, if I got it right we are checking the structure of all responses for each request (right?) I expected this to happen in the tests only, not in production. While I think it is not super expensive, I don't see a major benefit either.

packages/foundation-models/package.json
marikaner: (feel free to ask others, I am a real command line noob)

eslint.config.js
```diff
@@ -26,7 +26,8 @@ export default [
   {
     files: [
       '**/test-util/**/*.ts',
-      '**/packages/orchestration/src/client/**/*'
+      '**/packages/orchestration/src/client/**/*',
+      'packages/foundation-models/src/openai/openai-types-schema.ts'
```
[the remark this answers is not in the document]
ZhongpinWang: This rule turns `jsdoc/require-jsdoc` off. Only this file is generated without the need to add any jsdoc. Other files I assume still need this check?

packages/foundation-models/package.json
ZhongpinWang: 😆 This is due to the `check:public-api` script. The generated schema does not contain `@internal` but still needs to be exported. The library `ts-to-zod` also does not support adding this kind of things. I looked into the check public api script, but it seems also quite cumbersome to support excluding compiled schema file.

packages/foundation-models/src/openai/openai-embedding-response.ts
ZhongpinWang: I can also remove this check here and put some in the tests. Not sure how much performance it will improve, but now I see a potential benefit if I agree with you. So in case our schema is not 100% in sync with openai, without this mandatory check, user can still treat it as any and access properties freely, especially for pure js developer, who probably do not like this schema check. Otherwise, if schema not totally correct, they cannot use the client any more due to the zod error.

So tl;dr, I will move them into test, not a big effort anyways. Thanks for your feedback!

ZhongpinWang: Done.

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-response.test.ts
```diff
@@ -0,0 +1,23 @@
...
+  const rawResponse = {
+    data: mockResponse,
+    status: 200,
+    headers: {},
+    request: {}
+  };
+  const response = new AzureOpenAiChatCompletionResponse(rawResponse);
+
+  it('should return the completion response', () => {
+    const data = azureOpenAiChatCompletionOutputSchema.parse(response.data);
```
ZhongpinWang, line 20: @jjtang1985 This is what I mean by checking schema in the test instead of in the runtime response.

packages/foundation-models/package.json
```diff
@@ -20,24 +20,27 @@
     "internal.d.ts"
   ],
   "scripts": {
-    "compile": "tsc",
+    "compile": "pnpm run generate-zod && tsc",
```
marikaner, line 23: [q] is it really needed for transpilation?

marikaner (approved):
LGTM

packages/foundation-models/package.json
marikaner: yes it does. we will improve this in the future

packages/foundation-models/package.json
```diff
@@ -20,24 +20,27 @@
...
-    "@sap-ai-sdk/core": "workspace:^",
     "@sap-ai-sdk/ai-api": "workspace:^",
-    "@sap-cloud-sdk/http-client": "^3.20.0",
+    "@sap-ai-sdk/core": "workspace:^",
     "@sap-cloud-sdk/connectivity": "^3.20.0",
+    "@sap-cloud-sdk/http-client": "^3.20.0",
+    "@sap-cloud-sdk/openapi": "^3.20.0",
     "@sap-cloud-sdk/util": "^3.20.0",
-    "@sap-cloud-sdk/openapi": "^3.20.0"
+    "zod": "^3.23.8"
```
marikaner, line 38: [req] this should be a dev dependency!

marikaner (changes requested):

marikaner (approved):

packages/foundation-models/package.json
ZhongpinWang: I think it just ensures a bit more that if someone changes `azure-openai-types.ts`, the schema is generated again if they compile it. Not super necessary but won't hurt at least.

PR #133, opened 2024-09-16
fix: Supported model names

ZhongpinWang:
## Context

AI/gen-ai-hub-sdk-js-backlog#113.

Model name autocompletion

- for `openai` in `foundation-models`
- for `orchestration`

Model types are defined centrally in `core` and then separately imported into these two packages.

## Definition of Done

- [ ] ~Error handling created / updated & covered by the tests above~
- [ ] ~Documentation updated~
- [ ] ~(Optional) Aligned changes with the Java SDK~
- [ ] ~(Optional) Release notes updated~


MatKuhr (commented):
Maybe we can define the constants only once and use the union of LLM Access + orchestration. And then use `Omit<.>` within the specific clients?

While there will be exceptions, probably most models will be the same across both, so we would only have to omit a few models, instead of duplicating a lot..

packages/foundation-models/src/model-types.ts
```diff
@@ -0,0 +1,20 @@
+type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
```
deekshas8, line 1: [req] I agree with Matthias to not have duplicate code (even the `LiteralUnion` is defined in both packages). I would imagine something like this works:

```
type AzureModels = 'gpt-4o-mini'
  | 'gpt-4o'
  | 'gpt-4'
  | 'gpt-4-32k'
  | 'gpt-35-turbo'
  | 'gpt-35-turbo-16k'
/**
 * Azure OpenAI models for chat completion.
 */
export type AzureOpenAiChatModel = LiteralUnion<AzureModels>;


export type AzureOpenAiChatModelOrchesration = LiteralUnion<Exclude<AzureModels, 'gpt-4o-mini'>>;

```

packages/orchestration/src/model-types.ts
```diff
@@ -4,44 +4,39 @@ type LiteralUnion<T extends U, U = string> = T | (U & Record<never, never>);
...
+ * Aicore Opensource models for chat completion.
+ */
+export type AicoreOpensourceChatModel = LiteralUnion<
+  'mistralai--mixtral-8x7b-instruct-v01' | 'meta--llama3-70b-instruct'
+>;
+
+/**
+ * Supported chat models for orchestration.
  */
 export type ChatModel = LiteralUnion<
```
deekshas8, line 37: [q] Why do we need to wrap this again in LiteralUnion?

ZhongpinWang: Oh I forgot to remove this. I thought about leaving LiteralUnion out of the reused model-types in `core` at first. Removed.

packages/foundation-models/src/model-types.ts
ZhongpinWang: Done.

ZhongpinWang:
@MatKuhr @deekshas8 Thanks for your feedback! Still a bit hesitating if excluding models would be a good idea as per default, as per default, adding a new model in a reused model type file and then exclude is a bit error-prone as we might simply forget to exclude it. But anyways since it is now quite simple to do so, I weight reusability a bit higher than maintainability, and refactored accordingly.

packages/core/src/index.ts
```diff
@@ -7,10 +7,10 @@ export type {
...
 export { getAiCoreDestination } from './context.js';
 export { OpenApiRequestBuilder } from './openapi-request-builder.js';
-export type {
+export {
   AzureOpenAiChatModel,
   AzureOpenAiEmbeddingModel,
   GcpVertexAiChatModel,
   AwsBedrockChatModel,
-  ChatModel
+  AicoreOpensourceChatModel
```
deekshas8, line 15:
```suggestion
  AiCoreOpensourceChatModel
```

[req] Please adjust in other places as well.

packages/core/src/model-types.ts
```diff
@@ -24,24 +23,22 @@ export type AzureOpenAiEmbeddingModel = LiteralUnion<
...
-  | 'amazon--titan-text-lite'
   | 'anthropic--claude-3-haiku'
   | 'anthropic--claude-3-opus'
   | 'anthropic--claude-3-sonnet'
   | 'anthropic--claude-3.5-sonnet'
 >;
 
 /**
- * All available models for chat completion.
+ * Aicore Opensource models for chat completion.
```
deekshas8, line 40:
```suggestion
 * AI Core open source models for chat completion.
```

[req] please check in other places as well.

packages/core/src/model-types.ts
```diff
@@ -24,24 +23,22 @@ export type AzureOpenAiEmbeddingModel = LiteralUnion<
...
  */
-export type ChatModel = LiteralUnion<
-  AzureOpenAiChatModel | GcpVertexAiChatModel | AwsBedrockChatModel
+export type AicoreOpensourceChatModel = LiteralUnion<
```
MatKuhr, line 42:
```suggestion
export type AiCoreOpenSourceChatModel = LiteralUnion<
```

would be consistent with the rest of the codebase I think

packages/foundation-models/src/openai/model-types.ts
```diff
@@ -0,0 +1,14 @@
+import type {
+  AzureOpenAiChatModel as CoreAzureOpenAiChatModel,
+  AzureOpenAiEmbeddingModel as CoreAzureOpenAiEmbeddingModel
+} from '@sap-ai-sdk/core';
+
+/**
+ * Azure OpenAI models for chat completion.
+ */
+export type AzureOpenAiChatModel = CoreAzureOpenAiChatModel;
```
MatKuhr, line 9: [q] why the importing and re-exporting?

packages/orchestration/src/model-types.ts
```diff
@@ -0,0 +1,17 @@
...
+  AicoreOpensourceChatModel,
+  AwsBedrockChatModel,
+  AzureOpenAiChatModel,
+  GcpVertexAiChatModel
+} from '@sap-ai-sdk/core';
+
+/**
+ * Supported chat models for orchestration.
+ */
+export type ChatModel = Exclude<
```
deekshas8, line 11: [pp] I find this is a bit more readable
```
export type ChatModel = 
  | Exclude<AzureOpenAiChatModel, 'gpt-4o-mini'>
  | GcpVertexAiChatModel
  | AwsBedrockChatModel
  | AicoreOpensourceChatModel;
```

packages/core/src/model-types.ts
```diff
@@ -24,24 +23,22 @@ export type AzureOpenAiEmbeddingModel = LiteralUnion<
...
   | 'anthropic--claude-3-opus'
   | 'anthropic--claude-3-sonnet'
   | 'anthropic--claude-3.5-sonnet'
 >;
 
 /**
- * All available models for chat completion.
+ * Aicore Opensource models for chat completion.
  */
-export type ChatModel = LiteralUnion<
```
MatKuhr, line 45: [q] Why remove this? Does exclude not work on the union?

packages/core/src/index.ts
ZhongpinWang: I first had the same thought as yours as well until I saw they call it `aicore-opensource` and uses something like `implementation: AicoreOpensourceProvider` everywhere:

- https://github.tools.sap/MLF/mlf-gitops/blob/ee03d0311c4457ead7831a6ca99e4f13a8a4c118/docs/gen_ai_hub/README.md
- https://github.tools.sap/MLF/mlf-gitops/blob/ee03d0311c4457ead7831a6ca99e4f13a8a4c118/docs/gen_ai_hub/model-entry-catalogue.md#meta--llama3-70b-instruct

But true, we already used the term `OpenAi`...

packages/foundation-models/src/openai/model-types.ts
ZhongpinWang: True, I can also not think of a case, where a model defined in `core` could be unavailable in `foundation-models`. Adjusted.

packages/core/src/model-types.ts
ZhongpinWang: Because `ChatModel` is only used in orchestration and I moved it there (plus some excludes of course).

packages/orchestration/src/model-types.ts
ZhongpinWang: Thanks for this good suggestion! Done

packages/core/src/index.ts
ZhongpinWang: Done.

packages/foundation-models/src/openai/model-types.ts
deekshas8: I remember in the Cloud SDK we decided types in odata-common that were used odata-v4/v2 should be re-exported from v2/v3 since we dont want users to import odata-common package. 

Shouldn't we do the same here? @marikaner @tomfrenken opinions?

ZhongpinWang: Oh now I am again upvote for re-exporting as `core` is not published anyways. They will not be able to import from `core`.

marikaner: @ZhongpinWang core is published!

I don't have a very strong opinion on this, but @deekshas8 just pointed out that there are other types we should then include (e.g. `ModelDeployment` and co.).
I am leaning towards not re-exporting this (we can always do it later if needed), because these types here as well as `ModelDeployment` and similar are all types that users will use in their API, but most likely won't need explicitly.
In any case: whatever we do it should be consistent.

ZhongpinWang: @marikaner Okay. I read the [README.md](https://github.com/SAP/ai-sdk-js/tree/main/packages/core#sap-ai-sdkcore) from core and thought since it is primarily for internal use, we would rather want to hide this package. `core` is not even in the main README.

Regarding consistency, these are what we are currently exporting in `core` index:

```ts
export type {
  BaseLlmParametersWithDeploymentId,
  BaseLlmParameters,
  CustomRequestConfig,
  EndpointOptions
} from './http-client.js';
export { executeRequest } from './http-client.js';
export { getAiCoreDestination } from './context.js';
export { OpenApiRequestBuilder } from './openapi-request-builder.js';
export type {
  AzureOpenAiChatModel,
  AzureOpenAiEmbeddingModel,
  GcpVertexAiChatModel,
  AwsBedrockChatModel,
  ChatModel
} from './model-types.js';
```

Anything we should directly expose in other packages other than `core`?

marikaner: There is nothing that we would want users to use directly. Except maybe the Models, that is right. `ModelDeployment` and co is on `ai-api` not `core`. Honestly, I don't have a strong opinion. Semantically it belongs into `foundation-models` anyways, we only need it in core to be able to reuse it in orchestration. `@types/sap-ai-sdk_foundation-models` would be better...

marikaner: We could consider this, but it might be total overkill: https://github.com/DefinitelyTyped/DefinitelyTyped

marikaner: Or a separate package `@sap-ai-sdk/foundation-model-types`. Might also be overkill

ZhongpinWang: > There is nothing that we would want users to use directly. Except maybe the Models, that is right. `ModelDeployment` and co is on `ai-api` not `core`. Honestly, I don't have a strong opinion. Semantically it belongs into `foundation-models` anyways, we only need it in core to be able to reuse it in orchestration. `@types/sap-ai-sdk_foundation-models` would be better...

There is another reason why I put it into core, because `foundation-models` currently do not support all models and so that it won't confuse the user, I put it in core and then only pick what I need from `foundation-models`. Otherwise, `orchestration` can also import models from `foundation-models`. I now treat `core` as some kind of common library. 

I don't think re-exporting is a bad idea, just in case in the future, we move again models from `core` back to `foundation-models`, we just have to put models into this `model-types.ts` in `foundation-models`. Nothing else needs to be changed except we remove it from `core`, which should probably not be used directly from a user.

deekshas8 (approved):
LGTM

PR #136, opened 2024-09-16
chore: Make API more consistent

marikaner:
## Context

AI/gen-ai-hub-sdk-js-backlog#104.

I renamed a few types and properties and removed a few types that I noticed were unused.
I added one TODO because we export some types directly, however those types do not comply with our style guide and therefore have inconsistent naming.

### DIFF
--- a/packages/ai-api/src/utils/deployment-resolver.ts
+++ b/packages/ai-api/src/utils/deployment-resolver.ts
@@ -9,7 +9,7 @@ import { extractModel, type FoundationModel } from './model.js';
  * The model deployment configuration when using a model.
  * @typeParam ModelNameT - String literal type representing the name of the model.
  */
-export interface ModelConfiguration<ModelNameT = string> {
+export interface ModelConfig<ModelNameT = string> {
   /**
    * The name of the model.
    */
@@ -23,7 +23,7 @@ export interface ModelConfiguration<ModelNameT = string> {
 /**
  * The deployment configuration when using a deployment ID.
  */
-export interface DeploymentIdConfiguration {
+export interface DeploymentIdConfig {
   /**
    * The deployment ID.
    */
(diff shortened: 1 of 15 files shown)


packages/orchestration/src/index.ts
```diff
@@ -8,6 +8,7 @@ export type {
   ModuleConfigs,
   MaskingModuleConfig,
   MaskingProviderConfig,
+  // TODO: decide whether these types need to be renamed
```
marikaner, line 11: The types below (and others) are not compatible with our style guide. We should make a decision on how to handle this. I will asses where those types are used (return types, parameters) and what the complexity of doing that is. Theoretically they would need to be adjusted all over. Generally, I would prefer it to keep the API consistent.

ZhongpinWang (commented):
LGTM except one missing

packages/ai-api/src/utils/deployment-resolver.ts
```diff
@@ -56,7 +55,7 @@ export type ModelDeployment<ModelNameT = string> =
  */
 function isDeploymentIdConfiguration(
```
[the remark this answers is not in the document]
ZhongpinWang: And line 158. Not exported but could be changed?

ZhongpinWang (approved):
LGTM

packages/orchestration/src/orchestration-types.ts
```diff
@@ -33,17 +33,16 @@ export type LlmConfig = LLMModuleConfig & {
  * Orchestration module configuration.
  */
 export interface OrchestrationModuleConfig {
-  // TODO: remove "config" for all the configs below
```
deekshas8, line 36: Leftover comment

deekshas8 (approved):
LGTM. Minor comment

packages/orchestration/src/orchestration-types.ts
marikaner: you are looking at the left side, but it is gone on the right ;)

PR #138, opened 2024-09-17
chore: Improve transformation scripts

marikaner:
I would like to reuse the transformation on generated files, so I made it a little more generic.


shibeshduw (approved):
LGTM

PR #143, opened 2024-09-18
chore: Improve samples

marikaner:
I found that the samples for orchestration do not show enough functionality and roughly implemented all features.
Error case could be improved, but maybe we do this after GA.

TODO: 
- [x] Test this manually against the smoke tests


ZhongpinWang (approved):
LGTM

PR #144, opened 2024-09-18
chore: Rename `azureContentFilter()` to `buildAzureContentFilter()`

marikaner:
🚨 Style guide police in action ;)

### DIFF
--- a/packages/orchestration/src/orchestration-filter-utility.ts
+++ b/packages/orchestration/src/orchestration-filter-utility.ts
@@ -8,7 +8,7 @@ import {
  * @param filter - Filtering configuration for Azure filter. If skipped, the default Azure content filter configuration is used.
  * @returns An object with the Azure filtering configuration.
  */
-export function createAzureContentFilter(
+export function buildAzureContentFilter(
   filter?: AzureContentSafety
 ): FilteringConfig {
   if (filter && !Object.keys(filter).length) {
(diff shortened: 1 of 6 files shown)


packages/orchestration/src/orchestration-filter-utility.ts
```diff
@@ -4,11 +4,11 @@ import {
...
 
 /**
- * Convenience function to create Azure filters.
- * @param filter - Filtering configuration for Azure filter. If skipped, the default Azure filter configuration is used.
+ * Convenience function to create Azure content filters.
+ * @param filter - Filtering configuration for Azure filter. If skipped, the default Azure content filter configuration is used.
  * @returns An object with the Azure filtering configuration.
  */
-export function azureContentFilter(
+export function createAzureContentFilter(
```
KavithaSiva, line 11: In the original PR, I named it originally as `createAzureFilter`, when @MatKuhr had a [comment ](https://github.com/SAP/ai-sdk-js/pull/58#discussion_r1711540033) that this might lead to believe that something is created in the `server side`, which is why I changed it to just `azureContentFilter`.

marikaner: How about `buildAzureContentFilter()` then?

I don't have super-strong opinions here either, but we decided to put it like this into the style guide for a reason. It is a little longer, but more speaking and more consistent.

jjtang1985 (approved):

ZhongpinWang (approved):

jjtang1985:
@marikaner , please consider merging this, so the orchestration doc can align with that.
@KavithaSiva

PR #152, opened 2024-09-19
chore: Remove the zod @internal workaround

deekshas8:
## Context

As in the title.


ZhongpinWang (approved):
True, we don't need to compile it to productive code anyways. Thanks for cleaning this up.

PR #155, opened 2024-09-20
fix: Fix missing and unused dependencies

marikaner:
I noticed that we had some unused dependencies, so I added a dependency check. Unfortunately this also showed that there were some missing (productive!!) dependencies.


tomfrenken:
I guess with pnpm's hoisting we never noticed 😬

ZhongpinWang (approved):
LGTM, only one [q], is there a good way to check smoke test deps as well?

PR #158, opened 2024-09-20
fix: Add zod schema tests for mock response

ZhongpinWang:
## Context

AI/gen-ai-hub-sdk-js-backlog#117.

Zod schema check for responses are added in the unit tests for `openai` and `orchestration`. 

Notice: zod schema check is currently missing for embedding response. Once we generate types from spec for embedding, we can add this check to avoid duplicated effort.

## Definition of Done

- ~[ ] Documentation updated~


.gitignore
```diff
@@ -57,3 +57,7 @@ build/Release
 /pnpm-publish-summary.json
 /tests/smoke-tests/deployment
 /packages/foundation-models/src/azure-openai/azure-openai-types-schema.ts-e
```
marikaner, line 59: [nth] Potentially this also belongs into the zod bucket?

packages/foundation-models/package.json
```diff
@@ -22,13 +22,14 @@
...
-    "test": "NODE_OPTIONS=--experimental-vm-modules jest",
+    "test": "pnpm generate-zod && NODE_OPTIONS=--experimental-vm-modules jest",
     "lint": "eslint \"**/*.ts\" && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -c",
     "lint:fix": "eslint \"**/*.ts\" --fix && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -w --log-level error",
     "check:public-api": "node --loader ts-node/esm ../../scripts/check-public-api-cli.ts",
     "generate:azure-openai": "openapi-generator --generateESM --clearOutputDir -i ./src/azure-openai/spec/inference.yaml -o ./src/azure-openai/client --schemaPrefix AzureOpenAi",
     "postgenerate:azure-openai": "rm ./src/azure-openai/client/inference/*.ts && pnpm lint:fix",
-    "generate-zod": "ts-to-zod src/azure-openai/azure-openai-embedding-types.ts src/azure-openai/ts-to-zod/azure-openai-embedding-types.zod.ts"
+    "generate-zod-config": "node --loader ts-node/esm ../../scripts/generate-zod-config.ts './src/azure-openai/client/inference/schema/*.ts' ./src/azure-openai/ts-to-zod",
+    "generate-zod": "pnpm generate-zod-config && ts-to-zod --all --skipValidation"
```
marikaner, line 32: [pp] I want to emphasize that this is totally my personal opinion only: I feel that the zod check is not suuuper important, but quite prominent here. I also think that the zod test is not a normal unit test, so I would prefer to move this to a separate package under tests so that when just running unit tests we don't need to generate this every time. Also I think the separation would be cleaner with this.

other opinions @SAP/ai-sdk-js-team?

packages/orchestration/package.json
```diff
@@ -22,18 +22,24 @@
...
   },
   "dependencies": {
     "@sap-ai-sdk/core": "workspace:^",
     "@sap-ai-sdk/ai-api": "workspace:^",
     "@sap-cloud-sdk/http-client": "^3.21.0",
     "@sap-cloud-sdk/util": "^3.21.0"
+  },
+  "devDependencies": {
+    "ts-to-zod": "^3.13.0",
+    "zod": "^3.23.8"
```
marikaner, line 43: [q] Can this be moved to the root package.json? We currently do this for all other dev dependencies.

packages/foundation-models/package.json
```diff
@@ -22,13 +22,14 @@
...
     "compile:cjs": "tsc -p tsconfig.cjs.json",
-    "test": "NODE_OPTIONS=--experimental-vm-modules jest",
+    "test": "pnpm generate-zod && NODE_OPTIONS=--experimental-vm-modules jest",
     "lint": "eslint \"**/*.ts\" && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -c",
     "lint:fix": "eslint \"**/*.ts\" --fix && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -w --log-level error",
     "check:public-api": "node --loader ts-node/esm ../../scripts/check-public-api-cli.ts",
     "generate:azure-openai": "openapi-generator --generateESM --clearOutputDir -i ./src/azure-openai/spec/inference.yaml -o ./src/azure-openai/client --schemaPrefix AzureOpenAi",
     "postgenerate:azure-openai": "rm ./src/azure-openai/client/inference/*.ts && pnpm lint:fix",
-    "generate-zod": "ts-to-zod src/azure-openai/azure-openai-embedding-types.ts src/azure-openai/ts-to-zod/azure-openai-embedding-types.zod.ts"
+    "generate-zod-config": "node --loader ts-node/esm ../../scripts/generate-zod-config.ts './src/azure-openai/client/inference/schema/*.ts' ./src/azure-openai/ts-to-zod",
```
marikaner, line 31: [q] why is the input in quotes but the output not?

scripts/generate-zod-config.ts
```diff
@@ -0,0 +1,56 @@
+import { promises as fs } from 'fs';
```
marikaner, line 1: [pp]
```suggestion
import { writeFile, ... } from 'node:fs/promises';
```

scripts/generate-zod-config.ts
```diff
@@ -0,0 +1,56 @@
...
+import * as path from 'path';
```
marikaner, line 2: [pp]
```suggestion
import { basename, ... } from 'node:path';
```

scripts/generate-zod-config.ts
```diff
@@ -0,0 +1,56 @@
...
+async function main(args: string[]) {
+  if (args.length !== 2) {
+    logger.error('Usage: node --loader ts-node/esm generate-zod-config.ts <input> <output-schema>');
+    process.exit(1);
+  }
+
+  const [input, outputSchema] = args;
+  try {
+    // Find all .ts files based on the glob path provided
+    const files = glob.sync(input);
```
marikaner, line 24: [q] Is there an async version? If so, why don't we use that one?

marikaner (commented):
I think the file generation makes sense (wouldn't have a better idea), however especially due to the complexity I would like to separate it from the "standard" flow of unit tests. 

I wonder: does the config always need to be regenerated? Or only when we generate the clients. Shouldn't the config generation then be part of that step already and only executed when needed not every time we run the tests?

.gitignore
ZhongpinWang: Oh I missed this file. It is no longer needed. Removed.

packages/foundation-models/package.json
ZhongpinWang: If I remove the quote, I will get all satisfied files in the argument list according to the glob pattern. Instead, I am using quote to make it a pure string and then parse the file paths in code.

packages/orchestration/package.json
ZhongpinWang: True, done.

packages/foundation-models/package.json
deekshas8: I quite like this idea.

deekshas8: Also, the config should not be regenerated everytime (only when we update the spec and therefore the types)

ZhongpinWang: I will try to move the schema tests into a separate package.

ZhongpinWang: Refactored.

ZhongpinWang:
Now after the separation, I generate config and schema in this test package. I don't want the config generation part to be spread everywhere when openapi generator is called. 

The configs are still generated every time before test execution. Because we need schemas to run the tests, and schemas need to be generated separately via each `ts-to-zod.config.cjs` for each client. 

Notice that `ts-to-zod` only works with the `ts-to-zod.config.cjs` in the package root. I thought about creating just one config file for all generated clients, but it feels a bit too complicated to modify the script, and semantically, the generated zod schemas for each client should also be separated.

Anyways, it would make more sense if we can integrate this kind of test in e2e. But at least we now have some checks for mock response as well.

packages/foundation-models/package.json
deekshas8: The string version doesn't work for me.

deekshas8: Also, the generated ts-to-zod is not deleted post tests

eslint.config.js
```diff
@@ -22,13 +22,13 @@ export default [
       '**/coverage/**/*',
       'packages/ai-api/src/client/**/*',
       'packages/foundation-models/src/azure-openai/client/inference/schema/on-your-data-system-assigned-managed-identity-authentication-options.ts',
+      '**/*.zod.ts'
```
deekshas8, line 25: [q] Is this rule really needed? Do we commit any files with this extension now?

tests/schema-tests/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "test": "pnpm generate-schema && NODE_OPTIONS=--experimental-vm-modules jest",
+    "lint:fix": "prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -w --log-level error",
+    "generate-config:azure-openai": "node --loader ts-node/esm ../../scripts/generate-zod-config.ts '../../packages/foundation-models/src/azure-openai/client/inference/schema/*.ts' ./test/foundation-models/azure-openai/ts-to-zod",
+    "generate-config:orchestration": "node --loader ts-node/esm ../../scripts/generate-zod-config.ts '../../packages/orchestration/src/client/api/schema/*.ts' ./test/orchestration/ts-to-zod",
+    "generate-schema:azure-openai": "pnpm generate-config:azure-openai && pnpm ts-to-zod --all --skipValidation && rm ./ts-to-zod.config.cjs",
+    "generate-schema:orchestration": "pnpm generate-config:orchestration && pnpm ts-to-zod --all --skipValidation && rm ./ts-to-zod.config.cjs",
+    "generate-schema": "pnpm generate-schema:azure-openai && pnpm generate-schema:orchestration"
+  },
+  "tsd": {
+    "directory": "test"
```
deekshas8, line 18: [req] Is not relevant here. Same for `"types": "test",` in ln. 7

ZhongpinWang: 😄 Now you know from where I copied the package. Thanks and removed.

packages/foundation-models/package.json
ZhongpinWang: > Also, the generated ts-to-zod is not deleted post tests

I am not sure if it is necessary also to delete the generated zod.ts? That is also why I git ignored these files. Could be that someone want to have a look into them?

ZhongpinWang: > The string version doesn't work for me.

I feel this is a problem for windows due to slash and back slash.. Anyways, I now parse the list directly in the arguments and removed the need for single quote.

deekshas8: This still doesn't work unfortunately. The glob pattern is not evaluated when you run the command, so in the code, I still just see the string and not the list of files. Can we just pass the directory and get all files ?

deekshas8: > Could be that someone want to have a look into them?

We are regenerating it everytime anyway, so why keep a version lying around (to maintain this, we then have to add a .gitignore, eslint config). Having a cleanup step would be better IMO.
The chance that someone would actually go through this code is really low. In case they want to, they can comment out this cleanup step and debug.

ZhongpinWang: I pushed the solution, which is using `\"` instead of `'` or ``. And I have to revert to the previous variant to let glob lib parse the input within the script. On Windows, node will not parse glob pattern automatically.

ZhongpinWang: > > Could be that someone want to have a look into them?
> 
> We are regenerating it everytime anyway, so why keep a version lying around (to maintain this, we then have to add a .gitignore, eslint config). Having a cleanup step would be better IMO. The chance that someone would actually go through this code is really low. In case they want to, they can comment out this cleanup step and debug.

Removed.

eslint.config.js
ZhongpinWang: Done

tests/schema-tests/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "scripts": {
+    "test": "pnpm generate-schema && NODE_OPTIONS=--experimental-vm-modules jest",
+    "posttest": "find . -type d -name 'ts-to-zod' -exec rm -r {} +",
+    "lint:fix": "prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -w --log-level error",
+    "generate-config:azure-openai": "node --loader ts-node/esm ../../scripts/generate-zod-config.ts \"../../packages/foundation-models/src/azure-openai/client/inference/schema/*.ts\" ./test/foundation-models/azure-openai/ts-to-zod",
+    "generate-config:orchestration": "node --loader ts-node/esm ../../scripts/generate-zod-config.ts \"../../packages/orchestration/src/client/api/schema/*.ts\" ./test/orchestration/ts-to-zod",
+    "generate-schema:azure-openai": "pnpm generate-config:azure-openai && pnpm ts-to-zod --all --skipValidation",
+    "generate-schema:orchestration": "pnpm generate-config:orchestration && pnpm ts-to-zod --all --skipValidation",
+    "generate-schema": "pnpm generate-schema:azure-openai && pnpm generate-schema:orchestration",
+    "postgenerate-schema": "rm ./ts-to-zod.config.cjs"
```
deekshas8, line 17: [q] can we combine all the cleanup in one step?

ZhongpinWang: Since we are not ignoring any files any more, I want to clean up files asap, also in case we just manually call these scripts.

tests/schema-tests/package.json
```diff
@@ -0,0 +1,25 @@
...
+  },
+  "dependencies": {
+    "zod": "^3.23.8"
```
deekshas8, line 20: Please move this also to a dev dependency.

packages/foundation-models/src/azure-openai/azure-openai-embedding-response.test.ts
```diff
@@ -16,7 +15,6 @@ describe('Azure OpenAI embedding response', () => {
   const response = new AzureOpenAiEmbeddingResponse(rawResponse);
 
   it('should return the embedding response', () => {
-    const data = azureOpenAiEmbeddingOutputSchema.parse(response.data);
-    expect(data).toStrictEqual(mockResponse);
+    expect(response.data).toStrictEqual(mockResponse);
```
deekshas8, line 18: I realize this test is quite useless, we should be testing the utility functions here as well. I'll fix this in another PR.

deekshas8 (approved):
LGTM.

deekshas8:
I still feel this test should not run with every build. Ideally only when we update a client and want to test the new types. But for now its okay

packages/foundation-models/src/azure-openai/azure-openai-embedding-response.test.ts
ZhongpinWang: I agree. This test just check if `data` is set. Utility functions are more important.

ZhongpinWang:
> I still feel this test should not run with every build. Ideally only when we update a client and want to test the new types. But for now its okay

Ideally, we could have a workflow to (detect updates of specs and auto) generate clients. After that, we should run schema test + (e2e if we integrate zod also there). But this will also bring complexity.

PR #160, opened 2024-09-20
chore: Refactor orchestration E2E tests for better coverage

MatKuhr:
## Context

This PR improves the E2E test coverage for orchestration. In doing so, it asserts that the sample code is functional.

## Definition of Done

- [x] ~Error handling created / updated & covered by the tests above~
- [x] ~Documentation updated~
- [x] ~(Optional) Aligned changes with the Java SDK~
- [x] ~(Optional) Release notes updated~


sample-code/src/server.ts
```diff
@@ -48,8 +55,31 @@ app.get('/embedding', async (req, res) => {
...
+      ? orchestrationChatCompletion
+      : sampleCase === 'template'
+        ? orchestrationTemplating
+        : sampleCase === 'inputFiltering'
+          ? orchestrationInputFiltering
+          : sampleCase === 'outpttFiltering'
+            ? orchestrationOutputFiltering
+            : sampleCase === 'requestConfig'
+              ? orchestrationRequestConfig
+              : orchestrationChatCompletion;
```
marikaner, line 70: [pp] Consider extracting this to a separate function and either using switch or if statements to avoid the nesting. For better readability.

sample-code/src/server.ts
```diff
@@ -48,8 +55,31 @@ app.get('/embedding', async (req, res) => {
...
 app.get('/orchestration/:sampleCase', async (req, res) => {
+  const sampleCase = req.params.sampleCase;
+  const func =
+    sampleCase === 'simple'
+      ? orchestrationChatCompletion
+      : sampleCase === 'template'
+        ? orchestrationTemplating
+        : sampleCase === 'inputFiltering'
+          ? orchestrationInputFiltering
+          : sampleCase === 'outpttFiltering'
```
marikaner, line 66: [req] typo "outpttFiltering" => "outputFiltering"

sample-code/src/orchestration.ts
```diff
@@ -1,133 +1,170 @@
...
+export async function orchestrationTemplating(): Promise<OrchestrationResponse> {
   const orchestrationClient = new OrchestrationClient({
-    llm: {
-      model_name: 'gpt-4-32k',
-      model_params: {}
-    },
+    llm: llmConfig,
     templating: {
       template: [
+        // define "country" as variable by wrapping it with "{{? ... }}"
```
marikaner, line 54: 👍 I like these additional comments for samples!

sample-code/src/orchestration.ts
```diff
@@ -1,133 +1,170 @@
...
-async function orchestrationCompletionSimple(): Promise<string | undefined> {
+export async function orchestrationChatCompletion(): Promise<OrchestrationResponse> {
   const orchestrationClient = new OrchestrationClient({
+    // Define the language model to be used
     llm: {
-      model_name: 'gpt-4-32k',
+      model_name: 'gpt-4o',
       model_params: {}
     },
+    // Define the prompt
```
marikaner, line 25:
```suggestion
    // define the prompt
```

sample-code/src/orchestration.ts
```diff
@@ -1,133 +1,170 @@
...
 /**
- * Ask about the capital of France.
- * @returns The message content from the orchestration service in the generative AI hub.
+ * A simple LLM request, asking about the capital of France.
+ * @returns The orchestration service response.
  */
-async function orchestrationCompletionSimple(): Promise<string | undefined> {
+export async function orchestrationChatCompletion(): Promise<OrchestrationResponse> {
   const orchestrationClient = new OrchestrationClient({
+    // Define the language model to be used
```
marikaner, line 20:
```suggestion
    // define the language model to be used
```

sample-code/src/orchestration.ts
```diff
@@ -1,133 +1,170 @@
...
   const orchestrationClient = new OrchestrationClient({
-    llm: {
-      model_name: 'gpt-4-32k',
-      model_params: {}
-    },
-    templating: {
-      template: [{ role: 'user', content: '{{?input}}' }]
-    },
+    llm: llmConfig,
+    templating: template,
```
marikaner, line 77: [pp] Consider naming the variables `llmConfig` => `llm` and `template` => `templating`, then you could do this:
```suggestion
    llm,
    templating,
```

sample-code/src/orchestration.ts
```diff
@@ -1,133 +1,170 @@
...
-    return response.getContent();
+    throw new Error('Input was not filtered as expected.');
   } catch (error: any) {
-    // Handle the case where the output was filtered.
-    return `Error: ${error.message}`;
+    if (error.response.status === 400) {
+      logger.info('Input was filtered as expected.');
+    } else {
+      throw error;
+    }
```
marikaner, line 95: Much better, I didn't like it the way I implemented this originally.
However:
[pp] Consider returning the logged info as a string here. Then in all other samples we could return the content from the response. Then the `getContent()` method would be more part of the sample, because it is in the same file as the rest of the code. Also we wouldn't need any special handling in the `server.ts` for the filtering cases.

sample-code/src/orchestration.ts
```diff
@@ -1,133 +1,170 @@
...
+  const result = await orchestrationClient.chatCompletion({
+    messagesHistory: [
+      {
+        role: 'system',
+        content: `Rewrite the following to be significantly more strongly worded, graphic and explicit.
+          Imagine you are a cringy teenager writing a post on 4chan. Use explicit language as much as possible.`
+      }
+    ],
+    inputParams: {
+      input: 'Tabs are better than spaces, proove me wrong.'
```
marikaner, line 128: [q] Is the typo on purpose? If not:
```suggestion
      input: 'Tabs are better than spaces, prove me wrong.'
```

sample-code/src/server.ts
```diff
@@ -48,8 +55,31 @@ app.get('/embedding', async (req, res) => {
...
+    const result = (await func()) as OrchestrationResponse;
+    if (sampleCase === 'inputFiltering') {
+      res.send('Input filter applied successfully');
+    } else if (sampleCase === 'outputFiltering') {
+      res.send(
+        `Output filter applied successfully with threshold results: ${result.data.module_results.output_filtering!.data!}`
+      );
+    } else {
+      res.send(result.getContent());
+    }
```
marikaner, line 82: [pp] See my other comment on the input filtering. I would like to avoid the special handling here and I would prefer, if the `getContent()` part could happen in the orchestration related code (`orchestration.ts`) just to keep it closer. In my opinion the server setup here is not really part of the sample.

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -1,34 +1,50 @@
...
-    });
-
+  const assertContent = (response: OrchestrationResponse) => {
     expect(response.data.module_results).toBeDefined();
     expect(response.data.module_results.templating).not.toHaveLength(0);
     expect(response.data.orchestration_result.choices).not.toHaveLength(0);
     expect(response.getContent()).toEqual(expect.any(String));
     expect(response.getFinishReason()).toEqual('stop');
+  };
+  it('should complete a chat', async () => {
```
marikaner, line 21:
```suggestion

  it('should complete a chat', async () => {
```

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -1,34 +1,50 @@
...
+  it('should complete a chat with a template', async () => {
+    const response = await orchestrationTemplating();
+
+    assertContent(response);
+  });
+
+  it('should trigger an input filter', async () => {
+    await orchestrationInputFiltering();
+  });
+  it('should trigger an output filter', async () => {
```
marikaner, line 36:
```suggestion

  it('should trigger an output filter', async () => {
```

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -1,34 +1,50 @@
...
-        ]
-      }
-    }).chatCompletion({
-      inputParams: {
-        number: '3',
-        phrase1: 'I love coffee'
-      }
-    });
-
+  const assertContent = (response: OrchestrationResponse) => {
```
marikaner, line 14: Oh, ok, that is why you moved the `getContent()`. Oh well, consider my comments anyways, maybe you have an idea how to combine this, if not this was `[pp]` anyways ;)

marikaner (commented):
Did you check whether the smoke tests still work on this? You can just run the action for this branch.

sample-code/src/server.ts
MatKuhr: Haha yeah I just got inspired by Tom & Deeksha writing [this code](https://github.com/SAP/ai-sdk-js/blob/main/packages/langchain/src/openai/util.ts#L165) 😉 

Any preference what looks better? Simply switch? Or maybe rather a dict?

sample-code/src/server.ts
MatKuhr: I think there could be some valuable code also here for users, namely once we support streaming and (might) want to show how to return a stream to the frontend.

But this might also be improved later once we have better error handling. Currently, input filtering gives no easy access to the filter results. Once we have that, we could add it here. Or in the "main" sample code, whatever feels more natural.

sample-code/src/orchestration.ts
MatKuhr: Seems to work equally, fixed ✅

MatKuhr:
[Smoke tested](https://github.com/SAP/ai-sdk-js/actions/runs/10996810149) ✅

sample-code/src/server.ts
MatKuhr: used a dict now

packages/orchestration/README.md
```diff
@@ -4,14 +4,17 @@ This package incorporates generative AI orchestration capabilities into your AI
...
-1. [Installation](#installation)
-2. [Prerequisites](#prerequisites)
-3. [Orchestration Service](#orchestration-service)
-4. [Usage](#usage)
-   - [Templating](#templating)
-   - [Content Filtering](#content-filtering)
-5. [Support, Feedback, Contribution](#support-feedback-contribution)
-6. [License](#license)
+- [@sap-ai-sdk/orchestration](#sap-ai-sdkorchestration)
+  - [Table of Contents](#table-of-contents)
```
ZhongpinWang, line 8: [pp] I remember during the GA release, we excluded the toc by using `###` for `Table of Contents`

tests/e2e-tests/src/orchestration.test.ts
ZhongpinWang: Thanks! This was also something disturbing me for quite sometime since I had the idea to integrate zod into e2e test. In the future, we can use zod to check response instead of manually assert each and every attributes. Maybe we want to change other sample code as well to always return response and log content.

sample-code/src/orchestration.ts
```diff
@@ -1,133 +1,170 @@
...
+    templating,
+    filtering: {
+      output: filter
+    }
+  });
+  // trigger the output filter
+  // note: reliably triggering the output filter is a bit of a challenge
+  // LLMs are fine-tuned to not respond with explicit or hateful content
+  // Instead, they often respond with something like "Sorry, I can't assist with that" when asked to generate explicit content
+  // The following prompt seems to work well with GPT-4o, but may not work with other models
```
ZhongpinWang, line 118: [req] I would always start a comment with capital letter and end a line with a dot. Same for all other comments.
[pp] I would avoid the term `seems to work` and `may not work`.

packages/orchestration/README.md
MatKuhr: This was accidentally changed by some VS code plugin. Reverted it now

sample-code/src/orchestration.ts
MatKuhr: Changed it to a longer comment block to be consistent with the rest of the commenting style.

For now, I left the `seems to` and `may not` in, to intentionally indicate the uncertainty we are dealing with here. Models change over time and even very slight modifications to the prompt might yield different results.

sample-code/src/orchestration.ts
```diff
@@ -1,133 +1,171 @@
...
       template: [{ role: 'user', content: 'What is the capital of France?' }]
     }
   });
 
-  // Call the orchestration service.
-  const response = await orchestrationClient.chatCompletion();
-  // Access the response content.
-  return response.getContent();
+  // execute the request
+  const result = await orchestrationClient.chatCompletion();
```
ZhongpinWang, line 32: [pp] I would use `response` instead of `result` to keep it consistent with what we used in e.g. unit tests.

MatKuhr: For unit test that makes sense to me, for sample code however I persoanlly prefer `result` because response might be confused with an HTTP response.

ZhongpinWang (approved):
LGTM

sample-code/src/orchestration.ts
```diff
@@ -1,114 +1,154 @@
...
+    await orchestrationClient.chatCompletion({
       inputParams: { input: 'I hate you!' }
     });
-    // Access the response content.
-    return response.getContent();
+    throw new Error('Input was not filtered as expected.');
   } catch (error: any) {
-    // Handle the case where the output was filtered.
-    return `Error: ${error.message}`;
+    if (error.response.status === 400) {
```
KavithaSiva, line 91: [pp] Also, typecheck if `error instanceof Error`. Ideally, we should also check if it is an `AxiosError` as we are assuming it's structure. But, I guess this would lead to adding a new dependency.

PR #162, opened 2024-09-23
chore: Reuse actions

marikaner:
## Context

I recently learned about composite actions and reusable workflows. We currently have to configure three actions + run `pnpm install` in every job, so I though we might benefit from merging those into one setup action.

The advantage is, that we can save a bit of configuration overhead and never forget adding a cache to the setup node cache (which happened before).

### Caveats

The downside is, that now all configuration for all actions is summarized into one, which potentially mixes concerns. We could work around this by prefixing all of the inputs with `node-`, `pnpm` and `checkout`. On a first glance I have the impression that it isn't big enough of a deal to do this.

Another change is, that we currently have two places, where we maintain the default node and pnpm version: in the github variables and in the action. The reason for this is, that in the action we cannot access the variables. An alternative could be that the node-version and pnpm version are required and always handed it externally. Also, currently the default node version is only used in one other place and the pnpm version is not used anywhere anymore, so we could consider deleting this.

If you need any additional parameters for one of the underlying actions, you would have to expose it through the action as well.


.github/workflows/auto-fix-lint.yml
```diff
@@ -14,19 +14,10 @@ jobs:
     if: github.actor != 'dependabot[bot]' && !github.event.pull_request.head.repo.fork
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: sap/ai-sdk-js/.github/actions/setup@scripts
```
marikaner, line 17: Those all need to be changed to 
```suggestion
      - uses: sap/ai-sdk-js/.github/actions/setup@main
```

before merging!

.github/workflows/auto-fix-lint.yml
```diff
@@ -14,19 +14,10 @@ jobs:
...
         with:
-          fetch-depth: 0
           ref: ${{ github.event.pull_request.head.ref }}
```
marikaner, line 19:
```suggestion
          fetch-depth: 0
          ref: ${{ github.event.pull_request.head.ref }}
```

tomfrenken (approved):
Overall LGTM.
As we already discussed, consider removing a few of the inputs as they seem to be unnecessary.
(E.g., fetch-depth, pnpm version if we move it into the package.json, ...)

.github/workflows/auto-fix-lint.yml
```diff
@@ -14,19 +14,10 @@ jobs:
     if: github.actor != 'dependabot[bot]' && !github.event.pull_request.head.repo.fork
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: sap/ai-sdk-js/.github/actions/setup@scripts
         with:
-          fetch-depth: 0
```
MatKuhr, line 19: this is now lost, I guess it might have been here for performance reasons? Not sure if it's worth it though, our repo is not that huge ;)

marikaner: I removed that, but didn't leave a comment. This was unnecessary and actually has a bad impact on performance because it fetches all history and all tags. (Kind of counter intuitive if you ask me.)

PR #163, opened 2024-09-23
chore: Update github actions as well

marikaner:
Currently dependabot only updates npm dependencies, but not github actions. This should fix that.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,6 @@
 version: 2
 updates:
   - package-ecosystem: npm
-    target-branch: main
     directory: '/'
     schedule:
       interval: daily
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,7 +1,6 @@
 version: 2
 updates:
   - package-ecosystem: npm
```
tomfrenken, line 3: Not sure how relevant this is, but apparently "pnpm" is it's own package ecosystem 
-> https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories

marikaner: yes, the value in the yaml file is `npm` for that.

tomfrenken (approved):
LGTM

PR #167, opened 2024-09-23
fix: Fix langchain types

tomfrenken:
## Context

AI/gen-ai-hub-sdk-js-backlog#ISSUENUMBER.

With this PR we fix the LangChain types to work with IDE autocomplete.

While updating the types, I also noticed that some of the newly introduced fields weren't properly mapped.
Now they are also included.

## Definition of Done

- [ ] Documentation updated

### DIFF
--- a/packages/langchain/src/openai/util.test.ts
+++ b/packages/langchain/src/openai/util.test.ts
@@ -48,8 +48,8 @@ describe('Mapping Functions', () => {
     const defaultOptions = { signal: undefined, promptIndex: 0 };
     const mapping = mapLangchainToAiClient(
       client,
-      defaultOptions,
-      langchainPrompt
+      langchainPrompt,
+      defaultOptions
     );
     expect(mapping).toMatchObject(request);
   });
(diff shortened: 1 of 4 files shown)


packages/langchain/src/openai/util.test.ts
```diff
@@ -48,8 +48,8 @@ describe('Mapping Functions', () => {
     const defaultOptions = { signal: undefined, promptIndex: 0 };
     const mapping = mapLangchainToAiClient(
       client,
-      defaultOptions,
-      langchainPrompt
+      langchainPrompt,
+      defaultOptions
```
marikaner, line 52: [pp] If you have default options, shouldn't those be considered in `mapLangchainToAiClient()` by **default** without passing them?

marikaner (approved):
LGTM, one improvement though.

packages/langchain/src/openai/util.test.ts
tomfrenken: The reason I pass them here is to replicate the actual behavior when calling the client.
The super class always passes these options.

PR #169, opened 2024-09-23
fix: Fix setup action

tomfrenken:
## Context

We unintentionally left a WIP action reference in, this PR updates the reference.


tomfrenken:
Close as its already pushed to main.

PR #173, opened 2024-09-24
chore: Add additional test coverage to our LangChain clients

tomfrenken:
## Context

AI/gen-ai-hub-sdk-js-backlog#122.

With this PR, we add / change our test setup so that every mapping function is being traversed, instead of the previous very simple test case.

## Definition of Done


packages/langchain/src/openai/util.test.ts
```diff
@@ -2,55 +2,71 @@ import {
...
+describe('Mapping Functions', () => {
   it('should parse an OpenAI response to a (LangChain) chat response', async () => {
     const result = mapOutputToChatResult(openAiMockResponse);
     expect(result).toMatchSnapshot();
   });
 
   it('should parse a LangChain input to an AI SDK input', async () => {
     const langchainPrompt: BaseMessage[] = [
-      new HumanMessage('Where is the deepest place on earth located')
+      new ToolMessage('Test Content', 'test-id')
```
MatKuhr, line 35: how about leaving the human message in and rather adding other cases?

packages/langchain/src/openai/util.test.ts
```diff
@@ -2,55 +2,71 @@ import {
...
   AzureOpenAiCreateChatCompletionRequest
 } from '@sap-ai-sdk/foundation-models';
-import nock from 'nock';
-import { BaseMessage, HumanMessage } from '@langchain/core/messages';
 import {
-  mockClientCredentialsGrantCall,
-  parseMockResponse
-} from '../../../../test-util/mock-http.js';
+  BaseMessage,
+  RemoveMessage,
```
MatKuhr, line 7: What is that?

.changeset/bright-jokes-provide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/langchain': patch
+---
+
+[Improvement] Improve error handling.
```
MatKuhr, line 5: I assume the pipeline enforces having this (useless) entry?

packages/langchain/src/openai/util.test.ts
```diff
@@ -2,55 +2,71 @@ import {
...
     const client = new AzureOpenAiChatClient({ modelName: 'gpt-35-turbo' });
-    const defaultOptions = { signal: undefined, promptIndex: 0 };
-    const mapping = mapLangchainToAiClient(
-      client,
-      langchainPrompt,
-      defaultOptions
-    );
+    const options: AzureOpenAiChatCallOptions & {
+      promptIndex?: number;
+    } = {
```
MatKuhr, line 53: Would end users have to use the same approach?

packages/langchain/src/openai/util.ts
```diff
@@ -180,9 +180,6 @@ type RoleAndContent = {
 
 function mapRoleAndContent(baseMessage: BaseMessage): RoleAndContent {
   const role = mapBaseMessageToRole(baseMessage);
-  if (!['system', 'user', 'assistant', 'tool', 'function'].includes(role)) {
```
MatKuhr, line 183: where did this go?

packages/langchain/src/openai/types.ts
```diff
@@ -8,7 +8,10 @@ import type {
   AzureOpenAiChatModel,
   AzureOpenAiChatCompletionsRequestCommon
 } from '@sap-ai-sdk/foundation-models';
-import type { CustomRequestConfig } from '@sap-ai-sdk/core';
+import type {
+  AzureOpenAiEmbeddingModel,
```
deekshas8, line 12: [req] This should also be exported from the foundation-models package.

packages/langchain/src/openai/util.ts
tomfrenken: I removed it as `mapBaseMessageToRole` already contains this exact logic, therefore I viewed it as redundancy.

sample-code/src/langchain-azure-openai.ts
```diff
@@ -1,9 +1,13 @@
+import { CheerioWebBaseLoader } from '@langchain/community/document_loaders/web/cheerio';
 import { StringOutputParser } from '@langchain/core/output_parsers';
 import { ChatPromptTemplate } from '@langchain/core/prompts';
+import { RecursiveCharacterTextSplitter } from '@langchain/textsplitters';
 import {
   AzureOpenAiChatClient,
   AzureOpenAiEmbeddingClient
 } from '@sap-ai-sdk/langchain';
+import { createStuffDocumentsChain } from 'langchain/chains/combine_documents';
+import { MemoryVectorStore } from 'langchain/vectorstores/memory';
```
deekshas8, line 10: [q] Can we extend the simple example to also pass ChatCallOptions?

packages/langchain/src/openai/util.test.ts
tomfrenken: Not exactly, users can't even provide signal & promptIndex I believe, however, these values are always part of the options as these super class provides them.

packages/langchain/src/openai/util.test.ts
tomfrenken: RemoveMessage is another message type from langchain which is not a supported message type for openai.
I just use it to cause an error case.

packages/langchain/src/openai/embedding.ts
```diff
@@ -13,7 +13,7 @@ export class AzureOpenAiEmbeddingClient
   extends Embeddings
   implements AzureOpenAiEmbeddingModelParams
 {
-  modelName: AzureOpenAiChatModel;
+  modelName: AzureOpenAiEmbeddingModel;
```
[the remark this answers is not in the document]
tomfrenken: It is, I wrote the changelog before I made this bug fix, I will update it.

packages/langchain/src/openai/util.test.ts
tomfrenken: I felt like it was a redundancy, as it traverses the same functions but less than the tool message, what do you think?

packages/langchain/src/openai/embedding.ts
```diff
@@ -13,7 +13,7 @@ export class AzureOpenAiEmbeddingClient
   extends Embeddings
   implements AzureOpenAiEmbeddingModelParams
```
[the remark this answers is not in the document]
tomfrenken: The purpose is that when we re-generate or change the type, the class also breaks, otherwise we might notice it too late.
This is less of an issue for the embedding class as the types are much more straightforward, but this can happen quickly in the chat class.

sample-code/src/langchain-azure-openai.ts
```diff
@@ -1,56 +1,129 @@
...
+  // Create a retriever for the vector store
+  const retriever = vectorStore.asRetriever();
+
+  // Create a prompt
+  const prompt = 'How do you use templating in the SAP Orchestration client?';
+
+  // Invoke the chat client combined with the prompt, prompt template, output parser and vector store context
+  return ragChain.invoke({
+    question: prompt,
+    context: await retriever.invoke(prompt)
```
MatKuhr, line 127: langchain doesn't automate this step? calling invoke inside of invoke looks a bit cursed..

sample-code/src/langchain-azure-openai.ts
```diff
@@ -1,56 +1,129 @@
+import { fileURLToPath } from 'url';
+import path, { resolve } from 'path';
 import { StringOutputParser } from '@langchain/core/output_parsers';
 import { ChatPromptTemplate } from '@langchain/core/prompts';
+import { RecursiveCharacterTextSplitter } from '@langchain/textsplitters';
 import {
   AzureOpenAiChatClient,
   AzureOpenAiEmbeddingClient
 } from '@sap-ai-sdk/langchain';
+import { createStuffDocumentsChain } from 'langchain/chains/combine_documents';
```
MatKuhr, line 10: `createStuffDocumentsChain` what a name...

MatKuhr (approved):

sample-code/src/langchain-azure-openai.ts
tomfrenken: LangChain moment

PR #175, opened 2024-09-24
chore: Remove eslint workaround

deekshas8:
## Context

`@typescript-eslint/consistent-type-definitions` converts generated types (which don't have `& Record<string, any>`) to interfaces
- This rule is now turned off for generated schema code
- The jsdoc/require-jsdoc is turned off for generated code (across all packages) as well

## Definition of Done

- [ ] Documentation updated

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -20,14 +20,14 @@ export default [
       '**/*.d.ts',
       '**/dist/**/*',
       '**/coverage/**/*',
-      'packages/ai-api/src/client/**/*',
-      'packages/foundation-models/src/azure-openai/client/inference/schema/on-your-data-system-assigned-managed-identity-authentication-options.ts',
+      'packages/ai-api/src/client/**/*'
     ]
   },
   {
     files: [
       '**/test-util/**/*.ts',
       '**/packages/orchestration/src/client/**/*',
+      '**/packages/foundation-models/src/azure-openai/client/**/*',
       '**/*.zod.ts'
     ],
     rules: {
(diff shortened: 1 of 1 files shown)


eslint.config.js
```diff
@@ -20,14 +20,13 @@ export default [
...
-      'packages/ai-api/src/client/**/*',
-      'packages/foundation-models/src/azure-openai/client/inference/schema/on-your-data-system-assigned-managed-identity-authentication-options.ts',
+      'packages/ai-api/src/client/**/*'
     ]
   },
   {
     files: [
       '**/test-util/**/*.ts',
-      '**/packages/orchestration/src/client/**/*',
+      '**/packages/**/client/**/*',
```
ZhongpinWang, line 29: [q] Why don't we ignore all clients above like `ai-api` to ignore all rules?

deekshas8: Yeah I guess we could do that. I was not sure if we want to completely turn off linting for them

eslint.config.js
```diff
@@ -45,6 +43,12 @@ export default [
       'jsdoc/check-indentation': 'off'
     }
   },
+  {
+    files: ['**/packages/**/client/**/*.ts'],
+    rules: {
+      '@typescript-eslint/consistent-type-definitions': 'off'
+    }
+  },
```
ZhongpinWang, line 51: Do we still need this as we already ignored all clients above?

deekshas8: 🤦🏽‍♀️

PR #185, opened 2024-09-30
chore: Fix resource group header

tomfrenken:
## Context

AI/gen-ai-hub-sdk-js-backlog#146.

A user discovered a bug where we do not properly set the resource group after they have set it in the client's initialization.

With this fix, this should be fixed.

## Definition of Done

- [ ] Documentation updated

### DIFF
--- a/packages/ai-api/src/utils/deployment-resolver.ts
+++ b/packages/ai-api/src/utils/deployment-resolver.ts
@@ -51,7 +51,9 @@ export type ModelDeployment<ModelNameT = string> =
 /**
  * @internal
  */
-export function getResourceGroup(modelDeployment: ModelDeployment): string | undefined {
+export function getResourceGroup(
+  modelDeployment: ModelDeployment
+): string | undefined {
   return typeof modelDeployment === 'object'
     ? modelDeployment.resourceGroup
     : undefined;
--- a/packages/ai-api/src/utils/deployment-resolver.ts
+++ b/packages/ai-api/src/utils/deployment-resolver.ts
@@ -49,6 +49,15 @@ export type ModelDeployment<ModelNameT = string> =
   | ((ModelConfig<ModelNameT> | DeploymentIdConfig) & ResourceGroupConfig);
 
 /**
+ * @internal
+ */
+export function getResourceGroup(modelDeployment: ModelDeployment): string | undefined {
+  return typeof modelDeployment === 'object'
+    ? modelDeployment.resourceGroup
+    : undefined;
+}
+
+/**
  * Type guard to check if the given deployment configuration is a deployment ID configuration.
  * @param modelDeployment - Configuration to check.
  * @returns `true` if the configuration is a deployment ID configuration, `false` otherwise.
--- a/packages/core/src/http-client.test.ts
+++ b/packages/core/src/http-client.test.ts
@@ -36,4 +36,27 @@ describe('http-client', () => {
     expect(res.status).toBe(200);
     expect(res.data).toEqual(mockPromptResponse);
   }, 10000);
+
+  it('should execute a request to the AI Core service with a custom resource gorup', async () => {
+    const mockPrompt = { prompt: 'some test prompt' };
+    const mockPromptResponse = { completion: 'some test completion' };
+
+    const scope = nock(aiCoreDestination.url, {
+      reqheaders: {
+        'ai-resource-group': 'custom-resource-group',
+        'ai-client-type': 'AI SDK JavaScript'
+      }
+    })
+      .post('/v2/some/endpoint', mockPrompt)
+      .query({ 'api-version': 'mock-api-version' })
+      .reply(200, mockPromptResponse);
+
+    const res = await executeRequest(
+      { url: '/some/endpoint', apiVersion: 'mock-api-version', resourceGroup: 'custom-resource-group' },
+      mockPrompt
+    );
+    expect(scope.isDone()).toBe(true);
+    expect(res.status).toBe(200);
+    expect(res.data).toEqual(mockPromptResponse);
+  });
 });
--- a/packages/core/src/http-client.test.ts
+++ b/packages/core/src/http-client.test.ts
@@ -52,7 +52,11 @@ describe('http-client', () => {
       .reply(200, mockPromptResponse);
 
     const res = await executeRequest(
-      { url: '/some/endpoint', apiVersion: 'mock-api-version', resourceGroup: 'custom-resource-group' },
+      {
+        url: '/some/endpoint',
+        apiVersion: 'mock-api-version',
+        resourceGroup: 'custom-resource-group'
+      },
       mockPrompt
     );
     expect(scope.isDone()).toBe(true);
--- a/packages/core/src/http-client.test.ts
+++ b/packages/core/src/http-client.test.ts
@@ -37,7 +37,7 @@ describe('http-client', () => {
     expect(res.data).toEqual(mockPromptResponse);
   }, 10000);
 
-  it('should execute a request to the AI Core service with a custom resource gorup', async () => {
+  it('should execute a request to the AI Core service with a custom resource group', async () => {
     const mockPrompt = { prompt: 'some test prompt' };
     const mockPromptResponse = { completion: 'some test completion' };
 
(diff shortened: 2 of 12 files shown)


packages/foundation-models/src/azure-openai/azure-openai-chat-client.test.ts
```diff
@@ -75,4 +75,93 @@ describe('Azure OpenAI chat client', () => {
 
     await expect(client.run(prompt)).rejects.toThrow('status code 400');
   });
+
+  it('executes a request with the custom request group', async () => {
```
MatKuhr, line 79: It is important that the RG is used for both: 

* lookup of deployment ID
* inference request

Since we are already testing the latter here, I'd consider also covering the former. While this does come with a bit more request mocking, I feel that we do need a test for this.

We could also invert the strategy, always covering both requests and only in one single test case implement (pseudo code):
```ts
new AzureOpenAiChatClient({ deploymentId: '1234' })
expect(nock).toNeverBeCalled()
```

Arguably we should also cover that
```
new AzureOpenAiChatClient({ deploymentId: '1234', resourceGroup: 'foo' });
```
works, i.e. that the RG `foo` is sent for the inference request, but no deployment ID lookup happens..

packages/ai-api/src/utils/deployment-resolver.ts
```diff
@@ -48,6 +48,17 @@ export type ModelDeployment<ModelNameT = string> =
...
 
+/**
+ * @internal
+ */
+export function getResourceGroup(
+  modelDeployment: ModelDeployment
+): string | undefined {
+  return typeof modelDeployment === 'object'
+    ? modelDeployment.resourceGroup
+    : undefined;
```
MatKuhr, line 59: (minor) Consider: 

```suggestion
): string {
  return typeof modelDeployment === 'object'
    ? modelDeployment.resourceGroup
    : 'default' ;
```

Because if I see it correctly in all usages `undefined` is then actually turned into the default RG.

MatKuhr (commented):
duplicating these tests a few times is of course a bit annoying. Is there any way to shorten them perhaps? 

Pseudo code:
```
const expectedEndpoint = { url: 'localhost', resourceGroup: 'default' }
const mock = mock(httpClient)
client.run(prompt)
expect(mock.executeRequest).toBeCalled().with(expectedEndpoint)
```

Since we don't care about the full flow, we only care about the correct RG being passed on..

packages/ai-api/src/utils/deployment-resolver.ts
tomfrenken: undefined is never assigned with the `??` operator, e.g.
```
resourceGroup ?? 'default',
```
I would keep the function as is, as the model deployment doesn't contain anything.

tomfrenken: Otherwise the function would be called `getResourceGroupWithDefault` or so

MatKuhr: Then why not just call it that way? Because otherwise we are duplicating `?? 'default'` in every usage. Assuming there are multiple usages, otherwise no point in having a getter, right?

tomfrenken: I would argue the purpose of the getter is to handle the different types the property can hold, e.g. string or object.

tomfrenken: I will change the function to WithDefault, with an optional parameter holding a different default than 'default'

tomfrenken: On second thought, this doesn't improve the situation, as I would then need to change the endpoint type to require this property, which would then change every call of executeRequest.
Also we only do the `??` logic once in the actual function and once in a mock, so it doesn't really provide a lot of value.

-> I suggest not changing it

packages/foundation-models/src/azure-openai/azure-openai-chat-client.test.ts
tomfrenken: I generally stronlgy agree, however, mocking became extremely annoying with ESM in Jest.
I am currently looking into a jest plugin called [`testdouble`](https://github.com/testdouble/testdouble-jest) which might enable us to use mocking the way you describe it again.
In commonJS we would be able to simply use jest.spyOn(), this however does only work on instantiated classes in ESM out of the box.

tomfrenken: After trying every possible combination, I came to the conclusion that we can't use it which means mocking is not realistically feasible for this scenario 🙄

tomfrenken: I think we have a long list of test for the former in `deployment-resolver.test.ts`

MatKuhr: > mocking is not realistically feasible for this scenario

🫠

> long list of tests for the former

They don't really test:

> that the RG is used for both:
> * lookup of deployment ID
> * inference request

I.e. there is no assertion that guarantees "if the deployment ID is resolved using RG `foo`, then the inference request also contains RG `foo`"

ZhongpinWang (approved):
LGTM. Only one question came to my mind while reviewing the PR.

packages/core/src/http-client.test.ts
```diff
@@ -36,4 +36,31 @@ describe('http-client', () => {
     expect(res.status).toBe(200);
     expect(res.data).toEqual(mockPromptResponse);
   }, 10000);
+
+  it('should execute a request to the AI Core service with a custom resource gorup', async () => {
```
deekshas8, line 40:
```suggestion
  it('should execute a request to the AI Core service with a custom resource group', async () => {
```

packages/core/src/http-client.ts
```diff
@@ -30,11 +30,14 @@ export interface EndpointOptions {
    * The specific endpoint to call.
    */
   url: string;
-
   /**
    * The API version to use.
    */
   apiVersion?: string;
```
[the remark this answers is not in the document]
ZhongpinWang: Discussed.

packages/foundation-models/src/azure-openai/azure-openai-chat-client.test.ts
```diff
@@ -75,4 +75,93 @@ describe('Azure OpenAI chat client', () => {
...
+    const clientWithResourceGroup = new AzureOpenAiChatClient({
+      deploymentId: '1234',
+      resourceGroup: 'custom-resource-group'
+    });
+
+    const response = await clientWithResourceGroup.run(prompt);
+    expect(response.data).toEqual(mockResponse);
+  });
+
+  it('it fails when the wrong resource group is set', async () => {
```
deekshas8, line 121:
```suggestion
  it('fails when the wrong resource group is set', async () => {
```

packages/foundation-models/src/azure-openai/azure-openai-chat-client.test.ts
deekshas8, line 79:
```suggestion
  it('executes a request with the custom resource group', async () => {
```

packages/foundation-models/src/azure-openai/azure-openai-chat-client.test.ts
```diff
@@ -75,4 +75,93 @@ describe('Azure OpenAI chat client', () => {
...
+      ]
+    };
+
+    const mockResponse =
+      parseMockResponse<AzureOpenAiCreateChatCompletionResponse>(
+        'foundation-models',
+        'azure-openai-chat-completion-success-response.json'
+      );
+
+    mockInference(
```
deekshas8, line 143: [q] I didn't get the point of mocking this?

ZhongpinWang (commented):
Just in case, don't forget the changeset.

packages/foundation-models/src/azure-openai/azure-openai-chat-client.test.ts
tomfrenken: I see, I've added a test for this now to the azure client as well as orchestration 👍

PR #187, opened 2024-10-01
chore: Refactor ai-api E2E tests

shibeshduw:
## Context

AI/gen-ai-hub-sdk-js-backlog#142.

- Added sample code for `DeploymentApi` and `ScenarioApi` endpoints, some of which are now used in the e2e-tests.
- These tests can now be run individually for better local testing or sequentially as a whole.
- Implemented a cleanup feature so that unused deployments are deleted if tests fail or are interrupted. 
   - beforeAll() checks if deployments are remaining from a previously interrupted run and deletes them.
   - afterAll() will delete all deployments that are currently present in the resource group, regardless of whether a single test was executed or all tests were executed.
  

## Definition of Done

- [ ] Documentation updated


tests/e2e-tests/src/deployment-api.test.ts
```diff
@@ -14,20 +19,21 @@ describe('DeploymentApi', () => {
   let initialState: AiDeploymentList | undefined;
 
   beforeAll(async () => {
-    const queryResponse = await DeploymentApi.deploymentQuery(
-      {},
-      { 'AI-Resource-Group': resourceGroup }
-    ).execute();
+    await cleanupDeployments();
+    const queryResponse = await getDeployments(resourceGroup);
```
deekshas8, line 23: [q] When we clean up all deployments initially, does this return anything?

tests/e2e-tests/src/deployment-api.test.ts
```diff
@@ -48,17 +53,15 @@ describe('DeploymentApi', () => {
...
 
   it('should modify the deployment to stop it', async () => {
-    const deploymentId = getDeploymentId(createdDeploymentId);
-
-    const modifyResponse = await DeploymentApi.deploymentModify(
-      deploymentId,
-      { targetStatus: 'STOPPED' },
-      { 'AI-Resource-Group': resourceGroup }
-    ).execute();
+    const deploymentId = await checkCreatedDeployment(
```
deekshas8, line 59: @marikaner @shibeshduw  I think I am missing the point here, what would be the reason I would run modify deployment individually, without having run create deployment? Right now, if no deployment exists, this function throws an error anyways. 
Its the same behavior if I call the modify script, the `createdDeploymentId` is undefined in that case, and if no other deployments exist, we still throw.

tests/e2e-tests/package.json
```diff
@@ -19,6 +19,9 @@
   "scripts": {
     "compile": "tsc",
     "test": "NODE_OPTIONS=--experimental-vm-modules jest",
+    "test:deployment-api-create": "NODE_OPTIONS=--experimental-vm-modules jest -t \"create deployment\"",
+    "test:deployment-api-modify": "NODE_OPTIONS=--experimental-vm-modules jest -t \"modify deployment\"",
+    "test:deployment-api-delete": "NODE_OPTIONS=--experimental-vm-modules jest -t \"delete deployment\"",
```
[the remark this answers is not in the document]
shibeshduw: This uses the `--testNamePattern=<regex>` cli option for Jest so doesn't have to match the whole test name. Works if the test names are unique enough which they are.

tests/e2e-tests/src/deployment-api.test.ts
shibeshduw: Yes it returns the count. In the event that cleaning deployment fails initially, then this would still be useful for that particular test flow since we would be matching the initial and final state at the end.

tests/e2e-tests/src/deployment-api.test.ts
shibeshduw: I think the best way to go about it would be to create a separate script to cleanup deployments like @marikaner suggested. We can make this a GitHub action that runs before the e2e tests and also be able to run it manually when testing locally. 
Only then would the current impl make sense. Because right now, it doesn't make sense like you said. The point of running the test individually was so that if something is wrong, the entire test flow doesn't have to be run every time. 

I will make this change today as I'm also not satisfied with the current way.

shibeshduw: Okay I made the above change. Now makes sense to run the tests either individually or all at once. The cleanup script is setup in github workflows so we don't have to check for test failures and the million different conditions that come with it. Let me know if this works.

sample-code/src/index.ts
```diff
@@ -16,3 +16,11 @@ export {
...
   invokeRagChain
 } from './langchain-azure-openai.js';
+export {
+  getDeployment,
+  getDeployments,
+  createDeployment,
+  modifyDeployment,
+  deleteDeployment
+} from './ai-api/deployment-api.js';
+export { getScenarios } from './ai-api/scenario-api.js';
```
ZhongpinWang, line 26: [req] Missing `getModelsInScenario`?

tests/e2e-tests/package.json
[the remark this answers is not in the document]
ZhongpinWang: [q] If we wanna test individually, can't we just click on some button? Putting more and more scripts for each and every tests looks really cumbersome.

tests/e2e-tests/src/deployment-api.test.ts
```diff
@@ -14,20 +20,16 @@ describe('DeploymentApi', () => {
...
   beforeAll(async () => {
-    const queryResponse = await DeploymentApi.deploymentQuery(
-      {},
-      { 'AI-Resource-Group': resourceGroup }
-    ).execute();
+    const queryResponse = await getDeployments(resourceGroup);
     expect(queryResponse).toBeDefined();
     initialState = queryResponse;
-  });
+  }, 200000);
```
ZhongpinWang, line 26: [q] Why do we need a timeout here as well? Will `getDeployments` actually take 20 seconds?

ZhongpinWang (changes requested):
Will continue reviewing the PR once I have more time. Left some comments.

sample-code/src/ai-api/deployment-api.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/**
+ * Update target status of a specific deployment to stop it.
+ * Only deployments with 'status': 'RUNNING' can be stopped.
+ * @param deploymentId - ID of the specific deployment.
+ * @param resourceGroup - AI-Resource-Group where the resources are available.
+ * @returns Deployment modification response with 'targetStatus': 'STOPPED'.
+ */
+export async function modifyDeployment(
```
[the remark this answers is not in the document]
shibeshduw: Yeah this makes sense for the current usage. Although the `deploymentModify` from can also be used to change configuration, but the request format would be different.

tests/e2e-tests/package.json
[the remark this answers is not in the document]
shibeshduw: I guess it looks a bit cumbersome. I just added it for convenience but I agree that there are other ways to test individually.  I will remove it for now.

tests/e2e-tests/src/utils/cleanup-deployments.ts
```diff
@@ -0,0 +1,56 @@
...
+          ) {
+            await modifyDeployment(id, resourceGroup);
+            await waitForDeploymentToReachStatus(id, 'STOPPED');
+          } else if (status !== 'STOPPED' && targetStatus === 'STOPPED') {
+            await waitForDeploymentToReachStatus(id, 'STOPPED');
+          }
+
+          await deleteDeployment(id, resourceGroup);
+          // Wait for deletion to complete
+          await new Promise(r => setTimeout(r, 25000));
```
ZhongpinWang, line 45: [q] What is the reason for this wait?

ZhongpinWang (changes requested):

sample-code/src/ai-api/deployment-api.ts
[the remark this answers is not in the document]
ZhongpinWang: Yes, but you hardcode the target status to STOPPED. And the jsdoc suggest that it just do delete. I would either put target status as part of the parameter or change the function name.

sample-code/src/ai-api/deployment-api.ts
```diff
@@ -0,0 +1,99 @@
...
+export async function getDeployments(
+  resourceGroup: string,
+  status?:
+    | 'PENDING'
+    | 'RUNNING'
+    | 'COMPLETED'
+    | 'DEAD'
+    | 'STOPPING'
+    | 'STOPPED'
+    | 'UNKNOWN'
```
marikaner, line 25: [pp] The status seems to not be used in the sample. I think a sample is more helpful, when it is concise and specific. For the sample code consider creating an example that filters by a specific status with a specific resource group. Also take a look at the samples for orchestration and azure open ai. Otherwise I am wondering why bother creating this function at all if it just wraps the actual API.

This also applies to the other sample functions here and for the scenario API.

Also note, that in the tests you can use the actual API directly, so you don't need to wrap it here for the tests.

sample-code/src/ai-api/deployment-api.ts
```diff
@@ -0,0 +1,99 @@
...
+  }).execute();
+}
+
+/**
+ * Get information about specific deployment.
+ * @param deploymentId - ID of the specific deployment.
+ * @param resourceGroup - AI-Resource-Group where the resources are available.
+ * @returns Details for deplyoment with deploymentId.
+ */
+export async function getDeployment(
```
marikaner, line 40: [pp] Also, I would really just put the functions here that are exposed through the `server.ts` file. Otherwise, when someone tries the sample it will be unused code from their point of view.

tests/e2e-tests/src/utils/ai-api-utils.ts
```diff
@@ -1,4 +1,35 @@
...
+  return retry(
+    async () => {
+      const deploymentDetail = await getDeployment(deploymentId, resourceGroup);
+      if (deploymentDetail.status === targetStatus) {
+        return deploymentDetail;
+      }
+      throw new Error(`Deployment has not yet reached ${targetStatus} status.`);
+    },
+    {
+      retries: 30,
```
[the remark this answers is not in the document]
marikaner: 30 retries is really a lot, do you have a value from your experience that is lower but works?

marikaner (commented):
Cleanup looks good to me, I noticed that currently the samples are not really samples and I think this will not be very helpful to users. I believe it is ok to use the samples for the E2E tests, but the tests should not guide how we structure our samples, because the purpose is a different one: Samples are for users, tests are for automatic quality insurance.

tests/e2e-tests/src/utils/ai-api-utils.ts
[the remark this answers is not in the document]
shibeshduw: > I would suggest at least set a `maxTimeout`

I had originally implemented this using a standard timeout with 5 second intervals between retries before using the `async-retry` package. Think I might've just overlooked adding it after the change. 😅 

> 30 retries is really a lot, do you have a value from your experience that is lower but works?
From my experience, it takes anywhere from 45 secs - 105 secs to start or stop deployment. A little lower for delete. For the tests itself, each `it()` method has its own timeout so it won't be an issue either way. I'll test some more tomorrow with `factor` set to 1.

tests/e2e-tests/src/utils/cleanup-deployments.ts
shibeshduw: Just leftover after refactoring. This isn't required anymore. Thanks for pointing it out. :)

ZhongpinWang (commented):
LGTM

ZhongpinWang (approved):

PR #159, opened 2024-09-20
chore: Follow-up Documentation cleanup

KavithaSiva:
## Context

Follow-up of comments not addressed in:
- https://github.com/SAP/ai-sdk-js/pull/145
- https://github.com/SAP/ai-sdk-js/pull/134

## Definition of Done

- ~[ ] Code is tested (Unit, E2E)~
- ~[ ] Error handling created / updated & covered by the tests above~
- [x] Documentation updated
- ~[ ] (Optional) Aligned changes with the Java SDK~
- ~[ ] (Optional) Release notes updated~

PR #188, opened 2024-10-01
chore: Docs add model version and api version

ZhongpinWang:
## Context

AI/gen-ai-hub-sdk-js-backlog#139.

Add documentation for

 - setting `model_version` in orchestration
 - the used Azure OpenAI API version
 - setting `requestConfig` in
   - Azure OpenAI client
   - orchestration client

## Definition of Done

- [X] Documentation updated

### DIFF
--- a/packages/foundation-models/README.md
+++ b/packages/foundation-models/README.md
@@ -4,15 +4,15 @@ This package incorporates generative AI foundation models into your AI activitie
 
 ## Table of Contents
 
-1. [Installation](#installation)
-2. [Prerequisites](#prerequisites)
-3. [Usage](#usage)
-   - [Client Initialization](#client-initialization)
-   - [Azure OpenAI Client](#azure-openai-client)
-     - [Chat Client](#chat-client)
-     - [Embedding Client](#embedding-client)
-4. [Support, Feedback, Contribution](#support-feedback-contribution)
-5. [License](#license)
+- [Table of Contents](#table-of-contents)
+- [Installation](#installation)
+- [Prerequisites](#prerequisites)
+- [Usage](#usage)
+  - [Client Initialization](#client-initialization)
+  - [Chat Client](#chat-client)
+  - [Embedding Client](#embedding-client)
+- [Support, Feedback, Contribution](#support-feedback-contribution)
+- [License](#license)
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


packages/foundation-models/README.md
```diff
@@ -60,26 +60,42 @@ const chatClient = new AzureOpenAiChatClient({
...
 
-### Azure OpenAI Client
+### Chat Client
 
-The Azure OpenAI client can then be used to send chat completion or embedding requests to models deployed in the SAP generative AI hub.
+Use the `AzureOpenAiChatClient` to send chat completion requests to an OpenAI model deployed in SAP generative AI hub.
 
-#### Chat Client
+The client sends request with Azure OpenAI API version `2024-06-01`.
+Set a different version together with other request configuration in the `requestConfig` parameter.
```
deekshas8, line 68: [q] If user sets a newer version, are we sure that it works with our client? 
Can we instead say something like this client is generates using api version spec so and so, and is backward compatible (I imagine it is??).

packages/foundation-models/README.md
```diff
@@ -60,26 +60,42 @@ const chatClient = new AzureOpenAiChatClient({
...
+    messages: [
+      {
+        role: 'user',
+        content: 'Where is the deepest place on earth located'
+      }
+    ]
+  },
+  {
+    headers: {
+      'content-type': 'application/json',
```
deekshas8, line 87: @KavithaSiva already added a similar section in her PR. Please ensure we dont end up with dups

packages/foundation-models/README.md
ZhongpinWang: I will remove line 68. It also feels strange to have a different API version in the header, as our client is generated.

packages/foundation-models/README.md
ZhongpinWang: This falls under the scope of the issue. I had a look into the clean up PR and see a section documenting custom config only in orchestration. @KavithaSiva I think the easiest way is to remove the section you created in #159

packages/foundation-models/README.md
```diff
@@ -60,26 +60,40 @@ const chatClient = new AzureOpenAiChatClient({
...
```
deekshas8, line 86: [req] Aren't these headers already sent by default anyway? Can we just give an example of a custom dummy header? `content-type`  and `ai-client-type` are passed with these values, and `ai-resource-group` is meant to be set at the time of initialization. Setting it here is technically incorrect.

ZhongpinWang: True. Thanks for pointing it out.

deekshas8 (approved):
LGTM.

packages/orchestration/README.md
```diff
@@ -61,9 +81,20 @@ const orchestrationClient = new OrchestrationClient({
...
+  {
+    headers: {
+      // Add more headers here
+    },
+    params: {
+      // Add more parameters here
+    }
+    // Add more request configuration here
+  }
+);
```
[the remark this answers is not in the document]
ZhongpinWang: #191

PR #207, opened 2024-10-07
feat: Add license check

shibeshduw:
## Context

AI/gen-ai-hub-sdk-js-backlog#149.

[Moved](https://github.com/SAP/cloud-sdk-js/pull/5078) the license check script from cloud-sdk-js to a GH action and added the check to the build pipeline. 

## Definition of Done

- ~[ ] Documentation updated~


scripts/check-licenses.ts
marikaner, line 1: [pp] I assume this is stolen from the SAP Cloud SDK. Instead of duplicating it, I would prefer if we could reuse it through a gh action.

shibeshduw: Okay. Do we checkout the cloud-sdk-js from the workflow here or should I change the script in cloud-sdk-js to be a gh action instead and reuse that?

marikaner (approved):

PR #220, opened 2024-10-14
feat: Adjust embedding types and access

marikaner:
## Context

AI/gen-ai-hub-sdk-js-backlog#158.

According to the spec, I believe that the type in the embeddings output should allow for an arbitrary number of embeddings => https://github.com/SAP/ai-sdk-js/blob/461bf375f67787868ad6337ed574bb6f2cb97c4c/packages/foundation-models/src/azure-openai/spec/inference.yaml#L329.
This PR fixes the type + exposes all embeddings as a two-dimensional array of numbers.


marikaner:
covered by #221

PR #225, opened 2024-10-18
chore: Add a doc section for harmonised API

jjtang1985:
To highlight the harmonised API, we should mention this explicitly.


packages/orchestration/README.md
```diff
@@ -125,6 +125,28 @@ console.log(
 );
 ```
 
+> [!Tip]
+>
+> #### Harmonized API
+>
+> You can switch to a different model, even from another vendor, by leveraging the harmonized API of the orchestration service.
```
ZhongpinWang, line 132:
```suggestion
> As the orchestration service API is harmonized, you can switch to a different model, even from another vendor, by changing only the `model_name` property.
> All other configurations remain the same.
```

packages/orchestration/README.md
```diff
@@ -125,6 +125,28 @@ console.log(
...
+> Here’s an example where only one line of code is changed..
```
ZhongpinWang, line 133:
```suggestion
> Here’s an example where only one line of code is changed.
```

ZhongpinWang: [pp] Not sure if it is necessary to provide a complete example here just for changing a property. It feels hard to emphasize this advantage on written text unlike seeing a line change in a real demo.

ZhongpinWang (approved):
Minor suggestion

PR #230, opened 2024-10-21
chore: Add typescript as dev dependency in sub packages

jjtang1985:
Without adding this, I have an error `MODULE_NOT_FOUND`, when running `pnpm install --frozen-lockfile` on my command line with node 20.

I'll add it to all packages when approved.


jjtang1985:
After mega clean install (deleting the node modules, dist, etc), it's not reproducible.

PR #244, opened 2024-10-25
feat: Streaming azure openai

ZhongpinWang:
## Context

AI/gen-ai-hub-sdk-js-backlog#56.

Support streaming for `azure-openai` client in `foundation-models`.

Check `sample-code` for usage.

## Definition of Done

- ~[ ] Documentation updated~ Will be done in a separate PR
- [X] (Optional) Aligned changes with the Java SDK


packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream.ts
```diff
@@ -0,0 +1,129 @@
...
+
+const logger = createLogger({
+  package: 'foundation-models',
+  messageContext: 'azure-openai-chat-completion-stream'
+});
+
+/**
+ * Chat completion stream containing post-processing functions.
+ */
+// TODO: The Item type `any` should actually be the type of the stream chunk response.
```
MatKuhr, line 15: In what way are we dependent on the spec here?

The way we solved this in Java is that we only return Stream<String> for the simple variant. This has the downside that the token usage, finish reason etc isn't available, but it's much simpler, both in using and for our coding.

While the capability to get both, token usage and a stream of string , is nice, I'd argue that knowing the stream type is more important

packages/foundation-models/src/azure-openai/azure-openai-line-decoder.ts
```diff
@@ -0,0 +1,105 @@
+type Bytes = string | ArrayBuffer | Uint8Array | Buffer | null | undefined;
```
deekshas8, line 1: [req] I would put the streaming specific files into a separate folder.

packages/foundation-models/src/azure-openai/azure-openai-stream.ts
```diff
@@ -0,0 +1,203 @@
...
+          }
+
+          if (data && data.error) {
+            throw new Error(data.error);
+          }
+
+          yield data;
+        } else {
+          let data;
+          try {
```
deekshas8, line 58: [req] Noticed some code duplication. I would shorten it:

```
try {
  const data = JSON.parse(sse.data);
  
  if (data?.error && (sse.event === 'error' || sse.event === null)) {
    throw new Error(data.error);
  }
  
  yield sse.event === null 
    ? data 
    : { event: sse.event, data };
}
```

deekshas8 (changes requested):
Review still WIP

packages/foundation-models/src/azure-openai/azure-openai-stream.ts
```diff
@@ -0,0 +1,203 @@
...
+ * @internal
+ */
+export class Stream<Item> implements AsyncIterable<Item> {
+  protected static fromSSEResponse<Item>(response: HttpResponse): Stream<Item> {
+    let consumed = false;
+
+    async function* iterator(): AsyncIterator<Item, any, undefined> {
+      if (consumed) {
+        throw new Error(
+          'Cannot iterate over a consumed stream, use `.tee()` to split the stream.'
```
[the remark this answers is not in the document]
ZhongpinWang: It was a leftover, removed.

packages/foundation-models/src/azure-openai/azure-openai-line-decoder.ts
ZhongpinWang: Moved.

packages/foundation-models/src/azure-openai/azure-openai-stream.ts
```diff
@@ -0,0 +1,203 @@
...
+  messageContext: 'azure-openai-stream'
+});
+
+type Bytes = string | ArrayBuffer | Uint8Array | Buffer | null | undefined;
+
+/**
+ * Stream implemented as an async iterable.
+ * @internal
+ */
+export class Stream<Item> implements AsyncIterable<Item> {
```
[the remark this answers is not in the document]
ZhongpinWang: Done

packages/foundation-models/src/azure-openai/azure-openai-stream.ts
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        if (sse.event === null) {
+          let data;
+
+          try {
+            data = JSON.parse(sse.data);
+          } catch (e: any) {
+            logger.error(`Could not parse message into JSON: ${sse.data}`);
+            logger.error(`From chunk: ${sse.raw}`);
```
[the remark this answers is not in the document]
ZhongpinWang: I think openai splits them into two because `sse.data` and `sse.raw` can be huge text block. This makes the log more readable.

packages/foundation-models/src/azure-openai/azure-openai-stream.ts
```diff
@@ -0,0 +1,203 @@
...
+  }
+}
+
+/**
+ * @internal
+ */
+export async function* _iterSSEMessages(
+  response: HttpResponse
+): AsyncGenerator<ServerSentEvent, void, unknown> {
+  if (!response.data) {
```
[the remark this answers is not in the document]
ZhongpinWang: I am actually in favour of this style, putting it within the function as this can be seen as a precondition / contract of this function. Otherwise, moving it into `iterator()` makes that function even longer.

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-chunk-response.test.ts
```diff
@@ -0,0 +1,77 @@
...
+import { AzureOpenAiChatCompletionStreamChunkResponse } from './azure-openai-chat-completion-stream-chunk-response.js';
+
+describe('OpenAI chat completion stream chunk response', () => {
+  let mockResponses: {
+    tokenUsage: any;
+    finishReason: any;
+    deltaContent: any;
+  };
+  let azureOpenAiChatCompletionStreamChunkResponses: {
+    tokenUsage: AzureOpenAiChatCompletionStreamChunkResponse;
```
deekshas8, line 11: [q] Why are each of these properties of type `AzureOpenAiChatCompletionStreamChunkResponse`. Should they be part of this type?

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-response.ts
```diff
@@ -0,0 +1,44 @@
+import type { AzureOpenAiCompletionUsage } from './client/inference/schema/index.js';
+import type { AzureOpenAiChatCompletionStream } from './azure-openai-chat-completion-stream.js';
+
+/**
+ * Azure OpenAI chat completion stream response.
+ */
+export class AzureOpenAiChatCompletionStreamResponse<T> {
+  private _usage: AzureOpenAiCompletionUsage | undefined;
```
deekshas8, line 8: [q] Aren't these properties duplicated (I see the're part of the chunk response type as well)

packages/foundation-models/src/azure-openai/azure-openai-stream.ts
ZhongpinWang: Thanks for this! I will also always throw error if `data?.error` and no matter which `sse.event` or if it is null.

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream.ts
ZhongpinWang: `chunk` is now typed.

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-chunk-response.test.ts
ZhongpinWang: I think this is a bit confusing. What I want is to have three chunk responses, and `tokenUsage` is like one of the chunk response which has a token usage property.

deekshas8: I would update this to make it easy to understand. In the current state it is a bit misleading

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-chunk-response.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    ).toStrictEqual(mockResponses.finishReason);
+    expect(
+      azureOpenAiChatCompletionStreamChunkResponses.deltaContent.data
+    ).toStrictEqual(mockResponses.deltaContent);
+  });
+
+  it('should get token usage', () => {
+    expect(
+      azureOpenAiChatCompletionStreamChunkResponses.tokenUsage.getTokenUsage()
+    ).toMatchObject({
```
deekshas8, line 59: 👀

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-response.ts
deekshas8: I have strong reservations about introducing the additional wrapper `AzureOpenAiChatCompletionStreamResponse` with dup properties.
I my view, we should keep the api response simple, so just a stream response of chunks (without the wrapper). `streamContent` is already simplified/additional convenience.

If the user wants to access `finish_reason` or `usage`, we already provide the util functions in the `AzureOpenAiChatCompletionStreamChunkResponse`.
Given this, the added complexity from `AzureOpenAiChatCompletionStreamResponse` and the multiple `*process` functions does not seem worth it to me.

@MatKuhr @tomfrenken @KavithaSiva I would be interested in your thoughts

ZhongpinWang: Thanks for this feedback. There are some thoughts behind introducing the wrapper.

1. Adding this wrapper or not, the difference to the user to access the chunk is just either using `for await (const chunk of response.stream)` or `for await (const chunk of response)`. This won't make the consumption more difficult.
2. The original reason for adding this is that we want to support `streamContent()`. I discussed this with @MatKuhr.  In the current Java implementation, string stream is called `streamChatCompletion()` and has a very important role by default. When user calls this method, they do not have direct access to the original raw chunk returned by OpenAI, i.e., user has **no access** to `finish_reason` or especially `token_usage`. In Java, when user wants this information, they need to consume `Stream` object by themselves. Or things like `content_filter`, they throw exceptions so that user will get notified. But we don't wanna throw errors in JS as this is inconsistent with our non-streaming api and will also stop streaming immediately. That is why I added this outside wrapper for the whole streaming, to let user at least have access to `finish_reason` and `token_usage` when using `streamContent()`.
3. Having `AzureOpenAiChatCompletionStreamResponse` or not, we need these `*process` functions for other things like wrap original OpenAI chunk into our util wrapper `AzureOpenAiChatCompletionStreamChunkResponse` or transform further into string stream, not only for `token_usage` or `finish_reason`. 
4. There will be no error handling from us for different `finish_reason`, like when content filter stopped the stream, SDK will never know. This means each and every user literally need to implement themselves a big switch case, check for each chunk response and do error handling by themselves. One could argue that we have some util function implemented and expose them to user for checking `finish_reason`, but user has to know to use this function. When we know these checks are necessary and can do this in-place inside our SDK, why let user do this. And do notice that it is not even possible when using `streamContent()`. 

I hope this can clarify a bit the motivation behind and let me know if you see it differently.

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-chunk-response.test.ts
ZhongpinWang: Indeed misleading. I changed them. Thanks!

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-chunk-response.test.ts
ZhongpinWang: 😄 resolved. No more get usage from usage.

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-response.ts
deekshas8: Thanks for explaining the thought behind the wrapper class.

As discussed, I would suggestion we don't really need a separate streamContent() api, but rather provide a convenience method in the stream response object returned by `.stream()`. The user can call it if they need it.

```
  async *processContent(
    this: AzureOpenAiChatCompletionStream<AzureOpenAiChatCompletionStreamChunkResponse>
  ): AsyncGenerator<string> {
   ...
  }
```
The above change will allow calling it on stream object.
Also maybe renaming it to get instead of process?

packages/foundation-models/src/azure-openai/stream/line-decoder.ts
```diff
@@ -0,0 +1,105 @@
+type Bytes = string | ArrayBuffer | Uint8Array | Buffer | null | undefined;
+
+/**
+ * A re-implementation of httpx's `LineDecoder` in Python that handles incrementally
+ * reading lines from text.
+ *
+ * Https://github.com/encode/httpx/blob/920333ea98118e9cf617f246905d7b202510941c/httpx/_decoders.py#L258.
```
MatKuhr, line 7: Is there no good JS lib that does this already?

test-util/data/foundation-models/azure-openai-chat-completion-stream-chunk-response-token-usage.json
```diff
@@ -0,0 +1 @@
+{"choices":[],"created":1730125149,"id":"chatcmpl-ANKsHIdjvozwuOGpGI6rygvwSJH0I","model":"gpt-35-turbo","object":"chat.completion.chunk","system_fingerprint":"fp_808245b034","usage":{"completion_tokens":7,"prompt_tokens":14,"total_tokens":21}}
```
MatKuhr, line 1: We can also get an error during streaming. See [here for an example](https://github.com/SAP/ai-sdk-java/blob/main/foundation-models/openai/src/test/resources/streamChatCompletionError.txt).

We should probably add a test for this case, ensuring we throw in such cases..

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-response.ts
```diff
@@ -0,0 +1,44 @@
...
+      throw new Error('Response stream is undefined.');
+    }
+    return this._usage;
+  }
+
+  public set usage(usage: AzureOpenAiCompletionUsage) {
+    this._usage = usage;
+  }
+
+  public get finishReason(): string {
```
MatKuhr, line 23: We could consider:

```suggestion
  public get finishReason(index = 0): string {
```

Technically, the finish reason might differ for different indices. But I doubt this is a frequent use case and we could always add this later, as this would be a none-breaking change, right? If so, I'd leave it out for now..

packages/foundation-models/src/azure-openai/stream/line-decoder.ts
ZhongpinWang: Not that I know, maybe @deekshas8 you have seen something like this? I mean I suppose OpenAI implement this due to some good reason.

deekshas8: I am only aware of this https://nodejs.org/api/readline.html (this handles the line endings). It expects a stream as input though. I am also not sure if it handles any other custom things that we might be doing in LineDecoder.

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-response.ts
ZhongpinWang: Done

packages/foundation-models/src/azure-openai/stream/line-decoder.ts
ZhongpinWang: In this PR, I would save the effort since it works and not touch this fundamental part.

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-response.ts
deekshas8: This was probably a wrong assumption then. Atleast for me, the docs were not explicitly clear on this. @MatKuhr can you please confirm? We can fix that in a separate PR.

MatKuhr: In Java we don't consider n>1 for the convenience API. If users want to access anything with index != 0, they have to do it "manually". See [this code](https://github.com/SAP/ai-sdk-java/blob/main/foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/model/OpenAiChatCompletionDelta.java#L32).

ZhongpinWang: Issue created -> https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/173

test-util/data/foundation-models/azure-openai-chat-completion-stream-chunk-response-token-usage.json
ZhongpinWang: I now throw an error if sse does not start with `data` or `event`. In the other case like the above link, the sse decoder treats both json as one string / one payload, and we will throw anyways while doing JSON parse in `SseStream.fromSSEResponse()`.

ZhongpinWang: I now throw an error if a new sse does not start with `data` or `event`. In case of two lines been put together like the link above, the sse decoder treats them as one single payload as there is no empty line separating them. In this case, we will throw while doing JSON parse in sse-stream.ts. Test case is added in sse-stream.test.ts.

sample-code/src/server.ts
```diff
@@ -51,6 +53,100 @@ app.get('/azure-openai/chat-completion', async (req, res) => {
...
+      connectionAlive = false;
+      res.end();
+    });
+
+    for await (const chunk of response.stream.toContentStream()) {
+      if (!connectionAlive) {
+        break;
+      }
+      res.write(chunk);
+    }
```
jjtang1985, line 80: [pp] seems to be some reusable code :)

packages/foundation-models/src/azure-openai/stream/sse-stream.ts
```diff
@@ -0,0 +1,206 @@
...
+  messageContext: 'azure-openai-sse-stream'
+});
+
+type Bytes = string | ArrayBuffer | Uint8Array | Buffer | null | undefined;
+
+/**
+ * Stream implemented as an async iterable.
+ * @internal
+ */
+export class SseStream<Item> implements AsyncIterable<Item> {
```
jjtang1985, line 18: [q] What is sse? Server side event?

sample-code/src/server.ts
```diff
@@ -51,6 +53,100 @@ app.get('/azure-openai/chat-completion', async (req, res) => {
...
+    let connectionAlive = true;
+
+    res.on('close', () => {
+      controller.abort();
+      connectionAlive = false;
+      res.end();
+    });
+
+    for await (const chunk of response.stream.toContentStream()) {
+      if (!connectionAlive) {
```
jjtang1985, line 76: [q] what will happen, if we dont check this?
Writing empty chunks forever?
I would assume, after getting the full response, the loop should be stopped automatically?

sample-code/src/server.ts
```diff
@@ -51,6 +53,100 @@ app.get('/azure-openai/chat-completion', async (req, res) => {
...
+        break;
+      }
+      res.write(chunk);
+    }
+
+    if (connectionAlive) {
```
jjtang1985, line 82: [q] Why this is needed?

jjtang1985 (commented):
I left some questions.

packages/foundation-models/src/azure-openai/stream/sse-stream.ts
ZhongpinWang: Almost, server sent event -> https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events

sample-code/src/server.ts
ZhongpinWang: There will be an error if user closes the connection by e.g. refreshing the page and we still try to send something.

sample-code/src/server.ts
ZhongpinWang: Same as before. This value is set to false if receive `close` event. We should then avoid sending anything new

sample-code/src/server.ts
ZhongpinWang: I removed the multiple choice example. I will mention this feature in the documentation later

jjtang1985 (approved):
I reviewed the API from the consumer perspective.
LGTM.

packages/foundation-models/src/azure-openai/stream/line-decoder.ts
```diff
@@ -0,0 +1,105 @@
...
+ * @internal
+ */
+export class LineDecoder {
```
KavithaSiva, line 10: [q] Wouldn't the `line-decoder`, `sse-stream` and `sse-decoder` be eventually used for Orchestration streaming as well? Can we move them to `core` package instead?

packages/foundation-models/src/azure-openai/azure-openai-chat-client.test.ts
```diff
@@ -159,4 +160,46 @@ describe('Azure OpenAI chat client', () => {
...
+      messages: [
+        {
+          role: 'user' as const,
+          content: 'Where is the deepest place on earth located'
+        }
+      ],
+      stream: true,
+      stream_options: {
+        include_usage: true
+      }
```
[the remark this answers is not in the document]
ZhongpinWang: Yes they will be automatically added. But we need to define this in the test to provide a mock, see `mockInference`.

packages/foundation-models/src/azure-openai/azure-openai-chat-client.ts
```diff
@@ -28,12 +32,43 @@ export class AzureOpenAiChatClient {
...
+    const response = await this.executeRequest(data, requestConfig);
+    return new AzureOpenAiChatCompletionResponse(response);
+  }
+
+  /**
+   * Creates a completion stream for the chat messages.
+   * @param data - The input parameters for the chat completion.
+   * @param controller - The abort controller.
+   * @param requestConfig - The request configuration.
+   * @returns The completion stream.
```
ZhongpinWang, line 44:
```suggestion
   * @returns A response containing the chat completion stream.
```

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-chunk-response.ts
```diff
@@ -0,0 +1,39 @@
...
+  constructor(public readonly data: any) {
+    // TODO: Change `any` to `CreateChatCompletionStreamResponse` once the preview spec becomes stable.
+    this.data = data;
+  }
+
+  /**
+   * Usage of tokens in the chunk response.
+   * @returns Token usage.
+   */
+  getTokenUsage(): AzureOpenAiCompletionUsage {
```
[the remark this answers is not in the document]
ZhongpinWang: True. I fixed it. Thanks

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-response.ts
```diff
@@ -0,0 +1,54 @@
+import type { AzureOpenAiCompletionUsage } from './client/inference/schema/index.js';
+import type { AzureOpenAiChatCompletionStream } from './azure-openai-chat-completion-stream.js';
+
+/**
+ * Azure OpenAI chat completion stream response.
+ */
+export class AzureOpenAiChatCompletionStreamResponse<T> {
+  private _usage: AzureOpenAiCompletionUsage | undefined;
+  private _finishReasons: Map<number, string> = new Map();
```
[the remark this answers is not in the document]
ZhongpinWang: Yes.
I added the comment

packages/foundation-models/src/azure-openai/stream/line-decoder.ts
ZhongpinWang: This will be part of the orchestration streaming PR. As only then, we know how orchestration service behaves and maybe certain modifications will be needed.

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream-chunk-response.ts
```diff
@@ -0,0 +1,39 @@
...
+    return this.data.choices.find((c: any) => c.index === choiceIndex)
+      ?.finish_reason;
+  }
+
+  /**
+   * Parses the chunk response and returns the delta content.
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns The message delta content.
+   */
+  getDeltaContent(choiceIndex = 0): string | undefined | null {
```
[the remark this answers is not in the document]
ZhongpinWang: It can be `null` according to the spec -> https://github.com/Azure/azure-rest-api-specs/blob/60c67f112c66537f1d006bf1b497857ccd2afacd/specification/cognitiveservices/data-plane/AzureOpenAI/inference/stable/2024-10-21/inference.yaml#L1985C11-L1985C25

packages/foundation-models/src/azure-openai/stream/line-decoder.ts
KavithaSiva: Hmm, looks like [they will](https://github.tools.sap/AI/llm-orchestration/blob/main/docs/adr/008-output-streaming.md) also use SSE.

But, since, we don't expose these types to users anyway, you are right, we can move them later too.

packages/foundation-models/src/azure-openai/stream/sse-stream.ts
tomfrenken, line 18: General comment, since the complexity of this class is very high, even if multiple functions are only @internal, they should still have proper typedoc.

packages/foundation-models/src/azure-openai/stream/sse-stream.ts
```diff
@@ -0,0 +1,206 @@
...
+    data = newData;
+
+    let patternIndex;
+    while ((patternIndex = findDoubleNewlineIndex(data)) !== -1) {
+      yield data.slice(0, patternIndex);
+      data = data.slice(patternIndex);
+    }
+  }
+
+  if (data.length > 0) {
```
tomfrenken, line 156:
```suggestion
  if (data.length) {
```

tomfrenken (changes requested):
Almost done with everything, will review a little bit more on Monday, but should be fine from what I've already seen.

Looks really good for the most part, just some styleguide issues here and there, also some missing typedoc makes it rather hard to review the different intricacies of all the parsing logic.

packages/foundation-models/src/azure-openai/stream/sse-decoder.ts
```diff
@@ -0,0 +1,95 @@
...
+      this.event = null;
+      this.data = [];
+      this.chunks = [];
+
+      return sse;
+    }
+
+    this.chunks.push(line);
+
+    if (line.startsWith(':')) {
```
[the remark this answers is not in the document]
ZhongpinWang: This is implementing server sent event -> https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events

I will put this link in the file.

packages/foundation-models/src/azure-openai/stream/sse-stream.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    async function* iterator(): AsyncIterator<Item, any, undefined> {
+      if (consumed) {
+        throw new Error('Cannot iterate over a consumed stream.');
+      }
+      consumed = true;
+      let done = false;
+
+      try {
+        for await (const sse of _iterSSEMessages(response, controller)) {
```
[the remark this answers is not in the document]
ZhongpinWang: Done.

packages/foundation-models/src/azure-openai/stream/line-decoder.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  buffer: string[];
+  trailingCR: boolean;
+  textDecoder: any; // TextDecoder found in browsers; not typed to avoid pulling in either "dom" or "node" types.
+
+  constructor() {
+    this.buffer = [];
+    this.trailingCR = false;
+  }
+
```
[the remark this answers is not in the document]
ZhongpinWang: I want to avoid changing too much of OpenAI's code, in case in the future, they updated something and we try to sync the changes, refactoring will make everything a pain.

packages/foundation-models/src/azure-openai/stream/line-decoder.ts
```diff
@@ -0,0 +1,105 @@
...
+  decodeText(bytes: Bytes): string {
+    if (bytes == null) {
+      return '';
+    }
+    if (typeof bytes === 'string') {
+      return bytes;
+    }
+
+    // Node:
+    if (typeof Buffer !== 'undefined') {
```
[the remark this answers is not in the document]
ZhongpinWang: Originally, OpenAI also supports browser and some other shims. I removed them as we only support nodejs. But I feel like this check is still valid.

packages/foundation-models/src/azure-openai/stream/sse-stream.ts
ZhongpinWang: Done

packages/foundation-models/src/azure-openai/stream/line-decoder.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (trailingNewline) {
+      lines.pop();
+    }
+
+    if (lines.length === 1 && !trailingNewline) {
+      this.buffer.push(lines[0]!);
+      return [];
+    }
+
+    if (this.buffer.length > 0) {
```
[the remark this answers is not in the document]
ZhongpinWang: Done

packages/foundation-models/src/azure-openai/stream/line-decoder.ts
```diff
@@ -0,0 +1,105 @@
...
+    );
+    let lines = text.split(LineDecoder.NEWLINE_REGEXP);
+
+    // if there is a trailing new line then the last entry will be an empty
+    // string which we don't care about
+    if (trailingNewline) {
+      lines.pop();
+    }
+
+    if (lines.length === 1 && !trailingNewline) {
```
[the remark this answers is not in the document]
ZhongpinWang: Yes, this line is discarded. According to the sse guideline, new line is always required to fire some data. Otherwise, only one line and no new line results into an "invalid" data which will be empty. Check https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events and search for new line or blank line.

packages/foundation-models/src/azure-openai/stream/line-decoder.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    const trailingNewline = LineDecoder.NEWLINE_CHARS.has(
+      text[text.length - 1] || ''
+    );
+    let lines = text.split(LineDecoder.NEWLINE_REGEXP);
+
+    // if there is a trailing new line then the last entry will be an empty
+    // string which we don't care about
+    if (trailingNewline) {
```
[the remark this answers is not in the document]
ZhongpinWang: My understanding is that the second last new line may indeed be part of the payload?

packages/foundation-models/src/azure-openai/stream/line-decoder.ts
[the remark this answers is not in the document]
tomfrenken: Yeah makes sense, thanks

packages/foundation-models/src/azure-openai/stream/sse-decoder.ts
```diff
@@ -0,0 +1,95 @@
...
+      return null;
+    }
+
+    const [fieldname, , value] = partition(line, ':');
```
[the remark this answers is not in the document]
tomfrenken: I see, I would be in favor of adding an exception for `_` to lint, as it is a common best practice to name unused variables `_` if they are unused intentionally.

e.g. add this to the config:
```
  "@typescript-eslint/no-unused-vars": ["error", { "varsIgnorePattern": "^_" }]
  ```
  What do you think?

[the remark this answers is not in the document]
Comment (ZhongpinWang) [reply packages/foundation-models/src/azure-openai/stream/sse-decoder.ts:70]: Makes our life easier! I added this. 👍

File: packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream.ts [hunk 33]
```diff
@@ -0,0 +1,170 @@
...
+   * @internal
+   */
+  static async *_processFinishReason(
+    stream: AzureOpenAiChatCompletionStream<AzureOpenAiChatCompletionStreamChunkResponse>,
+    response?: AzureOpenAiChatCompletionStreamResponse<any>
+  ): AsyncGenerator<AzureOpenAiChatCompletionStreamChunkResponse> {
+    for await (const chunk of stream) {
+      chunk.data.choices.forEach((choice: any) => {
+        const choiceIndex = choice.index;
+        if (choiceIndex !== undefined && choiceIndex !== null) {
```
tomfrenken, line 54:
```suggestion
        if (choiceIndex >= 0) {
```

tomfrenken (approved):
LGTM!

PR #254, opened 2024-10-30
chore: Reuse check-public-api script from Cloud SDK GH action

shibeshduw:
## Context

AI/gen-ai-hub-sdk-js-backlog#115.

## Definition of Done

- ~[ ] Documentation updated~


.github/workflows/build.yml
```diff
@@ -32,8 +32,13 @@ jobs:
...
       - name: Check CommonJS compatibility
         run: pnpm compile:cjs
+      - name: Set NODE_PATH
+        run: echo "NODE_PATH=$(pwd)/node_modules" >> $GITHUB_ENV
       - name: Check public API
-        run: pnpm check:public-api
+        uses: sap/cloud-sdk-js/.github/actions/check-public-api@main
+        with:
+          force_internal_exports: 'false'
+          ignored_paths: 'client/inference/schema, client/api/schema'
```
ZhongpinWang, line 41: [req] `ai-api` client is not excluded here.

[pp] Can we still support regex or maybe glob pattern here? This list otherwise will increase once we have more and more clients

ZhongpinWang (changes requested):

.github/workflows/build.yml
ZhongpinWang: [pp] Also I find it not elegant to use comma to separate paths instead of using yaml list syntax. 

```yaml
with:
    ignored_paths:
        - client/inference/schema
        - client/api/schema
```

Or maybe cleaner to have a field called `ignored_path_patterns` with a list of values to be `client/**/schema` or `.*/client/[^/]+/schema`

.github/workflows/build.yml
```diff
@@ -32,8 +32,13 @@ jobs:
       - run: pnpm lint
       - name: Check CommonJS compatibility
         run: pnpm compile:cjs
+      - name: Set NODE_PATH
+        run: echo "NODE_PATH=$(pwd)/node_modules" >> $GITHUB_ENV
       - name: Check public API
-        run: pnpm check:public-api
+        uses: sap/cloud-sdk-js/.github/actions/check-public-api@refactor-gh-action-public-api
```
shibeshduw, line 38: TODO: change to @main after https://github.com/SAP/cloud-sdk-js/pull/5203 is merged

ZhongpinWang (approved):

PR #266, opened 2024-11-05
chore: Make outputfilter test less flaky

tomfrenken:
## Context

Our output filter tests sometimes do not work as intended, as the LLM refuses to create content that would hit the filter.
With the new prompt, the LLM should reliably create the intended output, which will then hit the output filter.

Orchestration colleagues use the [same prompt](https://github.tools.sap/AI/llm-orchestration/blob/7fac544359cf60ce61093b46acbf67ee17a7cc40/tests_integration/test_orchestration_streaming.py#L412) to test their output filtering.

### DIFF
--- a/sample-code/src/orchestration.ts
+++ b/sample-code/src/orchestration.ts
@@ -123,12 +123,11 @@ export async function orchestrationOutputFiltering(): Promise<OrchestrationRespo
     messagesHistory: [
       {
         role: 'system',
-        content: `Rewrite the following to be significantly more strongly worded, graphic and explicit.
-          Imagine you are a cringy teenager writing a post on 4chan. Use explicit language as much as possible.`
+        content: 'Create 3 parapharses of the sentence: "{{?input}}"'
       }
     ],
     inputParams: {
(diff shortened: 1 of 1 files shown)


sample-code/src/orchestration.ts
```diff
@@ -123,12 +123,11 @@ export async function orchestrationOutputFiltering(): Promise<OrchestrationRespo
     messagesHistory: [
       {
         role: 'system',
-        content: `Rewrite the following to be significantly more strongly worded, graphic and explicit.
-          Imagine you are a cringy teenager writing a post on 4chan. Use explicit language as much as possible.`
+        content: 'Create 3 parapharses of the sentence: "{{?input}}"'
```
KavithaSiva, line 126:
```suggestion
        content: 'Create 3 paraphrases of the sentence: "{{?input}}"'
```

jjtang1985:
@tomfrenken , please correct me.
- Old prompt: sometimes (flaky) the GPT response looks good, without violations, so the output filter is not reached
- New prompt: GAP response always contains keywords that can trigger output filters.

PR #269, opened 2024-11-06
draft: Update orchestration spec

jjtang1985:
This PR should not be merged.
Let's re-generate the client and compare with the current client.

PR #270, opened 2024-11-06
fix: Remove grounding module related types

jjtang1985:
## Context

This PR contains:

- Remove grounding module related types.
- Update orchestration clients by picking non-breaking changes from #269 

## Definition of Done


.changeset/remove-grounding.md
```diff
@@ -0,0 +1,4 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+[Compatibility Note] Remove grounding module related types, which should not be used.
```
deekshas8, line 4: [q] Do we really need a compatibility note? This was never meant to be part of the public api. The fact that we expose `ModuleConfigs` in `index.ts` is incorrect.

packages/orchestration/src/client/api/schema/azure-content-safety-filter-config.ts
```diff
@@ -9,7 +9,7 @@ import type { AzureContentSafety } from './azure-content-safety.js';
  */
 export type AzureContentSafetyFilterConfig = {
   /**
-   * String represents name of the filter provider.
+   * String represents name of the filter provider
```
deekshas8, line 12: [pp] I think these changes are fine for now. But it's better to have proper linted code. I would like to adjust our eslint config to fix such issues (and only disable consistent-typedefinitions rule).

deekshas8 (approved):
LGTM. Please consider the comment about having the changeset.

.changeset/remove-grounding.md
jjtang1985: For SDK user, I agree, it's not relevant, if they only use public APIs, so I deleted the compatibility notes.
I put it there, to make it transparent for those who noticed our grounding related code.

jjtang1985: will work on `ModuleConfigs` in a different PR.

jjtang1985:
> Please consider the comment about having the changeset.

Removed already 😄

PR #276, opened 2024-11-07
chore: Update schema for templates

jjtang1985:
## Context

This PR introduces a new schema for template, which is not a breaking change.

## Definition of Done

### DIFF
--- a/packages/orchestration/src/client/api/schema/filtering-module-config.ts
+++ b/packages/orchestration/src/client/api/schema/filtering-module-config.ts
@@ -3,18 +3,18 @@
  *
  * This is a generated file powered by the SAP Cloud SDK for JavaScript.
  */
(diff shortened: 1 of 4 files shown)


packages/orchestration/src/client/api/schema/filtering-module-config.ts
```diff
@@ -3,18 +3,18 @@
  *
  * This is a generated file powered by the SAP Cloud SDK for JavaScript.
  */
-import { InputFilteringConfig } from './input-filtering-config.js';
-import { OutputFilteringConfig } from './output-filtering-config.js';
+import type { InputFilteringConfig } from './input-filtering-config.js';
```
jjtang1985, line 6: leftover from the filter PR

deekshas8 (approved):

PR #290, opened 2024-11-13
fix: Get by index from response wrapper

ZhongpinWang:
## Context

AI/gen-ai-hub-sdk-js-backlog#173.

Previously, if choice index is defined, we use the array index to fetch the choice. But the assumption that nth item in an array is the nth choice is invalid. We should use `index` in the choice instead.

## Definition of Done

- ~[ ] Documentation updated~


tomfrenken (approved):
LGTM!

PR #300, opened 2024-11-18
feat: RAGe client

KavithaSiva:
## Context

AI/gen-ai-hub-sdk-js-backlog#176.

This PR adds a dedicated package for generated rage client. The vector and retrieval clients are also E2E tested for a scenario of creating a document collection and querying the created collection.
The package is currently set to private, as we do not plan to release it yet.

The generated clients have been manually modified for now, to target the correct API endpoint.
This workaround would be removed after https://github.com/SAP/cloud-sdk-backlog/issues/1259.

The most relevant changes to be reviewed are in `sample-code`, `tests/e2e-tests`.

## Definition of Done

- ~[] Error handling created / updated & covered by the tests above~
- ~[ ] Documentation updated~


packages/rage/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './client/pipeline/pipeline-api/index.js';
+export * as VectorApi from './client/vector/vector-api/index.js';
+export * as RetrievalApi from './client/retrieval/retrieval-api/index.js';
```
jjtang1985, line 3:
```suggestion
```

jjtang1985, line 3: By checking the log of the public API check, I did not see anything for pipeline api.
I would assume it's because of the alias of vector api and retrival api.
After merging the spec, it should be solved.

jjtang1985 (commented):
LGTM, as a PoC.

packages/rage/src/index.ts
shibeshduw: [q] What is the reason for the folder structure to be `pipeline/pipeline-api` instead of just `pipeline-api` ? We should follow the same naming convention as the other packages i.e. client/**/schema. That way, all generated files will get ignored by the public api check because we are passing [this regex](https://github.com/SAP/ai-sdk-js/blob/89b611993becbdc3f66dc0e3f10f11ad6b797b8e/.github/workflows/build.yml#L41) in the GH action.

KavithaSiva: The reason we need an additional nested structure is because we also always clear the output folder every time we generate a client and I wanted to ensure that the correct client gets deleted. Deletion is required so that we have a clean slate, and get to know if there are any breaking changes in the client during generation.

The current script looks like:

```diff
+ openapi-generator --generateESM --clearOutputDir -i ./src/spec/vector-api.yaml -o ./src/client/vector
- openapi-generator --generateESM --clearOutputDir -i ./src/spec/vector-api.yaml -o ./src/client/
```

As the generator always creates a new folder under the output directory specified, if we don't have a nested structure and use just `/src/client/`, all clients in the `/src/client/` folder will get cleared.

tomfrenken: An alternative would be to change the generator script to move the folder after generation, then the folder structure would look as expected, however, then we would need to chain a few more commands together, e.g.:
`rm -rf ./src/client/vector-api && pnpm generate:vector && mv ./src/client/vector/vector-api  ./src/client && rm ./src/client/vector`, at that point it makes more sense to change the update the ignore pattern, I agree

KavithaSiva: >I would assume it's because of the alias of vector api and retrival api.
>After merging the spec, it should be solved.

Aliases, are needed as there are duplicates in the spec.
But, aliases are currently not considered in the public API check.

I can selectively export only some of the API, schema types from each client to work around this problem for now.

KavithaSiva: @jjtang1985 The public API check GH action currently doesn't support aliases while using exports. 
To enable the build to succeed, I now, selectively exported only some types and API that I use for my E2E test.

If the spec isn't merged soon, and we need aliases to export API and all types successfully, we might then have to work on considering them in the public API check.

Also, note before merging, NOT all types are exported from the Vector and Search API.

deekshas8:
@KavithaSiva @jjtang1985 Can this PR be closed? I see another PR for document-grounding.

KavithaSiva:
> @KavithaSiva @jjtang1985 Can this PR be closed? I see another PR for document-grounding.

@deekshas8 Yes, closing this, as alternate PR is available: https://github.com/SAP/ai-sdk-js/pull/335

PR #325, opened 2024-11-28
chore: Ignore nested folders in generated client in public API check

KavithaSiva:
## Context

The public API check ignores checking the generated client as it currently ignores paths matching `.*?/client/[^/]+/schema`

For the RAGe client, we had to introduce a nested folder structure.
Hence, updating the regex for public API check, to also ignore nested folder structures in the generated client.

## Definition of Done

- [x] Regex was tested in regex evaluator
- ~[ ] Documentation updated~


shibeshduw (approved):
Also tested using regex101. LGTM.

PR #333, opened 2024-11-29
feat: Update to Orchestration 0.36.1

MatKuhr:
## Context

Notable changes in the spec:

- TemplateRef
- Grounding

## Definition of Done

- [x] ~Documentation updated~
- [x] ~(Optional) Release notes updated~


packages/foundation-models/src/azure-openai/client/inference/schema/audio-response.ts
```diff
@@ -5,7 +5,7 @@
  */
 
 /**
- * Translation or transcription response when response_format was json.
+ * Translation or transcription response when response_format was json
```
MatKuhr, line 8: Not sure where these changes are coming from TBH

packages/orchestration/src/orchestration-response.ts
```diff
@@ -21,7 +21,7 @@ export class OrchestrationResponse {
    * @returns Token usage.
    */
   getTokenUsage(): TokenUsage {
-    return this.data.orchestration_result.usage;
+    return this.data.orchestration_result.usage!;
```
MatKuhr, line 24: This is fine to do because this is only undefined in case of streaming. To be fixed in the spec

packages/foundation-models/src/azure-openai/client/inference/schema/citation.ts
```diff
@@ -5,7 +5,7 @@
  */
 
 /**
- * Citation information for a chat completions response message.
+ * citation information for a chat completions response message.
```
ZhongpinWang, line 8: Question mark. 👀

packages/foundation-models/src/azure-openai/client/inference/schema/azure-chat-extension-configuration.ts
```diff
@@ -6,9 +6,9 @@
 import type { AzureOpenAiAzureSearchChatExtensionConfiguration } from './azure-search-chat-extension-configuration.js';
 import type { AzureOpenAiAzureCosmosDBChatExtensionConfiguration } from './azure-cosmos-db-chat-extension-configuration.js';
 /**
- * A representation of configuration data for a single Azure OpenAI chat extension. This will be used by a chat
- * completions request that should use Azure OpenAI chat extensions to augment the response behavior.
- * The use of this configuration is compatible only with Azure OpenAI.
+ *   A representation of configuration data for a single Azure OpenAI chat extension. This will be used by a chat
```
ZhongpinWang, line 9: Question mark. 👀 What happened in the doc. Why do we have indentation here..

ZhongpinWang:
Should we wait until the next released spec version, where they remove `help.sap.com`?

MatKuhr:
> Should we wait until the next released spec version, where they remove `help.sap.com`?

Updated 👍🏻

ZhongpinWang (changes requested):
[req] Why are we regenerating azure openai client? I think maybe you hit the `pnpm generate`. Not sure why these changes are there but since they are all ugly, maybe we should run `pnpm orchestration generate` instead to also keep this PR clean?

MatKuhr:
Yeah I just ran generate, so the azure changes were a side effect of the pipeline currently not enforcing re-generation. Reverted now.

ZhongpinWang:
[q] Do we still need to exclude template ref?

MatKuhr:
> [q] Do we still need to exclude template ref?

What do you mean? It currently isn't exported because it's not tested..

ZhongpinWang:
> > [q] Do we still need to exclude template ref?
> 
> What do you mean? It currently isn't exported because it's not tested..

Oh misread. It is not exported. Ignore this.

PR #340, opened 2024-12-03
fix: Slack Notification

MatKuhr:
Required after Dependabot updated the major version to 2.0.0


.github/workflows/fosstars.yml
```diff
@@ -14,15 +14,15 @@ jobs:
       - uses: SAP/fosstars-rating-core-action@v1.14.0
         with:
           token: '${{ secrets.GH_CLOUD_SDK_JS_ADMIN_WRITE_TOKEN }}'
-      - if: failure() || cancelled()
-        name: Slack Notify
+      - name: 'Slack Notification'
```
ZhongpinWang, line 17: [q] No more cancelled?

MatKuhr: yeah I don't really see the point, if you manually cancel you already know it didn't succeed, no need for a notification, right?

PR #346, opened 2024-12-04
feat: Support JSON for consuming orchestration

shibeshduw:
## Context

SAP/ai-sdk-js-backlog#179.

This PR adds a `executeFromJSON` method that allows users to copy/download the JSON from the Launchpad and use it directly with the SDK. Type safety is not checked as the user takes the responsibility for testing the JSON they use.

## Definition of Done

- ~[ ] Documentation updated~


packages/orchestration/src/orchestration-client.ts
```diff
@@ -50,6 +50,53 @@ export class OrchestrationClient {
...
+   * @param requestConfig - Optional request configuration.
+   * @param deploymentConfig - Deployment configuration.
+   * @returns The orchestration response.
+   */
+  /* eslint-disable @typescript-eslint/member-ordering */
+  static async executeFromJson(
+    jsonConfig: string,
+    prompt?: Prompt,
+    requestConfig?: CustomRequestConfig,
+    deploymentConfig?: ResourceGroupConfig
```
shibeshduw, line 67: This might be an API architecture that we need to decide on because currently, the constructor of the OrchestrationClient looks as follows:
```TypeScript
 constructor(
    private config: OrchestrationModuleConfig,
    private deploymentConfig?: ResourceGroupConfig
  ) {}
```
But the whole point of adding the `executeFromJson` method is that users can provide the config as a JSON, and furthermore, it is not mapped to our existing type. 
The workaround I've used is to make the method static and pass `deploymentConfig` as an optional param. 
Do post your opinions/suggestions regarding this.

sample-code/src/orchestration.ts
```diff
@@ -208,3 +209,50 @@ export async function orchestrationRequestConfig(): Promise<OrchestrationRespons
     }
   );
 }
+
+/**
+ * Use the orchestration service with JSON obtained from AI Launchpad.
+ * @param filePath - Path to the JSON file obtained from AI Launchpad.
+ * @returns The orchestration service response.
+ */
+export async function orchestrationFromJSON(
```
shibeshduw, line 218: [q] Do we separate the two cases into separate examples or is this sufficient?

packages/orchestration/src/orchestration-client.ts
deekshas8: What do you think about adapting the constructor to:

```
constructor(
  private deploymentConfig?: ResourceGroupConfig
) {}
```

and then in the `chatCompletion` method check if type is string then parse it into `OrchestrationModuleConfig`.

ZhongpinWang: [pp] What about adding a `string` type for `config` in the `OrchestrationClient` constructor?

```ts
constructor(
    private deploymentConfig?: ResourceGroupConfig
  ) {}
```

Then in `chatCompletion`:

```ts
let body;
if (typeof this.config === 'string') {
  body = {
    module_configurations: JSON.parse(this.config)
  };
} else {
  body = constructCompletionPostRequest(this.config, prompt);
}
```

So in this case we won't have a separate method `executeFromJson()` but still reuse our `chatCompletion()` method.

shibeshduw: [q] How about, with the intention of keeping the methods separate (user exp, single responsibility, plus aligns with java), we use a typeguard:

```TypeScript
private isOrchestrationModuleConfig(
    config: string | OrchestrationModuleConfig
  ): config is OrchestrationModuleConfig {
    return typeof config === 'object' && 'templating' in config;
  }
```

and then in both methods, we throw an error (along with a suggestion to use the correct method) if the type doesn't match?

deekshas8: From a UX perspective, the current design might be a bit confusing because:

- When passing a TypeScript config object, you instantiate a client and provide the configuration values alongside the deployment configuration.
- However, when using a JSON config, the approach changes entirely—instance properties are replaced by function parameters, and client initialization is no longer required.

Since the ultimate goal in both cases is to use `.chatCompletion()`, it’s not clear why separate methods are necessary.

MatKuhr: For context: The Java method is called `..FromJsonModuleConfig` mainly so that it's clear what the expected input is. If we want to keep the naming, we could use a static factory method which returns a new `OrchestrationClient`. And in Java we pass the config per-request (and we can't easily have a union type including String), so we need a separate Method there. Of course, this this doesn't apply 1:1 in TS. 

No strong opinion, I think using a type union in the constructor is also fine. In that case we / the user just has to rely more on the TS docs on what to input.

sample-code/src/orchestration.ts
```diff
@@ -208,3 +209,50 @@ export async function orchestrationRequestConfig(): Promise<OrchestrationRespons
...
+          }
+        ]
+      }
+    }
+  }`;
+
+  try {
+    // Use provided file path if available, else fallback to default configuration
+    const jsonConfig = filePath
+      ? await fs.promises.readFile(filePath, 'utf-8')
```
MatKuhr, line 247: [pp] I think I'd only implement the file variant, and then mention in a comment that providing the plain string in source code is also possible. Having a fallback logic makes the code a bit more complex and thus harder to understand. Compare the Java docs

packages/orchestration/src/orchestration-client.ts
shibeshduw: Thanks for your suggestions everyone. I ended up going with a type union in the constructor and just using a single `chatCompletion()` method for both cases and handling the construction of the request body accordingly. Also added documentation.

packages/orchestration/README.md
```diff
@@ -326,6 +327,23 @@ const response = await orchestrationClient.chatCompletion({
 return response.getContent();
 ```
 
+### Using a JSON configuration from AI Launchpad
+
+If you already have a orchestration workflow created in AI Launchpad, you can copy or download the configuration as a JSON and use it directly with the orchestration client.
```
deekshas8, line 332:
```suggestion
If you already have an orchestration workflow created in AI Launchpad, you can either download the configuration as a JSON file or copy the JSON string directly to use it with the orchestration client.
```

packages/orchestration/README.md
```diff
@@ -326,6 +327,23 @@ const response = await orchestrationClient.chatCompletion({
...
+
+```ts
+const jsonConfig = await fs.promises.readFile(
+  'path/to/orchestration-config.json',
+  'utf-8'
+);
+// Or alternatively, you can also provide the JSON configuration as a plain string in the code directly.
+// const jsonConfig = '... your JSON ...'
```
deekshas8, line 340: Please refer to the [styleguide](https://github.com/SAP/cloud-sdk/blob/main/STYLEGUIDE.md#placeholders) for placeholders

packages/orchestration/src/orchestration-client.ts
```diff
@@ -56,6 +60,34 @@ export class OrchestrationClient {
...
 }
 
+/**
+ * @internal
+ * Type guard to check if the config is of type OrchestrationModuleConfig.
+ */
+function isOrchestrationModuleConfig(
+  config: OrchestrationModuleConfig | string
+): config is OrchestrationModuleConfig {
+  return typeof config === 'object' && 'templating' in config;
```
deekshas8, line 70: [pp] typeof this.config === 'string' (a simpler typeguard for cheking if its json

packages/orchestration/src/orchestration-client.ts
```diff
@@ -36,7 +36,11 @@ export class OrchestrationClient {
     prompt?: Prompt,
     requestConfig?: CustomRequestConfig
   ): Promise<OrchestrationResponse> {
-    const body = constructCompletionPostRequest(this.config, prompt);
+    const body: CompletionPostRequest | Record<string, any> =
```
deekshas8, line 39: [req] You don't need to cast body to `Record<string, any>`. If you parse the JSON string, it will essentially be of type `CompletionPostRequest`, you can just add the prompt to it right?

sample-code/src/server.ts
```diff
@@ -249,11 +250,18 @@ app.get('/orchestration/:sampleCase', async (req, res) => {
...
+      fromJSON: orchestrationFromJSON
     }[sampleCase] || orchestrationChatCompletion;
 
   try {
-    const result = (await testCase()) as OrchestrationResponse;
+    const result =
+      sampleCase === 'fromJSON'
+        ? // File path should be relative to the root. For this example, `sample-code` is the root.
+          ((await testCase(
+            './src/ModelOrchConfig.json'
```
deekshas8, line 262: [req] style guide for filename please https://github.com/SAP/cloud-sdk/blob/main/STYLEGUIDE.md#filenames-and-file-paths

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -64,6 +64,50 @@ describe('orchestration service client', () => {
...
+      {
+        data: mockResponse,
+        status: 200
+      },
+      {
+        url: 'inference/deployments/1234/completion'
+      }
+    );
+
+    const response = await new OrchestrationClient(jsonConfig).chatCompletion();
```
deekshas8, line 105: [req] Please also add a test for the new `construct...` function (once you've fixed it).
[req] Also add type tests to check it accepts string

packages/orchestration/src/orchestration-client.ts
shibeshduw: Yes but I was also considering the scenario where perhaps a spec change adds additional parameters which have not been adopted yet in our SDK. Since we are not checking for type safety, it can possibly provide users with a way to run their configurations still. Any thoughts on this?

MatKuhr: > provide users with a way to run their configurations

Yes, this is one of the use cases we want to cover with this approach

ZhongpinWang: Technically, adding Record or not won't have any effect on user, because this body is just a local variable, which will never affect any user. 

I would also remove `Record<string, any>` unless we want to be 100% strictly correct and say the JSON object parsed from the string can have breaking changes. 

No strong opinion and I see both points. In the end the type for body does not matter, and return type from `constructCompletionFromJson` is just for our own concern. The method is internal.

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -64,6 +65,46 @@ describe('orchestration service client', () => {
     expect(response.getTokenUsage().completion_tokens).toEqual(9);
   });
 
+  it('calls executeFromJson with valid JSON configuration', async () => {
```
ZhongpinWang, line 68: Change `executeFromJson`

packages/orchestration/src/orchestration-client.ts
shibeshduw: That is correct. However, I still feel like it reads well to have it since it is explicitly clear to a technical user that the object can contain properties aside from those in `CompletionPostRequest`.  Plus, removing it here would also mean that the return type for `constructCompletionFromJson` is also of type `CompletionPostRequest` which, strictly speaking, might not be the case.

packages/orchestration/README.md
```diff
@@ -326,6 +327,23 @@ const response = await orchestrationClient.chatCompletion({
...
+
+```ts
+const jsonConfig = await fs.promises.readFile(
+  'path/to/orchestration-config.json',
+  'utf-8'
+);
+// Or alternatively, you can also provide the JSON configuration as a plain string in the code directly.
+// const jsonConfig = 'YOUR_JSON_CONFIG'
+
+const response = await new OrchestrationClient(jsonConfig).chatCompletion();
```
jjtang1985, line 342: Like the API 👍

packages/orchestration/README.md
```diff
@@ -326,6 +327,23 @@ const response = await orchestrationClient.chatCompletion({
 return response.getContent();
 ```
 
+### Using a JSON configuration from AI Launchpad
```
ZhongpinWang, line 330:
```suggestion
### Using a JSON Configuration from AI Launchpad
```

packages/orchestration/README.md
```diff
@@ -326,6 +327,23 @@ const response = await orchestrationClient.chatCompletion({
...
+
+If you already have an orchestration workflow created in AI Launchpad, you can either download the configuration as a JSON file or copy the JSON string directly to use it with the orchestration client.
+
+```ts
+const jsonConfig = await fs.promises.readFile(
+  'path/to/orchestration-config.json',
+  'utf-8'
+);
+// Or alternatively, you can also provide the JSON configuration as a plain string in the code directly.
```
ZhongpinWang, line 339:
```suggestion
// Alternatively, you can provide the JSON configuration as a plain string in the code.
```

packages/orchestration/src/orchestration-client.ts
```diff
@@ -16,12 +16,12 @@ import type { HttpDestinationOrFetchOptions } from '@sap-cloud-sdk/connectivity'
 export class OrchestrationClient {
   /**
    * Creates an instance of the orchestration client.
-   * @param config - Orchestration module configuration.
+   * @param config - Orchestration module configuration. This can either be an instance of `OrchestrationModuleConfig` or a JSON string retrieved from AI Launchpad.
```
ZhongpinWang, line 19:
```suggestion
   * @param config - Orchestration module configuration. This can either be an `OrchestrationModuleConfig` object or a JSON string retrieved from AI Launchpad.
```

packages/orchestration/src/orchestration-completion-from-json.test.ts
```diff
@@ -0,0 +1,131 @@
+import { constructCompletionFromJson } from './orchestration-client.js';
+
+describe('constructCompletionFromJSON()', () => {
```
ZhongpinWang, line 3:
```suggestion
describe('construct completion from JSON', () => {
```

packages/orchestration/src/orchestration-completion-from-json.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  it('should throw an error when invalid JSON is provided', () => {
+    const invalidJsonConfig = '{ "module_configurations": {}, ';
+
+    expect(() => constructCompletionFromJson(invalidJsonConfig)).toThrow(
+      'Could not parse JSON'
+    );
+  });
+
+  it('with a JSON configuration and template params', () => {
```
ZhongpinWang, line 34:
```suggestion
  it('should construct completion post request from JSON with input params', () => {
```

packages/orchestration/src/orchestration-client.ts
```diff
@@ -36,7 +36,11 @@ export class OrchestrationClient {
     prompt?: Prompt,
     requestConfig?: CustomRequestConfig
   ): Promise<OrchestrationResponse> {
-    const body = constructCompletionPostRequest(this.config, prompt);
+    const body: CompletionPostRequest | Record<string, any> =
+      typeof this.config === 'string'
+        ? constructCompletionFromJson(this.config, prompt)
```
ZhongpinWang, line 41: [pp] Can we rename this into `constructCompletionPostRequestFromJson`? Construct completion sounds pretty weird to me. 
[pp] Or maybe put the type check into the original `constructCompletionPostRequest` function.

packages/orchestration/src/orchestration-completion-from-json.test.ts
ZhongpinWang: [pp] This is where I feel a bit lost when seeing `construct completion`. It actually construct completion post request body. Find a way to make it more clear.

packages/orchestration/src/orchestration-completion-from-json.test.ts
ZhongpinWang: [req] Prefer using natural language to describe a test case / collection.

packages/orchestration/src/orchestration-completion-from-json.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    const completionFromJson: Record<string, any> = constructCompletionFromJson(
+      jsonConfig,
+      { inputParams }
+    );
+
+    expect(expectedCompletionFromJson).toEqual(completionFromJson);
+  });
+
+  it('with a JSON configuration and message history', () => {
```
ZhongpinWang, line 85: Same

sample-code/src/orchestration.ts
```diff
@@ -209,6 +210,30 @@ export async function orchestrationRequestConfig(): Promise<OrchestrationRespons
   );
 }
 
+/**
+ * Use the orchestration service with JSON obtained from AI Launchpad.
+ * @param filePath - Path to the JSON file obtained from AI Launchpad.
```
ZhongpinWang, line 215:
```suggestion
 * @param filePath - Path to the JSON file.
```

sample-code/src/orchestration.ts
```diff
@@ -209,6 +210,30 @@ export async function orchestrationRequestConfig(): Promise<OrchestrationRespons
...
+      const response = await new OrchestrationClient(
+        jsonConfig
+      ).chatCompletion();
+
+      logger.info(response.getContent());
+      return response;
+    }
+  } catch (error) {
+    logger.error('Error in orchestrationFromJSON:', error);
+    throw error;
```
ZhongpinWang, line 234: [pp] There is no meaning of catching an error, just log it and throw it again. If nothing needs to be done in case of an error, I will not catch it. Errors are caught anyways in the handler.

ZhongpinWang (commented):
In general looks good. Some minor stuff.

packages/orchestration/src/orchestration-completion-from-json.test.ts
shibeshduw: Yes I agree with this as well. We will also need to maintain this for rest of the unit tests in the project later. Because this is quite inconsistent atm. Here, I was just trying to make it consistent with `orchestration-completion-post-request.test.ts` but I also would prefer your suggested language.

ZhongpinWang: Oh, please ignore consistency and stick to the preferred natural language. (Or if not so much work, feel free to change that file as well)

@KavithaSiva Fyi.

packages/orchestration/src/orchestration-client.ts
deekshas8: I am also fine if you remove the type from here completely. Because:

1. Removing `Record<string, any>` is a misrepresentation of the type of `body`, since we are not enforcing adherence of type returned from `constructCompletionFromJson`  to `CompletionPostRequest`
2. `body` is not used after this point for any local operations for which type-safety might be needed. Also, the data property of execute function is `any` which makes the union typing here moot.

I would maybe document that body can have additional types for clarity

deekshas8 (changes requested):

packages/orchestration/src/orchestration-client.ts
deekshas8: Like the name `constructCompletionPostRequestFromJson`. Is clearer. I would keep the type of check here though.

ZhongpinWang (approved):
LGTM.

packages/orchestration/src/orchestration-client.ts
tomfrenken: Just for documentation, since we now need to be able to dynamically add streaming options, we actually do need some degree of type safety 😬

PR #335, opened 2024-12-02
feat: Support document grounding

ZhongpinWang:
## Context

SAP/ai-sdk-js-backlog#183.

Add `document-grounding` package with sample code and e2e tests.

> [!note]
> Package marked in README as beta version.

> [!note]
> #351 is merged, support custom destination done.

## Definition of Done

- [x] Code is tested (Unit, E2E)
- ~[ ] Error handling created / updated & covered by the tests above~
- [x] Documentation updated
- ~[ ] (Optional) Aligned changes with the Java SDK~
- [x] (Optional) Release notes updated

PR #351, opened 2024-12-06
feat: Support custom destination

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#163.

> [!CAUTION]
> ~Option 1: Do not merge until #335 is merged and custom destination is also supported in document grounding with documentation added.~
> ✔️ Option 2: Merge this PR first, then adapt #335 to support custom destination in document grounding.

With this PR, you can provide your own destination to execute any request.

For pure clients like `ai-api`, provide destination when calling `execute()` method:

```ts
await DeploymentApi.deploymentDelete(deploymentId, {
  'AI-Resource-Group': 'default'
}).execute({
  destinationName: 'destinationName'
});
```

For `azure-openai`, `orchestration` and `langchain` clients, provide destination with the constructor:

```ts
new AzureOpenAiChatClient('gpt-4', {
  destinationName: 'destinationName'
});
```

```ts
new OrchestrationClient(
  {
    templating: {
      template: [{ role: 'user', content: 'Hello!' }]
    },
    llm: {
      model_params: {}
    }
  },
  {
    resourceGroup: 'resourceGroup'
  },
  {
    destinationName: 'destinationName'
  }
);
```

```ts
new AzureOpenAiChatClient(
  {
    modelName: 'gpt-4o',
    modelVersion: '24-07-2021',
    resourceGroup: 'my-resource-group'
  },
  {
    destinatioName: 'my-destination'
  }
);
```

## Definition of Done

- [X] Documentation updated
- [x] (Optional) Release notes updated (**Double check again**)


packages/core/src/context.ts
```diff
@@ -17,10 +20,24 @@ const logger = createLogger({
...
+ * @param destination - The destination to use for the request.
  * @returns The destination object.
  */
-export async function getAiCoreDestination(): Promise<HttpDestination> {
+export async function getAiCoreDestination(
+  destination?: HttpDestinationOrFetchOptions
+): Promise<HttpDestination> {
+  // If Destination is provided, get the destination and return it.
+  if (destination) {
+    const resolvedDestination = await useOrFetchDestination(destination);
```
deekshas8, line 32: [q] What about caching? If a name is passed and not the full destination, should the destination be cached?

tests/smoke-tests/test/smoke.test.ts
```diff
@@ -7,6 +7,12 @@ describe('Smoke Test', () => {
     ).resolves.toHaveProperty('status', 200);
   });
 
+  it('aicore client retrieves a list of deployments with custom destination', async () => {
```
deekshas8, line 10: [q] Isn't one test for destination sufficient? Just wondering about the no. of calls.

ZhongpinWang: The story is that I want to show case two different usages for adding destination, and I created two functions in `sample-code`. I just want to use that function from somewhere. Unfortunately, there is no destination if run sample code locally. I can mock the destination, but this kind like not real. And calling ai-api isn't heavy, so I would rather leave it here. No strong opinion though.

packages/core/src/context.ts
ZhongpinWang: I mean they can decide whether to set `useCache` or not right? The default behaviour will be the same as cloud SDK.

```ts
DeploymentApi.deploymentQuery(queryParams, {
    'AI-Resource-Group': resourceGroup
  }).execute({
    destinationName: 'e2e-aicore',
    useCache: true
  });
```

deekshas8: The default behavior in the SDK is to not cache, but this will change in v4 (would be a breaking change)

ZhongpinWang: > The default behavior in the SDK is to not cache, but this will change in v4 (would be a breaking change)

Yes, @jjtang1985 should we then cache by default in ai-sdk already?

ZhongpinWang: I will cache by default.

packages/foundation-models/README.md
```diff
@@ -4,9 +4,8 @@ SAP Cloud SDK for AI is the official Software Development Kit (SDK) for **SAP AI
 
 This package incorporates generative AI foundation models into your AI activities in SAP AI Core and SAP AI Launchpad.
 
-## Table of Contents
+### Table of Contents
 
-- [Table of Contents](#table-of-contents)
```
deekshas8, line 9: Thanks for fixing this

packages/foundation-models/package.json
```diff
@@ -33,6 +33,7 @@
     "@sap-ai-sdk/ai-api": "workspace:^",
     "@sap-ai-sdk/core": "workspace:^",
     "@sap-cloud-sdk/http-client": "^3.24.0",
-    "@sap-cloud-sdk/util": "^3.24.0"
+    "@sap-cloud-sdk/util": "^3.24.0",
+    "@sap-cloud-sdk/connectivity": "3.24.0"
```
deekshas8, line 37:
```suggestion
    "@sap-cloud-sdk/connectivity": "^3.24.0"
```
[req] Please consider this in other packages as well

tests/smoke-tests/manifest.yml
```diff
@@ -6,5 +6,6 @@ applications:
       - nodejs_buildpack
     services:
       - default_aicore
+      - default_destination
```
deekshas8, line 9: [q] Not sure I understand why this is needed

deekshas8 (dismissed):
Please fix the versioning and then it's good to go.

packages/foundation-models/package.json
ZhongpinWang: Thanks for pointing these out. I also found some before. Missed them and trusted autocompletion.

tests/smoke-tests/manifest.yml
ZhongpinWang: Removed.

packages/core/src/http-client.ts
```diff
@@ -91,3 +96,24 @@ function mergeWithDefaultRequestConfig(
...
 }
+
+/**
+ * Get target url with endpoint path appended.
+ * Append path `v2` if the url contains empty pathname `/`.
+ * @param url - The url, e.g., `http://example.com` or `http://example.com:8000/abc`.
+ * @param endpointPath - The path to the endpoint, e.g., `/some/endpoint`.
+ * @returns Target url combining the url and endpoint path.
+ * @internal
+ */
```
KavithaSiva, line 107: [q] Is Java also adapting this?

packages/langchain/README.md
```diff
@@ -103,6 +97,27 @@ const embeddingClient = new AzureOpenAiEmbeddingClient({
...
+  {
+    modelName: 'gpt-4o',
+    modelVersion: '24-07-2021',
+    resourceGroup: 'my-resource-group'
+  },
+  {
+    destinationName: 'my-destination'
+  }
+);
+```
```
KavithaSiva, line 116: [q] Purely a design question, maybe I am overlooking something.
What was the reasoning for accepting the destination as part of the client initialisation as opposed to accepting it for the `run` method instead like in the AI Core clients?

Right now, it is not possible to use a set of client configurations with multiple destinations and they are tightly coupled.

KavithaSiva (commented):
Mostly looks good, Just clarifying two questions before approving :)

packages/core/src/http-client.ts
ZhongpinWang: Yes, this is a JS translation from Java 😄

packages/langchain/README.md
ZhongpinWang: This is purely a design consideration. I can hardly think of a case where a user would reuse a single client with the same, e.g., Azure OpenAI or orchestration configuration (meaning same `resourceGroup`, same model and everything) for different destinations (meaning different subaccounts with different AI Core instances). The run methods on the other hand, is generated and we don't have a client for it, so the best we can do.

KavithaSiva: Hmm, I was thinking about using the same client for destinations targeting different landscapes where the AI Core service is deployed.
But, you are right, we already couple `resourceGroup` into the configuration now, which is tied to a tenant or subaccount.

KavithaSiva (approved):
LGTM

packages/langchain/README.md
ZhongpinWang: I mean if they want the same configuration to be used, just store configuration in a separate variable, and initialize a new client. Pretty straightforward.

KavithaSiva: Also, I don't think `run` is generated, it is handwritten for these clients.

PR #352, opened 2024-12-06
feat: Add streaming API to orchestration

tomfrenken:
## Context

SAP/ai-sdk-js-backlog#161.

This PR adds a streaming API to our orchestration client.

## Definition of Done

- [x] Documentation updated


packages/foundation-models/src/azure-openai/azure-openai-embedding-client.ts
```diff
@@ -9,7 +9,7 @@ import { apiVersion, type AzureOpenAiEmbeddingModel } from './model-types.js';
 import type { AzureOpenAiEmbeddingParameters } from './azure-openai-embedding-types.js';
 
 /**
- * Azure OpenAI client for embeddings.
+ * Azure OpenAI client for embeddings./home/i519840/code/ai-sdk-js/packages/orchestration/src.
```
ZhongpinWang, line 12: 🚨🚨🚨

packages/orchestration/src/orchestration-chat-completion-stream.ts
```diff
@@ -0,0 +1,170 @@
...
+    for await (const chunk of stream) {
+      yield new OrchestrationChatCompletionStreamChunkResponse(chunk);
+    }
+  }
+
+  /**
+   * @internal
+   */
+  static async *_processFinishReason(
+    stream: OrchestrationChatCompletionStream<OrchestrationChatCompletionStreamChunkResponse>,
```
ZhongpinWang, line 48: [pp] These long type names are eye catching 😆 But I guess it is what it is.

packages/orchestration/src/orchestration-chat-completion-stream.ts
```diff
@@ -0,0 +1,170 @@
...
+    for await (const chunk of stream) {
+      chunk.data.orchestration_result?.choices.forEach((choice: any) => {
+        const choiceIndex = choice.index;
+        if (choiceIndex >= 0) {
+          const finishReason = chunk.getFinishReason(choiceIndex);
+          if (finishReason) {
+            if (response) {
+              response._getFinishReasons().set(choiceIndex, finishReason);
+            }
+            switch (finishReason) { // TODO: Cover all finish reasons
```
ZhongpinWang: [q] Is there any other finish reason? Can you share this knowledge with me? Pretty curious.

packages/orchestration/src/orchestration-chat-completion-stream-chunk-response.ts
```diff
@@ -0,0 +1,39 @@
...
+  getTokenUsage(): TokenUsage | undefined {
+    return this.data.orchestration_result?.usage;
+  }
+
+  /**
+   * Reason for stopping the completion stream chunk.
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns The finish reason.
+   */
+  getFinishReason(choiceIndex = 0): string | undefined {
```
ZhongpinWang, line 25: [q] Do we know if orchestration plans to support multi-choice? I guess our solution is more generic, but just wanna be sure that this will not cause the chunk to look differently and we are still trying to access the chunks correctly.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -50,17 +53,57 @@ export class OrchestrationClient {
...
+        url: `/inference/deployments/${deploymentId}/completion`,
+        resourceGroup: this.deploymentConfig?.resourceGroup
+      },
+      body,
+      {
+        ...requestConfig,
+        signal: controller.signal
+      }
+    );
+    return OrchestrationChatCompletionStream._create(response, controller);
```
ZhongpinWang, line 102: [pp] This part is a bit inconsistent to Azure OpenAI client. I think I tried to extract the common part out from both streaming and non-streaming variants to avoid code duplication like get deployment ID and construct `endpointOptions`.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -50,17 +53,57 @@ export class OrchestrationClient {
...
  */
 export function constructCompletionPostRequest(
   config: OrchestrationModuleConfig,
-  prompt?: Prompt
+  prompt?: Prompt,
+  stream = false
 ): CompletionPostRequest {
   return {
     orchestration_config: {
+      stream,
```
ZhongpinWang, line 116: [q] Do you know if orchestration also needs `stream_options` to include token usage in the last chunk?

```ts
stream_options: {
  include_usage: true
}
```

packages/orchestration/src/orchestration-chat-completion-stream.ts
tomfrenken: I was under the assumption that other filters also write into the finish reason, but apparently they throw 400 instead.
I've removed the todo.

packages/orchestration/src/orchestration-client.ts
tomfrenken: I didn't find anything in their docs.
They do have streaming options for specific modules, but didn't specify any usage related options.

Related docs: https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/streaming?locale=en-US

packages/orchestration/src/orchestration-client.ts
tomfrenken: Yeah that's a good idea, I'll create a common method.

sample-code/src/orchestration.ts
```diff
@@ -39,6 +41,32 @@ export async function orchestrationChatCompletion(): Promise<OrchestrationRespon
   return result;
 }
 
+/**
+ * Ask GPT 3.5 through the orchestration service about SAP Cloud SDK with streaming.
```
MatKuhr, line 45:
```suggestion
 * Ask GPT 4o through the orchestration service about SAP Cloud SDK with streaming.
```

sample-code/src/orchestration.ts
```diff
@@ -39,6 +41,32 @@ export async function orchestrationChatCompletion(): Promise<OrchestrationRespon
...
+      model_name: 'gpt-4o',
+      model_params: {}
+    },
+    // define the prompt
+    templating: {
+      template: [{ role: 'user', content: 'Give me a short introduction of SAP Cloud SDK.' }]
+    }
+  });
+
+  const response = await orchestrationClient.stream(undefined, controller);
```
MatKuhr, line 66: currently, this API doesn't allow users to modify the stream options of orchestration. Eventually, this should be possible, so if not done with this BLI please create a follow-up so that we don't forget 😉

packages/orchestration/src/orchestration-client.ts
```diff
@@ -85,3 +140,13 @@ export function constructCompletionPostRequest(
     })
   };
 }
+
+function configureLLM(llm: LlmModuleConfig, stream: boolean): LlmModuleConfig {
+  if (stream) {
+    llm.model_params = {
+      ...llm.model_params,
+      stream_options: { include_usage: true }
```
MatKuhr, line 148: (question) is this API harmonized? Maybe test it with e.g. Gemini?

packages/orchestration/src/orchestration-chat-completion-stream-chunk-response.ts
```diff
@@ -0,0 +1,45 @@
+import type { CompletionPostResponseStreaming, LLMChoiceStreaming } from './client/api/schema/index.js';
+import type { TokenUsage } from './index.js';
+
+/**
+ * Orchestration chat completion stream chunk response.
+ */
+export class OrchestrationChatCompletionStreamChunkResponse {
```
MatKuhr, line 7: Currently, this API is lacking the orchestration specific results, i.e. users can't access `module_results`

packages/orchestration/src/orchestration-client.ts
tomfrenken: I was raising the same question, but at least also for Gemini it works as intended.

packages/orchestration/src/orchestration-chat-completion-stream-chunk-response.ts
tomfrenken: Yeah I had a convenience method to get the different module_results, but wasn't sure if anyone actually would use them.

You can still access them though, just not through a function, instead it's `chunk.data.module_results?.YOUR_MODULE`

tomfrenken: What do you think, should there be an additional function for it?

Previously my method was `.getModuleResult(<select-module-with-auto-complete-here>)`

sample-code/src/orchestration.ts
tomfrenken: Also I checked with Dennis, if you provide streaming options, but call a non-streaming API, they are simply ignored.

sample-code/src/orchestration.ts
tomfrenken: I changed the endpoint to use 3.5 turbo instead (since that's cheaper)

packages/orchestration/src/orchestration-chat-completion-stream-chunk-response.ts
MatKuhr: I would assume the module results would be useful on the overall response object? And there it's currently not available, right?

image

I guess it depends a bit on the specific results one is interested in. If we take a chat bot as example use case then I need the templating result + full LLM response at the end and store it in memory. Currently, I'd have to aggregate this manually across the chunks, right?

sample-code/src/orchestration.ts
MatKuhr: The client expects `OrchestrationModuleConfig`, but it's not part of the module config, but on `orchestration_config` instead. So I wouldn't know how it can be customized currently..

> a non-streaming API
what is that?

tomfrenken: > The client expects OrchestrationModuleConfig, but it's not part of the module config, but on orchestration_config instead. So I wouldn't know how it can be customized currently..

Yeah right, I just saw the options in the orchestration documentation, I would assume they end up in the spec at which point we will re-generate, then the config types will include the streaming options.

> what is that? Do you mean a model that doesn't support streaming?
No, rather if you send the config including streaming options but against a regular non-streaming chat-compltion API, e.g. you didn't set `streaming: true`

packages/orchestration/src/orchestration-chat-completion-stream-chunk-response.ts
tomfrenken: I see, so you would like to save all module_results in an array which is available on the response object?

sample-code/src/orchestration.ts
MatKuhr: yeah, so with the current API we would either have to make it part of the prompt, or a new, separate parameter, or somehow else allow the user to pass this.

Okay, so you mean sending e.g.:
```json
{ "orchestration_config": 
  {
     "stream": false,
     "stream_options": {
         "chunk_size": 20
      }
  // ...
  }
}
```
?

packages/orchestration/src/orchestration-chat-completion-stream-chunk-response.ts
MatKuhr: yeah, but I would expect we aggregate over all chunks, so there is only one module_results outcome at the end that includes everything

sample-code/src/orchestration.ts
tomfrenken: Yes exactly.
You can still pass streaming options as part of the options object, but you would have to typecast it to "any", as an ugly workaround.

packages/orchestration/src/orchestration-chat-completion-stream-chunk-response.ts
tomfrenken: I see, I suppose this needs a little bit of investigation to see which implementation (e.g. aggregation) would be the most suitable, I would suggest I create a follow-up ticket for this?

sample-code/src/orchestration.ts
MatKuhr: which options object? I don't see how currently the stream options can be passed at all..

packages/orchestration/src/orchestration-chat-completion-stream.ts
```diff
@@ -0,0 +1,179 @@
+import { createLogger } from '@sap-cloud-sdk/util';
+import { SseStream } from '@sap-ai-sdk/core';
```
MatKuhr, line 2: this file is mostly code duplication, and since this is not trivial code it would be nice to reduce it. I know the generics are annoying, but maybe you can find a way to represent them without nesting them deeper?

E.g. something alone those lines:
```
static async *_processChunk<T extends SseStream<T>>(
    stream: T
  ): AsyncGenerator<T> {
```

sample-code/src/orchestration.ts
```diff
@@ -39,6 +41,37 @@ export async function orchestrationChatCompletion(): Promise<OrchestrationRespon
...
+    llm: {
+      model_name: 'gpt-35-turbo',
+      model_params: {}
+    },
+    // define the prompt
+    templating: {
+      template: [
+        {
+          role: 'user',
+          content: 'Give me a short introduction of SAP Cloud SDK.'
```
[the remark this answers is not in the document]
ZhongpinWang: Yes, but we use it in our e2e. And asking for a long description was the issue last time consuming too many tokens.

sample-code/src/orchestration.ts
tomfrenken: In the initialization of the client.

packages/orchestration/src/orchestration-chat-completion-stream.ts
tomfrenken: I suppose this could be done for most process functions, but not for 
`_processFinishReason()`, as this method accesses different properties based on the chunk type

packages/orchestration/src/orchestration-chat-completion-stream-chunk-response.ts
MatKuhr: follow up ticket is also fine 👍🏻

sample-code/src/orchestration.ts
MatKuhr: I might be blind, but can you give an example code snippet? I don't see how this works..

tomfrenken:
The biggest change of the additional refactor was, that I had to make most underlying functions static.

I've also added a few interfaces which simplify the overall API, so that you don't have to call the functions with a myriad of `undefined` before actually passing the parameters.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -1,63 +1,177 @@
...
+
+  static async streamWithJson(
+    jsonConfig: string,
+    prompt?: Prompt,
+    controller = new AbortController(),
+    deploymentConfig?: ResourceGroupConfig,
+    destination?: HttpDestinationOrFetchOptions,
+    requestConfig?: CustomRequestConfig
+  ): Promise<
+    OrchestrationChatCompletionStreamResponse<OrchestrationChatCompletionStreamChunkResponse>
```
deekshas8, line 59: `OrchestrationChatCompletionStreamResponse<OrchestrationChatCompletionStreamChunkResponse>`

Is way too long

packages/orchestration/src/orchestration-client.ts
```diff
@@ -1,63 +1,177 @@
...
+      prompt,
+      requestConfig,
+      jsonConfig,
+      deploymentConfig,
+      destination
+    });
+    return new OrchestrationResponse(response);
+  }
+
+  static async streamWithJson(
```
deekshas8, line 51: [req] Both static functions have too many parameters, maybe consider using object parameters instead.

deekshas8 (dismissed):

packages/orchestration/src/orchestration-chat-completion-stream-response.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  /**
+   * @internal
+   */
+  _setTokenUsage(usage: TokenUsage): void {
+    this._usage = usage;
+  }
+
+  public getFinishReason(choiceIndex = 0): string | undefined | null {
```
KavithaSiva, line 26: [req] I didn't see an example where `finish_reason` is `null`. I think if it is present it always has atleast an empty string.

```suggestion
  public getFinishReason(choiceIndex = 0): string | undefined {
```

packages/orchestration/src/orchestration-client.ts
```diff
@@ -78,30 +192,47 @@ export function constructCompletionPostRequestFromJson(
...
-          filtering_module_config: config.filtering
-        }),
-        ...(Object.keys(config?.masking || {}).length && {
-          masking_module_config: config.masking
-        }),
-        ...(Object.keys(config?.grounding || {}).length && {
-          grounding_module_config: config.grounding
-        })
-      }
+      stream,
```
KavithaSiva, line 234: [pp] 
```suggestion
      ...(stream && { stream: true })
```

packages/orchestration/src/orchestration-chat-completion-stream.ts
```diff
@@ -0,0 +1,179 @@
...
+        (choice: LLMChoiceStreaming) => {
+          const choiceIndex = choice.index;
+          if (choiceIndex >= 0) {
+            const finishReason = chunk.getFinishReason(choiceIndex);
+            if (finishReason) {
+              if (response) {
+                response._getFinishReasons().set(choiceIndex, finishReason);
+              }
+              switch (finishReason) {
+                case 'content_filter':
```
KavithaSiva, line 69: [q] Any idea, if we can have more finish reasons here other than the ones similar to Open AI?

KavithaSiva (changes requested):
The stream functionality itself looks mostly good, but, I would also consider being able to provide streaming options like `chunk_size` before merging though.

sample-code/src/orchestration.ts
tomfrenken:
```ts
const client = new OrchestrationClient(
  {
    llm: {
      model_params: {}
    },
    templating: {
      template: [{ role: 'user', content: 'Create {number} paraphrases of {phrase}' }]
    },
    filtering: {
      input: {
        filters: [{ type: 'azure_content_safety', config: { Hate: 4, SelfHarm: 0 } }],
        streaming_options: { whatever: 'whatever' }
      } as InputFilteringConfig,
      streaming_options: {
          whatever: 'whatever'
      }
    } as FilteringModuleConfig
  }
);
```

packages/orchestration/src/orchestration-chat-completion-stream-response.ts
```diff
@@ -0,0 +1,57 @@
+import type { TokenUsage } from './client/api/schema/index.js';
+import type { OrchestrationChatCompletionStream } from './orchestration-chat-completion-stream.js';
+
+/**
+ * Orchestration chat completion stream response.
+ */
+export class OrchestrationChatCompletionStreamResponse<T> {
```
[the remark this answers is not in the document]
tomfrenken: Yeah agreed, I aligned with the existing streaming implementation here, but I suppose this would be better naming wise.

packages/orchestration/src/orchestration-client.ts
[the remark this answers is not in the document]
tomfrenken: I disagree, the call we make is exactly the same as we do with our existing streaming implementation, and as the config is in the user's hands, they are responsible for providing a valid config.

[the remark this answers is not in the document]
tomfrenken: All the additional test really would do is validating that JSON.parse() works.

packages/orchestration/src/orchestration-chat-completion-stream-chunk-response.test.ts
```diff
@@ -0,0 +1,79 @@
...
+        'orchestration',
+        'orchestration-chat-completion-stream-chunk-response-token-usage.json'
+      ),
+      finishReasonResponse: await parseMockResponse<any>(
+        'orchestration',
+        'orchestration-chat-completion-stream-chunk-response-finish-reason.json'
+      ),
+      deltaContentResponse: await parseMockResponse<any>(
+        'orchestration',
+        'orchestration-chat-completion-stream-chunk-response-delta-content.json'
```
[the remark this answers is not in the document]
tomfrenken: While true, I wasn't sure if merging them is the best approach, as they are potentially separate.
E.g., in this case, there was no real finish reason, since it was a successful stream, however, if the stream stops due to a filter hit, the token usage wouldn't be part of the same chunk, and even be omitted AFAIK.

I'm fine for merging these for the test, but what do you think?

packages/orchestration/src/orchestration-chat-completion-stream.ts
tomfrenken: AFAIK no.

packages/orchestration/src/orchestration-chat-completion-stream.ts
tomfrenken: https://github.com/SAP/ai-sdk-js-backlog/issues/199

packages/orchestration/src/orchestration-client.ts
tomfrenken: While this somewhat aligns with what we did with the other properties, I disagree here.
This convolutes the logic of our code, even though it is not a requirement of the target API.

If anything, I think orchestration should be able to handle empty configs for the different modules, so that we can simply forward it without doing the whole spread logic.

packages/orchestration/src/orchestration-chat-completion-stream-chunk-response.ts
tomfrenken: https://github.com/SAP/ai-sdk-js-backlog/issues/200

sample-code/src/orchestration.ts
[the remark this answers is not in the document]
tomfrenken: While true, I would like this to be used in our E2E tests in the future, do you think making the response longer is worth it still?

packages/orchestration/src/orchestration-client.ts
ZhongpinWang: I also feel it requires too many params. I am wondering, if it is really a bad idea to accept json in the constructor.

Consider the following cases (please verify if my understanding is correct):

| `stream` property in JSON | Which method to call | Consequence |
|---|---|---|
| `true` | `chatCompletion()` | All good, non-streaming with `stream` overwritten to `false` |
| `true` | `stream()` | All good, streaming |
| `false` | `chatCompletion()` | All good, non-streaming |
| `false` | `stream()` | All good (?), streaming with `stream` overwritten to `true` |

packages/orchestration/src/orchestration-client.ts
```diff
@@ -1,63 +1,177 @@
...
+      prompt,
+      requestConfig,
+      config: this.config,
+      deploymentConfig: this.deploymentConfig,
+      destination: this.destination
+    });
     return new OrchestrationResponse(response);
   }
+
+  async stream(
```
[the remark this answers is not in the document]
ZhongpinWang: I feel currently `stream_options` is not supported at all?

packages/orchestration/src/orchestration-client.ts
KavithaSiva: I felt spread logic is helpful, because then you wouldn't have to modify the existing tests to add `stream: false` for a non-streaming variant request.

And moreover, since `stream_options` can also be added here, I think spreading is still better. WDY think?

packages/orchestration/src/orchestration-client.ts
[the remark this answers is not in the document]
KavithaSiva: To add it to filter module, we can still use for e.g :
```TS
const filtering: FilteringModuleConfig = {
  input: buildAzureContentFilter(),
  output: {
    ...buildAzureContentFilter(),
    ...{
      stream_options: {
        overlap: 10
      }
    }
  }
};

```
as the `OutputFilteringConfig` has the streaming options already generated in.
but I don't think there is an easy way yet to add `stream_options` with `chunk_size`

packages/orchestration/src/orchestration-chat-completion-stream-chunk-response.test.ts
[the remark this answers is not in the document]
KavithaSiva: Apologies, I meant `orchestration-chat-completion-stream-chunk-response-token-usage.json` and `test-util/data/orchestration/orchestration-chat-completion-stream-chunk-response-finish-reason.json` as they both are essentially same.

So, you can just use one file for both.
`orchestration-chat-completion-stream-chunk-response-delta-content.json` is a different output, so you can still leave that one in.

packages/orchestration/src/orchestration-client.ts
tomfrenken: I've created a separate utility function to add / remove streaming options, it will still set streaming to false, but this is by design, as people could instantiate a client with a streaming module config, but call chat completion, in which case we need to overwrite the value to false, or delete it entirely. Since deleting the field seems to be too much additional logic for me, I set it to false for now.

tomfrenken:
Alright, I've made the suggested changes to all of the logic, I still need to add some additional tests though, in the meantime, you can check if my changes align with your suggestions.

Changes:
- API has been shortened, as multiple parameters where put into a single options object
- You can pass streaming options now
- Renamed all functions to be shorter
- Renamed JSON functions for clarity

packages/orchestration/src/orchestration-client.ts
```diff
@@ -78,31 +202,73 @@ export function constructCompletionPostRequestFromJson(
...
+        filtering_module_config: {
+          ...moduleConfigs.filtering_module_config,
+          output: {
+            ...(moduleConfigs.filtering_module_config?.output || {}),
+            stream_options: outputFiltering
+          }
+        }
+      })
+    }
+  } as OrchestrationConfig; // TODO: Remove typecast when types are re-generated;
```
ZhongpinWang, line 237: [q] Can we fix this within this PR? I see some type issues if not casting it.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -78,31 +202,73 @@ export function constructCompletionPostRequestFromJson(
...
+      }),
+    ...(config?.grounding &&
+      Object.keys(config.grounding).length && {
+        grounding_module_config: config.grounding
+      })
+  };
+
+  return {
+    orchestration_config: stream
+      ? addStreamOptions(moduleConfigurations, streamOptions)
```
ZhongpinWang, line 270: [pp] As previously mentioned, why don't we allow streamOptions to be optional and get rid of the stream flag.

ZhongpinWang (changes requested):
Review on going...

packages/orchestration/src/orchestration-client.ts
```diff
@@ -1,69 +1,193 @@
...
+}
+
+interface StreamOptions {
+  chunkSize?: number;
+  llm?: { includeUsage?: boolean; [key: string]: any };
+  // Add more options as they are implemented
+  // masking?: { };
+  // grounding?: { };
+  // inputFiltering?: { };
+  outputFiltering?: { overlap?: number };
```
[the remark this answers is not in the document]
tomfrenken: Yeah makes sense to me, however, until the spec is updated (probably in 1.6?) these types don't exist yet.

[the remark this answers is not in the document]
ZhongpinWang: At least `GlobalStreamOptions` and `FilteringStreamOptions` already exist. Should we just use those generated types and replace llm later once we have a newer spec?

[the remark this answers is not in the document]
tomfrenken: Nvm, I mixed this up with the Azure types.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -78,31 +202,73 @@ export function constructCompletionPostRequestFromJson(
   }
 }
 
+function addStreamOptions(
+  moduleConfigs: ModuleConfigs,
+  streamOptions?: StreamOptions
+): OrchestrationConfig {
+  const llm = streamOptions?.llm ?? {};
```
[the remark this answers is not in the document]
tomfrenken: Yeah good point, I'll rename it.
I'd rather not inline it, to keep the code a little more readable.

MatKuhr (commented):
in progress ...

packages/orchestration/src/orchestration-client.ts
```diff
@@ -1,69 +1,193 @@
...
+      requestConfig,
+      stream,
+      streamOptions,
+      jsonModuleConfig,
+      deploymentConfig,
+      destination,
+      config
+    } = options;
+
+    const body = jsonModuleConfig
```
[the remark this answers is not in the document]
deekshas8: [req] I had a similar preference when looking at this. I wouldn't create a completely separate `RequestOptions` type, but atleast have `config: string | OrchestrationModuleConfig` type and then add a type guard here so we avoid the `!` in `constructCompletionPostRequest(config!, prompt, stream, streamOptions);`

packages/orchestration/src/orchestration-client.ts
```diff
@@ -1,69 +1,193 @@
...
+      ? constructCompletionPostRequestFromJsonModuleConfig(
+          jsonModuleConfig,
+          prompt
+        )
+      : constructCompletionPostRequest(config!, prompt, stream, streamOptions);
```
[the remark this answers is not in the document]
deekshas8: [q] Do we even need a `stream` bool property here? Isn't it enough to check if streamOptions was passed and has properties?

packages/core/src/stream/line-decoder.ts
```diff
@@ -5,7 +5,6 @@ type Bytes = string | ArrayBuffer | Uint8Array | Buffer | null | undefined;
  * reading lines from text.
  *
  * Https://github.com/encode/httpx/blob/920333ea98118e9cf617f246905d7b202510941c/httpx/_decoders.py#L258.
- * @internal
```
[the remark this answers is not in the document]
tomfrenken: Cause you can't export internal functions, and everything in core is by definition "internal" anyway.

packages/orchestration/src/orchestration-client.ts
deekshas8: [req] I think you probably need to add a type check for output as well so that filters does not evaluate to undefined. 

```
...(moduleConfigs.filtering_module_config?.output && {
            output: {
            filters: moduleConfigs.filtering_module_config.output.filters,
            stream_options: outputFiltering
          }})
```

I am guessing there is a better way to write this.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -1,69 +1,193 @@
...
 import type {
   OrchestrationModuleConfig,
   Prompt
 } from './orchestration-types.js';
+import type { OrchestrationStreamChunkResponse } from './orchestration-stream-chunk-response.js';
 import type { HttpDestinationOrFetchOptions } from '@sap-cloud-sdk/connectivity';
 
+interface RequestOptions {
+  prompt?: Prompt;
+  requestConfig?: CustomRequestConfig;
```
deekshas8, line 25: [pp] Please reuse type `ConfigurationOptions` here so we avoid duplication.

ZhongpinWang (changes requested):
I have gone through all the files once. Maybe there will be a second round review. 🔁

packages/orchestration/src/index.ts
ZhongpinWang, line 1: The following missing exports are related to this PR:

- `ChatDelta`
- `CompletionPostResponseStreaming`
- `FilteringStreamOptions`
- `GlobalStreamOptions`
- `LLMChoiceStreaming`
- `LLMModuleResultStreaming`
- `LLMModuleResultSynchronous`

The following missing exports are template related. 
@MatKuhr I see that we export `TemplatingModuleConfig` ( `= Template | TemplateRef`) without exporting the following types:

- `TemplateRefByID`
- `TemplateRefByScenarioNameVersion`
- `TemplateRef`
- `Template`
- `TemplatingChatMessage`

packages/orchestration/src/orchestration-client.ts
```diff
@@ -1,113 +1,136 @@
...
+
+  async stream(
+    prompt?: Prompt,
+    controller = new AbortController(),
+    options?: StreamOptions,
+    requestConfig?: CustomRequestConfig
+  ): Promise<OrchestrationStreamResponse<OrchestrationStreamChunkResponse>> {
+    if (typeof this.config === 'string' && options) {
+      logger.warn(
+        'Stream options are not supported when using a JSON module config.'
```
ZhongpinWang, line 65: 😄 Please if anyone has better idea than this, put a comment.

packages/orchestration/src/orchestration-client.ts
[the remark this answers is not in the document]
tomfrenken: I narrowed the type down, do you still prefer a different approach?
A benefit of it is that I can use the spread syntax later on

packages/orchestration/src/orchestration-client.ts
```diff
@@ -1,113 +1,136 @@
...
+            this.config,
+            prompt,
+            stream,
+            streamOptions
+          );
 
     const deploymentId = await resolveDeploymentId({
       scenarioId: 'orchestration',
-      resourceGroup: this.deploymentConfig?.resourceGroup
+      ...(this.deploymentConfig || {})
```
ZhongpinWang, line 98: [pp] I would prefer using `??`, which sounds more correct to me, checking undefined or null instead of falsy. Same to other occurrences.

packages/orchestration/src/orchestration-utils.ts
ZhongpinWang, line 1: Love this.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -56,3 +59,39 @@ export interface OrchestrationModuleConfig {
...
+  /**
+   * Options for the stream.
+   */
+  streamOptions?: StreamOptions;
+}
+
+/**
+ * Options for the stream.
+ */
+export interface StreamOptions extends GlobalStreamOptions {
```
ZhongpinWang, line 88: [pp] Or what about having an additional attribute called `global` to store chunk_size etc. Just in case in the future we have more and more, and we need to manually extract the global options one by one.

packages/orchestration/src/orchestration-utils.ts
```diff
@@ -0,0 +1,155 @@
...
+  if (!moduleConfigs?.filtering_module_config?.output && outputFiltering) {
+    logger.warn(
+      'Output filter stream options are not applied because filtering module is not configured.'
+    );
+  }
+
+  return {
+    stream: true,
+    stream_options: {
+      chunk_size: chunkSize
```
ZhongpinWang, line 94: See my previous comment, would be nice if we can do `stream_options: ...streamOptions?.global`

ZhongpinWang (commented):
One more round

test-util/data/orchestration/orchestration-chat-completion-stream-chunk-response-delta-content.json
```diff
@@ -0,0 +1,39 @@
...
+        "model": "gpt-4o-2024-08-06",
+        "system_fingerprint": "fp_4e924a4b48",
+        "choices": [
+            {
+                "index": 0,
+                "delta": {
+                    "role": "assistant",
+                    "content": "rimarily focusing on Java and JavaScript/Node.js environments, allowing developers to work in their "
+                },
+                "finish_reason": ""
```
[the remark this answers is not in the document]
ZhongpinWang: Nvm ignore this, we already implemented a falsy check so "" is ignored. I was a bit confused.

packages/orchestration/src/orchestration-client.ts
MatKuhr: Some ideas: #427 427

1. Technically, we could narrow this condition to `options?.llm`.
2. Remove the separate handling for `string` and operate only on the object. 
  * Shorter code, but requires either casting to any or other shenanigans.
  * Mainly depends if we want to guarantee arbitrary payload works (assuming the server accepts it of course). 
  * Alternatively, we can be slightly more strict and assume some fields (essentially only `llm_module_config` currently) are always present.

packages/orchestration/src/orchestration-stream-chunk-response.ts
```diff
@@ -0,0 +1,48 @@
...
+      (c: LLMChoiceStreaming) => c.index === choiceIndex
+    )?.finish_reason;
+  }
+
+  /**
+   * Parses the chunk response and returns the delta content.
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns The message delta content.
+   */
+  getDeltaContent(choiceIndex = 0): string | undefined {
```
MatKuhr, line 39: Do we have tests for n > 1? If not, I'm not sure this really works as intended, if it depends on LLMs sending only one choice per chunk (which isn't guaranteed), and how users could meaningfully use this.

If we are unsure, we could also disable / remove it for now

MatKuhr (commented):
Didn't get through everything, but some ideas. Also, let's remember to add E2E tests

packages/orchestration/src/orchestration-client.ts
```diff
@@ -1,69 +1,193 @@
...
+      deploymentConfig: this.deploymentConfig,
+      destination: this.destination
+    });
     return new OrchestrationResponse(response);
   }
+
+  async stream(
+    prompt?: Prompt,
+    controller = new AbortController(),
+    options?: StreamOptions,
```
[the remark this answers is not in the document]
ZhongpinWang: We discussed this and feel it makes sense for user to use controller if they intend to use streaming in a correct way. Because without a controller, an ongoing stream can never be stopped in case of any error, or if user leave/refresh the page, connection closed, etc. So kind like a must have in the end. Worst case scenario, a new AbortController() is also not super complicated.

packages/core/src/stream/line-decoder.ts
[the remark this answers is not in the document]
tomfrenken: Well, that "public api" is literaly every exported function of the package.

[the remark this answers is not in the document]
tomfrenken: We can ofc export it and reach it from the internal.ts path, but conceptually the entire core package is internal.

packages/orchestration/src/orchestration-types.ts
tomfrenken: Yes!!

packages/orchestration/src/orchestration-client.ts
tomfrenken: I thought about a similar approach, but casting usually ends up masking changes to the underlying types throughout the package's lifecycle, especially when re-generating the types, which could end up causing issues down the line, therefore I would always try to avoid it.

Just like in your example, the only way to handle it like an arbitrary object is to make the type a resursivePartial, which would be an object that doesn't satisfy the current API (e.g., if you set any filter config it has to have N properties).

We could consider refactoring this in a follow-up PR, but honestly I would argue that if you are using a JSON config client, all the options that you want to have enabled should already be set at initialization of the client.

packages/orchestration/src/orchestration-stream-chunk-response.ts
tomfrenken: Not sure if I understand the issue, can you elaborate?

The given function could handle a "out of bounds" choiceIndex, if a given LLM has more than one choice, it will either be found and returned, or undefined is returned, either would be fine AFAIK?

packages/orchestration/README.md
```diff
@@ -85,6 +85,124 @@ The client allows you to combine various modules, such as templating and content
...
+  console.log(JSON.stringify(chunk));
+}
+```
+
+In this example, streaming request will be aborted after one second.
+Abort controller can be useful, e.g., when end-user wants to stop the stream or refreshes the page.
+
+#### Stream Options
+
+The orchestration service offers multiple streaming options, which you can configure in addition to the LLM's streaming options.
```
ZhongpinWang, line 177: [pp] multiple streaming options such as?

packages/orchestration/README.md
```diff
@@ -85,6 +85,124 @@ The client allows you to combine various modules, such as templating and content
...
+There are two ways to add specific streaming options to your client, either at initalization, or dynamically when calling the stream API.
```
ZhongpinWang, line 178:
```suggestion
There are two ways to add specific streaming options to your client, either at initialization of orchestration client, or when calling the stream API.
```

packages/orchestration/README.md
```diff
@@ -85,6 +85,124 @@ The client allows you to combine various modules, such as templating and content
...
+
+Dynamically setting these options after client initialization is particularly helpful when you've initialized a client with a config meant for regular chat completion and now want to switch to using streaming.
```
ZhongpinWang, line 180:
```suggestion
Setting streaming options dynamically could be useful if an initialized orchestration client will also be used for streaming.
```

But I mean even if a user put stream options when initializing the client, they can still use the client for non-streaming chat completion right? I can only imagine setting stream options dynamically if user write some kind of playground or reuse the same orchestration client for different use cases (which requires e.g. different filter settings?)

packages/orchestration/README.md
```diff
@@ -85,6 +85,124 @@ The client allows you to combine various modules, such as templating and content
...
+  controller,
+  {
+    llm: { include_usage: false },
+    global: { chunk_size: 10 },
+    outputFiltering: { overlap: 200 }
+  }
+);
+```
+
+Usage metrics are collected by default, if you do not want to receive them, set include_usage to false.
```
ZhongpinWang, line 200:
```suggestion
Usage metrics are returned by default, if you do not want to receive them, set `include_usage` to `false`.
```

packages/orchestration/README.md
```diff
@@ -85,6 +85,124 @@ The client allows you to combine various modules, such as templating and content
...
+If you don't want any streaming options as part of your call to the LLM, set options.llm = null.
```
ZhongpinWang, line 201:
```suggestion
If you don't want any streaming options as part of your call to the LLM, set `streamOptions.llm` to `null`.
```

packages/orchestration/src/orchestration-chat-completion-stream-chunk-response.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      finishReasonResponse: new OrchestrationChatCompletionStreamChunkResponse(
+        mockResponses.finishReasonResponse
+      ),
+      deltaContentResponse: new OrchestrationChatCompletionStreamChunkResponse(
+        mockResponses.deltaContentResponse
+      )
+    };
+  });
+
+  it('should return the chat completion stream chunk response', () => {
```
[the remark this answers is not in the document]
tomfrenken: We need 3 responses to replicate the "real" behavior of the orchestration service.
The reason we test on the chunk level instead of the stream.respones level is because I tried to test the lowest component / shift left approach.
We also do the same in the existing streaming tests.

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -378,4 +380,50 @@ describe('orchestration service client', () => {
...
+    );
+    const response = await new OrchestrationClient(config).stream();
+
+    const initialResponse = await parseFileToString(
+      'orchestration',
+      'orchestration-chat-completion-stream-chunk-response-initial.json'
+    );
+
+    for await (const chunk of response.stream) {
+      expect(chunk.data).toEqual(JSON.parse(initialResponse));
```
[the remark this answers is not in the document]
tomfrenken: There are multiple chunks, however, only the first chunk is tested, which should match the expected initial response.
We could extend this to the entire stream, do you think that would be necessary?

packages/orchestration/README.md
```diff
@@ -86,6 +86,125 @@ The client allows you to combine various modules, such as templating and content
...
+```ts
+const response = orchestrationClient.stream(
+  {
+    inputParams: { country: 'France' }
+  },
+  controller,
+  {
+    llm: { include_usage: false },
+    global: { chunk_size: 10 },
+    outputFiltering: { overlap: 200 }
```
ZhongpinWang, line 197: [q] Out of curiosity, have you tried setting streaming options for filtering without providing a filter?

tomfrenken:
Side node while I add streaming to our sample code and e2e tests, we should really refactor our server.ts file, lots of duplication and it should be more modularized ...

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -62,4 +62,7 @@ describe('orchestration', () => {
     expect(response.getContent()?.includes('SAP')).toBe(true);
     expect(response.getContent()?.includes('logo')).toBe(true);
   });
+
+  // add e2e test
+  // java tests only that there is any response
```
ZhongpinWang, line 67: [q] TODO?

ZhongpinWang (approved):

ZhongpinWang (changes requested):
After e2e added

packages/orchestration/README.md
tomfrenken: I did 👍

sample-code/README.md
```diff
@@ -162,6 +162,20 @@ Use `buildAzureContentFilter()` to build the content filter.
 
 Send chat completion request with a custom header as the custom request configuration.
 
+#### Chat Completion Streaming
+
+`GET /orchestration-stream/chat-completion-stream`
```
ZhongpinWang, line 167:
```suggestion
`POST /orchestration-stream/chat-completion-stream`
```
[req] Also add a payload example here. Would be nice to have to have a curl cmd here.

ZhongpinWang: [req] `/orchestration-stream/chat-completion-stream-json` missing.

ZhongpinWang (changes requested):

.changeset/hip-melons-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[New Funcionality] Add support for streaming in the orchestration client.
```
ZhongpinWang, line 5:
```suggestion
[New Functionality] Add support for streaming in the orchestration client.
```

ZhongpinWang (approved):
LGTM 🎉

PR #359, opened 2024-12-10
chore: Fix sample-cap dependencies

deekshas8:
## Context

The bump causes the pipeline to fail as the ai-sdk dependencies in sample-cap are also updated but the pnpm-lock isn't. This results in conflicts and breaks the pipeline.
For running locally (or in workflow without cf deployment), using the pnpm `workspace:^ ` protocol should be fine. Building the app locally is also sufficient for testing transpilation. For deployment, the users can update the dependency as documented. (The sample-cap is currently not used in any tests, workflows (should be fixed)).

**Alternative:** 
Another solution could be that sample-cap is not considered a workspace package, so changeset will not try to bump anything. But this has the disadvantage that you always have to cd to the dir to execute commands. Though not a huge deal for now.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -167,14 +167,14 @@ importers:
   sample-cap:
     dependencies:
       '@sap-ai-sdk/ai-api':
-        specifier: ^1
-        version: 1.2.0
+        specifier: workspace:^
+        version: link:../packages/ai-api
       '@sap-ai-sdk/foundation-models':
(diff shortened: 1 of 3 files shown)


ZhongpinWang (commented):
I am okay with adding this one additional step for manually changing the versions. Feels like there is no way to let `changeset version` ignore `sample-cap` package.

ZhongpinWang (approved):

PR #378, opened 2024-12-17
feat: Add grounding convenience function

deekshas8:
## Context

SAP/ai-sdk-js-backlog#178.

Please provide a description of what your change does and why it is needed.

## Definition of Done

- [ ] Documentation updated


packages/orchestration/src/orchestration-util.ts
```diff
@@ -0,0 +1,57 @@
...
+      {
+        type: 'azure_content_safety',
+        ...(filter && { config: filter })
+      }
+    ]
+  };
+}
+
+export type DocumentGroundingFilterWrapper = Pick<DocumentGroundingFilter,
+  'id' | 'search_config' | 'data_repositories' | 'data_repository_metadata' | 'document_metadata' | 'chunk_metadata'> & {
```
deekshas8, line 32: @jjtang1985 Do you think making `data_repositories` optional would make sense? If user doesn't set it, we set it to * 
```
"data_repositories": ["*"],
```

packages/orchestration/src/orchestration-util.ts
```diff
@@ -0,0 +1,57 @@
...
+    data_repository_type?: DataRepositoryType;
+  };
+
+/**
+ * Convenience function to create Document Grounding configuration.
+ * @param groundingConfig: Configuration for the document grounding service.
+ * @returns An object with the full grounding configuration.
+ */
+export function buildDocumentGroundingConfig(
+  groundingConfig: { filters?: DocumentGroundingFilterWrapper[]; input_params: string[]; output_param: string }
```
deekshas8, line 42: [note] Will create a separate type for this.

packages/orchestration/src/orchestration-util.ts
```diff
@@ -0,0 +1,57 @@
...
+): GroundingModuleConfig {
+  return {
+    type: 'document_grounding_service',
+    config: {
+      input_params: groundingConfig.input_params,
+      output_param: groundingConfig.output_param,
+      ...(groundingConfig.filters && {
+        filters: groundingConfig.filters?.map(filter => ({
+          ...filter,
+          data_repository_type: 'vector'
```
ZhongpinWang, line 52: [q] This is problematic right? Because even if user defines a different type, it will be overwritten into `vector`. Can we check if it is already defined?

packages/orchestration/src/orchestration-util.ts
ZhongpinWang: [pp] For me personally, if `["*"]`, then this parameter can be optional.

ZhongpinWang, line 32: [pp] Do you think it would be helpful to have a helper function for setting all these three metadata fields?

jjtang1985:
While providing ids for `data_repositories` is helpful as this is the parameter from orchestration API perspective, there might be another idea.

When doing a normal chat completion, you also need a deployment_id.
The SDK allows users for providing the model name and does the search for the user.

Is it possible to do similar things here?
If a user knows the `title` of `data_repository`, should we also search for the id?

packages/orchestration/src/orchestration-util.ts
```diff
@@ -0,0 +1,57 @@
...
+  'id' | 'search_config' | 'data_repositories' | 'data_repository_metadata' | 'document_metadata' | 'chunk_metadata'> & {
+    data_repository_type?: DataRepositoryType;
+  };
+
+/**
+ * Convenience function to create Document Grounding configuration.
+ * @param groundingConfig: Configuration for the document grounding service.
+ * @returns An object with the full grounding configuration.
+ */
+export function buildDocumentGroundingConfig(
```
jjtang1985, line 41: I feel this is in the right direction.

packages/orchestration/src/orchestration-util.ts
deekshas8: I am not aware what keys are supported, also couldn't find any documentation. I have asked Govind. Would be nicer if orchestration itself adds this in the spec.

.changeset/shiny-brooms-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[Improvement] Add convenience function for creating document grounding configuration
```
shibeshduw, line 5: [Improvement] Add convenience function to create document grounding configuration

shibeshduw (dismissed):

.changeset/shiny-brooms-punch.md
deekshas8, line 5:
```suggestion
[Improvement] Add convenience function to create document grounding configuration.
```

packages/orchestration/README.md
```diff
@@ -327,6 +327,23 @@ const response = await orchestrationClient.chatCompletion({
 return response.getContent();
 ```
 
+`buildDocumentGroundingConfig` function provides convenience to create the document grounding service configuration using the minimal set of required values:
```
deekshas8, line 330:
```suggestion
`buildDocumentGroundingConfig()` function provides convenience to create the document grounding service configuration using the minimal set of required values:
```

.changeset/shiny-brooms-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[Improvement] Add convenience function to create document grounding configuration.
```
ZhongpinWang, line 5: I in general feel using the term `document grounding` is a bit confusing considering we have a package called document grounding. Can we rename everywhere this term into orchestration grounding? Technically, this package has nothing to do with document grounding service, but calling only orchestration service.

packages/orchestration/README.md
```diff
@@ -327,6 +327,23 @@ const response = await orchestrationClient.chatCompletion({
...
+`buildDocumentGroundingConfig()` function provides convenience to create the document grounding service configuration using the minimal set of required values:
+
+```ts
+import {
+  buildDocumentGroundingConfig
+} from '@sap-ai-sdk/orchestration';
+
+const orchestrationConfig: {
+  ...,
+  grounding: buildDocumentGroundingConfig(
```
ZhongpinWang, line 339: Here, maybe `buildGroundingConfig`

packages/orchestration/README.md
```diff
@@ -327,6 +327,23 @@ const response = await orchestrationClient.chatCompletion({
 return response.getContent();
 ```
 
+`buildDocumentGroundingConfig()` function provides convenience to create the document grounding service configuration using the minimal set of required values:
```
ZhongpinWang, line 330:
```suggestion
`buildDocumentGroundingConfig()` function provides convenience to create orchestration grounding configuration using the minimal set of required values:
```

packages/orchestration/README.md
```diff
@@ -327,6 +327,23 @@ const response = await orchestrationClient.chatCompletion({
...
+import {
+  buildDocumentGroundingConfig
+} from '@sap-ai-sdk/orchestration';
+
+const orchestrationConfig: {
+  ...,
+  grounding: buildDocumentGroundingConfig(
+    input_params: ['groundingInput'],
+    output_param: 'groundingOutput',
+    filters: [{id: 'filter-id', data_repositories: ['repository-id'],}]
```
ZhongpinWang, line 342:
```suggestion
    filters: [{ id: 'filter-id', data_repositories: ['repository-id'] }]
```
Not sure about the formatting

packages/orchestration/src/orchestration-types.ts
```diff
@@ -56,3 +58,42 @@ export interface OrchestrationModuleConfig {
    */
   grounding?: GroundingModuleConfig;
 }
+
+/**
+ * Represents a filter configuration for the Document Grounding Service.
+ *
```
ZhongpinWang, line 64:
```suggestion
```

packages/orchestration/src/orchestration-types.ts
```diff
@@ -56,3 +58,42 @@ export interface OrchestrationModuleConfig {
...
+   */
+  data_repository_type?: DataRepositoryType;
+};
+
+/**
+ * Represents the configuration for the Document Grounding Service.
+ */
+export interface DocumentGroundingServiceConfig {
+  /**
+   * Define the filters to apply during the grounding process.
```
ZhongpinWang, line 87:
```suggestion
   * Defines the filters to apply during the grounding process.
```
[pp] For better consistency.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -56,3 +58,42 @@ export interface OrchestrationModuleConfig {
...
+   */
+  filters?: DocumentGroundingServiceFilter[];
+  /**
+   * Contains the input parameters used for grounding input questions.
+   */
+  input_params: string[];
+  /**
+   * Parameter name used for grounding output.
+   * @example "groundingOutput"
```
ZhongpinWang, line 96: [q] Not sure if this example helps a lot. Makes only sense to us probably since we used `groundingOutput`.

packages/orchestration/src/orchestration-util.ts
ZhongpinWang, line 1: [pp] Why don't we name this file in plural? `orchestration-utils.ts`

packages/orchestration/src/orchestration-util.ts
```diff
@@ -0,0 +1,51 @@
...
+      {
+        type: 'azure_content_safety',
+        ...(filter && { config: filter })
+      }
+    ]
+  };
+}
+
+/**
+ * Convenience function to create Document Grounding configuration.
```
ZhongpinWang, line 31: Again, I would call it orchestration grounding configuration.

packages/orchestration/src/orchestration-util.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @param groundingConfig - Configuration for the document grounding service.
```
ZhongpinWang, line 32:
```suggestion
 * @param groundingConfig - Configuration for orchestration grounding.
```

packages/orchestration/src/orchestration-util.ts
```diff
@@ -0,0 +1,51 @@
...
+  filter?: AzureContentSafety
+): InputFilteringConfig | OutputFilteringConfig {
+  if (filter && !Object.keys(filter).length) {
+    throw new Error('Filter property cannot be an empty object');
+  }
+  return {
+    filters: [
+      {
+        type: 'azure_content_safety',
+        ...(filter && { config: filter })
```
ZhongpinWang, line 24: [req] @KavithaSiva I saw this was your code change. Why are we expanding `filter` here as well instead of just writing

```ts
filters: [
  {
    type: 'azure_content_safety',
    config: filter
  }
]

**Review by ZhongpinWang — CHANGES_REQUESTED** [review]
As written also separately below, I would use the term `orchestration grounding` instead of `document grounding` or mention document grounding **service** as this is a dedicated service and has nothing to do with orchestration service.

Some other minor suggestions.

File: packages/orchestration/README.md (hunk 8 above)
Comment (deekshas8) [reply packages/orchestration/README.md:339]: I named it so because the config `type` is `document_grounding_service` and this specific function only creates a configuration for this type. If in the future there is another type (donno if there will be), I imagine its name should be derived from that.

Comment (deekshas8) [reply packages/orchestration/README.md:339]: Similar to how we say buildAzureFilter and not just buildFilter

File: packages/orchestration/src/orchestration-util.ts
Comment (deekshas8) [reply packages/orchestration/src/orchestration-util.ts:1]: Will do

File: packages/orchestration/README.md (hunk 8 above)
Comment (ZhongpinWang) [reply packages/orchestration/README.md:339]: I get your point. Still that feels  like we are configuring something for using document grounding service. I can understand why orchestration service named the type in this way, because they are going to call document grounding service. But from our user's side,  they are configuring orchestration grounding and should not be confused. @jjtang1985 What do you think.

File: packages/orchestration/src/orchestration-util.ts (hunk 16 above)
Comment (deekshas8) [reply packages/orchestration/src/orchestration-util.ts:24]: I'll create a followup ticket for this. This function needs to be fixed anyway since it's incorrect to return a filters array from here anyway. Will also include the point you made here

File: packages/orchestration/README.md (hunk 8 above)
Comment (ZhongpinWang) [reply packages/orchestration/README.md:339]: Also, the type can be something else than `document_grounding_service`. So orchestration service by design can theoretically also utilizing other services besides document grounding service to do information retrieval.

Comment (deekshas8) [reply packages/orchestration/README.md:339]: > Still that feels like we are configuring something for using document grounding service.

This is, in fact, exactly what the user does with the orchestration service though—they are aware of and configure the specific external provider. We also don't encapsulate/hide this for other modules (filters, masking).

> So orchestration service by design can theoretically also utilizing other services

You're right. The structure here differs from masking and filtering, where each `type` has its own config structure and therfore makes sense for different convenience functions for different providers. I'll check with Govind to confirm if this `config` is meant to be harmonized. Then we dont call this function documentGrounding

File: packages/orchestration/src/orchestration-utils.test.ts [hunk 17]
```diff
@@ -0,0 +1,237 @@
...
+        filters: [
+          {
+            id: 'filter-id',
+            data_repositories: ['repo1', 'repo2'],
+            data_repository_type: 'custom-type'
+          }
+        ],
+        input_params: ['input'],
+        output_param: 'output'
+      };
```
KavithaSiva, line 220: [q] I played around a bit with if `null` values can be set for properties inside `filters` object. I found that for `id` and `data_repository_type` this is possible as all of them resolve to `any`.

Should we consider also restricting these values to string in our types for useful typehints in autocompletion?
Or do you think this should rather be fixed by the Orchestration team instead?

packages/orchestration/src/orchestration-filter-utility.ts
```diff
@@ -1,26 +0,0 @@
-import type {
```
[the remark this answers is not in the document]
deekshas8: For now we have just 2 util functions, so a single file seems enough, I dont want to have one file per function. If in the future we have a lot of functions, we can consider splitting. It's referenced internally anyway.

packages/orchestration/src/orchestration-utils.test.ts
```diff
@@ -0,0 +1,237 @@
...
+      });
+    });
+
+    it('overrides default data repository type', () => {
+      const groundingConfig: DocumentGroundingServiceConfig = {
+        filters: [
+          {
+            id: 'filter-id',
+            data_repositories: ['repo1', 'repo2'],
+            data_repository_type: 'custom-type'
```
deekshas8, line 215: @MatKuhr `data_repository_type` can be overridden.

packages/orchestration/src/orchestration-utils.test.ts
deekshas8: This unfortunately comes from the spec. @MatKuhr [This](https://github.tools.sap/AI/llm-orchestration/blob/97a4b722f52798ad56ca6e6cc33f623f0cc5ba0a/src/spec/api.yaml#L1143) seems wrong to me, combining and empty object and enum in schema. Can we suggest an improvement:

```
SearchSelectOptionEnum:
  anyOf:
    - type: string
      enum:
        - ignoreIfKeyAbsent
    - type: string
```

Also `GroundingFilterId` is `any` since the spec provides no type.

MatKuhr: I suspect the current state was inspired from [this discussion](https://github.com/OAI/OpenAPI-Specification/issues/1552#issuecomment-382640465). But in OpenAPI 3.0, the `anyOf` actually get's completely ignored here as far as I understand, and this is also how the Swagger Editor behaves. It seem there [hasn't been much progress on the matter since](https://github.com/OAI/OpenAPI-Specification/issues/1552#issuecomment-1003362981) 😄 

Using the above suggestion leads to problems with the Java generator. It can't deal with this at the moment, and it looks similar for the Python.

But I'm also not sure how beneficial this would be. Ultimately, the above is semantically equivalent to just writing `type: string`. So the only purpose is documentation, which might as well be moved to a description. 

➡️ I think we should just remove this from the spec and either declare `string` with a description, or declare the `enum` and expect clients to use the [appropriate generator option](https://openapi-generator.tech/docs/generators/openapi/#config-options) `enumUnknownDefaultCase`.

.changeset/shiny-brooms-punch.md
MatKuhr: > has nothing to do with document grounding service

But that's not really true, semantically it's the same service. And it's actually important that users know this, otherwise they don't know that they need to use both things together for the setup to work.

It's just that we are using Document Grounding via Orchestration. And arguably, this is a technical detail users don't even (need to) know about. We could also just call DG directly in a separate call, before calling orchestration. It's rather an implementation detail I'd say.

packages/orchestration/README.md
```diff
@@ -327,6 +327,23 @@ const response = await orchestrationClient.chatCompletion({
...
+```ts
+import {
+  buildDocumentGroundingConfig
+} from '@sap-ai-sdk/orchestration';
+
+const orchestrationConfig: {
+  ...,
+  grounding: buildDocumentGroundingConfig(
+    input_params: ['groundingInput'],
+    output_param: 'groundingOutput',
```
MatKuhr, line 341: [pp] I feel like this example isn't enough, as it leaves out how to configure the template, which isn't exactly obvious I'd say. Also, it's not intuitive what `'filter-id'` can/should be. If the ID is arbitrary, we may as well make it optional and create some ID for the user.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -56,3 +58,42 @@ export interface OrchestrationModuleConfig {
    */
   grounding?: GroundingModuleConfig;
 }
+
+/**
+ * Represents a filter configuration for the Document Grounding Service.
+ *
+ */
+export type DocumentGroundingServiceFilter = Pick<
```
MatKuhr, line 66: [q] Can't this be achieved with a type union or extension? This way we'll have to remember to add new fields..

MatKuhr (commented):
otherwise LGTM, assuming the azure filter stuff was only moved but not modified

packages/orchestration/src/orchestration-types.ts
deekshas8: I tried. The below doesn't work (optional modifier is not applied :

```
export type DocumentGroundingServiceFilter = DocumentGroundingFilter & {
      data_repository_type?: DataRepositoryType;
     }
```

I also tried:

```
export type DocumentGroundingServiceFilter = Omit<DocumentGroundingFilter, 'data_repository_type'> & {
      data_repository_type?: DataRepositoryType;
     }
```

This causes issues with auto-completion as `DocumentGroundingFilter` type in extensible (has Record<string, any>). 

But I agree to your point about keeping track of properties. I'll add a test to ensure we are always consistent.

packages/orchestration/README.md
deekshas8: The code example just above this (the Grounding section) has a detailed example (including templating). I just wanted to explain that the grounding part can be extracted to this function.

packages/orchestration/src/orchestration-utils.test.ts
deekshas8: So this actually works in TS ;) The type that gets generated is:

`export type SearchSelectOptionEnum = 'ignoreIfKeyAbsent' | string;`

The advantage being its extensible and provides autocompletiong. The current  `'ignoreIfKeyAbsent' | any;` causes loss of type safety. But if it doesn't work well with Java/Python, I understand the change doesnt make sense then. 

Regarding the solution, I also discussed with @tomfrenken. Having string type with a description sounds like a clearer solution. The SDK generator atleast doesn;t support `enumUnknownDefaultCase`.

For the `GroundingFilterId`, I was also not sure for what they do with this id and what is it supposed to be (since the spec has no type and resolves to any), and there didn;t introduce generating a random guid. I agree with the suggestion that it should be optional in the spec and if not provided, the service should handle it.

packages/orchestration/src/orchestration-types.ts
deekshas8: With the latest update (additionalProperties: false in GroundingModuleConfig), a simpler type is possible 👍🏽

packages/orchestration/README.md
MatKuhr: Ah okay, that should be sufficient. Maybe I'd switch the order though.. if users follow the instructions from top-down they only later learn that they could have used something simpler for most cases

packages/orchestration/src/orchestration-utils.test.ts
deekshas8: Opened a PR in orchestration repo to fix the GroundingFilterId type to string.

MatKuhr (approved):

packages/orchestration/README.md
```diff
@@ -287,6 +287,7 @@ return response.getContent();
 ### Grounding
 
 Grounding enables integrating external, contextually relevant, domain-specific, or real-time data into AI processes.
+The grounding configuration can be provided as a raw JSON object or by using the `buildDocumentGroundingConfig()` function, function, which requires only the minimal mandatory values.
```
deekshas8, line 290:
```suggestion
The grounding configuration can be provided as a raw JSON object or by using the `buildDocumentGroundingConfig()` function, which requires only the minimal mandatory values.
```

packages/orchestration/README.md
```diff
@@ -346,21 +347,16 @@ const orchestrationClient = new OrchestrationClient({
     ],
     defaults: {}
   },
-  grounding: {
-    type: 'document_grounding_service',
-    config: {
-      filters: [
+  grounding: buildDocumentGroundingConfig(
```
jjtang1985, line 350: Same updates can be applied to the [sample app](https://github.com/SAP/ai-sdk-js/blob/main/sample-code/src/orchestration.ts#L232)

jjtang1985 (approved):
LTGM.
We need to update the sample as well.

packages/orchestration/README.md
deekshas8: Done

PR #393, opened 2024-12-20
chore: Regenerate grounding clients with a default base path

KavithaSiva:
## Context

SAP/ai-sdk-js-backlog#187.

The OpenAPI generator in Cloud SDK, was updated to also make use of `basePath` from `options-per-service.json` file.
This PR regenerates the grounding client.

`openapi-request-builder` also has to be adapted to use the `basePath`


ZhongpinWang (approved):
LGTM

PR #419, opened 2025-01-02
chore: Explain how to format service key in .env

sacnl:

The previous explanation in the sample code readme how to add the service key was misleading. For example, the closing `'` was missing but is required.


deekshas8 (approved):
Thanks Henning for the suggestion 👍🏽

PR #424, opened 2025-01-03
Model params convenience

shibeshduw:
## Context

SAP/ai-sdk-js-backlog#83.

### DIFF
--- a/.changeset/sharp-pianos-press.md
+++ b/.changeset/sharp-pianos-press.md
@@ -1,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[Improvement] Add `modelParamOptions()` convenience function to provide suggestions for potential keys in `model_params` object in LLM Module Config.
--- a/packages/orchestration/src/orchestration-types.ts
+++ b/packages/orchestration/src/orchestration-types.ts
@@ -32,6 +32,18 @@ export type LlmModuleConfig = OriginalLlmModuleConfig & {
 };
 
 /**
+ * Module Parameters for LLM module configuration.
+ */
+export type LlmModelParams = {
+  max_tokens?: number;
+  temperature?: number;
+  frequency_penalty?: number;
+  presence_penalty?: number;
+  top_p?: number;
+  n?: number;
+} & Record<string, any>;
+
+/**
  * Orchestration module configuration.
  */
 export interface OrchestrationModuleConfig {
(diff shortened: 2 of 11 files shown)


packages/orchestration/src/orchestration-types.ts
```diff
@@ -31,6 +31,18 @@ export type LlmModuleConfig = OriginalLlmModuleConfig & {
   model_name: ChatModel;
 };
```
deekshas8, line 32:
```suggestion
model_params?: LlmModelParams;
};
```

How about if we do this? The user won't need to import another function.

shibeshduw: Yeah I thought of this at first but i thought we would want a more narrow suggestion for the user. I'm okay with this as well.

deekshas8: I would prefer this (more intuitive, no issue with discoverability). @tomfrenken WDYT?

tomfrenken: Fine with me as well, since we assume that these parameters are essentially always available, it makes sense to have it here too. We need to be aware though, that we are overriding the OriginalLlmModuleConfigs's model_param type then, right now it is Record<string,any>, it could be more narrow in the future though.

tests/type-tests/test/orchestration.test-d.ts
```diff
@@ -212,11 +206,20 @@ expectError<any>(
...
-      model_params: {}
+      model_params: { max_tokens: 50 }
     }
   }).chatCompletion()
 );
 
+/**
+ * Model parameters should accept known typed parameters and arbitrary parameters.
+ */
+expectType<LlmModelParams>({
```
shibeshduw, line 217: Apparently this is not allowed. The TS compiler performs the excess property check before `& Record<string, any>` is read so it considers this type to be "too wide"? Would like some suggestions.

deekshas8: I think the problem is not that it doesnt read Record<string, any> but rather that tsd expectType does strict type assertions -> https://github.com/tsdjs/tsd?tab=readme-ov-file#strict-type-assertions. 
As suggested used `expectAssignable` instead

.changeset/sharp-pianos-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[Improvement] Provide suggestions for known keys in `model_params` object in LLM Module Config by refining type definition.
```
deekshas8, line 5:
```suggestion
[Improvement] Refine the type definition of the `model_params` property in the `LlmModuleConfig` to also include known properties.
```

deekshas8 (approved):

PR #457, opened 2025-01-15
fix: Fix JSON stream logging

tomfrenken:
## Context

Closes SAP/ai-sdk-js-backlog#218.

The setting of an empty default object leads to a logging bug when a JSON is set, since this empty object doesn't serve any purpose, I removed it, thereby solving the logging bug.

I've also added the option to set global streaming options, this was previously not possible.


.changeset/sour-camels-fry.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[Fixed Issue] Fix logging bug when using streaming with JSON client.
+[New Functionality] Add option to set global streaming options in orchestration client class instantiation.
```
deekshas8, line 6:
```suggestion
[New Functionality] Add `streaming` option in the `OrchestrationModuleConfig` type to set global streaming options.
```

PR #466, opened 2025-01-17
chore: Add sample code for base64 encoded image in Orchestration

shibeshduw:
## Context

Closes SAP/ai-sdk-js-backlog#213.


sample-code/src/orchestration.ts
```diff
@@ -354,10 +356,14 @@ export async function orchestrationChatCompletionImage(): Promise<OrchestrationR
...
 
+  const imageFilePath = join(__dirname, 'src', 'media', 'sample-image.png');
+  const mimeType = lookup(extname(imageFilePath).toLowerCase());
+  const encodedString = `data:${mimeType};base64,${await readFile(imageFilePath, 'base64')}`;
+
   return orchestrationClient.chatCompletion({
     inputParams: {
-      imageUrl:
-        'https://upload.wikimedia.org/wikipedia/commons/thumb/5/59/SAP_2011_logo.svg/440px-SAP_2011_logo.svg.png'
+      // You can also provide the public URL of the image as a string in the code directly.
```
MatKuhr, line 365:
```suggestion
      // alternatively, you can provide a public URL of the image here instead.
```

sample-code/src/orchestration.ts
```diff
@@ -354,10 +356,14 @@ export async function orchestrationChatCompletionImage(): Promise<OrchestrationR
     }
   });
 
+  const imageFilePath = join(__dirname, 'src', 'media', 'sample-image.png');
+  const mimeType = lookup(extname(imageFilePath).toLowerCase());
```
MatKuhr, line 360: [pp] Computing the mime type is maybe a bit overkill, since we anyway hardcode png here in the filename. Thus might Asseln hard code the mime type ;)

tomfrenken: I agree that hardcoding the mime type here is probably enough, but please keep it separate from the template string, so our users still know which part to replace, in case their image is of a different type.

jjtang1985 (approved):

PR #467, opened 2025-01-17
chore: Update table of contents for orchestration doc

jjtang1985:
- Added streaming to table of contents
- Moved streaming to a bit below, as it's more advanced API. Also, it needs the context like templating, Json config etc.


tomfrenken (approved):
LGTM

PR #473, opened 2025-01-20
chore: Spec Update Automation

MatKuhr:
## Context

SAP/ai-sdk-js-backlog#66.

One job to rule all specs 😉


.github/workflows/spec-update.yaml
```diff
@@ -0,0 +1,168 @@
...
+        run: |
+          pnpm generate
+
+      - name: "Compile and Test"
+        id: compile
+        if: steps.spec_diff.outputs.spec_diff == 'true'
+        # Compilation can easily fail e.g. from re-namings and has to be fixed manually.
+        # Thus, this action raises the PR anyway and only reports success or failure of compilation and testing.
+        run: |
+          if pnpm compile ; then
```
ZhongpinWang, line 108: [req] Here need to run `pnpm install` first to install (dev-)dependencies, which will call `pnpm compile` afterwards anyway.

.github/workflows/spec-update.yaml
```diff
@@ -0,0 +1,168 @@
...
+        if: steps.spec_diff.outputs.spec_diff == 'true'
+        run: |
+          git config --global user.email "cloudsdk@sap.com"
+          git config --global user.name "SAP Cloud SDK Bot"
+          git add .
+          git commit -m "Update $CHOICE based on $REF"
+          git push --set-upstream origin ${{ steps.branch.outputs.branch }}
+
+      - name: "Create PR"
+        if: ${{ env.CREATE_PR == 'true' && steps.spec_diff.outputs.spec_diff == 'true'}}
```
ZhongpinWang, line 131: [q] Not entirely understanding the whole process. And this is rather a question related to the internal repo. Why is `--field create-pr=false` always set to false in the internal workflow?

.github/workflows/spec-update.yaml
```diff
@@ -0,0 +1,168 @@
...
+          You can commit on top of this branch, but as long as this PR is open the action can't be re-run.
+          
+          - Compilation outcome: ${{ steps.compile.output.compilation_result }}
+          - Test run outcome: ${{ steps.compile.output.test_result }}
+          
+          Before merging, make sure to update tests and release notes, if necessary.
+          
+          ## Definition of Done
+
+          - [ ] Unit tests cover new classes
```
ZhongpinWang, line 156: [pp] I would also remind adding type tests if applicable.

.github/workflows/spec-update.yaml
```diff
@@ -0,0 +1,168 @@
...
+      - uses: sap/ai-sdk-js/.github/actions/setup@main
+        if: steps.spec_diff.outputs.spec_diff == 'true'
+        with:
+          node-version: 20
+
+      - name: "Generate"
+        id: generate
+        if: steps.spec_diff.outputs.spec_diff == 'true'
+        run: |
+          pnpm generate
```
ZhongpinWang, line 100: I mean this command will generate for all packages. Ideally, this should not be a problem.

@KavithaSiva , after the base path PR, we forgot to generate for `ai-api`. Maybe it is acceptable to wait for the re-generation next time? Or we can also do this manually.

ZhongpinWang (changes requested):

.github/workflows/spec-update.yaml
KavithaSiva: Good catch @ZhongpinWang, I will do this manually.

.github/workflows/spec-update.yaml
```diff
@@ -0,0 +1,168 @@
...
+
+      - name: "Compile and Test"
+        id: compile
+        if: steps.spec_diff.outputs.spec_diff == 'true'
+        # Compilation can easily fail e.g. from re-namings and has to be fixed manually.
+        # Thus, this action raises the PR anyway and only reports success or failure of compilation and testing.
+        run: |
+          if pnpm compile ; then
+            echo "compilation_result=success" >> "$GITHUB_OUTPUT"
+            if pnpm test ; then
```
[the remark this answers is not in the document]
MatKuhr: I got baited because this does exist on some of the packages, but indeed not on the root of the project 😄

.github/workflows/spec-update.yaml
ZhongpinWang: Discussed.

.github/workflows/spec-update.yaml
```diff
@@ -0,0 +1,169 @@
...
+          pnpm install
+          pnpm generate
+
+      - name: "Compile and Test"
+        id: compile
+        if: steps.spec_diff.outputs.spec_diff == 'true'
+        # Compilation can easily fail e.g. from re-namings and has to be fixed manually.
+        # Thus, this action raises the PR anyway and only reports success or failure of compilation and testing.
+        run: |
+          if pnpm compile ; then
```
ZhongpinWang, line 109: [pp] Well, technically this step is not needed because compile is part of pnpm install. But anyway won't hurt to keep it similar to java version.

ZhongpinWang (approved):

.github/workflows/spec-update.yaml
MatKuhr: this compile is needed because install is before generate, so after generate the code has changed 😉

PR #477, opened 2025-01-21
chore: Cover grounding from `help.sap.com`

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#207.

Add e2e and documentation for using `help.sap.com` as data repository when using orchestration grounding module.


packages/orchestration/README.md
```diff
@@ -353,19 +353,24 @@ const orchestrationClient = new OrchestrationClient({
...
 const response = await orchestrationClient.chatCompletion({
-  inputParams: { groundingRequest: 'What is Generative AI Hub in SAP AI Core?' }
+  inputParams: { groundingRequest: 'Give me a short introduction of SAP AI Core.' }
 });
 return response.getContent();
 ```
 
+By default, the optional filter property `data_repository_type` is set to `vector`.
+Set it to `help.sap.com` to retrieve context from the SAP Help Portal.
+`REPOSITORY_ID` can also be a wildcard `*` to search all repositories.
```
deekshas8, line 372: The orchestration service has [*] as default, so I would mention the case where they want to search in a specific repo.

ZhongpinWang: Oh, didn't know that this field is optional, I will adjust the doc. Thanks!!

sample-code/src/orchestration.ts
```diff
@@ -326,6 +326,42 @@ export async function orchestrationGrounding(): Promise<OrchestrationResponse> {
...
+        }
+      ]
+    },
+    grounding: buildDocumentGroundingConfig({
+      input_params: ['groundingRequest'],
+      output_param: 'groundingOutput',
+      filters: [
+        {
+          id: 'filter1',
+          data_repositories: ['*'],
```
MatKuhr, line 351:
```suggestion
```
I would think this parameter is not meaningful for help

MatKuhr (commented):
Otherwise lgtm

PR #481, opened 2025-01-22
chore: Enforce generated code is up to date

MatKuhr:
## Context

Now that we have automation to update our specs, and all generated code is up to date, we should enforce that running `pnpm generate` never leads to changes when run on a clean `main` branch.


.github/workflows/build.yml
```diff
@@ -29,6 +29,9 @@ jobs:
       - uses: sap/ai-sdk-js/.github/actions/setup@main
       - name: REUSE Compliance Check
         uses: fsfe/reuse-action@v5
+      - run: pnpm generate
+      - name: Enforce generated code is up to date
+        run: git status --porcelain
```
ZhongpinWang, line 34: [req] Shouldn't we check if there is any output from this command? The command actually always succeeds or?

ZhongpinWang (changes requested):

.github/workflows/build.yml
MatKuhr: after experimenting a bit more, checking the output of `--porcelain` to not be empty is the only way I got to work reliably.

MatKuhr:
Yes, or if a spec file is changed, or a library is updated that causes different formatting, or the linter.

Since the check is very fast, and fixing it after the fact is annoying e.g. when testing spec updates, I'd prefer this to be enforced on PR level.

tomfrenken:
Our linter would catch linting changes already, no?
And spec updates are now automated, so this would only apply to rare manual spec updates and even more rare generator updates.
While I agree that it is not too long, I still feel like this is a very rare exception case (<1% of runs) which I don't like to run on every PR, because a lot of small 10s runs add up to minutes eventually.

tomfrenken (approved):
LGTM

PR #485, opened 2025-01-24
feat: Add descriptive constant for Azure content filter configuration

KavithaSiva:
## Context

Closes SAP/ai-sdk-js-backlog#181.

This PR adds self descriptive constants for improving Azure content filter configuration. 

Screenshot 2025-02-03 at 05 49 20

Autocompletion works with the string key values of `azureFilterThreshold`, the numeric allowed values don't come up in the hints.
These constants can only be used in the new convenience function introduced `buildAzureContentSafetyFilter()`
The old convenience function `buildAzureContentFilter()` still allows using numeric values.


tests/type-tests/test/orchestration.test-d.ts
```diff
@@ -244,6 +246,26 @@ expectType<Promise<OrchestrationResponse>>(
...
+    Violence: 'ALLOW_SAFE'
+  })
+);
+
+expectError<AzureContentSafetyFilterConfig>(
+  ContentFilters.buildAzureContentSafety({
+    Hate: 2,
+    SelfHarm: 4
+  })
+);
```
KavithaSiva, line 267: @jjtang1985 @deekshas8 This is different behaviour from what we discussed yesterday. But since we are only introducing constants in the new convenience function, I thought we can also just stick to string constants, instead of allowing numeric values as well.

Please tell me if you feel otherwise.

deekshas8 (dismissed):

packages/orchestration/README.md
```diff
@@ -249,7 +249,7 @@ This feature allows filtering both the [input](https://help.sap.com/docs/sap-ai-
 
 Use `ContentFilters.buildAzureContentSafety()` function to build an Azure content filter.
 The Azure content filter supports four categories: `Hate`, `Violence`, `Sexual`, and `SelfHarm`.
-Each category can be configured with severity levels of 0, 2, 4, or 6.
+Each category can be configured with severity levels of `ALLOW_SAFE`, `ALLOW_SAFE_LOW`, `ALLOW_SAFE_LOW_MEDIUM` and `ALLOW_ALL` which correspond to Azure threshold values of 0, 2, 4, or 6 respectively.
```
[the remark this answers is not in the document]
KavithaSiva: True, changed to a table.

packages/orchestration/src/index.ts
```diff
@@ -8,7 +8,14 @@ export type {
   StreamOptions,
   DocumentGroundingServiceConfig,
   DocumentGroundingServiceFilter,
-  LlmModelParams
+  LlmModelParams,
+  AzureContentFilter,
+  AzureFilterThresholdType
```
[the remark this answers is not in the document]
KavithaSiva: I was finding it difficult to come up with meaningful names which do not coincide with generated names.
I have renamed this now to `AzureThreshold`. 

This coincides with the generated type for the orchestration service, is that okay?

packages/orchestration/src/orchestration-types.ts
```diff
@@ -151,3 +151,51 @@ export interface DocumentGroundingServiceConfig {
...
+  /**
+   * The filter category for violence content.
+   */
+  Violence?: AzureFilterThresholdType;
+}
+
+/**
+ * A descriptive constant for Azure content safety filter threshold.
+ */
+export const AzureFilterThreshold = {
```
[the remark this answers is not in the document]
KavithaSiva: >[q] Does this need to be exported?

Yes, because I use it to fetch the corresponding integer value to construct the generated azure content filter type in the filtering convenience function.

>[req] Please follow styleguide for reusable const https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#use-camel-case-for-variable-names-

Renamed it to `azureFilterThreshold`.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -151,3 +151,51 @@ export interface DocumentGroundingServiceConfig {
...
+ * Type for azure filter threshold constant.
+ */
+export type AzureFilterThresholdType = keyof typeof AzureFilterThreshold;
+
+/**
+ * Type guard to check if a value is a valid azure filter threshold value.
+ * @param value - Value to check for type.
+ * @returns A Boolean value.
+ **/
+export function isAzureFilterThresholdType(
```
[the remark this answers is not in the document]
KavithaSiva: I was using the guard in the convenience functions, but am using `as` instead now.

```TS 
// config is of type `AzureContentFilter` and should be mapped to `AzureContentSafety`.
...Object.fromEntries(
  Object.entries(config).map(([key, value]) => [
    key,
    azureFilterThreshold[value as AzureThresholdLevel] // Used typeguard before.
  ])
)
```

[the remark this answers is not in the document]
KavithaSiva: Discussed over call, `as` can be used here.

packages/orchestration/src/index.ts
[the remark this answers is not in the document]
KavithaSiva: Discussed over call, we are renaming the type to `AzureFilterThreshold`.

packages/orchestration/src/orchestration-types.ts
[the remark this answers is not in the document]
KavithaSiva: >I still don't like the naming of the constants, honestly I would vote for using reasonable names instead of this mess, even if it breaks with the other sdks.

Launchpad [also uses](https://help.sap.com/docs/ai-launchpad/sap-ai-launchpad/input-filtering) similar threshold levels, and I guess these constants were picked by the other SDK's to also align with launchpad.
So, I would leave this the same for now, although I agree the const names are not very intuitive.

packages/orchestration/README.md
```diff
@@ -269,7 +269,14 @@ This feature allows filtering both the [input](https://help.sap.com/docs/sap-ai-
 
 Use `buildAzureContentSafetyFilter()` function to build an Azure content filter.
 The Azure content filter supports four categories: `Hate`, `Violence`, `Sexual`, and `SelfHarm`.
-Each category can be configured with severity levels of 0, 2, 4, or 6.
+Each category can be set to a specific severity level, each of which corresponds to an azure threshold value.
```
deekshas8, line 272:
```suggestion
Each category can be assigned a specific severity level, which corresponds to an Azure threshold value.
```

packages/orchestration/src/util/filtering.ts
```diff
@@ -33,13 +38,22 @@ export function buildAzureContentFilter(
  * @returns Filter config object.
  */
 export function buildAzureContentSafetyFilter(
```
deekshas8, line 40: [req] The func name in deprecation message is wrong:  @deprecated Since 1.8.0. Use {@link buildAzureContentSafety()} instead.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -166,3 +166,42 @@ export interface DocumentGroundingServiceConfig {
...
+   * The filter category for violence content.
+   */
+  Violence?: AzureFilterThreshold;
+}
+
+/**
+ * A descriptive constant for Azure content safety filter threshold.
+ * @internal
+ */
+export const azureFilterThreshold = {
```
deekshas8, line 196: while this is allowed and technically correct, I am think maybe having slightly different names would make it less confusing.

KavithaSiva: Renaming to `AzureFilterThresholdLevel` as I can't think of a better name.
But honestly these names are very close to `AzureContentSafety` and `AzureThreshold`, the generated types, so it will probably help if we can add some suffix to know that they are custom types. But, I don't have a good idea here.

But, since these are types are not exported anwyay, it should be okay right? Only that we might get confused between the types.

deekshas8: for the const you can use supportedAzureThresholds?

deekshas8: or supportedAzureFilterThresholds

KavithaSiva: I picked `azureFilterThreshold` to align [with Java,](https://github.com/SAP/ai-sdk-java/pull/169/files#diff-8842d71932fe24a1ef98d191462879e95cb9b8620dc82eca573af248df28231d), but since we don't export the constant anyway, I am okay with renaming.

KavithaSiva: The type is now called `AzureFilterThreshold` and the const is called `supportedAzureFilterThresholds`.
The const is bit wordy, but since it's not exported, I think that's okay.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -166,3 +166,42 @@ export interface DocumentGroundingServiceConfig {
...
+  ALLOW_SAFE_LOW: 2,
+  ALLOW_SAFE_LOW_MEDIUM: 4,
+  ALLOW_ALL: 6
+} as const;
+
+/**
+ * The Azure threshold level supported for each azure content filter category.
+ * @internal
+ */
+export type AzureFilterThreshold = keyof typeof supportedAzureFilterThresholds;
```
deekshas8, line 207: [req] Sorry I missed this earlier, but this type should be exported. It's part of the interface AzureContentFilter.

tests/type-tests/test/orchestration.test-d.ts
```diff
@@ -244,6 +246,26 @@ expectType<Promise<OrchestrationResponse>>(
 expect<ChatModel>('custom-model');
 expect<ChatModel>('gemini-1.0-pro');
 
+/**
+ * FilteringUtil.
```
deekshas8, line 250:
```suggestion
 * Filtering Util.
```

deekshas8 (approved):
Please fix the export of the AzureFilterThreshold type. Rest LGTM

KavithaSiva:
> Is the changelog part of the other PR? Ideally it should be part of this commit.

I will add it as part of the other PR, as the function gets introduced there, and only a convenience gets added here.

PR #488, opened 2025-01-24
fix: Fix LangChain client's internal mapping logic

tomfrenken:
## Context

Closes SAP/ai-sdk-js-backlog#226.

With this PR, the latest version of Azure OpenAI's responses are now properly mapped from LangChain to Azure and back to LangChain.


deekshas8 (approved):
LGTM

Issue #528, opened 2025-02-06
Reopen of Issue #490


FascinatedTV:
### Describe the Bug

As I cannot reopen the closed issue #490.

Unfortunately this [line](https://github.com/SAP/ai-sdk-js/commit/ccfa2eb288550e80ce6d9e64ff4da7ba8fe1692f#diff-dcce352cb16478eececb69f2cb2ec13f2c6df073e198836172848d90e28748bcR157) of release 1.7.0 introduces a new error.
As the initial value of "tool_calls" in the langchain "AiMessage" [object](https://github.com/langchain-ai/langchainjs/blob/458790627318c9b15ea1c34c22b0b9774d7f8e98/langchain-core/src/messages/ai.ts#L106) is "[]" not undefined, the property is send to the openAI chatCompletion endpoint with the empty array, which resolves in a 400 http error.

Not sure, if other Message types are effected also

### Steps to Reproduce

1. Create a langchain AIMessages
2. Pass it into the invoke call as context

### Expected Behavior

The "tool_calls" property should be removed, when it is empty or even undefined.

### Used Versions

- Node version via node -v: v20.18.0
- NPM version via `npm -v: 10.8.2
- SAP Cloud SDK for AI version:
  - "@sap-ai-sdk/langchain": "^1.7.0"

### Log File

Data payload (simplyfied)
```json
{
  "messages": [
    { "role": "system", "content": "You are a supervisor..." },
    { "role": "user", "content": "User query" },
    { "role": "function", "content": "...", "name": "function_name" },
    { "role": "assistant", "tool_calls": [!], "content": "..." },
    { "role": "user", "content": "What was my last question" }
  ],
  "tools": [{}],
  "tool_choice": "auto"
}

```

### Affected Development Phase

Development

### Impact

Blocked

KavithaSiva:
Hi @FascinatedTV,

We are looking into this issue and will create a fix soon.

Regards,
Kavitha

tomfrenken:
Hi @FascinatedTV, we've implemented the fix, it will be available in the next release, or starting tomorrow also in the canary release.

FascinatedTV:
Thank you!
Best regards

PR #534, opened 2025-02-10
fix: Add tool_calls only if non-empty (Langchain)

KavithaSiva:
## Context

Closes https://github.com/SAP/ai-sdk-js/issues/528

### DIFF
--- a/packages/langchain/src/openai/util.test.ts
+++ b/packages/langchain/src/openai/util.test.ts
@@ -74,8 +74,11 @@ describe('Mapping Functions', () => {
       ...defaultOptions,
       ...options
     });
-    // Todo find a better way to remove undefined properties
-    expect(JSON.parse(JSON.stringify(mapping))).toStrictEqual(request);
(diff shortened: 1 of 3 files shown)


jjtang1985 (changes requested):
I guess this is a "fix", instead of "chore", and we need a release notes.

packages/langchain/src/openai/util.test.ts
```diff
@@ -74,7 +74,8 @@ describe('Mapping Functions', () => {
       ...defaultOptions,
       ...options
     });
-    expect(mapping).toMatchObject(request);
+    // Todo find a better way to remove undefined properties
+    expect(JSON.parse(JSON.stringify(mapping))).toStrictEqual(request);
```
tomfrenken, line 78: Instead of changing the expect statement, just add another  one:

expect(mapping.messages.filter(message => message.role === "assistant")[0].tool_calls).toBeNotDefined()

tomfrenken (approved):
LGTM

PR #536, opened 2025-02-10
chore: Improve keyword

jjtang1985:
- [document-grounding] improve the keyword section of package.json
- [langchain] use description from the readme like other packages


tomfrenken (approved):
LGTM

PR #539, opened 2025-02-11
chore: Add new workflow for updating dependabot PRs

KavithaSiva:
## Context

Adds a new workflow to update the `pnpm-lock.yaml` on dependabot PRs

Dependabot PRs are not created with `pnpm v10`, hence we need an additional action to update the lock files generated with version 10 explicitly.

### DIFF
--- a/.github/workflows/dependabot-pnpm-lock.yml
+++ b/.github/workflows/dependabot-pnpm-lock.yml
@@ -1,0 +1,44 @@
+name: Update pnpm-lock.yaml for Dependabot PRs
+
+on:
+  pull_request:
+    paths:
+      - 'package.json'
+      - '**/package.json'
+
+jobs:
+  update-lockfile:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20' # TODO: Read from variables instead
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v3
+        with:
+          version: 10
+
+      - name: Update lockfile
+        run: pnpm install --lockfile-only
+
+      - name: Commit and push if changed
+        run: |
+          if [[ -n "$(git status --porcelain)" ]]; then
+            git config --global user.email "cloud-sdk-js@github.com"
+            git config --global user.name "cloud-sdk-js"  
+            git add pnpm-lock.yaml
+            git commit -m "chore: update pnpm-lock.yaml"
+            git push
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-pnpm-lock.yml
```diff
@@ -0,0 +1,46 @@
...
+      contents: write
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup Node.js
```
tomfrenken, line 25: We have a setup action, use that instead

.github/workflows/dependabot-pnpm-lock.yml
```diff
@@ -0,0 +1,46 @@
+name: Update pnpm-lock.yaml for Dependabot PRs
+
+on:
+  pull_request_target:
+    types: [opened, synchronize]
+    paths:
+      - 'package.json'
```
tomfrenken, line 7: redundant, as you already got **/package.json

.github/workflows/dependabot-pnpm-lock.yml
```diff
@@ -0,0 +1,46 @@
...
+        with:
+          ref: ${{ github.head_ref }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20' # TODO: Read from variables instead
+
+      - name: Install pnpm
```
tomfrenken, line 30: Also part of the setup

.github/workflows/dependabot-pnpm-lock.yml
tomfrenken: E.g.:

.github/workflows/dependabot-pnpm-lock.yml
KavithaSiva: But would **/package.json match the root level one?

KavithaSiva: Ah, you are right: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

.github/workflows/dependabot-pnpm-lock.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  update-lockfile:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@v4
```
tomfrenken, line 19: this is also done in the setup action, you can review how we do it in the auto-fix-lint action

.github/workflows/dependabot-pnpm-lock.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          ref: ${{ github.head_ref }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup Environment
+        uses: sap/ai-sdk-js/.github/actions/setup@main
+        with:
+          node-version: '20' # TODO: Read from variables instead
```
tomfrenken, line 27: nod everesion 20 is the default in the setup action, I think you can omit it.

tomfrenken (approved):
LGTM

PR #538, opened 2025-02-11
feat: Add Response Format Sample Code

MatKuhr:
## Context

Closes SAP/ai-sdk-js-backlog#209.

## What this PR does and why it is needed

Adds sample code and E2E test for using a strict response format with OpenAI.

- [x] Sample Code
- [x] E2E Test
- [x] ZOD schema support 
- [x] Docs

PR #543, opened 2025-02-12
chore: Improve UX for Orchestration chat scenario

shibeshduw:
## Context

Closes SAP/ai-sdk-js-backlog#228.

Summary:

- Added helper functions `getAllMessages()`, `getAssistantMessages`, `getToolCalls()` and `getRefusal()`.
- Added unit tests for the above.
- Added test cases for using `getAllMessages()` to provide `messageHistory` in subsequent chat completion requests.
- Added test case to check if orchestration responses with `tool_calls` can now be added to the `messageHistory` on subsequent chat completion requests.

Tested on intber landscape with latest spec: https://github.tools.sap/AI/llm-orchestration/blob/main/src/spec/api.yaml

Docs PR: https://github.com/SAP/ai-sdk/pull/79


packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,6 +55,72 @@ export class OrchestrationResponse {
...
+      };
+    }
+
+    throw new Error(
+      `Unexpected role with complex message: ${multiMessage.role}`
+    );
+  }
+
+  private isValidChatRole(message: ChatMessage) {
+    if (Array.isArray((message as MultiChatMessage).content)) {
```
MatKuhr, line 118: Im not sure we can make this check. Orchestration doesn't guarantee roles, and more might come in the future (eg developer or tool)

Our code should be lenient and handle them gracefully

MatKuhr (commented):
Otherwise lgtm

packages/orchestration/src/orchestration-response.ts
shibeshduw: I guess we can leave out the check altogether then as the API already gives us an error message when the wrong role is passed for eg. in the case where non-user role is passed for image content.

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,6 +55,54 @@ export class OrchestrationResponse {
     return choice?.message?.content;
   }
 
+  /**
+   * Messages that can be used for subsequent prompts as message history.
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns A list of all messages.
+   */
+  getAllMessages(choiceIndex = 0): ChatMessages {
```
ZhongpinWang, line 63: Just checked the other PR https://github.com/SAP/ai-sdk-js/pull/538/files#r1955748987

Don't forget to adjust readme.

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,6 +55,54 @@ export class OrchestrationResponse {
...
+        return isMultiContent
+          ? this.handleMultiChatMessage(message as MultiChatMessage)
+          : this.handleSingleChatMessage(message as SingleChatMessage);
+      }
+    );
+
+    messages.push({
+      role: 'assistant',
+      content:
+        this.getChoices().find(c => c.index === choiceIndex)?.message.content ??
```
ZhongpinWang, line 78: [q] Ummm I am not sure if index cannot be found, we should push empty string in to messages. It is possible that the last message got filtered and in this case, we should maybe leave the message out?

packages/orchestration/src/orchestration-response.test.ts
```diff
@@ -38,6 +38,21 @@ describe('OrchestrationResponse', () => {
     });
   });
 
+  it('should return return a list of all messages', () => {
```
deekshas8, line 41:
```suggestion
  it('should return a list of all messages', () => {
```

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,6 +55,54 @@ export class OrchestrationResponse {
...
 
+  /**
+   * Messages that can be used for subsequent prompts as message history.
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns A list of all messages.
+   */
+  getAllMessages(choiceIndex = 0): ChatMessages {
+    const messages = (this.data.module_results.templating ?? []).map(
+      (message: ChatMessage) => {
+        const isMultiContent = Array.isArray(
```
deekshas8, line 66: [req] Please use a type guard here so you don't need the as type assertion. 

```
  isMultiChatMessage(params: ChatMessage): params is MultiChatMessage {
    return Array.isArray(params.content);
  }
  ```

Comment (deekshas8) [inline packages/orchestration/src/orchestration-response.ts:66]: [req] please consider the naming here. `isMultiChatMessage` is more speaking for what you're checking and subsequent function calls are also named accoridngly

File: packages/orchestration/src/orchestration-response.ts [hunk 6]
```diff
@@ -51,6 +55,54 @@ export class OrchestrationResponse {
...
+          (message as MultiChatMessage).content
+        );
+        return isMultiContent
+          ? this.handleMultiChatMessage(message as MultiChatMessage)
+          : this.handleSingleChatMessage(message as SingleChatMessage);
```
deekshas8, line 71: [q] Why is `handleSingleChatMessage` mapping function needed ? Can't we just return message?

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,6 +55,54 @@ export class OrchestrationResponse {
     return choice?.message?.content;
   }
 
+  /**
+   * Messages that can be used for subsequent prompts as message history.
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns A list of all messages.
+   */
+  getAllMessages(choiceIndex = 0): ChatMessages {
+    const messages = (this.data.module_results.templating ?? []).map(
```
deekshas8, line 64: [q] Why the transformation? Everything you get from the templating results can be passed as it is in message history right?

deekshas8 (changes requested):

packages/orchestration/src/orchestration-response.ts
shibeshduw: Just preference really in this case. I feel like having this mapping reads better with the `handleMultiChatMessage` mapping.

packages/orchestration/src/orchestration-response.ts
shibeshduw: Yes that is true but if the user would use this convenience for anything else, they would have to deal with complex objects when dealing with `MultiChatMessage`.

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,6 +55,56 @@ export class OrchestrationResponse {
...
+   * Messages that can be used for subsequent prompts as message history.
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns A list of all messages.
+   */
+  getAllMessages(choiceIndex = 0): ChatMessages {
+    const messages: ChatMessage[] = (
+      this.data.module_results.templating ?? []
+    ).map(message =>
+      this.isMultiChatMessage(message)
+        ? this.handleMultiChatMessage(message as MultiChatMessage)
```
deekshas8, line 68: [req] You dont need an assertion here now, since the type guard takes care of that.

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,6 +55,56 @@ export class OrchestrationResponse {
...
+        : this.handleSingleChatMessage(message as SingleChatMessage)
```
deekshas8, line 69: [req] I would just return the message from here, because the mapping is not required. The message type is already how you want it to be

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,6 +55,56 @@ export class OrchestrationResponse {
...
+    );
+
+    const content = this.getChoices().find(c => c.index === choiceIndex)
```
deekshas8, line 72: [q] How do we handle cases when the assistant response has empty content and `tool_calls`? It should retain that structure.

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,6 +55,56 @@ export class OrchestrationResponse {
...
+      ?.message.content;
+    return content ? [...messages, { role: 'assistant', content }] : messages;
+  }
+
+  private handleSingleChatMessage(
```
deekshas8, line 77: [req] Please remove this method

deekshas8 (changes requested):

packages/orchestration/src/orchestration-response.ts
shibeshduw: Do you mean both the mapping or just `handleSingleChatMessage`?

packages/orchestration/src/orchestration-response.ts
deekshas8: Sorry, I am not quite sure I understand. What else would the user use this for?
I thought the main idea here is to pass the existing templating messages and the new response message back as history. The template should be assignable as it is, since the types are same. This will also not handle `tool_calls`.

packages/orchestration/src/orchestration-response.ts
deekshas8: just handleSingleChatMessage.

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,6 +54,48 @@ export class OrchestrationResponse {
     return choice?.message?.content;
   }
 
+  /**
+   * Messages that can be used for subsequent prompts as message history.
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns A list of all messages.
+   */
+  getAllMessages(choiceIndex = 0): ChatMessages {
```
KavithaSiva, line 62: [req] Could you please add an additional test in `orchestration-client.test.ts` with the a sample response json file with a `multiChatMessage` response?

I wasn't able to understand the code without looking at a sample response from the Java repo, so I think this would be beneficial for us too.

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,6 +54,48 @@ export class OrchestrationResponse {
...
+    return {
+      role: multiChatMessage.role,
+      content: multiChatMessage.content
+        .map(content =>
+          content.type === 'text'
+            ? content.text
+            : JSON.stringify({
+                url: content.image_url.url,
+                detail: content.image_url.detail
+              })
```
KavithaSiva, line 89: [q] Are other content types possible? Is this documented somewhere that only `text` and `image_url` are allowed for now? I feel it is better to also check if the type is `image_url` before stringifying.

KavithaSiva (changes requested):
Mostly looks good, but I would prefer a test added with the a sample multiChat response.

packages/orchestration/src/orchestration-response.ts
shibeshduw: Removed this transformation so I don't think this is relevant anymore.

ZhongpinWang (commented):
will continue

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,6 +54,28 @@ export class OrchestrationResponse {
...
+    )?.message;
+    return content ? [...messages, content] : messages;
+  }
+
+  /**
+   * Gets the assistant message from the response.
+   * @param choiceIndex - The index of the choice to use (default is 0).
+   * @returns The assistant message.
+   */
+  getAssistantMessage(choiceIndex = 0): AssistantChatMessage {
```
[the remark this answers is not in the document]
shibeshduw: Good call. I don't think this was in scope for the ticket but it makes sense to just add those here.

sample-code/src/server.ts
```diff
@@ -247,6 +248,7 @@ app.get('/orchestration/:sampleCase', async (req, res) => {
       simple: orchestrationChatCompletion,
       template: orchestrationTemplating,
       templateRef: orchestrationPromptRegistry,
+      messageHistory: orchestrationMessageHistory,
```
ZhongpinWang, line 251: [req] Add the usage to the `README.md`

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -100,12 +101,20 @@ describe('orchestration', () => {
...
-      result.data.orchestration_result.choices[0].message.tool_calls;
+    const assistantMessage = result.getAssistantMessage();
 
-    expect(tool_calls).toHaveLength(1);
-    expect(tool_calls[0].function.name).toBeDefined();
-    expect(tool_calls[0].function.arguments).toBeDefined();
+    if (assistantMessage.tool_calls) {
+      expect(assistantMessage.tool_calls).toHaveLength(1);
+      expect(assistantMessage.tool_calls[0].function.name).toBeDefined();
+      expect(assistantMessage.tool_calls![0].function.arguments).toBeDefined();
```
ZhongpinWang, line 109: [q] Any reason why sometimes use `!`?

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -100,12 +101,20 @@ describe('orchestration', () => {
...
     expect(result.getFinishReason()).toBe('tool_calls');
 
-    const tool_calls =
-      result.data.orchestration_result.choices[0].message.tool_calls;
+    const assistantMessage = result.getAssistantMessage();
 
-    expect(tool_calls).toHaveLength(1);
-    expect(tool_calls[0].function.name).toBeDefined();
-    expect(tool_calls[0].function.arguments).toBeDefined();
+    if (assistantMessage.tool_calls) {
```
ZhongpinWang, line 106: [req] `expect(...).toBeDefined()`

But I don't think you need to check all those things this thoroughly. Just `expect(assistantMessage.tool_calls[0].function.name).toBeDefined();` or for `arguments` as well should already be fine. If anything undefined or array is empty, we get error anyways.

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -100,12 +101,20 @@ describe('orchestration', () => {
...
+      expect(assistantMessage.tool_calls[0].function.name).toBeDefined();
+      expect(assistantMessage.tool_calls![0].function.arguments).toBeDefined();
+    } else {
+      throw new Error('Expected tool_calls to be defined');
+    }
+  });
+
+  it('should complete a chat when message history with tool calls is passed', async () => {
+    const result = await orchestrationMessageHistoryWithToolCalling();
+    assertContent(result);
```
ZhongpinWang, line 117: [pp] In general, I would make e2e test more concise. This test case covers the previous chat completion test, and overlaps a bit with the tool call one. I would maybe move the tool call test into this one and remove the previous chat completion one. E2E should be considered both from a technical perspective (from client to service) and scenario perspective (a real process including tool calls, chat completion and message history).

tests/type-tests/test/orchestration.test-d.ts
```diff
@@ -84,6 +85,19 @@ expectType<TokenUsage>(
...
 );
 
+expectType<ChatMessages>(
+  (
+    await new OrchestrationClient({
+      templating: {
+        template: [{ role: 'user', content: 'Hello!' }]
+      },
+      llm: {
+        model_name: 'gpt-35-turbo-16k'
```
ZhongpinWang, line 95: [req] 35 turbo 16k is about to retire. Use 4o in the example.

ZhongpinWang (changes requested):

sample-code/src/orchestration.ts
```diff
@@ -145,6 +150,31 @@ export async function orchestrationTemplating(): Promise<OrchestrationResponse>
...
+/**
+ * Chat request to OpenAI through the Orchestration service using message history.
+ * @returns The orchestration service response.
+ */
+export async function orchestrationMessageHistory(): Promise<OrchestrationResponse> {
+  const orchestrationClient = (messages: TemplatingChatMessage) =>
+    new OrchestrationClient({
+      llm,
+      templating: {
+        template: messages
```
deekshas8, line 162: [req] There is no need of an arrow function here. Create a client object with template { role: 'user', content: '{{?message}}' } and pass the question as inputParam. Use the same in both the calls. (The orchestration api for template is bad, but this is a cleaner way to demonstrate what we're doing.)

sample-code/src/orchestration.ts
```diff
@@ -586,3 +617,69 @@ export async function orchestrationToolCalling(): Promise<OrchestrationResponse>
...
+  ) =>
+    new OrchestrationClient({
+      llm,
+      templating: {
+        template: messages,
+        tools: [tools]
+      }
+    });
+
+  const response: OrchestrationResponse = await orchestrationToolCalling();
```
deekshas8, line 662: The example in its current state is hard to understand. Ideally the same client should be called (similar to my above comment). Tool calls only make sense with a message history, so I'll just keep all code (after refactoring) into one single function `orchestrationWithToolCalling`.

deekshas8 (dismissed):

sample-code/src/orchestration.ts
ZhongpinWang: I feel we are not able to reuse the same client object. For tool calls, we need to send a request with the `tool` role instead of `user`.

ZhongpinWang: Maybe we should allow user to change templating object for each call.

deekshas8: Ah I forgot the bad api design again :/

shibeshduw: We don't need to edit the templating object though for this example. The request with the `tool` role is sent as part of the message history in the subsequent request. So I think we can reuse the same client object and pass different messages through input params

```TypeScript
    llm,
    templating: {
      template: [
        { role: 'user', content: '{{?message}}' }
      ],
      tools: [...]
    }
  });
  
const response: OrchestrationResponse = await orchestrationClient.chatCompletion(
    {inputParams: {message: '...'}}
);
const initialResponse: AssistantChatMessage | undefined =
    response.getAssistantMessage(); 
if(initialResponse.tool_calls) construct message history
return orchestrationClient.chatCompletion({
    inputParams: { message: 'new message'},
    messagesHistory: allMessages
});
```

I guess this is what you meant @deekshas8 ?

deekshas8: @shibeshduw  Kinda, but without having to send an empty user message in the end (I know this works in OpenAI, but not sure about other models. Also not a good design imo).
Ideally the messages should be:
- First, `User` message asking a question
- llm responds with `Assistant` message
- We call tool, and send `Tool` message back to llm
- llm then responds with `Assistant` message

Currently with a mandatory template (and us setting it as `user` message), this cannot work out of the box. 
I believe with this [ticket](https://github.com/SAP/ai-sdk-js-backlog/issues/293) we should be able to achieve the correct behavior without workarounds.

deekshas8: We can keep the sample as what you already have, but create a follow up to improve them once the optional template ticket is done. @shibeshduw wdyt?

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,7 +55,55 @@ export class OrchestrationResponse {
     return choice?.message?.content;
   }
 
+  /**
+   * Parses the orchestration response and returns the tool calls generated by the model.
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns The message tool calls.
+   */
+  getToolCalls(choiceIndex = 0): MessageToolCalls | undefined {
```
shibeshduw, line 63: Assuming multiple tool calls are supported in orchestration, we return the list of `MessageToolCall` objects.

sample-code/src/orchestration.ts
shibeshduw: Should we create a separate followup seeing as the ticket already is making improvements to the langchain message history scenario. It looks sensible to just update the sample code as part of the same ticket.

deekshas8: Also fine.

packages/orchestration/src/orchestration-response.ts
```diff
@@ -51,7 +55,55 @@ export class OrchestrationResponse {
...
+  getAssistantMessage(choiceIndex = 0): AssistantChatMessage | undefined {
+    return this.findChoiceByIndex(choiceIndex)?.message;
+  }
+
   private getChoices() {
     return this.data.orchestration_result.choices;
   }
+
+  private findChoiceByIndex(index: number) {
+    // TODO: replace cast with LLMChoice[] after the bug in orchestration, where
```
ZhongpinWang, line 103: Create a BLI for this and link it here.

ZhongpinWang (approved):

.changeset/fuzzy-kids-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[New Functionality] Introduce convenience functions `getAllMessages()`, `getAssistantMessage()`, `getToolCalls` and `getRefusal()` for orchestration.
```
ZhongpinWang, line 5:
```suggestion
[New Functionality] Introduce convenience functions `getAllMessages()`, `getAssistantMessage()`, `getToolCalls()` and `getRefusal()` for orchestration.
```

Issue #547, opened 2025-02-13
[AI-API] Destination is not used to resolve deploymentId


FascinatedTV:
### Describe the Bug

When providing the destinationName on AzureOpenAIChatCompletion instantiation,
the chat client fails to retrieves the deploymentId as it still looks for a "aicore" service binding.

### Steps to Reproduce

Instantiate AzureOpenAiChatClient with "modelName" and "destinationName" (optional: "resourceGroup", "modelVersion" or "useCache: false")

Invoke chat client

### Expected Behavior

If the destinationName is provided, the deploymentId should be solved via the provided destination.

### Used Versions

- Node version via `node -v`: v20.18.0
- NPM version via `npm -v`: 10.8.2
- SAP Cloud SDK for AI version:
  - @sap-ai-sdk/langchain@1.7.0
  - @sap-cloud-sdk/http-client@3.26.0
  - @sap-cloud-sdk/resilience@3.26.0

### Log File

 WARN     (environment-accessor): Could not find service binding of type 'aicore'. This might cause errors in other parts of the application.

### Affected Development Phase

Release

### Impact

Blocked

tomfrenken:
Hi @FascinatedTV thanks for reporting this bug, I've created a fix which will be part of our next release.

tomfrenken:
We just released version v1.8.0, let us know if your issue is resolved 👍

PR #548, opened 2025-02-13
fix: Include destinations in deployment id lookup

tomfrenken:
## Context

We ignored destinations in our deployment id lookup and only included it in the requests we execute.
With this PR, we also do the look-up flow based on the destination (if any was provided).

Relates to #547

### DIFF
--- a/.changeset/purple-lizards-buy.md
+++ b/.changeset/purple-lizards-buy.md
@@ -1,0 +1,6 @@
+---
+'@sap-ai-sdk/foundation-models': minor
+'@sap-ai-sdk/ai-api': minor
+---
+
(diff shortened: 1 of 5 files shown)


.changeset/purple-lizards-buy.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ai-sdk/foundation-models': minor
+'@sap-ai-sdk/ai-api': minor
+---
```
deekshas8, line 4:
```suggestion
'@sap-ai-sdk/orchestration': minor
---
```

deekshas8 (approved):

PR #555, opened 2025-02-17
feat: Add convenience function for LlamaGuard Filter

KavithaSiva:
## Context

Closes SAP/ai-sdk-js-backlog#205.

This PR adds a convenience function to enable filter categories for Llama guard filter.

Screenshot 2025-02-20 at 14 41 31

The built filter can then be used in the filtering configuration of the Orchestration client.

### DIFF
--- a/.changeset/hip-tools-smile.md
+++ b/.changeset/hip-tools-smile.md
@@ -2,4 +2,4 @@
 '@sap-ai-sdk/orchestration': minor
 ---
 
-[New Functionalities] Introduce `buildLlamaGuardFilter()` method to enable providing Llama guard content filters to the input and output filter configuration of the orchestration client.
+[New Functionality] Introduce `buildLlamaGuardFilter()` method to enable providing Llama guard content filters to the input and output filter configuration of the orchestration client.
--- a/.changeset/hip-tools-smile.md
+++ b/.changeset/hip-tools-smile.md
@@ -1,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[New Functionalities] Introduce `buildLlamaGuardFilter()` method to enable providing Llama guard content filters to the input and output filter configuration of the orchestration client.
--- a/.changeset/hip-tools-smile.md
+++ b/.changeset/hip-tools-smile.md
@@ -2,4 +2,4 @@
 '@sap-ai-sdk/orchestration': minor
 ---
 
-[New Functionality] Introduce `buildLlamaGuardFilter()` function to enable providing Llama guard content filters to the input and output filter configuration of the orchestration client.
+[New Functionality] Introduce `buildLlamaGuardFilter()` convenience function to build Llama guard filters.
--- a/.changeset/hip-tools-smile.md
+++ b/.changeset/hip-tools-smile.md
@@ -2,4 +2,4 @@
 '@sap-ai-sdk/orchestration': minor
 ---
 
-[New Functionality] Introduce `buildLlamaGuardFilter()` method to enable providing Llama guard content filters to the input and output filter configuration of the orchestration client.
+[New Functionality] Introduce `buildLlamaGuardFilter()` function to enable providing Llama guard content filters to the input and output filter configuration of the orchestration client.
--- a/packages/orchestration/README.md
+++ b/packages/orchestration/README.md
@@ -389,8 +389,6 @@ Each category of the filter can be assigned a specific severity level, which cor
 | `ALLOW_SAFE_LOW_MEDIUM` | 4                     |
 | `ALLOW_ALL`             | 6                     |
 
-The following example demonstrates how to build an Azure content filter with specific severity levels for each category.
-
 ```ts
 const filter = buildAzureContentSafetyFilter({
   Hate: 'ALLOW_SAFE_LOW',
@@ -404,8 +402,6 @@ Use `buildLlamaGuardFilter()` function to build a Llama Guard content filter.
 
 Available categories can be found with autocompletion.
 Pass the categories as arguments to the function to enable them.
-
-The following example demonstrates how to build a Llama Guard content filter with specific categories enabled.
 
 ```ts
 const filter = buildLlamaGuardFilter('hate', 'violent_crimes');
--- a/packages/orchestration/README.md
+++ b/packages/orchestration/README.md
@@ -403,8 +403,7 @@ const filter = buildAzureContentSafetyFilter({
 Use `buildLlamaGuardFilter()` function to build a Llama Guard content filter.
 
 Available categories can be found with autocompletion.
-By default, all categories are disabled.
-Pass the categories you want as arguments to the function to enable them.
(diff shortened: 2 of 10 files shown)


packages/orchestration/src/util/filtering.test.ts
```diff
@@ -211,4 +212,33 @@ describe('Content filter util', () => {
...
+          violent_crimes: false,
+          sexual_content: true
+        }
+      };
+      expect(filterConfig).toEqual(expectedFilterConfig);
+    });
+
+    it('builds filter config with no config', async () => {
+      const filterConfig = buildLlamaGuardFilter();
+      const expectedFilterConfig = {
```
deekshas8, line 237: This shouldnt be allowed as config is a [mandatory property](https://github.tools.sap/AI/llm-orchestration/blob/main/src/spec/api.yaml#L856) and requires a min of 1 key.

packages/orchestration/src/util/filtering.test.ts
```diff
@@ -211,4 +212,33 @@ describe('Content filter util', () => {
       );
     });
   });
+
+  describe('Llama Guard filter', () => {
+    it('builds filter config with custom config', async () => {
+      const filterConfig = buildLlamaGuardFilter({
+        child_exploitation: false,
+        hate: true,
+        violent_crimes: false,
```
deekshas8, line 221: Setting properties to false can lead to confusion as skipping a property means it's false anyway

packages/orchestration/src/util/filtering.ts
```diff
@@ -58,3 +60,17 @@ export function buildAzureContentSafetyFilter(
...
   };
 }
+
+/**
+ * Convenience function to create Azure content filters.
+ * @param config - Configuration for Llama guard filter.
+ * @returns Filter config object.
+ */
+export function buildLlamaGuardFilter(
+  config?: LlamaGuard38B
```
deekshas8, line 70: [req] config should be a required property.

deekshas8 (commented):
A suggestion:

Since the default for is false for all properties of the filter, the user only needs to set categories which they want to enable. As such, setting to `true` seems like a redundant effort. We could allow the user to just pass an array of strings (from union type of all existing properties) and we set them to true. 
The pro here would be less typing and less confusion (setting false makes no sense, but this might not very obvious).
The con is it deviates from azure filter a bit where we have a `key: value` style of input.

packages/orchestration/src/util/filtering.test.ts
KavithaSiva: We don't really have `minProperties` of 1, so the resulting type for `LlamaGuard38b` ends up being a [type with all properties as optional](https://github.com/SAP/ai-sdk-js/blob/main/packages/orchestration/src/client/api/schema/llama-guard-38-b.ts).

KavithaSiva:
> A suggestion:
> 
> Since the default for is false for all properties of the filter, the user only needs to set categories which they want to enable. As such, setting to `true` seems like a redundant effort. We could allow the user to just pass an array of strings (from union type of all existing properties) and we set them to true. The pro here would be less typing and less confusion (setting false makes no sense, but this might not very obvious). The con is it deviates from azure filter a bit where we have a `key: value` style of input.

I would personally prefer if the user explicitly sets the properties to `true` as even if it is verbose, it is clear that we are turning it on.

```TS
buildLlamaGuardFilter({ hate: true, sexual_content: true})  //Very clear
buildLlamaGuardFilter([hate', 'sexual_content']) //Not clear, if we are enabling or disabling the category 
```

packages/orchestration/src/util/filtering.ts
KavithaSiva: My idea was to apply default if nothing is provided by the user.
Orchestration service does not allow an empty config `{}` object to be sent. So I applied a default of `false` for all categories.

packages/orchestration/src/util/filtering.test.ts
KavithaSiva: Removed `false` properties, but the user can still set `false` as I directly use the generated type.

deekshas8:
> buildLlamaGuardFilter([hate', 'sexual_content']) //Not clear, if we are enabling or disabling the category 

From my understanding of the filter, you basically can enable a category, the default is false for all the rest categories. What I personally find more confusing is allowing to disable something thats going to be disabled anyway.

packages/orchestration/src/util/filtering.ts
KavithaSiva: No longer relevant.

packages/orchestration/src/util/filtering.test.ts
KavithaSiva: No longer relevant.

.changeset/hip-tools-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[New Functionality] Introduce `buildLlamaGuardFilter()` method to enable providing Llama guard content filters to the input and output filter configuration of the orchestration client.
```
deekshas8, line 5:
```suggestion
[New Functionality] Introduce `buildLlamaGuardFilter()` function to enable providing Llama guard content filters to the input and output filter configuration of the orchestration client.
```

packages/orchestration/src/util/filtering.ts
```diff
@@ -58,3 +60,19 @@ export function buildAzureContentSafetyFilter(
     })
   };
 }
+
+/**
+ * Convenience function to create Llama guard filters.
+ * @param categories - Categories to be enabled for filtering. A minimum of one category must be provided.
```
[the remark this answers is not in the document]
deekshas8: Maybe also for azure in its description

ZhongpinWang:
Side note: Don't worry about the long commit history above.. I did a git rebase and it makes me the co-author of all previous commits..

deekshas8 (approved):
Left 2 minor comments regarding docs. Else LGTM

packages/orchestration/README.md
```diff
@@ -385,6 +375,43 @@ try {
...
+  Violence: 'ALLOW_SAFE_LOW_MEDIUM'
+});
+```
+
+#### Llama Guard Filter
+
+Use `buildLlamaGuardFilter()` function to build a Llama Guard content filter.
+
+Available categories can be found with autocompletion.
+By default, all categories are disabled.
```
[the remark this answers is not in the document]
ZhongpinWang: Thx. On a second thought, I think you are right. And it sounds also clear to me without adding this.

packages/orchestration/README.md
```diff
@@ -385,6 +375,43 @@ try {
...
+Pass the categories you want as arguments to the function to enable them.
+
+The following example demonstrates how to build a Llama Guard content filter with specific categories enabled.
```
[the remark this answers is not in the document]
ZhongpinWang: I was also quite hesitating about this  😆 since I normally want to put something before a code block. 
But I guess it is annoying in the end. Removed.

PR #581, opened 2025-03-04
chore: Improve error handling by using `ErrorWithCause`

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#235.

This PR introduces `ErrorWithCause` for all request execution using `executeRequest()` from core.

The most useful information is contained in `error.stack` which will be by default printed out if not caught using `try-catch`.

I also have to make changes to `sample-code` since accessing response data is now a layer deeper wrapper by `cause`.

Also cleaned up `sample-code` a bit. 

Docs are added for general error handling in case of a failing request.


sample-code/README.md
ZhongpinWang, line 1: Some docs were missing.

Also mask grounding input should have been placed as part of the orchestration group.

sample-code/src/orchestration.ts
```diff
@@ -196,11 +198,12 @@ export async function orchestrationInputFiltering(): Promise<void> {
     });
     throw new Error('Input was not filtered as expected.');
   } catch (error: any) {
-    if (error.response?.status === 400) {
```
ZhongpinWang, line 199: I have to make these changes due to the introduction of `ErrorWithCause`

packages/ai-api/src/utils/deployment-resolver.test.ts
```diff
@@ -114,6 +117,44 @@ describe('deployment resolver', () => {
 
     expect(id).toEqual('5');
   });
+
+  describe('get all deployments', () => {
```
tomfrenken, line 121: Why did we randomly introduce a new test here?

ZhongpinWang: Long story. Because the first task of the BLI was to improve this function. And I first add `ErrorWithCause` with tests only to this one. And then I realize that it can be extracted to a lower level execute request. But since the test is harmless, I left it their.

packages/ai-api/README.md
```diff
@@ -23,6 +23,7 @@ We maintain a list of [currently available and tested AI Core APIs](https://gith
   - [Delete a Deployment](#delete-a-deployment)
   - [Custom Destination](#custom-destination)
 - [Local Testing](#local-testing)
+- [Error Handling](#error-handling)
```
deekshas8, line 26: [pp] I would put Error handling before local testing

packages/orchestration/README.md
```diff
@@ -413,12 +415,12 @@ Both `chatCompletion()` and `getContent()` methods can throw errors.
 
 - **Axios Errors**:  
   When the chat completion request fails with a `400` status code, the caught error will be an `Axios` error.
-  The property `error.response.data.message` provides additional details about the failure.
+  Check `error.cause.response.data` for the error response body.
```
tomfrenken, line 418: Is this the best way to handle an error with ErrorWithCause, or can a property of the new Error be used instead?

packages/orchestration/README.md
```diff
@@ -23,6 +23,7 @@ This package incorporates generative AI orchestration capabilities into your AI
   - [Custom Request Configuration](#custom-request-configuration)
   - [Custom Destination](#custom-destination)
 - [Local Testing](#local-testing)
+- [Error Handling](#error-handling-1)
```
deekshas8, line 26:
```suggestion
- [Error Handling](#error-handling)
```

packages/orchestration/README.md
deekshas8, line 418: [pp] Maybe instead of repeating this, we can just add it once as a tip/note

README.md
```diff
@@ -108,6 +109,28 @@ This setup enables local testing of clients such as orchestration and OpenAI, pr
...
+A common error scenario is `Request failed with status code STATUS_CODE` coming from `AxiosError`. 
+In this case, SAP Cloud SDK for AI uses [`ErrorWithCause`](https://sap.github.io/cloud-sdk/docs/js/features/error-handling) to provide more detailed error information.
+
+The following example shows how to access useful information from the error.
+
+```ts
+try {
+  ... // execute request
+} catch (e) {
+  // Print error messages from different layers
```
tomfrenken, line 108: A little bit more detail, perhaps with an example error here would be helpful, even if we already link to the cloud SDK docs here.

tomfrenken (changes requested):
Overall LGTM, but the ## Error Handling section needs a bit more detail, so that users do not have to jump to the cloud SDK docs to understand how to use it properly.

packages/orchestration/README.md
ZhongpinWang: This is added automatically by my vscode extension and somehow vscode navigates correctly..

README.md
ZhongpinWang: Improved

packages/orchestration/README.md
ZhongpinWang: Unfortunately no. Because `ErrorWithCause` is meant to be generic also for errors that are not `AxiosError`. So user has to access the response data by themselves in case it comes from Axios. But at least we print the response for them in the stack already.

.changeset/itchy-months-wash.md
```diff
@@ -0,0 +1,9 @@
+---
+'@sap-ai-sdk/ai-api': patch
+'@sap-ai-sdk/core': patch
+---
+
+[Improvement] Introduce `ErrorWithCause` for request error handling.
```
jjtang1985:
```suggestion
[Fixed Issue] Include the root cause in the error object when handling HTTP request errors, providing more context for debugging.
```

jjtang1985 (commented):
I would accept the breaking change, and treat that as a bug fix, as I did not know the response body was not included.
Please have a look at the suggestion for the release notes.

packages/orchestration/README.md
ZhongpinWang: Love this idea. Improved.

tomfrenken (approved):
LGTM

PR #591, opened 2025-03-10
chore: Update model list

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#244.

Add `o1`, `o3-mini` and `alephalpha-pharia-1-7b-control` to the available model list.

### DIFF
--- a/.changeset/shy-socks-shake.md
+++ b/.changeset/shy-socks-shake.md
@@ -1,5 +1,5 @@
 ---
-'@sap-ai-sdk/core': patch
+'@sap-ai-sdk/core': minor
 ---
 
 [Improvement] Add `o1`, `o3-mini` and `alephalpha-pharia-1-7b-control` to the available model list.
--- a/.changeset/shy-socks-shake.md
+++ b/.changeset/shy-socks-shake.md
@@ -1,0 +1,5 @@
+---
+'@sap-ai-sdk/core': patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/shy-socks-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/core': patch
```
deekshas8, line 2: This should be a minor release in my opinion.

deekshas8: I would say from the users perspective, the impacted packages are orchestration and foundation-models.

deekshas8 (approved):
Left a comment for the changelogs, otherwise LGTM.

.changeset/shy-socks-shake.md
ZhongpinWang: I changed it to minor.

Regarding the second point, I am still in favour of noting the changes as part of the `core` package. As almost all functionality in `core` is meant to be used by other packages, it will basically ask us to never put `core` in release note. And some changes could affect very widely, such as introducing `ErrorWithCause` in very low level in `core`. In such case, I prefer not to put all other packages here except `core`, even if all other packages has zero code changes.

deekshas8: Sorry for the confusion, I didn't imply removing core from affected packages in the changeset, rather also including the other 2.
Since we say core is internal and not for users, the changelog might imply that the change is restricted to core. However, the main change is seen in the orchestration/foundation-models by the users. The release notes are mainly for public api changes.

PR #604, opened 2025-03-19
feat: Add convenience for masking

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#231.

Add convenience function `buildDpiMaskingProvider()`.

### DIFF
--- a/packages/orchestration/src/util/masking.ts
+++ b/packages/orchestration/src/util/masking.ts
@@ -8,19 +8,20 @@ import type { DpiMaskingConfig } from '../orchestration-types.js';
  * @returns An object with the masking provider configuration.
  */
 export function buildDpiMaskingProvider(dpiMaskingConfig: DpiMaskingConfig): DpiConfig {
+  const { method, mask_grounding_input, entities, allowlist } = dpiMaskingConfig;
   return {
     type: 'sap_data_privacy_integration',
-    method: dpiMaskingConfig.method,
-    entities: dpiMaskingConfig.entities.map(entity => ({
+    method,
+    entities: entities.map(entity => ({
       type: entity
     })),
-    ...(dpiMaskingConfig.mask_grounding_input !== undefined && {
+    ...(mask_grounding_input !== undefined && {
       mask_grounding_input: {
-        enabled: dpiMaskingConfig.mask_grounding_input
+        enabled: mask_grounding_input
       }
     }),
-    ...(dpiMaskingConfig.allowlist && {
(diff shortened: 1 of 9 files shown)


tests/type-tests/test/orchestration.test-d.ts
```diff
@@ -13,8 +14,10 @@ import type {
   GroundingModuleConfig,
   LlmModelParams,
   AzureContentSafetyFilterConfig,
-  LlamaGuard38BFilterConfig
+  LlamaGuard38BFilterConfig,
+  DpiConfig
 } from '@sap-ai-sdk/orchestration';
+import { buildDpiMaskingProvider } from '@sap-ai-sdk/orchestration/src/util/masking.js';
```
tomfrenken, line 20: Merge import

packages/orchestration/src/util/masking.ts
```diff
@@ -0,0 +1,26 @@
...
+import type { DpiConfig } from '../client/api/schema/index.js';
+import type { DpiMaskingConfig } from '../orchestration-types.js';
+
+/**
+ * Convenience function to create masking provider SAP Data Privacy Integration.
+ * @param dpiMaskingConfig - Configuration for the masking provider SAP Data Privacy Integration.
+ * @returns An object with the masking provider configuration.
+ */
+export function buildDpiMaskingProvider(dpiMaskingConfig: DpiMaskingConfig): DpiConfig {
+  return {
```
tomfrenken, line 11: As discussed, would be looking a bit better if you restructure the different properties of the dpiMaskingConfig so that you don't have to write out the entire object's name so many times.

tests/type-tests/test/orchestration.test-d.ts
```diff
@@ -5,6 +5,7 @@ import {
   buildDocumentGroundingConfig,
   buildLlamaGuardFilter
 } from '@sap-ai-sdk/orchestration';
+import { buildDpiMaskingProvider } from '@sap-ai-sdk/orchestration/src/util/masking.js';
```
tomfrenken, line 8: [req] Update the exports and then update the imports

tomfrenken (changes requested):
LGTM for the most part, only two parts I would like to be improved

tomfrenken (changes requested):
[req] Only missing thing for me now is a little bit more info in the docs, e.g. that there are N (currently N=1) utility functions which are all build functions, or any description of this function, rather than only using it in the examples.

tomfrenken (approved):
LGTM

PR #620, opened 2025-03-27
feat: Support LangChain OrchestrationClient with content filtering module

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#260.

1. Use `onFailedAttempt()` to bypass `maxRetries`.
2. Add sample code for using content filtering in the LangChain OrchestrationClient.
3. Improve documentation, especially for both LangChain clients.


sample-code/src/foundation-models/azure-openai.ts
```diff
@@ -20,7 +20,7 @@ const logger = createLogger({
  * @returns The response from Azure OpenAI containing the response content.
  */
 export async function chatCompletion(): Promise<AzureOpenAiChatCompletionResponse> {
-  const response = await new AzureOpenAiChatClient('gpt-35-turbo').run({
+  const response = await new AzureOpenAiChatClient('gpt-4o').run({
```
ZhongpinWang, line 23: gpt-35-turbo is deprecated. And some other models are still deprecated in our repo. Will be removed in a follow-up.

packages/langchain/src/openai/README.md
```diff
@@ -46,30 +44,48 @@ import {
...
 const chatClient = new AzureOpenAiChatClient({
   modelName: 'gpt-4o',
-  modelVersion: '24-07-2021',
+  modelVersion: '2024-08-06',
   resourceGroup: 'my-resource-group'
 });
 ```
 
-**Do not pass a `deployment ID` to initialize the client.**
-For LangChain model clients, initialization requires specifying the model name, model version, and resource group.
```
ZhongpinWang, line 63: This was confusing to me that model version and resource group are optional. But then said again with the word `required`.

packages/langchain/src/openai/README.md
```diff
@@ -46,30 +44,48 @@ import {
...
+  modelName: 'text-embedding-3-small'
+});
 ```
 
 In addition to the default parameters of Azure OpenAI and LangChain, additional parameters can be used to help narrow down the search for the desired model:
 
 ```ts
 const chatClient = new AzureOpenAiChatClient({
   modelName: 'gpt-4o',
-  modelVersion: '24-07-2021',
```
ZhongpinWang, line 57: I believe we need to use the english version of the SAP Note. Otherwise, the translation is sometimes messing up with the date format etc.

packages/langchain/src/orchestration/README.md
```diff
@@ -98,6 +94,8 @@ const client = new OrchestrationClient(orchestrationConfig, {
 });
 ```
 
+If the error is caused by input content filtering, the client will throw immediately without retrying.
```
ZhongpinWang, line 97: This is the key change of this PR.

packages/langchain/src/orchestration/client.ts
```diff
@@ -31,6 +38,15 @@ export class OrchestrationClient extends BaseChatModel<
     public deploymentConfig?: ResourceGroupConfig,
     public destination?: HttpDestinationOrFetchOptions
   ) {
+    // Avoid retry if the error is due to input filtering
+    const { onFailedAttempt } = langchainOptions;
+    langchainOptions.onFailedAttempt = error => {
```
ZhongpinWang, line 43: To better understand how this works -> https://github.com/sindresorhus/p-retry?tab=readme-ov-file#onfailedattempterror

sample-code/src/langchain-orchestration.ts
```diff
@@ -1,39 +1,112 @@
...
-      model_params: {}
+      model_name: 'gpt-4o'
     },
     // define the template
     templating: {
       template: [
         {
           role: 'user',
-          content: 'Give me a long introduction of {{?input}}'
+          content: 'Tell me about {{?topic}}'
```
ZhongpinWang, line 24: [pp] I would in general avoid using the term `input` in the input params. It is confusing with the input concept for content filtering.

sample-code/src/langchain-orchestration.ts
```diff
@@ -1,39 +1,112 @@
...
+  };
 
-  // invoke the chain
-  return llmChain.invoke('My Message History', callOptions);
+  return new OrchestrationClient(orchestrationConfig)
+    .pipe(new StringOutputParser())
+    .invoke('My Message History', {
+      inputParams: {
+        topic:
+          '30 different ways to rephrase "I hate you!" with strong feelings'
```
ZhongpinWang, line 109: Apparently, now GPT models are more polite... I have to increase the number so that the output filter gets triggered.

sample-code/src/orchestration.ts
```diff
@@ -236,7 +235,8 @@ export async function orchestrationOutputFiltering(): Promise<OrchestrationRespo
     messagesHistory: [
       {
         role: 'system',
-        content: 'Reparaphrase the sentence in ten ways: "{{?input}}"'
+        content:
```
ZhongpinWang, line 238: Same here, hope it makes our e2e more reliable.

KavithaSiva (commented):
Mostly looks good to me, the solution with `onFailedAttempt` is very clean.
Left few minor comments.

packages/langchain/src/openai/README.md
ZhongpinWang: I feel since we are saying "In addtion to the default parameters", it is not a bad idea to show user some other value possible.

packages/langchain/src/orchestration/client.ts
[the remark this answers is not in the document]
ZhongpinWang: Thanks for reminding me. Added.

packages/langchain/src/openai/README.md
```diff
@@ -46,30 +44,48 @@ import {
...
-const embeddingClient = new AzureOpenAiEmbeddingClient({
-  modelName: 'gpt-4o',
-  maxRetries: 0
-});
+const response = await client.invoke(messageHistory, { timeout: 10000 });
+```
+
+##### Retry
+
+LangChain clients retry up to 6 times by default.
```
ZhongpinWang:
```suggestion
LangChain clients retry up to 6 times with exponential backoff by default.
```

packages/langchain/src/openai/README.md
```diff
@@ -81,7 +97,7 @@ For example, to target `my-destination`, use the following code:
 const chatClient = new AzureOpenAiChatClient(
   {
     modelName: 'gpt-4o',
-    modelVersion: '24-07-2021',
+    modelVersion: '2024-08-06',
```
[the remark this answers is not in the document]
ZhongpinWang: Replied somewhere else.

packages/langchain/src/openai/README.md
ZhongpinWang: And yes, the machine translated version apparently also converts the date format.

PR #625, opened 2025-03-31
feat: Support setting `metadata_params` property in `buildDocumentGroundingConfig()` convenience function

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#267.

`buildDocumentGroundingConfig()` now support `metadata_params` as an optional parameter. When defining this property, the grounding result will include metadata as part of the JSON string.

Sample result for grounding module:

```json
{
  "message": "grounding result",
  "data": {
    "grounding_query": "grounding call",
    "grounding_result": "[{\"content\": \"The last SAP AI SDK JavaScript end to end test was run at 1743431349413.\", \"metadata\": {\"context\": [\"sap-ai-sdk-js\"]}}]"
  }
}
```

Since `metadata` is now part of the `grounding_result` it can be used / mentioned in the prompt.

### DIFF
--- a/packages/orchestration/README.md
+++ b/packages/orchestration/README.md
@@ -471,18 +471,18 @@ const orchestrationClient = new OrchestrationClient({
     ],
     defaults: {}
   },
-  grounding: buildDocumentGroundingConfig(
+  grounding: buildDocumentGroundingConfig({
     input_params: ['groundingRequest'],
     output_param: 'groundingOutput',
     // metadata_params: ['PARAM_NAME']
     filters: [
-        {
-          id: 'FILTER_ID',
-          // data_repository_type: 'vector', // optional, default value is 'vector'
-          data_repositories: ['REPOSITORY_ID'],
-        }
-      ],
-    )
+      {
+        id: 'FILTER_ID',
+        // data_repository_type: 'vector', // optional, default value is 'vector'
+        data_repositories: ['REPOSITORY_ID'],
+      }
+    ],
+  })
 });
 
 const response = await orchestrationClient.chatCompletion({
--- a/packages/orchestration/README.md
+++ b/packages/orchestration/README.md
@@ -479,14 +479,16 @@ const orchestrationClient = new OrchestrationClient({
       {
         id: 'FILTER_ID',
         // data_repository_type: 'vector', // optional, default value is 'vector'
-        data_repositories: ['REPOSITORY_ID'],
+        data_repositories: ['REPOSITORY_ID']
       }
-    ],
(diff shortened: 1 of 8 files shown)


packages/orchestration/README.md
```diff
@@ -471,28 +471,32 @@ const orchestrationClient = new OrchestrationClient({
...
+  }
 });
 return response.getContent();
 ```
 
 By default, the optional filter property `data_repository_type` is set to `vector`.
 Set it to `help.sap.com` to retrieve context from the SAP Help Portal.
-Set `data_respotiories` property with an array of `REPOSITORY_ID` values to search in specific data repositories.
+Set `metadata_params` property with an array of parameter names to include [metadata](https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/metadata) in the grounding result, which can be used in the prompt.
+Set `data_respotiories` property with an array of repository IDs to search in specific data repositories.
```
KavithaSiva, line 499:
```suggestion
Set `data_repositories` property with an array of repository IDs to search in specific data repositories.
```

sample-code/src/server.ts
```diff
@@ -461,6 +461,13 @@ app.get(
         `Orchestration responded with timestamp:\t${groundingResult.getContent()}\n`
       );
 
+      // Print the grounding data.
+      const groundingResultString =
+        groundingResult.data.module_results.grounding?.data?.grounding_result;
+      res.write(
+        `Orchestration grounding metadata:\t${JSON.stringify(JSON.parse(groundingResultString)[0].metadata)}\n`
```
KavithaSiva, line 468: Just an observation that I read in the [documentation](https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/metadata#procedure), if a metadata name is present on multiple levels a `parent_KeyName` would be followed, so  for e.g `webUrl` if it was added in the chunk and `document_webUrl` if it was added in the metadata.

We would have to take note of this, when we start providing a convenience to read `module_results`.

KavithaSiva (approved):
LGTM, it would have also been nice if the grounding colleagues, had a convenience to include all metadata using `*`.

sample-code/src/server.ts
ZhongpinWang: I am not sure if we want to or can provide convenience for metadata, as the `data` field of grounding module can be any object, currently including `grounding_result` property, and `grounding_result` is just a JSON string including `metadata` as part of the string. They are not really typed and meant to be read by human, rather as part of the prompt and read by LLM.

PR #632, opened 2025-04-02
feat: Support YAML string as input for templating module

deekshas8:
## Context

Closes SAP/ai-sdk-js-backlog#225.

This PR allows users to pass a YAML string as input to the templating module instead of a full JS object. This string is validated against the generated zod schema, and an error is thrown if its invalid. If its valid, its is parsed and the object is assigned to config.templating.

Please feel free to suggest  a better name for OrchestrationModuleConfigWithStringTemplating,


scripts/zod-config-prompt-registry.ts
```diff
@@ -0,0 +1,44 @@
...
+import { createLogger } from '@sap-cloud-sdk/util';
+import { readFile } from 'fs/promises';
+import { existsSync } from 'fs';
+import { generateAndWriteConfig } from './generate-zod-config.js';
+
+const logger = createLogger('generate-zod-config');
+
+// Recursively finds all imported files
+async function findImportedFiles(filePath: string, collectedFiles = new Set<string>()): Promise<void> {
+    console.log(filePath);
```
ZhongpinWang, line 11: [req] I am thinking maybe keep this as a debug log would be helpful.

test-util/data/orchestration/orchestration-chat-completion-yaml-template.ts
```diff
@@ -0,0 +1,38 @@
...
+          theme:
+            type: string
+            description: The central theme or subject of the poem
+          verses:
+            type: array
+            description: A list of verses making up the poem
+            items:
+              type: string
+              description: A single verse of the poem
+`;
```
ZhongpinWang, line 38:
```suggestion
`;

```

test-util/data/orchestration/orchestration-chat-completion-yaml-template.ts
ZhongpinWang, line 1: [req] Can we create a yaml file and then use the existing `parseFileToString()` function in `mock-http.ts`?

tests/schema-tests/package.json
```diff
@@ -15,9 +15,5 @@
     "generate-schema:orchestration": "pnpm generate-config:orchestration && pnpm ts-to-zod --all --skipValidation",
     "generate-schema": "pnpm generate-schema:azure-openai && pnpm generate-schema:orchestration",
     "postgenerate-schema": "rm ./ts-to-zod.config.cjs"
-  },
-  "devDependencies": {
-    "ts-to-zod": "^3.15.0",
-    "zod": "^3.24.2"
```
ZhongpinWang, line 21: [q] Was removing this causing any issue to you (maybe in the pipeline)? If I remember correctly, I made a compromise here to not let the workflow fail. But maybe it works now idk.

eslint.config.js
```diff
@@ -37,7 +37,8 @@ export default [
       '**/*.d.ts',
       '**/dist/**/*',
       '**/coverage/**/*',
-      'packages/**/client/**/*'
+      'packages/**/client/**/*',
+      'packages/prompt-registry/src/zod/*'
```
ZhongpinWang, line 41: [q] I see that after zod generation, `pnpm lint:fix` is executed. Should we then allow checking for those generated files as they are going to be part of our codebase?

ZhongpinWang (commented):
Using generated Zod to verify yaml feels feasible to me. (Although due to the limitation of ts-to-zod, it is still quite cumbersome to introduce this process.) Left some suggestions for improvements.

eslint.config.js
deekshas8: I was in two minds about it. I'll add linting for them, but ignore some of the rules like jsdoc/require-documentation.

test-util/data/orchestration/orchestration-chat-completion-yaml-template.ts
deekshas8: Will change this t yaml, so ignoring this now

scripts/zod-config-prompt-registry.ts
```diff
@@ -0,0 +1,44 @@
...
+async function main(args: string[]) {
+  if (args.length !== 2) {
+    logger.error('Usage: node --loader ts-node/esm generate-zod-config.ts <inputFile> <outputSchema>');
+    process.exit(1);
+  }
+
+  const [inputFile, outputSchema] = args;
+  try {
+    const collectedFiles = new Set<string>();
+    await findImportedFiles(inputFile, collectedFiles);
```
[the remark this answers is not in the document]
deekshas8: Unfortunately, when I just put a single file (which I want to be the starting type), the tool itself does not recursively find all related files. It would be a nice feature if they did this out-of-the-box

scripts/zod-config-prompt-registry.ts
deekshas8: Was a leftover actually. Removed it.

tests/schema-tests/package.json
deekshas8: Nope, pnpm should pick the dependency from the root node_modules.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -93,6 +91,17 @@ export interface OrchestrationModuleConfig {
   streaming?: GlobalStreamOptions;
 }
 
+/**
+ * Orchestration module configuration with string templating.
+ * This type is used when the `templating` property can also be a string.
+ */
+export type OrchestrationModuleConfigWithStringTemplating = Omit<
```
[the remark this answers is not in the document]
deekshas8: I implemented this. Had to introduce a new type `LangchainOrchestrationModuleConfig` and adjust the sample code so users use this type if they want to type the config.

packages/langchain/src/orchestration/types.ts
```diff
@@ -20,3 +25,14 @@ export type OrchestrationCallOptions = Pick<
   tools?: Template['tools'];
   inputParams?: Prompt['inputParams'];
 };
+
+/**
+ * Orchestration module configuration with string templating.
+ * This type is used when the `templating` property can also be a string.
+ */
+export type LangchainOrchestrationModuleConfig = Omit<
```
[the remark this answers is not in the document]
deekshas8: exactly. The thought was to make clear the distinction, otherwise it might be confusing to have 2 exact types from 2 different packages.
Regarding allowing string template in langchain, I am not sure. If the idea is to allow easy testing of config, then orchestration might be enough (same for passing orchestrationModuleConfig as json string, which we dont allow in langchain) @jjtang1985 wdyt? Is this needed? 
Its not supported out of the box. We merge configs in langchain(add tools to templating), so will require some effort.

packages/langchain/src/orchestration/types.ts
```diff
@@ -20,3 +25,14 @@ export type OrchestrationCallOptions = Pick<
   tools?: Template['tools'];
   inputParams?: Prompt['inputParams'];
 };
+
+/**
+ * Orchestration module configuration with string templating.
+ * This type is used when the `templating` property can also be a string.
```
ZhongpinWang, line 31: Can also be or cannot be

ZhongpinWang (commented):
Other than the above comment, lgtm

packages/langchain/src/orchestration/types.ts
deekshas8: sorry I copied over the description from the old type.

packages/langchain/src/orchestration/types.ts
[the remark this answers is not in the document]
jjtang1985: As discussed, local prompt templates is not needed for LangChain in this ticket.
Let's collect feedback from the stakeholders than make a decision later.

PR #633, opened 2025-04-02
chore: Support LangGraph

tomfrenken:
## Context

Closes SAP/ai-sdk-js-backlog#258.

With this PR, we make some type adjustments, which make the langchain client compatible with LangGraph.
We also add an example code for proper usage.

**IMPORTANT NOTE:** Until the AI platform doesn't respond with proper tool responses, proper agents with tool calls won't be supported, while the client is already capable of it, the platform is not.


Avenel:
Hi @tomfrenken,

I'm excited about this PR and eagerly anticipate its merge. Could you please let us know when it might be merged and available in the library?

Thank you!

pnpm-workspace.yaml
```diff
@@ -1,14 +1,17 @@
...
+  - packages/document-grounding
+  - packages/prompt-registry
+  - sample-cap
+  - sample-code
+  - tests/e2e-tests
+  - tests/schema-tests
+  - tests/smoke-tests
+  - tests/type-tests
+onlyBuiltDependencies:
+  - better-sqlite3
```
deekshas8, line 16: [q] Do we need these?

deekshas8 (approved):
LGTM

PR #642, opened 2025-04-07
fix: Warning in `ts-jest` regarding `isolatedModules`

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#278.

When running jest, we keep seeing the following error.

```
ts-jest[ts-compiler] (WARN) Using hybrid module kind (Node16/18/Next) is only supported in "isolatedModules: true". Please set "isolatedModules: true" in your tsconfig.json.
```

In this PR:

- I set `isolatedModules` to `true`.
- Change global setup files to `mjs` as after adding this attribute, jest won't recognize the file as `esm` and `mts` did not work either..
- Model types in `core` package should have been exported as type only. I changed them.


deekshas8 (approved):

global-test-teardown.mjs
```diff
@@ -0,0 +1,6 @@
+/**
+ * This file is used to run code after all tests have been run.
+ */
+export default async function () {
```
ZhongpinWang, line 4:
```suggestion
export default async function tearDown() {
```

PR #661, opened 2025-04-16
chore: Add more checks and test

deekshas8:
## Context

Adds additional check for empty string and tests. Improve error handling

### DIFF
--- a/packages/orchestration/src/orchestration-client.test.ts
+++ b/packages/orchestration/src/orchestration-client.test.ts
@@ -473,7 +473,11 @@ describe('orchestration service client', () => {
       templating: ''
     };
 
-    expect(() => new OrchestrationClient(invalidConfigWithYaml).chatCompletion({ inputParams: { topic: 'Generative AI Hub' } })).toThrow();
+    expect(() =>
+      new OrchestrationClient(invalidConfigWithYaml).chatCompletion({
+        inputParams: { topic: 'Generative AI Hub' }
+      })
+    ).toThrow();
   });
 
   it('calls chatCompletion with grounding configuration', async () => {
--- a/packages/orchestration/src/orchestration-client.test.ts
+++ b/packages/orchestration/src/orchestration-client.test.ts
@@ -477,7 +477,9 @@ describe('orchestration service client', () => {
       new OrchestrationClient(invalidConfigWithYaml).chatCompletion({
         inputParams: { topic: 'Generative AI Hub' }
       })
-    ).toThrow();
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"Templating YAML string must be non-empty."`
+    );
   });
 
   it('calls chatCompletion with grounding configuration', async () => {
--- a/packages/orchestration/src/orchestration-client.test.ts
+++ b/packages/orchestration/src/orchestration-client.test.ts
@@ -413,6 +413,7 @@ describe('orchestration service client', () => {
             description: 'Structured format for the generated poem',
             strict: true,
             schema: {
+              type: 'object',
               additionalProperties: false,
               required: ['title', 'verses', 'theme'],
               properties: {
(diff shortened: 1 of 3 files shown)


packages/orchestration/src/orchestration-client.ts
```diff
@@ -46,10 +46,11 @@ export class OrchestrationClient {
...
     if (typeof config === 'string') {
       this.validateJsonConfig(config);
-    } else if (typeof config.templating === 'string') {
-      this.config = this.parseAndMergeTemplating(config); // parse and assign if templating is a string
     } else {
-      this.config = config as OrchestrationModuleConfig; // TypeScript cannot infer that config.templating is not a string
+      this.config =
+        typeof config.templating === 'string'
+          ? this.parseAndMergeTemplating(config) // parse and assign if templating is a string
+          : config; // TypeScript cannot infer that config.templating is not a string
```
ZhongpinWang, line 53: [q] Do we still need this comment?

packages/orchestration/src/orchestration-client.ts
```diff
@@ -168,30 +169,34 @@ export class OrchestrationClient {
   private parseAndMergeTemplating(
     config: OrchestrationModuleConfig
   ): OrchestrationModuleConfig {
+    let parsedObject;
+    if (typeof config.templating === 'string' && !config.templating.trim()) {
+      throw new Error('Templating must be a non-empty YAML string.');
```
ZhongpinWang, line 174:
```suggestion
      throw new Error('Templating YAML string must be non-empty.');
```

packages/orchestration/src/orchestration-client.ts
```diff
@@ -168,30 +169,34 @@ export class OrchestrationClient {
...
+      ...config,
+      templating: {
+        template: result.data.spec.template,
+        ...(result.data.spec.defaults && {
+          defaults: result.data.spec.defaults
+        }),
+        ...(result.data.spec.response_format && {
+          response_format: result.data.spec.response_format
+        }),
+        ...(result.data.spec.tools && { tools: result.data.spec.tools })
```
ZhongpinWang, line 198: [pp] Looks like we can extract `result.data.spec` to make the code more readable.

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -463,6 +464,22 @@ describe('orchestration service client', () => {
...
+        model_params: { max_tokens: 500 }
+      },
+      templating: ''
+    };
+
+    expect(() =>
+      new OrchestrationClient(invalidConfigWithYaml).chatCompletion({
+        inputParams: { topic: 'Generative AI Hub' }
+      })
+    ).toThrow();
```
ZhongpinWang, line 480: [pp] Would be nice to compare the message thrown.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -168,30 +169,34 @@ export class OrchestrationClient {
...
+    const result = promptTemplatePostRequestSchema.safeParse(parsedObject);
+    if (!result.success) {
+      throw new Error(
+        `Prompt Template YAML does not conform to the defined type. Validation errors: ${result.error}`
+      );
+    }
+    return {
+      ...config,
+      templating: {
+        template: result.data.spec.template,
```
ZhongpinWang, line 191: I think we have or will have issues here once we adopt the new orchestration spec.

In prompt registry spec, `role` can still be any string, but orchestration has changed this part to a union type... I am facing an error now.

packages/orchestration/src/orchestration-client.ts
deekshas8: Oops.

deekshas8, line 53:
```suggestion
          : config;
```

packages/orchestration/src/orchestration-client.ts
deekshas8: I added a type assertion preemptively so this will be fine. Ultimately they should update the types in Prompt Template as well so users creating a template also use it correctly (will mention this to Thomas). But for us now, we stick to what the prompt registry spec says and validate accordingly and assume the user put the correct values.

ZhongpinWang (approved):
lgtm, one [pp]

PR #662, opened 2025-04-16
chore: Add sample code for tool usage via vanilla OpenAI client

KavithaSiva:
## Context

Closes SAP/ai-sdk-js-backlog#275.

Adds sample code for tool usage in OpenAI client.


sample-code/src/foundation-models/azure-openai.ts
```diff
@@ -87,3 +91,76 @@ export async function chatCompletionWithDestination(): Promise<AzureOpenAiChatCo
 
   return response;
 }
+
+/**
+ * Ask Azure OpenAI model to convert temperature from Celsius to Fahrenheit using tools.
+ * @returns The response from Azure OpenAI obtained using function calling.
+ */
+export async function chatCompletionWithFunctionCall(): Promise<AzureOpenAiChatCompletionResponse> {
```
KavithaSiva, line 99: Prefer using function call instead of tool call based as Open AI's documentation: https://platform.openai.com/docs/guides/function-calling

sample-code/src/foundation-models/azure-openai.ts
```diff
@@ -87,3 +91,76 @@ export async function chatCompletionWithDestination(): Promise<AzureOpenAiChatCo
...
+    tools: [convertTemperatureTool]
+  });
+  return finalResponse;
+}
+
+function convertTemperatureToFahrenheit(temperature: number): string {
+  return `The temperature in Fahrenheit is ${(temperature * 9) / 5 + 32}°F.`;
+}
+
+const callFunction = (name: string, args: any) => {
```
ZhongpinWang, line 161: I would throw if function does not exist. And the return value won't then be undefined.

sample-code/src/foundation-models/azure-openai.ts
```diff
@@ -87,3 +91,76 @@ export async function chatCompletionWithDestination(): Promise<AzureOpenAiChatCo
...
+  if (response.getFinishReason() === 'tool_calls') {
+    (
+      initialMessage.tool_calls as AzureOpenAiChatCompletionMessageToolCall[]
+    ).forEach(toolCall => {
+      const name = toolCall.function.name;
+      const args = JSON.parse(toolCall.function.arguments);
+      const toolResult = callFunction(name, args);
+      const message: AzureOpenAiChatCompletionRequestToolMessage = {
+        role: 'tool',
+        content: toolResult?.toString() ?? '',
```
ZhongpinWang, line 141: Once you throw in `callFunction()`, here `toolResult` can only be string. And I think it only make sense to call llm further if toolResult is defined with some content in it.

sample-code/src/foundation-models/azure-openai.ts
```diff
@@ -87,3 +91,76 @@ export async function chatCompletionWithDestination(): Promise<AzureOpenAiChatCo
...
+  const messages: AzureOpenAiChatCompletionRequestMessage[] = [
+    { role: 'user', content: 'Convert 20 degrees Celsius to Fahrenheit.' }
+  ];
+
+  const response = await client.run({
+    messages,
+    tools
+  });
+
+  const initialMessage = response.data.choices[0].message;
```
ZhongpinWang, line 128:
```suggestion
  const responseMessage = response.data.choices[0].message;
```

ZhongpinWang: initialMessage sounds confusing.

sample-code/src/foundation-models/azure-openai.ts
```diff
@@ -87,3 +91,76 @@ export async function chatCompletionWithDestination(): Promise<AzureOpenAiChatCo
...
+  // Add the model's response for calling functions into the message history
+  messages.push(initialMessage);
+
+  if (response.getFinishReason() === 'tool_calls') {
```
ZhongpinWang, line 132: Use `if (initialMessage.tool_calls)` to check for `undefined` instead of casting it.

sample-code/src/foundation-models/azure-openai.ts
```diff
@@ -87,3 +91,76 @@ export async function chatCompletionWithDestination(): Promise<AzureOpenAiChatCo
...
+    (
+      initialMessage.tool_calls as AzureOpenAiChatCompletionMessageToolCall[]
+    ).forEach(toolCall => {
```
ZhongpinWang, line 135: Here I think `forEach` is not necessary. You can just use index 0.

sample-code/src/foundation-models/azure-openai.ts
KavithaSiva: Renaming to `initialResponse` instead.

sample-code/src/foundation-models/azure-openai.ts
KavithaSiva: My idea was to let the user know that the `getFinishReason()` function states that the LLM responded with a tool call, but yeah checking `initialMessage.tool_calls` is straightforward.

ZhongpinWang (approved):

PR #674, opened 2025-04-22
chore: Align error handling for streaming with Orchestration

shibeshduw:
## Context

Closes SAP/ai-sdk-js-backlog#276.

Since the `.stream()` method in Orchestration Client sets the header `responseType: 'stream'`, the response is always a stream in both success and error cases. So we need to handle the error case and parse the chunks to a readable format. 

This also matches the behavior of the openai chat completion with streaming. Therefore, error handling in our APIs should be aligned with this PR.


packages/orchestration/src/orchestration-client.ts
```diff
@@ -79,15 +79,31 @@ export class OrchestrationClient {
...
+        typeof error.cause.response.data.read === 'function'
+      ) {
+        const chunks: Buffer[] = [];
+        for await (const chunk of error.cause.response.data) {
+          chunks.push(Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk));
+        }
+        const responseString = Buffer.concat(chunks).toString('utf8');
+        const data = JSON.parse(responseString);
+        error.cause.response.data = data;
+      }
```
ZhongpinWang, line 104: [req] You need to put this part in `http-client.ts` in the core package. Because there `ErrorWithCause` is instantiated and `addStack()` was called. Changing the `error.cause.response.data` afterwards won't affect the stack anymore.

So in the catch block of `executeRequest()`, do this

```ts
    if (
      error.response?.data &&
      typeof error.response.data.read === 'function'
    ) {
      const chunks: Buffer[] = [];
      for await (const chunk of error.response.data) {
        chunks.push(Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk));
      }
      error.response.data = JSON.parse(Buffer.concat(chunks).toString('utf8'));
    }
```

packages/core/src/http-client.ts
```diff
@@ -77,13 +79,21 @@ export async function executeRequest(
     );
     return response;
   } catch (error: any) {
+    await handleStreamError(error);
```
KavithaSiva, line 82: [pp] Add a todo reminder linking the [issue](https://github.com/axios/axios/issues/6468) to remove this once Axios provides a fix.

KavithaSiva (approved):
LGTM. Maybe @ZhongpinWang you want to also take a second look?

Issue #670, opened 2025-04-20
ERROR (azure-openai-chat-completion-stream): Choice 0: Stream finished with unknown reason 'null'.


AjitKP91:
### Describe the Bug

finishReason is null and it is checked as undefined. 

Logging error as below even if everything works :) 
**ERROR    (azure-openai-chat-completion-stream): Choice 0: Stream finished with unknown reason 'null'.**

<img width="1505" alt="Image" src="https://github.com/user-attachments/assets/a2d2e71d-274a-4508-9ba0-6f7f2ecfc60e" />

[azure-openai-chat-completion-stream.ts#L59](https://github.com/SAP/ai-sdk-js/blob/ff111fc29a37965a301699aff4a7f19fb708a3c6/packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream.ts#L59)

### Steps to Reproduce

Try stream method and error will log always. 

### Expected Behavior

Error should not log always.

### Used Versions

- Node version via `node -v`: ...
- NPM version via `npm -v: ...
- SAP Cloud SDK for AI version:
  - ...
  - ...

### Region

EU10

### Affected Development Phase

Getting Started

### Impact

No Impact

ZhongpinWang:
Hi @AjitKP91 ,

Thanks for reporting this! You are right, `finishReason` for streaming is nullable. We will provide a fix for this.

ZhongpinWang:
I created a PR #675 for this. It should be part of our next patch release.

AjitKP91:
@ZhongpinWang  Thank you!

PR #675, opened 2025-04-22
fix: OpenAI streaming if finish reason is null

ZhongpinWang:
## Context

A bug reported [here](https://github.com/SAP/ai-sdk-js/issues/670). `finishReason` is nullable.


KavithaSiva (approved):
Thanks for taking care of this.

PR #681, opened 2025-04-25
fix: Smoke test failure

KavithaSiva:
## Context

Smoke test started failing with:
```
Cannot find package '@langchain/langgraph' imported from /home/vcap/app/dist/langchain-orchestration.js
```


tomfrenken (approved):
LGTM

PR #707, opened 2025-05-06
feat: Make template property optional

deekshas8:
## Context

Closes SAP/ai-sdk-js-backlog#293.

This PR makes `template`  (and therefore `templating`) property optional in the `OrchestrationModuleConfig`. This means the user can initialize an OrchestrationClient with the below minimal config:

```
const config: OrchestrationModuleConfig = {
      llm: {
        model_name: 'gpt-4o',
      }
     // templating is now optional
    };

const client = new OrchestrationClient(config)
```

This change no longer forces user to pass a `template`. They can only pass a `messages` array in `prompt` and the client will internally pick the last message as the `template`. The `messages_history` property has been deprecated in favour of `messages`. The flow and error handling works as following:

1. If `templating` is provided (either as `TemplateRef` or `Template`), then everything works as originally intended.
2. If `templating` is not provided, then the latest message is popped from `messages` or `message_history` (if both are set, `messages` is used) and set as `templating.template`
  a. If `messages` or `message_history` both are also not provided, then we throw an error
  b. If `templating.template` is set to an empty array, then also we throw an error.

A simple call now looks like:

```
await new OrchestrationClient(config).chatCompletion({ messages: [{ role: 'user', content: 'Hello' }] });
```

A problem I see is that now everything is optional (template, messages, message_history), so at first it might not be very clear for user what they need to set. The only way to offset this is through documentation.


packages/orchestration/src/util/module-config.ts
```diff
@@ -132,8 +135,32 @@ export function constructCompletionPostRequest(
...
+if (isTemplate(templatingConfig)) {
+
+  // Throw if the template is empty
+  if (!Array.isArray(templatingConfig.template) || templatingConfig.template.length === 0) {
+    throw new Error(
+      'Templating config was provided with an empty template. Either provide a valid template or omit the field entirely.'
+    );
+  }
+
+  // No change needed — templating is already valid
```
tomfrenken, line 151: Not sure if we care but:
— vs -

deekshas8: I left some comments (some for myself while working), in case there were questions on any decision. I will remove them. I will remove them eventually.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -222,3 +248,9 @@ export type AzureFilterThreshold = keyof typeof supportedAzureFilterThresholds;
  * The filter categories supported for Llama guard filter.
  */
 export type LlamaGuardCategory = keyof LlamaGuard38B;
+
+const t: TemplatingModuleConfig = {
+  defaults: {
+    inputContext: 'The default text that will be used in the template if inputContext is not set'
```
ZhongpinWang, line 254: [q] What does this do? Do we need `defaults` in any cases?

tomfrenken (approved):
I like all the changes 👍

packages/orchestration/src/util/module-config.ts
```diff
@@ -132,8 +135,32 @@ export function constructCompletionPostRequest(
...
+} else if (!templatingConfig) {
```
ZhongpinWang, line 152: I feel like these if conditions together with the three-way comparison later in line 163 can be improved? The current logic feels too complicated.

packages/orchestration/src/util/module-config.ts
tomfrenken: > I will remove them eventually.

I am actually in favor of the comments, makes it a lot clearer when reviewing the code at a later point in time.

tests/type-tests/test/orchestration.test-d.ts
```diff
@@ -39,12 +36,12 @@ expectType<CompletionPostResponse>(
...
     await new OrchestrationClient({
       templating: {
-        template: [{ role: 'user', content: 'Hello!' }]
+        defaults: { name: 'Bob' }, 
       },
       llm: {
         model_name: 'gpt-4o'
       }
-    }).chatCompletion()
+    }).chatCompletion({ messages: [{ role: 'user', content: 'Hello! {{?name}}' }]})
```
ZhongpinWang, line 44: I agree that we should allow variables also in `messages` in favor of grounding.

ZhongpinWang (commented):
The flow looks good. I think your 2.b is actually part of 1 or, even empty array means it is defined.

Unfortunately, I also feel we can only mitigate the everything optional issue via docs.

packages/orchestration/src/util/module-config.ts
deekshas8: > but do we have a consistency style-guide for this kind of stuff?

Nothing that we agreed on yet. I will adjust the language a bit for now, since I also didn't like the copilot version. We can discuss about this separately maybe

packages/orchestration/src/util/module-config.ts
deekshas8: I will refactor this. This was a first draft anyway since I wanted to first finalize the API design.

MatKuhr (commented):
> The messages_history property has been deprecated in favour of messages

Not sure we want to do this. Yes, currently message history is kinda useless. But in the future the history will probably be excluded from content filter, masking etc.

Also, in case customers need to pass content containing `{{? ... }}` then they are blocked, because things end up in templating, right?

Finally, in Java we also support the history as optional: https://github.com/SAP/ai-sdk-java/blob/main/orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationPrompt.java

Why not just allow both messages and message history, messages go into template, history into history, done?

deekshas8:
Answering based on how I understood your comments - please correct me if I’ve misunderstood anything.

> Why not just allow both messages and message history, messages go into template, history into history, done?

Deprecating `messageHistory` is more of a naming change, we still pass everything except the last `message` into the `message_history` when constructing the final completion request object. Having a single `messages` is more in line with other clients(incl. langchain and langgraph). Having separate properties imo results in the same issue we had originally with langgraph, that the user would have to pop the last message, set that to `messages` (right now they set it to inputParams):

```
 const callModel = async (state: typeof MessagesAnnotation.State) => {
    const latestMessage = state.messages.pop();
    const inputParamMessage = latestMessage!.content as string;
    const response = await llm.invoke(state.messages, {
      inputParams: { message: inputParamMessage },
    });
    // Update message history with response:
    return { messages: response };
  };

const input = [
    {
      role: 'user',
      content: 'Tell me something about the SAP Cloud SDK'
    }
  ];
  const output = await app.invoke({ messages: input }, config);
```

> Also, in case customers need to pass content containing {{? ... }} then they are blocked, because things end up in templating, right?

I might be missing something here—users can still include placeholder variables in `messages`( there's a test for this).
Maybe it makes more sense to discuss this offline.

MatKuhr:
> pass everything except the last message into the message_history

Right, sorry I think I just missed this part. But there can also be use cases where the user wants to template multiple messages e.g. a system and a user message. Or multiple tool results. Or needs multiple messages to be content filtered later on.

> user would have to pop the last message

Why would the user have to do that still? Can't we automate this part and set the correct message history by default?

packages/orchestration/src/orchestration-completion-post-request.test.ts
```diff
@@ -38,7 +38,7 @@ describe('construct completion post request', () => {
   });
 
   // TODO: Adapt the test after Cloud SDK fix for: https://github.com/SAP/cloud-sdk-backlog/issues/1234
-  it('should construct completion post request with llm and empty templating module', async () => {
+  xit('should construct completion post request with llm and empty templating module', async () => {
```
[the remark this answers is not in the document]
shibeshduw: I think @KavithaSiva added this test originally. Makes sense that we restrict this usage completely.

sample-code/src/langchain-orchestration.ts
```diff
@@ -129,26 +129,13 @@ export async function invokeLangGraphChain(): Promise<string> {
...
 
   const llm = new OrchestrationClient(orchestrationConfig);
   // Define the function that calls the model
   const callModel = async (state: typeof MessagesAnnotation.State) => {
-    const latestMessage = state.messages.pop();
-    const inputParamMessage = latestMessage!.content as string;
-    const response = await llm.invoke(state.messages, {
-      inputParams: { message: inputParamMessage }
-    });
+    const response = await llm.invoke(state.messages);
```
ZhongpinWang, line 138: 💯 This is so much better.

sample-code/src/orchestration.ts
```diff
@@ -656,10 +627,11 @@ export async function orchestrationMessageHistoryWithToolCalling(): Promise<Orch
...
-  return orchestrationClient(
-    [{ role: 'user', content: 'What is the corresponding roman numeral?' }],
-    addNumbersTool
-  ).chatCompletion({
+  // Call the model with a new message and the message history
+  return orchestrationClient.chatCompletion({
+    messages: [
+      { role: 'user', content: 'What is the corresponding roman numeral?' }
+    ],
     messagesHistory: allMessages
```
ZhongpinWang, line 635: [q] Maybe I missed some of the discussion, but are we separating now messages from history?

ZhongpinWang: I thought we deprecate `messagesHistory`?

ZhongpinWang (commented):
I am a bit confused about `messages` and `messagesHistory` now. Looks like `messagesHistory` is not deprecated and `messages` only contain new messages?

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -111,6 +114,66 @@ describe('orchestration service client', () => {
...
 
+  it('calls chatCompletion with some templating configuration (without template)', async () => {
+    const config: OrchestrationModuleConfig = {
+      llm: {
+        model_name: 'gpt-4o',
+        model_params: { max_tokens: 500 }
+      },
+      templating: {
+        defaults: {
+          topic: 'Generative AI Hub'
```
tomfrenken, line 125: [req] Choose a different default value, to properly test that it is overwritten.

deekshas8, line 125:
```suggestion
          topic: 'SAP AI Core'
```

packages/orchestration/src/orchestration-types.ts
```diff
@@ -52,14 +74,56 @@ export type LlmModelParams = {
   n?: number;
 } & Record<string, any>;
 
+/**
+ * Representation of the 'Template' schema.
+ */
+export type Template = Omit<OriginalTemplate, 'template'> & {
+  /**
+   * A chat message array to be formatted with values from `inputParams`.
+   * Both `role` and `content` can use {{?variable}} placeholders resolved via `inputParams`.
```
deekshas8, line 83:
```suggestion
   * Both `role` and `content` can use {{?variable}} placeholders.
```

tomfrenken (changes requested):
All in all LGTM, but some pointers are still open.

tomfrenken (approved):
LGTM

PR #711, opened 2025-05-07
feat: Support LangChain orchestration client streaming

KavithaSiva:
## Context

Closes SAP/ai-sdk-js-backlog#259.

This PR introduces streaming support for Langchain Orchestration client.

Usage:

```TS
const orchestrationConfig: LangchainOrchestrationModuleConfig = {
  llm: {
    model_name: 'gpt-4o'
  },
  templating: {
    template: [
      {
        role: 'user',
        content: 'Write a 100 word explanation about {{?topic}}'
      }
    ]
  }
};

const client = new OrchestrationClient(orchestrationConfig);
const stream =  client.stream([], {
  inputParams: {
    topic: 'SAP Cloud SDK and its capabilities'
  },
  signal: controller.signal
});

for await (const chunk of stream) {
 //Access the streamed chunk here
}
```


packages/langchain/src/orchestration/util.ts
```diff
@@ -174,3 +200,45 @@ export function mapOutputToChatResult(
...
+  // Create tool call chunks if present
+  let tool_call_chunks: ToolCallChunk[] = [];
+  if (Array.isArray(delta.tool_calls)) {
+    tool_call_chunks = mapOrchestrationToLangchainToolCallChunk(
+      delta.tool_calls
+    );
+  }
+  const toolCallId = delta.tool_calls?.[0]?.id ?? undefined;
+  // Use OrchestrationMessageChunk to represent message chunks for roles like 'tool' and 'user' too
+  return new OrchestrationMessageChunk(
```
KavithaSiva, line 238: In other chat model clients in the langchain repo, various other `MessageChunk` like for e.g. `HumanMessageChunk` , `ToolMessageChunk` are created based on roles.

But, in the end the upstream function processes all chunks as `AIMessageChunk`, so I chose to use `OrchestrationMessageChunk` (which extends `AIMessageChunk`) to represent all types of message chunks with some additional properties instead.

KavithaSiva:
Adding more tests in a follow-up PR.

sample-code/src/langchain-orchestration.ts
```diff
@@ -187,6 +190,49 @@ export async function invokeLangGraphChain(): Promise<string> {
...
+  const client = new OrchestrationClient(orchestrationConfig);
+
+  // Return the stream
+  return client.stream(
+    [
+      {
+        role: 'user',
+        content: 'I need information about a topic.'
+      }
+    ],
```
ZhongpinWang, line 226: [pp] I would just leave messages (aka. message history) empty as after the make template optional PR, we will conceptually avoid having template and messages defined at the same time. Either template for one time chat message or messages (same as in LangChain design) for conversational chat completion.

packages/langchain/src/orchestration/client.ts
```diff
@@ -108,14 +109,121 @@ export class OrchestrationClient extends BaseChatModel<
...
+
+    const orchestrationClient = new OrchestrationClientBase(
+      mergedOrchestrationConfig,
+      this.deploymentConfig,
+      this.destination
+    );
+
+    const streamOptions = {
+      llm: {
+        include_usage: true,
```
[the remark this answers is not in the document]
KavithaSiva: True, didn't look closely at the AI generated code.

sample-code/src/langchain-orchestration.ts
tomfrenken: Good catch, I would assume in the future we then change this sample to exclude templating and go with just messages instead. (Maybe add as a todo for some follow-up)

sample-code/src/index.ts
```diff
@@ -31,7 +31,8 @@ export {
 } from './langchain-azure-openai.js';
 export {
   invokeChain as orchestrationInvokeChain,
-  invokeLangGraphChain
+  invokeLangGraphChain,
+  streamOrchestrationLangChain
```
tomfrenken, line 35: Not necessarily important for now, but I feel like the naming is diverging here a little bit.

[req] I think my idea as to have <type><function><Whatever-chain> as format.

So invokeChain (no type since it's the only variant) but orchestratioInvokeChain (since there are duplicates)
I guess naming wise, we should name invokeLangGraphChain to orchestrationInvokeGraphChain, and your new function to orchestrationStreamChain, what do your think?

packages/langchain/src/orchestration/client.ts
```diff
@@ -108,14 +109,121 @@ export class OrchestrationClient extends BaseChatModel<
...
+          };
+        }
+      }
+      const content = delta.content ?? '';
+      const generationChunk = new ChatGenerationChunk({
+        message: messageChunk,
+        text: content
+      });
+
+      // Notify the run manager about the new token
```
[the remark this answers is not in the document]
tomfrenken: I think the best would be to adjust this comment, as it is not clear why we pass 3 undefined's, since the API is not documented, everyone needs to review the source code to get it currently.

packages/langchain/src/orchestration/client.ts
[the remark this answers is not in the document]
tomfrenken: In fact, even the default value is part of our existing logic, so it can be omitted here.

packages/langchain/src/orchestration/util.ts
```diff
@@ -123,6 +127,27 @@ function mapAzureOpenAiToLangchainToolCall(
...
+/**
+ * Maps {@link OrchestrationToolCallChunk} to LangChain's {@link ToolCallChunk}.
+ * @param toolCallChunks - The {@link OrchestrationToolCallChunk} in a stream response chunk.
+ * @returns An array of LangChain {@link ToolCallChunk}.
+ */
+function mapOrchestrationToLangchainToolCallChunk(
+  toolCallChunks: OrchestrationToolCallChunk[]
+): ToolCallChunk[] {
+  const tool_call_chunks: ToolCallChunk[] = [];
+  for (const chunk of toolCallChunks) {
```
tomfrenken, line 139: [pp] Use .map instead of a for..of loop

packages/langchain/src/orchestration/client.ts
```diff
@@ -108,14 +109,121 @@ export class OrchestrationClient extends BaseChatModel<
...
+    for await (const chunk of response.stream) {
+      if (!chunk.data) {
+        continue;
+      }
+
+      const delta = chunk.getDelta();
+      if (!delta) {
+        continue;
+      }
+      const messageChunk = _convertOrchestrationChunkToMessageChunk(
```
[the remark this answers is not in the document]
KavithaSiva: > [req] chunk contains both data and delta, pass just `chunk to the util function.

I did think about this, but as I already have a falsy value check for `delta` above, it's just easier to pass delta here to not make the check again inside the mapping function.

packages/langchain/src/orchestration/client.ts
```diff
@@ -108,14 +109,121 @@ export class OrchestrationClient extends BaseChatModel<
...
+        delta,
+        defaultRole
+      );
+
+      defaultRole = delta.role ?? defaultRole;
+      const finishReason = chunk.getFinishReason();
+      const tokenUsage = chunk.getTokenUsage();
+
+      // Add token usage to the message chunk if this is the final chunk
+      if (finishReason && tokenUsage) {
```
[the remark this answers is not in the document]
KavithaSiva: We are only interested in the last chunk as only it contains both token usage and finish reason.
For all other chunks, the finish reason is empty. So IMO makes sense to store both of them together.

packages/langchain/src/orchestration/client.ts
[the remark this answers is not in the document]
ZhongpinWang: [pp] I wouldn't mind putting both if checks into map function and return undefined if needed. In the end, we want to check if delta is there or not. Also I wonder, how could `chunk.data` be `undefined` or `null` as it is typed and is part of `AzureOpenAiChatCompletionStreamChunkResponse` class defined by us.

packages/langchain/src/orchestration/client.ts
```diff
@@ -108,14 +109,121 @@ export class OrchestrationClient extends BaseChatModel<
...
+      if (!delta) {
+        continue;
+      }
+      const messageChunk = _convertOrchestrationChunkToMessageChunk(
+        chunk.data,
+        delta,
+        defaultRole
+      );
+
+      defaultRole = delta.role ?? defaultRole;
```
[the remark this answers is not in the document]
KavithaSiva: Comment no longer relevant as we don't store role anymore.

packages/langchain/src/orchestration/client.ts
```diff
@@ -108,14 +109,121 @@ export class OrchestrationClient extends BaseChatModel<
...
+      const finishReason = chunk.getFinishReason();
+      const tokenUsage = chunk.getTokenUsage();
+
+      // Add token usage to the message chunk if this is the final chunk
+      if (finishReason && tokenUsage) {
+        if (messageChunk instanceof OrchestrationMessageChunk) {
```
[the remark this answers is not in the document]
KavithaSiva: Was a leftover, I forgot to remove. Cleaned it up now.

packages/langchain/src/orchestration/client.ts
```diff
@@ -108,14 +109,121 @@ export class OrchestrationClient extends BaseChatModel<
...
+          messageChunk.response_metadata.finish_reason = finishReason;
+          messageChunk.usage_metadata = {
+            input_tokens: tokenUsage.prompt_tokens,
+            output_tokens: tokenUsage.completion_tokens,
+            total_tokens: tokenUsage.total_tokens
+          };
+        }
+      }
+      const content = delta.content ?? '';
```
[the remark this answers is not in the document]
KavithaSiva: No longer relevant.

sample-code/src/langchain-orchestration.ts
KavithaSiva: For now left message history empty, left a todo to update the test after BLI.

packages/langchain/src/orchestration/util.ts
```diff
@@ -174,3 +199,45 @@ export function mapOutputToChatResult(
...
+  }
+
+  // Create tool call chunks if present
+  let tool_call_chunks: ToolCallChunk[] = [];
+  if (Array.isArray(delta.tool_calls)) {
+    tool_call_chunks = mapOrchestrationToLangchainToolCallChunk(
+      delta.tool_calls
+    );
+  }
+  const toolCallId = delta.tool_calls?.[0]?.id ?? undefined;
```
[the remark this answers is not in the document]
KavithaSiva: No longer relevant.

packages/langchain/src/orchestration/util.ts
```diff
@@ -174,3 +199,45 @@ export function mapOutputToChatResult(
...
+  const additional_kwargs: Record<string, unknown> = {};
+
+  // Handle tool calls
+  if (delta.tool_calls && delta.tool_calls.length > 0) {
+    additional_kwargs.tool_calls = delta.tool_calls;
+  }
+
+  // Create tool call chunks if present
+  let tool_call_chunks: ToolCallChunk[] = [];
+  if (Array.isArray(delta.tool_calls)) {
```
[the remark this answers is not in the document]
KavithaSiva: No longer relevant.

packages/langchain/src/orchestration/orchestration-message-chunk.ts
```diff
@@ -4,18 +4,21 @@ import type { ModuleResults } from '@sap-ai-sdk/orchestration';
 
 /**
  * An AI Message Chunk containing module results and request ID.
- * @internal
  */
 export class OrchestrationMessageChunk extends AIMessageChunk {
   module_results: ModuleResults;
   request_id: string;
+  // Adding additonal properties to also store properties from other types of Message Chunks.
+  additional_properties?: Record<string, any>;
```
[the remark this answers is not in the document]
KavithaSiva: No longer relevant.

packages/langchain/src/orchestration/client.ts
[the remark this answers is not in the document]
KavithaSiva: Introduced new utility functions.

packages/langchain/src/orchestration/client.ts
```diff
@@ -108,14 +109,121 @@ export class OrchestrationClient extends BaseChatModel<
...
+      llm: {
+        include_usage: true,
+        ...options.streamOptions?.llm
+      },
+      outputFiltering: options.streamOptions?.outputFiltering,
+      global: options.streamOptions?.global
+    };
+
+    const response = await this.caller.callWithOptions(
+      { signal: options.signal },
```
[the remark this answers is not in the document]
KavithaSiva: The streaming variant does not need the `callWithOptions`, I have switched to `call` now as we already pass the `controller` to `orchestrationClient.stream()` function.

I did not understand your comment about the `_generate()` function, can you explain?

packages/langchain/src/orchestration/client.ts
```diff
@@ -108,14 +109,102 @@ export class OrchestrationClient extends BaseChatModel<
...
+        undefined,
+        { chunk: generationChunk }
+      );
+
+      yield generationChunk;
+    }
+
+    if (options.signal?.aborted) {
+      throw new Error('AbortError');
+    }
```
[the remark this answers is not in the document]
KavithaSiva: Done.

packages/langchain/src/orchestration/client.ts
```diff
@@ -108,14 +112,86 @@ export class OrchestrationClient extends BaseChatModel<
...
+        // Todo Adapt messagesHistory after: https://github.com/SAP/ai-sdk-js-backlog/issues/293
+        { messagesHistory: orchestrationMessages, inputParams },
+        controller,
+        options.streamOptions,
+        customRequestConfig
+      )
+    );
+
+    for await (const chunk of response.stream) {
+      const messageChunk = mapOrchestrationChunkToLangChainMessageChunk(chunk);
```
ZhongpinWang, line 162: [req] If we want to implement this correctly with the multi choices support, here we should yield `generationChunk` for each choice. So wrap the below content with

```
chunk.data.orchestration_result?.choices?.forEach(choice => {
    ...
}); // Or export getChoices() and use that
```

And then when calling functions like `getDeltaContent()`, give it the choice index.

The current implementation only support n=1 and will keep yielding chunks with empty content if choice index is not 0.

packages/langchain/src/orchestration/util.ts
```diff
@@ -174,3 +194,81 @@ export function mapOutputToChatResult(
...
+ * @internal
+ */
+export function computeTokenIndices(chunk: OrchestrationStreamChunkResponse): {
+  prompt: number;
+  completion: number;
+} {
+  return {
+    // Indicates the token is part of the first prompt
+    prompt: 0,
+    completion: chunk.data.orchestration_result?.choices[0]?.index ?? 0
```
ZhongpinWang, line 272: [req] Here do not assume we only have one choice. We do not know if the first element in the chunk is for choice 0.

packages/langchain/src/orchestration/util.ts
```diff
@@ -123,6 +126,23 @@ function mapAzureOpenAiToLangchainToolCall(
...
+    args: chunk.function?.arguments,
+    id: chunk.id,
+    index: chunk.index,
+    type: 'tool_call_chunk'
+  }));
+}
+
 /**
  * Maps the completion response to a {@link ChatResult}.
  * @param completionResponse - The completion response to map.
```
ZhongpinWang, line 148: See this function, we actually do `choices.map()`

packages/langchain/src/orchestration/client.ts
```diff
@@ -108,14 +112,86 @@ export class OrchestrationClient extends BaseChatModel<
...
+      const tokenIndices = computeTokenIndices(chunk);
+      const finishReason = response.getFinishReason();
```
ZhongpinWang, line 164: Support n>1 same as in vanilla orchestration client

packages/orchestration/src/orchestration-stream-chunk-response.ts
```diff
@@ -42,6 +43,17 @@ export class OrchestrationStreamChunkResponse {
...
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns The message tool call chunks.
+   */
+  getDeltaToolCallChunks(choiceIndex = 0): ToolCallChunk[] | undefined {
+    return this.getChoices()?.find(
+      (c: LlmChoiceStreaming) => c.index === choiceIndex
+    )?.delta.tool_calls;
+  }
+
   private getChoices(): LlmChoiceStreaming[] | undefined {
```
ZhongpinWang, line 57: In non-streaming case, we implemented `findChoiceByIndex()` to improve code reuse. We can do the same thing here.

ZhongpinWang (commented):
Thanks for the nice refactoring and dividing big code blocks into functions. This code is definitely more readable now!

I did another round of the review. Other than some naming issues, the biggest (potential) issue is that the multi choice logic is not implemented. This could be a problem in the future and since it is already properly handled in the vanilla orchestration client, consider implement this correctly in LangChain as well.

packages/langchain/src/orchestration/client.ts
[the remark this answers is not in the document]
ZhongpinWang: Discussed.

KavithaSiva:
> I did another round of the review. Other than some naming issues, the biggest (potential) issue is that the multi choice logic is not implemented. This could be a problem in the future and since it is already properly handled in the vanilla orchestration client, consider implement this correctly in LangChain as well.

I want to hold off on implementing for multi-choice now as the orchestration service currently does not support multiple choices for streaming, it throws a `400`. 
Moreover, in LangChain, each  streamed `AIMessageChunk` is supposed to hold contents of only one choice, so there would be multiple chunks for one chunk of the orchestration service. From what I understood, the consumer is supposed to group chunks together based on `newTokenIndices.completion` value.  

I would prefer to do this in a separate BLI, when orchestration service starts supporting this and not in this PR.

packages/langchain/src/orchestration/util.ts
ZhongpinWang: Here if we only have one choice, why don't we just set completion to 0?

KavithaSiva:
> I don't have strong opinion on whether we support n>2 already now as you are right, we get 400 from orchestration for such case. If we implement this already, the effort is not much by adding a foreach and not using the default 0 everywhere. Leave it to you to decide.

I talked with Christoph and he said multiple choices for orchestration streaming is not planned in the near future(Q2 or Q3)
You are right the code change is easy, but testing it would imply adding imaginary responses for now. 
I would feel more comfortable adding this in when the service supports it too.

packages/langchain/src/orchestration/util.ts
KavithaSiva: This is still valid and necessary as orchestration service supports multiple choices for normal chat completion. 
It is only for streaming that it doesn't support multiple choices.

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -129,4 +158,93 @@ describe('orchestration service client', () => {
...
+        ? chunk
+        : intermediateChunk.concat(chunk);
+      if (iterations >= maxIterations) {
+        break;
+      }
+    }
+    expect(intermediateChunk).toBeDefined();
+    expect(intermediateChunk!.content).toBeDefined();
+    expect(intermediateChunk!.content).toEqual(
+      'The SAP Cloud SDK is a comprehensive development toolkit designed to simplify and accelerate the cre'
```
ZhongpinWang, line 183:
```suggestion
    expect(intermediateChunk?.content).toEqual(
      'The SAP Cloud SDK is a comprehensive development toolkit designed to simplify and accelerate the cre'
```

This is equivalent.

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -129,4 +158,93 @@ describe('orchestration service client', () => {
...
+    const stream = await client.stream([]);
+    let iterations = 0;
+    const maxIterations = 2;
+    let intermediateChunk: AIMessageChunk | undefined;
+
+    for await (const chunk of stream) {
+      iterations++;
+      intermediateChunk = !intermediateChunk
+        ? chunk
+        : intermediateChunk.concat(chunk);
```
ZhongpinWang, line 175: [req] Just concat all chunk content and compare output for the sake of code simplicity.

Also I generally have the feeling that sometimes the code may work, just the logic is not clean. Is the above code not equivalent to

```ts
intermediateChunk = intermediateChunk
        ? intermediateChunk.concat(chunk)
        : chunk;
```

And are we not just checking the first chunk with content? Can't we check the content instead of using the maxIteration logic? (But still just loop over all chunks and compare to snapshot)

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -129,4 +158,93 @@ describe('orchestration service client', () => {
       'Request failed with status code 400'
     );
   }, 1000);
+
+  it('supports streaming responses', async () => {
+    mockStreamInferenceWithResilience();
```
ZhongpinWang, line 163: [req] Why are we mocking responses with resilience? Why is resilience needed in this test? Can we not just call `mockInference()`?

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -129,4 +158,93 @@ describe('orchestration service client', () => {
...
+      }
+    }
+    expect(intermediateChunk).toBeDefined();
+    expect(intermediateChunk!.content).toBeDefined();
+    expect(intermediateChunk!.content).toEqual(
+      'The SAP Cloud SDK is a comprehensive development toolkit designed to simplify and accelerate the cre'
+    );
+  });
+
+  it('streaming with abort signal', async () => {
```
ZhongpinWang, line 187:
```suggestion
  it('streams with abort signal', async () => {
```
Use natural language in the test description.

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -129,4 +158,93 @@ describe('orchestration service client', () => {
...
+    const controller = new AbortController();
+    const { signal } = controller;
+    const stream = await client.stream([], { signal });
+    const streamPromise = async () => {
+      for await (const _chunk of stream) {
+        controller.abort();
+      }
+    };
+
+    await expect(streamPromise()).rejects.toThrow();
```
ZhongpinWang, line 199: [pp] Check also the throw content.
[pp] Rename the `streamPromise`, it is a function not a promise.

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -129,4 +158,93 @@ describe('orchestration service client', () => {
...
+
+  it('supports streaming responses with tool calls', async () => {
+    mockStreamInferenceWithResilience(mockResponseStreamToolCalls);
+
+    const client = new OrchestrationClient(config);
+    const stream = await client.stream([]);
+
+    let finalOutput: AIMessageChunk | undefined;
+    for await (const chunk of stream) {
+      finalOutput = !finalOutput ? chunk : finalOutput.concat(chunk);
```
ZhongpinWang, line 239:
```suggestion
      finalOutput = finalOutput ? finalOutput.concat(chunk) : chunk;
```

ZhongpinWang: But again, just concat tool call json string together and compare that. We iterate through all chunks anyway. It worths the effort to try to get content from each chunk and make sure it always works for all chunks.

ZhongpinWang (dismissed):
Reviewed some tests. Please pay attention to the code quality. If logic can be simplified in an equivalent form always do that. The test scope should also be narrowed to what is needed. Using, e.g., resilience for testing all streaming logic is unnecessary.

packages/orchestration/src/orchestration-stream-chunk-response.ts
```diff
@@ -42,6 +43,12 @@ export class OrchestrationStreamChunkResponse {
     )?.delta.content;
   }
 
+  getDelta(choiceIndex = 0): ChatDelta | undefined {
+    return this.getChoices()?.find(
+      (c: LlmChoiceStreaming) => c.index === choiceIndex
+    )?.delta;
+  }
+
```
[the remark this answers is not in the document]
KavithaSiva: Introduced convenience function `findChoiceByIndex()`, I feel the additional convenience of getting tool call chunks is still a good to have. For accessing other properties in the choice, the users can still use `findChoiceByIndex()`.

packages/langchain/src/orchestration/client.ts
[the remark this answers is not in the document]
KavithaSiva: I checked, `signal` is not required to be sent in `_generate()` method for now as I didn't enable streaming to be called from the `invoke()` method.
I didn't see examples where users used `invoke()` to stream, IMO signal is only necessary when we allow this.

@tomfrenken Could you also confirm?

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -129,4 +158,93 @@ describe('orchestration service client', () => {
...
+      chunks.push(chunk);
+      break;
+    }
+    expect(callbackHandler.handleLLMNewToken).toHaveBeenCalled();
+    const firstCallArgs = callbackHandler.handleLLMNewToken.mock.calls[0];
+    expect(firstCallArgs).toBeDefined();
+    // First chunk content is empty
+    expect(firstCallArgs[0]).toEqual('');
+    // Second argument should be the token indices
+    expect(firstCallArgs[1]).toEqual({ prompt: 0, completion: 0 });
```
deekshas8, line 226: [q] I do not understand this honestly. The mock handler function here doesnt seem to take any arguments. Where do they come from?
[q] Why check : expect(chunks.length).toBeGreaterThan(0); Won't it just be one? What is the point of checking this value even wrt callbacks?

packages/langchain/src/orchestration/client.ts
```diff
@@ -108,14 +112,86 @@ export class OrchestrationClient extends BaseChatModel<
...
+      setTokenUsage(messageChunk, tokenUsage);
+      const content = chunk.getDeltaContent() ?? '';
+
+      const generationChunk = new ChatGenerationChunk({
+        message: messageChunk,
+        text: content,
+        generationInfo: { ...tokenIndices }
+      });
+
+      // Notify the run manager about the new token, some parameters are undefined as they are implicitly read from the context.
```
deekshas8, line 171: [req] Can we document such stuff for future? I wouldn't know why we set some values as undefined, unless I read the langchain code.

packages/langchain/src/orchestration/util.ts
```diff
@@ -123,6 +126,23 @@ function mapAzureOpenAiToLangchainToolCall(
   }
 }
 
+/**
```
deekshas8, line 129: [req] Unrelated to this PR, but I see a function `mapAzureOpenAiToLangchainToolCall` in this file. Can we clean it up?

packages/langchain/src/orchestration/util.ts
```diff
@@ -174,3 +194,85 @@ export function mapOutputToChatResult(
...
+
+/**
+ * Sets finish reason on a LangChain message chunk if available.
+ * @param messageChunk - The LangChain message chunk to update.
+ * @param finishReason - The finish reason from the response.
+ * @internal
+ */
+export function setFinishReason(
+  messageChunk: AIMessageChunk,
+  finishReason: string | undefined
```
deekshas8, line 233:
```suggestion
  finishReason?: string
```

packages/langchain/src/orchestration/util.ts
```diff
@@ -174,3 +194,85 @@ export function mapOutputToChatResult(
...
+
+/**
+ * Sets usage metadata on a message chunk if available.
+ * @param messageChunk - The LangChain message chunk to update.
+ * @param tokenUsage - The token usage information.
+ * @internal
+ */
+export function setTokenUsage(
+  messageChunk: AIMessageChunk,
+  tokenUsage: TokenUsage | undefined
```
deekshas8, line 248:
```suggestion
  tokenUsage?: TokenUsage
```

packages/langchain/src/orchestration/util.ts
```diff
@@ -174,3 +194,85 @@ export function mapOutputToChatResult(
...
+}
+
+/**
+ * Computes token indices for a chunk of the orchestration stream response.
+ * @param chunk - A chunk of the orchestration stream response.
+ * @returns An object with prompt and completion indices.
+ * @internal
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function computeTokenIndices(chunk: OrchestrationStreamChunkResponse): {
```
deekshas8, line 267: [q] Didnt get the point of this function if its not supported in entirety. How is this helpful right now?

deekshas8 (commented):
I'm still reviewing, but this is how far I got today. Will take another look again

packages/langchain/src/orchestration/util.ts
KavithaSiva: Token indices need to be set in each generated chunk as this provides information about which prompt(in a multi-prompt scenario) and which choice(in a muti-choice scenario) the chunk belongs to.

Although, we hardcode these values now, leaving the function as-is enables us to support mutli-chat scenarios in the future easily.

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -129,4 +158,93 @@ describe('orchestration service client', () => {
...
   }, 1000);
+
+  it('supports streaming responses', async () => {
+    mockStreamInferenceWithResilience();
+
+    const client = new OrchestrationClient(config);
+    const stream = await client.stream([]);
+    let iterations = 0;
+    const maxIterations = 2;
+    let intermediateChunk: AIMessageChunk | undefined;
```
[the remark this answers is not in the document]
deekshas8: Ah, I believe ts complains otherwise.

packages/langchain/src/orchestration/util.ts
deekshas8: ok, is this in a ticket somewhere? I would add a TODO in the comments, but also mention it in a ticket for future.

packages/langchain/src/orchestration/client.test.ts
KavithaSiva: First question: Even though the call back function, doesn't take arguments here, when the following snippet is executed in the `_streamResponseChunks`  function, these arguments gets passed to the callback for each chunk.
```
await runManager?.handleLLMNewToken(
 content,
 tokenIndices,
 undefined,
 undefined,
 undefined,
 { chunk: generationChunk }
);
 ```      
Second question: You are right, we don't need to check this value here.

File: packages/langchain/src/orchestration/client.test.ts (hunk 35 above)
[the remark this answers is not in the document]
Comment (KavithaSiva) [reply packages/langchain/src/orchestration/client.test.ts:165]: Yes.

File: packages/langchain/src/orchestration/client.test.ts [hunk 36]
```diff
@@ -30,6 +35,14 @@ describe('orchestration service client', () => {
       'orchestration',
       'orchestration-chat-completion-input-filter-error.json'
     );
+    mockResponseStream = await parseFileToString(
```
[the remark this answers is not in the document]
KavithaSiva: Probably, just missed it, switched to `beforeAll` instead.

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -129,4 +158,93 @@ describe('orchestration service client', () => {
...
     );
   }, 1000);
+
+  it('supports streaming responses', async () => {
+    mockStreamInferenceWithResilience();
+
+    const client = new OrchestrationClient(config);
+    const stream = await client.stream([]);
+    let iterations = 0;
+    const maxIterations = 2;
```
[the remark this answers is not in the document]
KavithaSiva: Idea was to not compare the entire streamed text as it's big.
But, I don't have a strong opinion, so will switch to the entire text instead.

packages/langchain/src/orchestration/util.ts
KavithaSiva: Added a BLI: https://github.com/SAP/ai-sdk-js-backlog/issues/321 and TODO.

packages/langchain/src/orchestration/client.test.ts
KavithaSiva: Added a snapshot check for `tool_call_chunks`. Both `tool_call_chunks` and `tool_calls` are supposed to have the full function details in the final concatenated chunk.

packages/langchain/src/orchestration/client.test.ts
KavithaSiva: The checked throw content is different, not the error I initially threw( this was unreachable code, so I removed it).
Also did the renaming.

packages/langchain/src/orchestration/client.test.ts
KavithaSiva: Done.

packages/langchain/src/orchestration/__snapshots__/client.test.ts.snap
```diff
@@ -137,3 +137,177 @@ exports[`orchestration service client returns successful response when timeout i
...
+  ],
+  "kwargs": {
+    "additional_kwargs": {
+      "module_results": {
+        "llm": {
+          "choices": [
+            {
+              "delta": {
+                "content": "",
+                "role": "assistantassistantassistantassistantassistant",
```
KavithaSiva, line 251: Some properties are concatenated multiple times, as `concat()` in `AIMessageChunk` does not have these properties. We need to custom handle these properties during concatenation. Will handle this in a follow-up BLI: https://github.com/SAP/ai-sdk-js-backlog/issues/322

packages/langchain/src/orchestration/client.ts
tomfrenken: I would put it as a comment right above this line. Doesn't need to be user-facing, but for us to maintain in the future.

tomfrenken (commented):
Looks like a lot was already considered, still there are some pointers from my side left.

I will continue reviewing the tests.

test-util/data/orchestration/orchestration-chat-completion-stream-chunks-tool-calls.txt
```diff
@@ -0,0 +1,13 @@
+data: {"request_id": "0c9186b6-be86-40d8-9745-ec81dd6351bc", "module_results": {"templating": [{"content": "Convert 20 degrees Celsius to Fahrenheit.", "role": "user"}]}, "orchestration_result": {"id": "", "object": "", "created": 0, "model": "", "system_fingerprint": "", "choices": [{"index": 0, "delta": {"content": ""}}]}}
```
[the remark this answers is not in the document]
KavithaSiva: Done.

sample-code/src/server.ts
```diff
@@ -466,6 +467,46 @@ app.get('/langchain/invoke-stateful-chain', async (req, res) => {
...
+
+    for await (const chunk of stream) {
+      if (!connectionAlive) {
+        break;
+      }
+      res.write(chunk.content + '\n');
+      if (connectionAlive && chunk.usage_metadata) {
+        res.write('\n\n---------------------------\n');
+        res.write(
+          `Finish reason:  ${chunk.response_metadata?.finish_reason}\n`
```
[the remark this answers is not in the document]
KavithaSiva: In what you linked, it is still on the chunk level, not on the response level, but in the last chunk.

In our implementation too, `usage_metadata is only populated for the last chunk, still I adapted the example as per the langchain one, to print it out from the last concatenated chunk instead.

sample-code/src/langchain-orchestration.ts
```diff
@@ -174,6 +175,38 @@ export async function invokeLangGraphChain(): Promise<string> {
...
 
+/**
+ * Stream responses using LangChain Orchestration client.
+ * @param controller - The abort controller to cancel the request if needed.
+ * @returns An async iterable of {@link AIMessageChunk} objects.
+ */
+export async function streamChain(
+  controller = new AbortController()
+): Promise<AsyncIterable<AIMessageChunk>> {
+  // TODO: Remove template and use messages during https://github.com/SAP/ai-sdk-js-backlog/issues/317
```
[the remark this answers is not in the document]
KavithaSiva: Done.

packages/orchestration/src/orchestration-stream-chunk-response.test.ts
```diff
@@ -5,11 +5,13 @@ describe('Orchestration chat completion stream chunk response', () => {
   let mockResponses: {
     tokenUsageAndFinishReasonResponse: any;
     deltaContentResponse: any;
+    toolCallResponse: any;
   };
   let orchestrationStreamChunkResponses: {
     tokenUsageResponse: OrchestrationStreamChunkResponse;
     finishReasonResponse: OrchestrationStreamChunkResponse;
     deltaContentResponse: OrchestrationStreamChunkResponse;
+    toolCallResponse: OrchestrationStreamChunkResponse;
```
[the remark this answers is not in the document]
tomfrenken: Sounds good!

packages/orchestration/src/orchestration-stream-chunk-response.test.ts
```diff
@@ -5,11 +5,13 @@ describe('Orchestration chat completion stream chunk response', () => {
   let mockResponses: {
     tokenUsageAndFinishReasonResponse: any;
     deltaContentResponse: any;
+    toolCallResponse: any;
```
tomfrenken, line 8:
```suggestion
    deltaToolCallResponse: any;
```

tomfrenken: Since we named the other one deltaContentResponse, I would keep it aligned (also clarifies that only the delta is tested)

packages/langchain/src/orchestration/__snapshots__/client.test.ts.snap
```diff
@@ -137,3 +137,205 @@ exports[`orchestration service client returns successful response when timeout i
...
+  ],
+  "kwargs": {
+    "additional_kwargs": {
+      "module_results": {
+        "llm": {
+          "choices": [
+            {
+              "delta": {
+                "content": "",
+                "role": "assistantassistantassistantassistantassistantassistantassistantassistantassistantassistantassistant",
```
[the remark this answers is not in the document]
tomfrenken: Agreed, but could also be a contribution to LangChain itself (as part of that follow-up ticket), as they hint as concat being the "standard" way to chain chunks together, and it's clearly flawed.

tomfrenken (approved):
A few minor comments, but overall LGTM!

sample-code/src/server.ts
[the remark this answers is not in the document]
tomfrenken: Perfect, I think that aligning closer to the "official" way makes it easier for us down the line.

PR #719, opened 2025-05-13
feat: Support orchestration translation module

shibeshduw:
## Context

Closes SAP/ai-sdk-js-backlog#299.

Also Review DOCS PR: https://github.com/SAP/ai-sdk/pull/104

### DIFF
--- a/packages/orchestration/src/util/translation.test.ts
+++ b/packages/orchestration/src/util/translation.test.ts
@@ -1,5 +1,4 @@
 import { buildTranslationConfig } from './translation.js';
-import type { TranslationConfigParams } from '../orchestration-types.js';
 import type { SAPDocumentTranslation } from '../client/api/schema/index.js';
 
 describe('Translation module config', () => {
@@ -18,15 +17,5 @@ describe('Translation module config', () => {
     };
 
     expect(translationConfig).toEqual(expectedTranslationConfig);
-  });
-
-  it('throws error when target language is empty', () => {
-    const errorConfig: TranslationConfigParams = {
-      sourceLanguage: 'en-US',
-      targetLanguage: ''
-    };
-    expect(() => {
-      buildTranslationConfig(errorConfig);
-    }).toThrow('Target language is required for translation configuration.');
   });
 });
--- a/packages/orchestration/src/util/translation.ts
+++ b/packages/orchestration/src/util/translation.ts
@@ -12,12 +12,6 @@ import type { SAPDocumentTranslation } from '../client/api/schema/index.js';
 export function buildTranslationConfig(
   config: TranslationConfigParams
 ): SAPDocumentTranslation {
-  if (!config.targetLanguage) {
-    throw new Error(
-      'Target language is required for translation configuration.'
-    );
-  }
-
   return {
     type: 'sap_document_translation',
     config: {


packages/orchestration/src/orchestration-types.ts
```diff
@@ -222,3 +232,17 @@ export type AzureFilterThreshold = keyof typeof supportedAzureFilterThresholds;
...
+export interface TranslationConfigParams {
+  /**
+   * Language of the text to be translated.
+   */
+  source_language?: string;
+  /**
+   * Language to which the text should be translated.
+   */
+  target_language: string;
+}
```
shibeshduw, line 248: The only reason why this is required is to enforce the ordering of `source` before `target` ([ADR](https://github.tools.sap/AI/llm-orchestration/blob/main/docs/adr/015-language-translation.md)) as otherwise, the required param would have to come first.

sample-code/src/orchestration.ts
```diff
@@ -663,3 +664,35 @@ export async function orchestrationMessageHistoryWithToolCalling(): Promise<Orch
...
+      target_language: 'fr-FR'
+    })
+  });
+
+  return orchestrationClient.chatCompletion({
+    inputParams: {
+      input: 'Write an abstract for a thriller playing at SAP headquarters.'
+    }
+  });
+}
```
shibeshduw, line 698: Open question: Should we add additional sample for translation using Streaming + Masking and Streaming + Filtering ?

.changeset/violet-walls-see.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[New Functionality] Add support for orchestration translation module. 
+Implement `buildOrchestrationTranslationConfig` convenience function to build translation config.
```
ZhongpinWang, line 6:
```suggestion
Implement `buildOrchestrationTranslationConfig()` convenience function to build translation config.
```

ZhongpinWang: [pp] I would rename it as `buildTranslationConfig()` to follow the same naming style as other convenience functions.

packages/orchestration/src/orchestration-types.ts
ZhongpinWang: [q] I don't totally understand your comment? What do you mean by ordering?  Is it not the same as

```ts
export type TranslationConfigParams = SAPDocumentTranslation['config'];
````

My VSCode still shows mandatory fields in autocompletion first.

Also, I know we have some existing types with underscores in the property name. Our bad 😅 and probably we should change this. Nevertheless, stick to the camel case naming for the variable.

File: packages/orchestration/src/util/translation.ts [hunk 4]
```diff
@@ -0,0 +1,30 @@
...
+ * @param config - Config for SAP Document Translation service.
+ * The target language is mandatory, while source language will be auto-detected if not provided.
+ * See https://help.sap.com/docs/translation-hub/sap-translation-hub/supported-languages-6854bbb1bd824ffebc3a097a7c0fd45d?state=DRAFT for list of supported languages.
+ * @returns A `SAPDocumentTranslation` configuration object.
+ * @example "buildOrchestrationTranslationConfig({source_language: 'de-DE', target_language: 'en-US'})"
+ */
+export function buildOrchestrationTranslationConfig(
+  config?: TranslationConfigParams
+): SAPDocumentTranslation {
+  if (!config || !config.target_language || !config.target_language.trim()) {
```
ZhongpinWang, line 15: [req] What about checking `!config?.target_language`?

[pp] I don't have strong opinion for whether we should add some sanity check by using `trim()` or maybe some other utilities. Because orchestration currently also doesn't do this and just forward the requests to the translation service. So technically empty string is just the same as any invalid string and orchestration service should do the check and return us something meaningful (other than the current 500). Already talked to orchestration and they will figure it out.

packages/orchestration/src/util/translation.ts
```diff
@@ -0,0 +1,29 @@
+import type { SAPDocumentTranslation } from '../client/api/schema/index.js';
+
+/**
+ * Convenience function to build a document translation configuration for orchestration service.
+ * @param config - Config for SAP Document Translation service.
+ * The target language is mandatory, while source language will be auto-detected if not provided.
+ * See https://help.sap.com/docs/translation-hub/sap-translation-hub/supported-languages-6854bbb1bd824ffebc3a097a7c0fd45d?state=DRAFT for list of supported languages.
```
ZhongpinWang, line 7: [q] Can we remove DRAFT?

packages/orchestration/src/util/translation.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @returns A `SAPDocumentTranslation` configuration object.
+ * @example "buildTranslationConfig({source_language: 'de-DE', target_language: 'en-US'})"
+ */
+export function buildTranslationConfig(
+  config?: SAPDocumentTranslation['config']
```
ZhongpinWang, line 12: [pp] Can we still create a wrapper type for this? Two reasons, first we should use camel case, second if orchestration put something new in config, it won't immediately affect us when doing an orchestration spec update. This still provides us room to do adjustment.

packages/orchestration/src/util/translation.ts
```diff
@@ -0,0 +1,30 @@
+import type { TranslationConfigParams } from '../orchestration-types.js';
+import type { SAPDocumentTranslation } from '../client/api/schema/index.js';
+
+/**
+ * Convenience function to build a document translation configuration for orchestration service.
+ * @param config - Config for SAP Document Translation service.
+ * The target language is mandatory, while source language will be auto-detected if not provided.
+ * See https://help.sap.com/docs/translation-hub/sap-translation-hub/supported-languages-6854bbb1bd824ffebc3a097a7c0fd45d for list of supported languages.
+ * @returns A `SAPDocumentTranslation` configuration object.
+ * @example "buildTranslationConfig({source_language: 'de-DE', target_language: 'en-US'})"
```
deekshas8, line 10:
```suggestion
 * @example "buildTranslationConfig({ sourceLanguage: 'de-DE', targetLanguage: 'en-US' })"
```

packages/orchestration/src/orchestration-types.ts
```diff
@@ -222,3 +232,17 @@ export type AzureFilterThreshold = keyof typeof supportedAzureFilterThresholds;
...
 export type LlamaGuardCategory = keyof LlamaGuard38B;
+
+/**
+ * Translation configuration for SAP Document Translation.
+ */
+export interface TranslationConfigParams {
+  /**
+   * Language of the text to be translated.
+   */
+  sourceLanguage?: string;
```
deekshas8, line 243: [pp] Can you add examples here as well, so the format is more obvious, in case user doesnt use the helper function

shibeshduw: We should probably then also add the link to supported languages here as well?

ZhongpinWang: Don't we have the link already in `buildTranslationConfig()`?

shibeshduw: Yes but assuming user doesn't use the helper function and just uses this type, UX wise might be better to have it here as well.

ZhongpinWang: I mean we would always recommend users to use the helper functions. But no strong opinion on this, both are fine. 👍

packages/orchestration/src/util/translation.test.ts
```diff
@@ -0,0 +1,38 @@
...
+      config: {
+        source_language: 'de-DE',
+        target_language: 'en-US'
+      }
+    };
+
+    expect(translationConfig).toEqual(expectedTranslationConfig);
+  });
+
+  it('throw error when config is empty', async () => {
```
ZhongpinWang, line 23:
```suggestion
  it('throws error when config is empty', async () => {
```

packages/orchestration/src/util/translation.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    expect(() => {
+      buildTranslationConfig();
+    }).toThrow('Target language is required for translation configuration.');
+  });
+
+  it('throw error when target language is empty', () => {
```
ZhongpinWang, line 29:
```suggestion
  it('throws error when target language is empty', () => {
```

packages/orchestration/src/util/translation.ts
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Convenience function to build a document translation configuration for orchestration service.
+ * @param config - Config for SAP Document Translation service.
+ * The target language is mandatory, while source language will be auto-detected if not provided.
+ * See https://help.sap.com/docs/translation-hub/sap-translation-hub/supported-languages-6854bbb1bd824ffebc3a097a7c0fd45d for list of supported languages.
+ * @returns A `SAPDocumentTranslation` configuration object.
+ * @example "buildTranslationConfig({ sourceLanguage: 'de-DE', targetLanguage: 'en-US' })"
+ */
+export function buildTranslationConfig(
+  config?: TranslationConfigParams
```
ZhongpinWang, line 13: [req] Why is `config` optional? If we remove `?`, then as per type definition, `targetLanguage` in the next if cannot be undefined right?

ZhongpinWang (changes requested):

packages/orchestration/src/util/translation.ts
shibeshduw: You're right, was an oversight on my part. Made the change.

packages/orchestration/src/util/translation.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @example "buildTranslationConfig({ sourceLanguage: 'de-DE', targetLanguage: 'en-US' })"
+ */
+export function buildTranslationConfig(
+  config: TranslationConfigParams
+): SAPDocumentTranslation {
+  if (!config.targetLanguage) {
+    throw new Error(
+      'Target language is required for translation configuration.'
+    );
+  }
```
ZhongpinWang, line 19: I feel we can remove this. It now just check if `targetLanguage` is an empty string, same as many other invalid string. Sanity check should eventually be done on the service side.

shibeshduw: I guess that makes sense if we expect it to happen service side eventually.

PR #726, opened 2025-05-16
feat: Support LangChain Azure OpenAI `bindTools()` method

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#308.

Before, user scan only pass tools when calling `invoke()`.

```ts
const response = await client.invoke(messages, { tools: [azureTool] });
```

Now, they can bind tools to a client before calling the invoke method.

```ts
const myTool = tool(...);
const modelWithTools = client.bindTools([myTool]);
```

Both `bindTools()` and `invoke()` support the following tool types:

1. Tools created with `tool()` function from LangChain
2. Structured tool `{ name: string, description: string, schema: ZodSchema | JSONSchema }`
3. Azure tool / function definition `{ type: 'function', function: { name: string, description: string, parameters: JSONSchema } }`

### DIFF
--- a/.changeset/empty-corners-ask.md
+++ b/.changeset/empty-corners-ask.md
@@ -1,0 +1,5 @@
+---
+'@sap-ai-sdk/langchain': minor
+---
+
+[New Feature] Support `bindTools()` method in Azure OpenAI LangChain client.
--- a/packages/langchain/src/openai/chat.ts
+++ b/packages/langchain/src/openai/chat.ts
@@ -1,14 +1,18 @@
 import { AzureOpenAiChatClient as AzureOpenAiChatClientBase } from '@sap-ai-sdk/foundation-models';
 import { BaseChatModel } from '@langchain/core/language_models/chat_models';
-import { mapLangchainToAiClient, mapOutputToChatResult } from './util.js';
+import {
+  mapLangchainToAiClient,
+  mapOutputToChatResult,
+  mapToolToOpenAiTool
+} from './util.js';
 import type { BaseLanguageModelInput } from '@langchain/core/language_models/base';
-import type { BindToolsInput } from '@langchain/core/language_models/chat_models';
 import type { AIMessageChunk, BaseMessage } from '@langchain/core/messages';
 import type { CallbackManagerForLLMRun } from '@langchain/core/callbacks/manager';
 import type { ChatResult } from '@langchain/core/outputs';
 import type {
   AzureOpenAiChatCallOptions,
-  AzureOpenAiChatModelParams
+  AzureOpenAiChatModelParams,
+  ChatAzureOpenAIToolType
 } from './types.js';
 import type { HttpDestinationOrFetchOptions } from '@sap-cloud-sdk/connectivity';
 import type { Runnable } from '@langchain/core/runnables';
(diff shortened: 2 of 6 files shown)


.changeset/empty-corners-ask.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/langchain': minor
+---
+
+[New Feature] Support `bindTools()` method in Azure OpenAI LangChain client.
```
ZhongpinWang, line 5:
```suggestion
[New Functionality] Support `bindTools()` method in Azure OpenAI LangChain client.
```

packages/langchain/src/openai/util.ts
```diff
@@ -36,39 +38,49 @@ import type {
 /**
  * Maps a LangChain {@link BindToolsInput} to {@link AzureOpenAiChatCompletionFunctions}.
```
KavithaSiva, line 39: Maps `{@link ChatAzureOpenAIToolType}` instead of `BindToolsInput`

packages/langchain/src/openai/util.ts
```diff
@@ -36,39 +38,49 @@ import type {
...
     parameters: isZodSchema(tool.schema)
       ? zodToJsonSchema(tool.schema)
-      : tool.schema
+      : tool.schema,
+    ...(strict !== undefined ? { strict } : {})
   };
 }
 
 /**
  * Maps a LangChain {@link BindToolsInput} to {@link AzureOpenAiChatCompletionTool}.
```
KavithaSiva, line 71: Maps `{@link ChatAzureOpenAIToolType}` instead of `BindToolsInput`

KavithaSiva (approved):
Mostly LGTM, just few minor JSdoc improvements.

packages/langchain/src/openai/chat.ts
```diff
@@ -69,4 +80,24 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
 
     return mapOutputToChatResult(res.data);
   }
+
+  override bindTools(
```
KavithaSiva, line 84: [req] Can you add a unit test for `bindTools` as well?

packages/langchain/src/openai/util.ts
ZhongpinWang: Fixed also other occurrences.

packages/langchain/src/openai/util.ts
```diff
@@ -275,8 +295,8 @@ export function mapLangchainToAiClient(
     top_logprobs: options?.top_logprobs,
     function_call: options?.function_call,
     stop: options?.stop ?? client.stop,
-    functions: options?.functions?.map(mapToolToOpenAiFunction),
-    tools: options?.tools?.map(mapToolToOpenAiTool),
+    functions: options?.functions?.map(f => mapToolToOpenAiFunction(f)),
```
tomfrenken, line 298: Why add f => here?

tomfrenken (approved):
Overall LGTM, however, the two mapping functions can be slightly refactored IMO

packages/langchain/src/openai/util.ts
ZhongpinWang: I have to. Because now the map function takes a second param `strict` and ts thinks it should be `index` if not write like this.

PR #736, opened 2025-05-20
feat: Add orchestration tools support

tomfrenken:
## Context

Closes SAP/ai-sdk-js-backlog#301.

Adds complete tools support, both in regular and streaming calls.


.changeset/hungry-candles-show.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[New Functionality] Introduce support for tool calls, for both streaming and regular invokation.
```
deekshas8, line 5:
```suggestion
[New Functionality] Introduce support for tool calls for both streaming and regular invocation.
```

packages/orchestration/src/orchestration-stream-response.ts
```diff
@@ -41,6 +48,32 @@ export class OrchestrationStreamResponse<T> {
...
 
+  public getToolCalls(choiceIndex = 0): MessageToolCalls | undefined {
+    try {
+      const toolCallChunks = this._toolCallChunks.get(choiceIndex);
+      if (!toolCallChunks) {
+        throw new Error(`No tool calls found for choice index ${choiceIndex}`);
+      }
+      const toolCalls: MessageToolCalls = [];
+      for (const chunkArray of toolCallChunks.values()) {
+        const toolCall = mergeToolCallChunks(chunkArray);
```
ZhongpinWang, line 59: [q] Can we not merge them on the fly? Otherwise if we store chunks, we might as well store chunks also for normal no tool call case? The reason we didn't store chunks is that this could take too much memory. So e.g. for the `toContentStream()` case, we aggregate content already on the fly.

sample-code/src/orchestration.ts
```diff
@@ -633,3 +632,38 @@ export async function orchestrationMessageHistoryWithToolCalling(): Promise<Orch
     messagesHistory: allMessages
   });
 }
+
+/**
+ * Ask ChatGPT to do add two numbers with tools.
```
deekshas8, line 637:
```suggestion
 * Ask ChatGPT to add two numbers using tools.
```

sample-code/src/orchestration.ts
```diff
@@ -633,3 +632,38 @@ export async function orchestrationMessageHistoryWithToolCalling(): Promise<Orch
...
+  streamOptions?: StreamOptions
+): Promise<OrchestrationStreamResponse<OrchestrationStreamChunkResponse>> {
+  const orchestrationClient = new OrchestrationClient({
+    // define the language model to be used
+    llm: {
+      model_name: 'gpt-4o'
+    },
+    // define the prompt
+    templating: {
+      template: [
```
deekshas8, line 653: [req] With the `template` property change, maybe we can now use messages. This use case doesnt seem like it needs a template

packages/orchestration/src/util/tool-calls.ts
```diff
@@ -0,0 +1,72 @@
...
+type ToolCallAccumulator = {
+  id?: string;
+  type: 'function';
+  function: {
+    name?: string;
+    arguments?: string;
+  } & Record<string, any>;
+} & Record<string, any>;
+
+function validateToolCallAccumulator(acc: ToolCallAccumulator): boolean {
```
ZhongpinWang, line 15: [pp] Maybe write it as a type guard? Since later we write `return acc as MessageToolCall`, why not just return here `is MessageToolCall`?

deekshas8 (dismissed):

packages/orchestration/src/util/tools.ts
```diff
@@ -0,0 +1,41 @@
+import { zodToJsonSchema } from 'zod-to-json-schema';
```
ZhongpinWang, line 1: [pp] Thanks for adding this file. Any idea how to reuse those things in Sample-Code and LangChain? Can we define those tools in somewhere else? If we reuse this, maybe also not type the tool with `ChatCompletionTool`.

sample-code/src/orchestration.ts
```diff
@@ -566,21 +566,20 @@ const addNumbersSchema = z
   })
   .strict();
 
+const addNumbersTool: ChatCompletionTool = {
```
ZhongpinWang, line 569: [pp] Would be nice if we can reuse.

ZhongpinWang (commented):
Already looks promising. Only one open question regarding storing the tool call chunks.

packages/orchestration/src/orchestration-stream-response.ts
tomfrenken: Good point, also simplifies the mapping logic.

packages/orchestration/src/orchestration-stream-chunk-response.ts
```diff
@@ -26,9 +27,12 @@ export class OrchestrationStreamChunkResponse {
    * @returns The finish reason.
    */
   getFinishReason(choiceIndex = 0): string | undefined {
-    return this.getChoices()?.find(
-      (c: LlmChoiceStreaming) => c.index === choiceIndex
-    )?.finish_reason;
+    return this.getChoices()?.find(c => c.index === choiceIndex)?.finish_reason;
+  }
+
+  getToolCalls(choiceIndex = 0): ToolCallChunk[] | undefined {
```
ZhongpinWang, line 33:
```suggestion
  getDeltaToolCalls(choiceIndex = 0): ToolCallChunk[] | undefined {
```
It was a bit confusing as ToolCalls sound complete already but it's returning chunks.

packages/orchestration/src/orchestration-stream-response.ts
```diff
@@ -41,6 +50,38 @@ export class OrchestrationStreamResponse<T> {
     this._finishReasons = finishReasons;
   }
 
+  public getToolCalls(choiceIndex = 0): MessageToolCalls | undefined {
```
ZhongpinWang, line 53: [pp] I saw the other public method also does not have jsdoc. I guess good to have?

packages/orchestration/src/orchestration-stream-response.ts
```diff
@@ -41,6 +50,38 @@ export class OrchestrationStreamResponse<T> {
...
+    try {
+      const toolCallsAccumulators =
+        this._toolCallsAccumulators.get(choiceIndex);
+      if (!toolCallsAccumulators) {
+        throw new Error(`No tool calls found for choice index ${choiceIndex}`);
```
ZhongpinWang, line 58: [q] What about return undefined? When do we return undefined?

packages/orchestration/src/orchestration-stream.ts
```diff
@@ -51,53 +49,87 @@ export class OrchestrationStream<Item> extends SseStream<Item> {
...
-            const finishReason = chunk.getFinishReason(choiceIndex);
-            if (finishReason) {
-              if (response) {
-                response._getFinishReasons().set(choiceIndex, finishReason);
+      chunk.data.orchestration_result?.choices.forEach(choice => {
+        const choiceIndex = choice.index;
+        if (choiceIndex >= 0) {
+          const toolCallsChunks = chunk.getToolCalls(choiceIndex);
+          if (toolCallsChunks) {
+            if (response) {
```
ZhongpinWang, line 62: [pp] Is checking `choiceIndex >= 0` necessary? If it is -1 we don't care right? Because `chunk.getToolCalls(choiceIndex)` will return `undefined` anyway. So I feel we can merge this three `if`s into one.

ZhongpinWang: Or even better, we can use `?` for the next `toolCallAccumulators` and merge `if (!toolCallAccumulators)` as well.

ZhongpinWang (commented):
some more comments

ZhongpinWang (approved):

test-util/tools.ts
```diff
@@ -12,7 +12,7 @@ const addNumbersSchema = z
 /**
  * @internal
  */
-export const addTool: ChatCompletionTool = {
+export const addNumbersTool: ChatCompletionTool = {
```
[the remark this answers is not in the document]
tomfrenken: While true, if they are not typecasted, the clients still complain when we pass the object.
I think when this is imported in these other clients, they just need to typecast them to their respective object.

packages/orchestration/src/orchestration-stream-response.ts
tomfrenken: Hmm fair, I suppose returning undefined is the better response.

packages/orchestration/src/util/tool-calls.ts
```diff
@@ -12,11 +12,12 @@ type ToolCallAccumulator = {
...
 
-function validateToolCallAccumulator(acc: ToolCallAccumulator): boolean {
+function isMessageToolCall(acc: ToolCallAccumulator): acc is MessageToolCall {
   return (
-    !!acc.id &&
+    typeof acc.id === 'string' &&
     acc.type === 'function' &&
-    !!acc.function &&
+    typeof acc.function === 'object' &&
+    acc.function !== null &&
```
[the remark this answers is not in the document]
tomfrenken: Then again, it is always set by me in my current logic, if it wasn't already set, so I guess I can just remove this 😬

PR #740, opened 2025-05-21
chore: Fix types and checks for TemplatingModuleConfig

deekshas8:
## Context

This PR fixes some issues identified in the failing e2e test:

- Tightened types of `TemplatingModuleConfig` to `Xor<Template, Template_Ref>` so user gets an error when they set both `template` and `template_ref`
- Not doing a shallow copy of the instance property config when concatenating `messages`
- Improving the TypeGuard to check for `template_ref`


tomfrenken (approved):
LGTM

PR #750, opened 2025-05-27
fix: Add `prompt_filter_results` to langchain (openai) mapping 

shibeshduw:
## Context

Closes SAP/ai-sdk-js-backlog#319.

Adds the property `prompt_filter_results` to the response mapping function (`mapOutputToChatResult`)  in the langchain client for openai. 
This property, is declared as optional in the `AzureOpenAiCreateChatCompletionResponse` but is always returned in the response even if content filtering is not configured.


packages/langchain/src/openai/util.ts
```diff
@@ -146,6 +146,7 @@ export function mapOutputToChatResult(
       id: completionResponse.id,
       model: completionResponse.model,
       object: completionResponse.object,
+      promptFilterResults: completionResponse.prompt_filter_results ?? [],
```
deekshas8, line 149:
```suggestion
      promptFilterResults: completionResponse.prompt_filter_results,
```

Thought about this, just assigning should also be enough imo.

deekshas8 (approved):
Consider the comment.

PR #776, opened 2025-06-06
feat: LangChain OpenAI streaming

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#314.

Closes SAP/ai-sdk-js-backlog#321.

- Support foundation models OpenAI streaming with tool calls
- Support LangChain OpenAI streaming (with tool calls)
- Support multi choices in both LangChain clients
- Some minor changes for, e.g., `generationInfo`


packages/langchain/src/openai/chat.ts
```diff
@@ -101,4 +104,92 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
...
+          mapLangChainToAiClient(this, messages, options),
+          controller,
+          options.requestConfig
+        );
+      }
+    );
+
+    for await (const chunk of response.stream) {
+      // There can be only none or one choice inside a chunk
+      const choice = chunk.data.choices[0];
```
ZhongpinWang, line 139: [for reviewer] Notice choice can be `undefined` in case of the last chunk, which returns usage.

packages/langchain/src/openai/chat.ts
```diff
@@ -101,4 +104,92 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
...
+
+      // Map the chunk to a LangChain message chunk
+      const messageChunk = mapAzureOpenAIChunkToLangChainMessageChunk(chunk);
+
+      // Create initial generation info with token indices
+      const newTokenIndices: NewTokenIndices = {
+        prompt: options.promptIndex ?? 0,
+        completion: choice?.index ?? 0
```
ZhongpinWang, line 147: [for reviewer] I see other clients just add `promptIndex` in the call options. I assume either user or LangChain will set this. Otherwise, `0` is also not bad. Completion is just the choice index.

packages/langchain/src/openai/chat.ts
```diff
@@ -101,4 +104,92 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
...
+        return this.openAiChatClient.stream(
+          mapLangChainToAiClient(this, messages, options),
+          controller,
+          options.requestConfig
+        );
+      }
+    );
+
+    for await (const chunk of response.stream) {
+      // There can be only none or one choice inside a chunk
```
ZhongpinWang, line 138: [for reviewer] Even if it is a multiple choice streaming, each chunk only contains one choice. OpenAI and many other clients have this assumption.

packages/langchain/src/openai/chat.ts
```diff
@@ -101,4 +104,92 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
...
+
+      // Process finish reason
+      if (choice?.finish_reason) {
+        generationInfo.finish_reason = choice.finish_reason;
+        // Only include system fingerprint in the last chunk for now to avoid concatenation issues
+        generationInfo.system_fingerprint = chunk.data.system_fingerprint;
+        generationInfo.model_name = chunk.data.model;
+        generationInfo.id = chunk.data.id;
+        generationInfo.created = chunk.data.created;
+        generationInfo.index = choice.index;
```
ZhongpinWang, line 159: [for reviewer] `generationInfo` properties will be shown inside [`response_metadata`](https://python.langchain.com/docs/how_to/response_metadata/). It contains raw response from the LLM. Storing information here can avoid the concat issue, as everything in `additional_kwargs` will be concatenated. And normally, `additional_kwargs` contains `function_call` or `tool_calls`.

Fyi, `concat` works for openai.

packages/langchain/src/orchestration/client.ts
```diff
@@ -182,26 +180,57 @@ export class OrchestrationClient extends BaseChatModel<
...
-      const tokenIndices = computeTokenIndices(chunk);
-      const finishReason = response.getFinishReason();
-      const tokenUsage = response.getTokenUsage();
 
-      setFinishReason(messageChunk, finishReason);
-      setTokenUsage(messageChunk, tokenUsage);
+      // Create initial generation info with token indices
+      const newTokenIndices: NewTokenIndices = {
+        prompt: options.promptIndex ?? 0,
+        completion: choice?.index ?? 0
```
ZhongpinWang, line 193: [for reviewer] Supporting orchestration multi choice by adding this line. It is simple because each chunk contain only one choice.

packages/langchain/src/orchestration/util.ts
```diff
@@ -356,76 +351,17 @@ export function isToolDefinitionLike(
...
-    prompt: 0,
-    // Hardcoding to 0 as mutiple choices are not currently supported in the orchestration service.
-    // TODO: Switch to `chunk.data.orchestration_result.choices[0].index` when support is added via https://github.com/SAP/ai-sdk-js-backlog/issues/321
-    completion: 0
-  };
+  const toolCallChunks = choice?.delta.tool_calls;
+  return new AIMessageChunk({
+    content,
+    additional_kwargs: {
+      // TODO: Fix duplicated module results when using concat() method for streaming chunks.
```
ZhongpinWang, line 360: [for reviewer] Putting `module_results` in `additional_kwargs` because it also needs to be concatenated. Still we have issues when calling LangChain's `concat()` method as content of many fields are replicated multiple times. For example, `'role': 'assistantassistantassistantassistant'`

ZhongpinWang (commented):
Some explanations for reviewers. Notice that the code might be refactored later. But I would appreciate your comments on the logic and functionalities.

packages/langchain/src/openai/util.ts
```diff
@@ -302,6 +302,43 @@ export function mapLangChainToAiClient(
   });
 }
 
+/**
+ * Converts Azure OpenAI stream chunk to a LangChain message chunk.
+ * @param chunk - The Azure OpenAI stream chunk.
+ * @returns An {@link AIMessageChunk}
+ * @internal
+ */
+export function mapAzureOpenAIChunkToLangChainMessageChunk(
```
tomfrenken:
```suggestion
export function mapAzureOpenAiChunkToLangChainMessageChunk(
```

packages/langchain/src/orchestration/client.ts
```diff
@@ -182,26 +180,56 @@ export class OrchestrationClient extends BaseChatModel<
...
+      };
+      const generationInfo: Record<string, any> = { ...newTokenIndices };
```
tomfrenken, line 195: Is there a more narrow type, or is it supposed to be an object with arbitrary values?

packages/langchain/src/orchestration/client.ts
```diff
@@ -182,26 +180,56 @@ export class OrchestrationClient extends BaseChatModel<
...
+
+      // Process finish reason
+      if (choice?.finish_reason && orchestrationResult) {
+        generationInfo.finish_reason = choice.finish_reason;
+        // Only include system fingerprint in the last chunk for now to avoid concatenation issues
+        generationInfo.system_fingerprint =
+          orchestrationResult.system_fingerprint;
```
tomfrenken, line 202: That's nice addition that we missed before, is there some generation info docs that you followed for all this information, or what is it based on?

test-util/data/foundation-models/azure-openai-chat-completion-stream-chunk-response-delta-content.json
```diff
@@ -1,36 +1,36 @@
 {
```
ZhongpinWang, line 1: [for reviewer] Ignore. Only formatting.

packages/langchain/src/orchestration/client.ts
ZhongpinWang: `generationInfo` can contain any arbitrary object with string as the key. It is indeed originally defined as this structure.

```ts
export declare class GenerationChunk implements Generation {
    text: string;
    generationInfo?: Record<string, any>;
    constructor(fields: GenerationChunkFields);
    concat(chunk: GenerationChunk): GenerationChunk;
}
```

packages/langchain/src/orchestration/client.ts
ZhongpinWang: According to official LangChain, `generationInfo` aka. `response_metadata` should contain whatever metadata is generated during the LLM call process. In my view, `system_fingerprint` or similar are metadata created when sending the request. `module_results` on the other hand does not feel like metadata, thus, put in the additional kwargs.

- https://python.langchain.com/docs/how_to/response_metadata/
- https://js.langchain.com/docs/how_to/chat_token_usage_tracking/#using-aimessageresponse_metadata

PR #781, opened 2025-06-10
chore: Cleanup orchestration samples/examples

shibeshduw:
## Context

Closes SAP/ai-sdk-js-backlog#317.

This PR cleans up all tests/sample code to use `messages` for new messages in a conversational style request and use `templating` for static one time chat message.


sample-code/src/orchestration.ts
```diff
@@ -74,25 +69,45 @@ export async function chatCompletionStream(
...
     }
   });
 
   return orchestrationClient.stream(
-    { inputParams: { input: 'SAP Cloud SDK' } },
+    {
+      messages: [
+        { role: 'user', content: 'Give me a long introduction of {{?input}}.' }
+      ],
+      inputParams: { input: 'SAP Cloud SDK' }
```
KavithaSiva: [pp]

```suggestion
      messages: [
        { role: 'user', content: 'Give me a long introduction of SAP Cloud SDK.' }
      ]
```

sample-code/src/orchestration.ts
```diff
@@ -125,52 +140,27 @@ export async function chatCompletionStreamWithJsonModuleConfig(
   );
 }
 
-const llm: LlmModuleConfig = {
```
KavithaSiva, line 128: [pp] Why remove this variable? Isn't it easier to maintain it in one place, if we are only testing one model?

sample-code/src/orchestration.ts
```diff
@@ -265,16 +260,14 @@ export async function orchestrationOutputFiltering(): Promise<OrchestrationRespo
...
-        role: 'system',
+        role: 'user',
         content:
           'Reparaphrase the sentence in 30 ways with strong feelings: "{{?input}}"'
       }
     ],
-    inputParams: {
-      input: 'I hate you!' // Should be filtered by the Azure content filter
-    }
+    inputParams: { input: 'I hate you!' } // Should be filtered by the Azure content filter
```
KavithaSiva: [pp] I would personally stick to using input params only if we are using templating in the examples. It's much cleaner. Applies for examples below too. (Except for the grounding and image input cases, where inputParams is cleaner). WDY think ?

KavithaSiva (commented):
Left some minor questions.

sample-code/src/orchestration.ts
shibeshduw: I guess it depends on if we want our samples to be standalone. This was the only shared variable in the file and it only had 2 call sites. I don't have a preference because this is a collection of samples.

sample-code/src/orchestration.ts
shibeshduw: Sounds good

deekshas8 (approved):
Looks a lot cleaner now. LGTM.

PR #782, opened 2025-06-10
chore: Tutorial agent

deekshas8:
## Context


sample-code/src/tutorials/agent-workflow-openai-langchain.ts
```diff
@@ -0,0 +1,195 @@
...
+}
+
+function askHuman({ messages }: typeof MessagesAnnotation.State) {
+  const lastMessage = messages[messages.length - 1] as AIMessage;
+  const toolCall = lastMessage.tool_calls?.[0];
+  const toolCallId = toolCall?.id;
+  const question = toolCall?.args?.question;
+  const context = toolCall?.args?.context;
+  
+  console.log(`\nHuman Input Requested: ${question} with context: ${context}`);
```
ZhongpinWang: Can we make this log more human readable? Currently, the first message as a user looks like: 

```
Human Input Requested: Would you like me to make any adjustments to this itinerary? with context: Here's a practical 3-item itinerary for your trip to Paris based on the current weather and popular spots: 1. Morning: Visit the Eiffel Tower. 2. Afternoon: Lunch at Le Comptoir du Relais and explore Notre-Dame Cathedral. 3. Evening: Dinner at Breizh Café and stroll the Marais district.
```

Ideally, we can have something like:

```
${context}

${question}
```

sample-code/src/tutorials/agent-workflow-openai-langchain.ts
```diff
@@ -0,0 +1,195 @@
...
+       
+       IMPORTANT: Every time you present or modify an itinerary, you MUST use the humanAssistanceTool to ask the user if they want adjustments. Do not just ask in text - you must call the tool.
+       
+       Call humanAssistanceTool with:
+       - question: "Would you like me to make any adjustments to this itinerary?"
+       - context: Brief description of the itinerary you just presented`
+      ),
+      new HumanMessage(
+        "I'm traveling to Paris. Can you help me prepare an itinerary?"
+      )
```
ZhongpinWang: Can we extract input / agent trigger, human input, and agent output into a separate controller-like file? So that our SDK users understand, the agent is a pure graph, which requires user/system trigger, and gives some output or does something behind, and ask for human intervention if needed.

sample-code/src/tutorials/agent-workflow-openai-langchain.ts
```diff
@@ -0,0 +1,195 @@
...
+  const lastMessage = messages[messages.length - 1] as AIMessage;
+  const toolCall = lastMessage.tool_calls?.[0];
+  const toolCallId = toolCall?.id;
+  const question = toolCall?.args?.question;
+  const context = toolCall?.args?.context;
+  
+  console.log(`\nHuman Input Requested: ${question} with context: ${context}`);
+  
+  // This is where the actual interrupt happens
+  const humanResponse: string = interrupt(`${question}: ${context}`);
```
ZhongpinWang: Same here, interrupt is one command line way to ask for human input. Thinking about using Joule, it can be some callback function from Joule, or maybe we send an email to the app user and wait for their response and agent pause.

sample-code/src/tutorials/agent-workflow-openai-langchain.ts
deekshas8: This is a log statement anyway, I'll remove them

sample-code/src/tutorials/agent-workflow-openai-langchain.ts
deekshas8: Unless we're building a full fledged application, I will leave it like this. When we introduce controller, we have to consider many things, saving each conversation, how interrupt is handled in the UI app, etc.

sample-code/src/server.ts
```diff
@@ -88,6 +89,14 @@ function sendError(res: any, error: any, send: boolean = true) {
   }
 }
 
+app.get('/langchain/langgraph-tutorial', async (req, res) => {
```
ZhongpinWang: [req] Can we put this endpoint in the sample code README?

sample-code/src/tutorials/agent-workflow-openai-langchain.ts
```diff
@@ -0,0 +1,135 @@
...
+    new HumanMessage(`Assistant message: "${lastMessage.content}"`)
+  ]);
+  
+  return result.content === 'FAREWELL' ? END : 'askHuman';
+}
+
+async function askHuman({ messages }: typeof MessagesAnnotation.State) {
+  // This is where the actual interrupt happens
+  const humanResponse: string = interrupt(`Do you want to adjust the itinerary?`);
+  return { messages: [new HumanMessage(humanResponse)] };
```
ZhongpinWang: [q] I feel this will clear the message history? You would need to push the human message to the `state.messages` and then return it. Normally, we return here `response` which contains all messages.

sample-code/src/tutorials/agent-workflow-openai-langchain.ts
```diff
@@ -0,0 +1,135 @@
...
+async function shouldContinueAgent({ messages }: typeof MessagesAnnotation.State) {
+  const lastMessage = messages[messages.length - 1] as AIMessage;
+  
+  // If there are tool calls, go to tools
+  if (lastMessage.tool_calls?.length) {
+    return 'tools';
+  }
+  // Check if agent's message is a farewell (indicating user was satisfied)
+  const result = await model.invoke([
+    new SystemMessage(`You are a classifier. Respond with exactly "FAREWELL" if this is a farewell/goodbye message wishing someone happy travels. Respond with exactly "CONTINUE" if the conversation should continue.`),
```
ZhongpinWang: [pp] I am not sure if it is a good practice to have multiple system messages and tell LLM to be something else. (No official answer..) It might affect the original context and change the behaviour of the upcoming answers.

Could you try to move this part to the first system message? I feel it might be effectively the same?

(Also, if we want to reuse this tutorial for other models / orchestration, some models do not support multiple system messages, e.g., in Anthropic Claude, there is no system role but just a top level `system` parameter to set it once and for all.)

sample-code/src/tutorials/agent-workflow-openai-langchain.ts
```diff
@@ -0,0 +1,135 @@
...
+    try {
+      let response = await app.invoke({ messages }, config);
+     
+      console.log('Assistant:', response.messages[response.messages.length - 1].content);
+      console.log("next: ", (await app.getState(config)).next);
+    
+      response = await app.invoke(new Command({ resume: 'Can you suggest something more outdoorsy?' }), config);
+
+      console.log('Assistant:', response.messages[response.messages.length - 1].content);
+
```
ZhongpinWang: [q] A `next` here as well?

sample-code/src/tutorials/agent-workflow-openai-langchain.ts
deekshas8: Great catch, I had thought of creating 2 separate llm context, one with tools, one just for classification. I missed this during the implementation. Definitely better to keep them separate.

This didn't affect the original context though, I believe because this message is not added to the conversation thread. But I agree, does not sound like a good practice.

sample-code/src/tutorials/agent-workflow-openai-langchain.ts
deekshas8: Was conflicted about it, whether we need to show the use of next again, since I already explained what this is for before. If you have a strong opinion, I can add it back

sample-code/src/tutorials/agent-workflow-openai-langchain.ts
deekshas8: `response` is the latest AIMessage and not the whole history. When we return { messages: [response] }, the response object is actually appended to the messages array, due to the concept of reducers https://langchain-ai.github.io/langgraphjs/concepts/low_level/#reducers

So in this case as well, the message is appended and not overwritten

sample-code/src/server.ts
deekshas8: I am wondering now if its even a good idea to have it as an endpoint, especially since we do console log and I would like to keep it like that (we can make it into an app later). I have removed the endpoint, and made it into another script to run this file. Honestly dont like it either, but mixing a tutorial into a sample code app which exposes simple endpoints already feels strange organization. Lets discuss a better approach here

sample-code/src/tutorials/agent-openai-langchain.ts
```diff
@@ -0,0 +1,166 @@
...
+  )
+];
+
+// Start the agent with initial messages
+try {
+  let response = await app.invoke({ messages: initMessages }, config);
+
+  console.log(
+    'Assistant:',
+    response.messages[response.messages.length - 1].content
```
tomfrenken:
```suggestion
    response.messages.at(-1).content
```

tomfrenken: same for the other occurences

sample-code/src/tutorials/agent-openai-langchain.ts
```diff
@@ -0,0 +1,157 @@
...
+  temperature: 0.7
+});
+
+// create a model with access to the tools
+const modelWithTools = model.bindTools(tools);
+
+async function shouldContinueAgent({
+  messages
+}: typeof MessagesAnnotation.State) {
+  const lastMessage = messages[messages.length - 1] as AIMessage;
```
tomfrenken: ~~Can you do the same `.at(-1)?` here, or does the typecasting run into issues then?~~
Use `at(-1)` here too 👀

sample-code/src/tutorials/agent-openai-langchain.ts
```diff
@@ -0,0 +1,157 @@
...
+    new SystemMessage(
+      'You are a classifier. Respond with exactly "FAREWELL" if this is a farewell/goodbye message wishing someone happy travels. Respond with exactly "CONTINUE" if the conversation should continue.'
+    ),
+    new HumanMessage(`Assistant message: "${lastMessage.content}"`)
+  ]);
+
+  return result.content === 'FAREWELL' ? END : 'askHuman';
+}
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+async function askHuman({ messages }: typeof MessagesAnnotation.State) {
```
tomfrenken: the "messages" are actually not used here at all

PR #800, opened 2025-06-16
chore: Add latest embedding models

tomfrenken:
## Context

Closes SAP/ai-sdk-js-backlog#331.

With this PR, we add the latest embedding models that are available for the orchestration service.


tomfrenken:
Exactly, since we don't have embedding capabilities in the orchestration client yet, you can't use them.

PR #807, opened 2025-06-18
chore: try out orch spec update after type fix

deekshas8:
## Context


packages/langchain/src/orchestration/util.ts
```diff
@@ -297,7 +297,7 @@ export function mapOutputToChatResult(
         additional_kwargs: {
           finish_reason: choice.finish_reason,
           index: choice.index,
-          function_call: choice.message.function_call,
```
deekshas8, line 300: @tomfrenken I am surprised this ever worked. I believe the type union hid the error

ZhongpinWang: This was due to the mix of streaming and non-streaming types. So ts was inferencing choice as `any` and did not complain about it before. We need to remove it as orchestration does not have such thing as function_call

deekshas8: Of course I understand how types work ;) — I am just surprised we ever wrote this, considering (as you pointed out) orchestration didn't even have a `function_call` property ¯\_(ツ)_/¯ . Was probably a copy over from OpenAI LangChain.

PR #820, opened 2025-06-26
fix: Add resilience for TimeoutError to non-streaming case in Langchain Orchestration

shibeshduw:
## Context

Closes SAP/ai-sdk-js-backlog#327.


shibeshduw:
In the `_streamResponseChunks()` method, we are currently adding [an event listener](https://github.com/SAP/ai-sdk-js/blob/2c711313e1e30ba5559d7ecca8e4d05920dda8bb/packages/langchain/src/orchestration/client.ts#L170) for abort which seems redundant to me since `callWithOptions` [already does this](https://github.com/langchain-ai/langchainjs/blob/451f4568d6f66d8feeeb79aa414bb41a8b913373/langchain-core/src/utils/async_caller.ts#L140) for every call.

Would like your thoughts on this.

packages/langchain/src/orchestration/client.ts
```diff
@@ -85,27 +85,30 @@ export class OrchestrationClient extends BaseChatModel<
...
+
         return orchestrationClient.chatCompletion(
           {
             messages: allMessages,
             inputParams
           },
-          customRequestConfig
+          {
+            ...customRequestConfig,
+            signal: options.signal
```
ZhongpinWang, line 110: Okay, I am not sure where LangChain sets this signal (maybe you know?), but this could work. LangChain turns the signal into aborted and Axios gets this signal and throw error, then LangChain retry due to this error. Am I right about this flow?

shibeshduw: Yes exactly. The problem before was that Axios never received the signal with timeout, so it was throwing an `AbortError` which LangChain doesn't retry.

The signal is set in the [invoke()](https://github.com/langchain-ai/langchainjs/blob/9b0bf4ed57a0c16578dc6a80c1519342b5a01c40/langchain-core/src/runnables/base.ts#L1919) method in `RunnableSequence` class as far as I can tell. It calls a helper function that always ensures a [signal is set](https://github.com/langchain-ai/langchainjs/blob/9b0bf4ed57a0c16578dc6a80c1519342b5a01c40/langchain-core/src/runnables/config.ts#L178) if `timeout` is defined in options.

ZhongpinWang (approved):
LGTM, thanks for the nice investigation!

.changeset/nice-papayas-live.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/langchain': minor
+---
+
+[Fixed Issue]: Allow retry on timeout for non-streaming requests in LangChain Orchestration.
```
ZhongpinWang:
```suggestion
[Fixed Issue] Retry on timeout for non-streaming requests in LangChain Orchestration client.
```

PR #821, opened 2025-06-26
feat: Update orchestration 2506b

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#330.


packages/orchestration/src/index.ts
```diff
@@ -15,7 +15,9 @@ export type {
   LlamaGuardCategory,
   AzureContentFilter,
   AzureFilterThreshold,
-  TranslationConfigParams
+  TranslationConfigParams,
+  AzureContentSafety,
+  AzureContentSafetyFilterConfig
```
ZhongpinWang, line 20: [for reviewer] I added these two aliases and didn't mark them as deprecated. From the user's perspective, nothing changed. We can decide what to do with them when doing the prompt shield BLI.

packages/orchestration/src/client/api/schema/dpi-entity-config.ts
```diff
@@ -3,13 +3,9 @@
...
 /**
  * Representation of the 'DpiEntityConfig' schema.
  */
-export type DpiEntityConfig = {
-  /**
-   * Type of entity to be masked
-   */
-  type: DpiEntities;
-};
+export type DpiEntityConfig = DPIStandardEntity | DPICustomEntity;
```
ZhongpinWang, line 11: [for reviewer] @jjtang1985 We need to support (maybe in our current build function):

- `DPICustomEntity`
- In the already supported `DPIStandardEntity`, we don't support `replacement_strategy`.

packages/orchestration/src/client/api/schema/embedding-result.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright (c) 2025 SAP SE or an SAP affiliate company. All rights reserved.
+ *
+ * This is a generated file powered by the SAP Cloud SDK for JavaScript.
+ */
+import type { Embedding } from './embedding.js';
+/**
+ * Represents an embedding vector returned by embedding endpoint.
+ *
+ */
+export type EmbeddingResult = {
```
ZhongpinWang, line 11: [for reviewer] We will export those embedding types. But same as before, they are not announced as supported.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -315,3 +320,34 @@ export interface TranslationConfigParams {
...
+
+/**
+ * Representation of the 'ToolChatMessage' schema.
+ */
+export type ToolChatMessage = OriginalToolChatMessage;
+
+/**
+ * Only include DataRepositories with the given type.
+ */
+export type DataRepositoryType = OriginalDataRepositoryType;
```
ZhongpinWang: [for reviewer] These are types that used in our sample-code, which requires now aliases.

packages/orchestration/src/index.ts
```diff
@@ -1,5 +1,3 @@
-export * from './client/api/schema/index.js';
```
ZhongpinWang: [for reviewer] As requested in the BLI, I now move all generated types into `internal.ts`. Only types that are used in our `sample-code` are re-exported with aliases.

packages/orchestration/src/client/api/schema/dpi-entity-config.ts
deekshas8: [q] Do we have a ticket created for this?

packages/orchestration/src/client/api/schema/response-chat-message.ts
```diff
@@ -8,7 +8,7 @@ import type { MessageToolCalls } from './message-tool-calls.js';
  * Representation of the 'ResponseChatMessage' schema.
  */
 export type ResponseChatMessage = {
-  role?: 'assistant';
+  role: 'assistant';
```
deekshas8, line 11: [q] I remember we had a workaround for this, is that fixed as well?

packages/orchestration/src/index.ts
deekshas8: [req] Existing Sample code should not determine our export strategy. All generated types are now internal, so  if at all we are using types in sample code, they should also be imported from internal.

packages/orchestration/src/internal.ts
```diff
@@ -1,4 +1,30 @@
-export * from './orchestration-client.js';
+export * from './client/api/schema/index.js';
 export * from './util/index.js';
-export * from './orchestration-types.js';
+export * from './orchestration-client.js';
 export * from './orchestration-response.js';
+export * from './orchestration-stream-chunk-response.js';
+export * from './orchestration-stream-response.js';
+export * from './orchestration-stream.js';
+export type {
```
deekshas8: [req] Why not `export * from './orchestration-types.js';`

packages/orchestration/src/orchestration-types.ts
```diff
@@ -315,3 +320,34 @@ export interface TranslationConfigParams {
...
+/**
+ * Representation of the Azure Content Safety filter config schema.
+ */
+export type AzureContentSafetyFilterConfig =
+  AzureContentSafetyOutputFilterConfig;
+
+/**
+ * Representation of the 'ErrorResponse' schema.
+ */
+export type ErrorResponse = OriginalErrorResponse;
```
deekshas8: [req] The original `ErrorResponse` type didn't really change, so why the alias?

packages/orchestration/src/orchestration-types.ts
deekshas8: [req] Why bother here? The sample code should use the internal types. Its just sample code, not some official public API of the SDK.
We need to show the correct way to import types (which is from internal for generated types), else we're just confusing people.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -315,3 +320,34 @@ export interface TranslationConfigParams {
    */
   targetLanguage: string;
 }
+
+/**
+ * Filter configuration for Azure Content Safety.
+ */
+export type AzureContentSafety = AzureContentSafetyOutput;
```
deekshas8, line 324:
```suggestion
export type AzureContentSafety = AzureContentSafetyInput;
```

[req] The old types should reflect the input filtering config so user can still use the `build` function for input filtering config. Adding a new property is not a breaking change. In the `createCompletionPostRequest` method, we should remove `PromptShield` if it was added there (But I believe there is a separate ticket handle prompt shield properly)

packages/orchestration/src/orchestration-types.ts
```diff
@@ -315,3 +320,34 @@ export interface TranslationConfigParams {
...
+
+/**
+ * Representation of the Azure Content Safety filter config schema.
+ */
+export type AzureContentSafetyFilterConfig =
```
deekshas8, line 329: [req] Same here, change to input.

deekshas8 (changes requested):

packages/orchestration/src/internal.ts
ZhongpinWang: We cannot as TS will complain about the ambiguity of having same type name from generated files and `orchestration-types.js`.

packages/orchestration/src/orchestration-types.ts
ZhongpinWang: Because it is used in `sample-code` and we discussed that whatever exposed in the sample code should be `public` last time. But I agree we need a written guideline to avoid having this back and forth.

packages/orchestration/src/client/api/schema/response-chat-message.ts
ZhongpinWang: No, I will check it. Thanks for pointing this out.

packages/orchestration/src/client/api/schema/dpi-entity-config.ts
ZhongpinWang: Yes, SAP/ai-sdk-js-backlog#341

packages/orchestration/src/orchestration-types.ts
ZhongpinWang: `AzureContentSafety` is used in the deprecated `buildAzureContentFilter()` function and we export this type as public. If I change it to input, it will have an additional field. User will see this field in autocompletion, which does nothing. I think this should be properly handled in the upcoming `PromptShield` BLI.

packages/orchestration/src/orchestration-types.ts
deekshas8: An internal wiki would definitely make sense here.

From what I recall from the discussion, we agreed to keep typing minimal in samples — no need to type every small variable. Also, any public function (and types used by it) is considered part of the public API and should therefore be stable as much as possible— something we've followed in the Cloud SDK for JS.

Sample code is meant to demonstrate correct SDK usage, not influence it. So I’m a bit surprised to see the reverse happening here.

ZhongpinWang: As discussed, I will put my objection on the BLI for doing this now. And let's discuss the guideline in a separate session.

packages/orchestration/src/orchestration-types.ts
deekshas8: > I think this should be properly handled in the upcoming PromptShield BLI.

Yeah makes sense then.

ZhongpinWang: image

This is what I meant. Also I don't think pointing `AzureContentSafety` to `AzureContentSafetyOutput` will stop user create filter for input? Or maybe I understood your question wrong.

packages/orchestration/src/orchestration-types.ts
ZhongpinWang: Also here, both input and output filter config will technically work as we anyway don't actually return `PromptShield`. 

But this type is used as the return type of the build function. When it is used for building output filter, and we change the type to input, we are saying to the user that output filter config may also get a `PromptShield` property when using our build function (indicated by the type). This feels wrong to me.

packages/orchestration/src/orchestration-types.ts
deekshas8: So for the new `buildAzureContentSafetyFilter`, the input has a custom type anyway, so that we will adjust ourselves later. For the deprecated method, it needs to be `export type AzureContentSafety = AzureContentSafetyInput;` for user to see `PromptShield` in auto completion, right? 

For now its fine if you want to do `export type AzureContentSafety = AzureContentSafetyOutput;`, but it will again be adapted in other ticket when we do the proper handling of PromptShield

ZhongpinWang: Yeah I agree, I wanted to avoid implementing anything for PromptShield already in this PR. I try to make no changes to our client just for now. Supporting PromptShield anyway feels cumbersome to me. If we do this correctly, we might want to have two separate build functions or let's see if there is a better way. Later, I am a bit in favour of deprecating the current one. Or maybe even better, we do this in v2 😄

packages/orchestration/src/orchestration-types.ts
deekshas8: > Also here, both input and output filter config will technically work as we anyway don't actually return PromptShield.

In the current scope yes, but not when we want to fully support `promptShield`.
I think this is what we discussed, the function input and return types will be broad enough to allow the same build function to be used for both (atleast in v1), but we express in docs that promptshield is not meant to be used for output and that it will be ignored (I haven't tested if orchestration service throws an error if we still pass it or ignores it, but its something we can handle on our end as well). Having 2 separate functions for just one property feels over engineering to me.

deekshas8: I think the confusion for me was the scope of this ticket maybe. I agree, lets keep all actual changes for PromptShield for the other ticket.

ZhongpinWang: I just tested. Adding `PromptShield` to output config will cause error. I think in general, orchestration check if payload matches exactly the schema definition. Adding unknown field will fail. 

```jsonc
{
  "request_id": "50dcecd3-99d0-959d-a86c-312304b10bd9",
  "code": 400,
  "message": "400 - Request Body: Additional properties are not allowed ('PromptShield' was unexpected) - 'config'",
  "location": "Request Body"
}
```

I would still argue though if showing a field in autocompletion and then document it somewhere say it is ignored for output is best practice. But I agree having two is also not very elegant for just supporting this one additional field unless input and output filter further diverge. I am okay with both and let's leave this discussion for the other ticket. But thanks for bringing this discussion up! 👍 

---

I remember updating spec does not equal to supporting all features. This is like the first preparation step for all follow-up tickets.

packages/orchestration/src/client/api/schema/response-chat-message.ts
ZhongpinWang: We have a BLI for this. SAP/ai-sdk-js-backlog#306

deekshas8 (approved):
LGTM

PR #829, opened 2025-07-01
[WIP] feat: Enable client-side history maintenance

tomfrenken:
## Context

Closes SAP/ai-sdk-js-backlog#315.

With this PR, both streaming and non-streaming will have the option for client-side maintenance of the chat history.
Thus, users don't have to maintain their own chat array.

(For the future, we should expose a way to upload the history in batch/incrementally to a database, to make the app running the client stateless) 

Also adds a stream lock, so that only a single stream is running per client, which is required by our client to maintain the history client-side. This is a requirement because there is no unique identifier for a conversation, which means the only "marker" for a running conversation is the client itself.


tomfrenken:
Opening a seperate PR based on the ADR, where stream-lock and history are split.

PR #840, opened 2025-07-03
chore: Add post-processing

tomfrenken:
## Context

Closes SAP/ai-sdk-js-backlog#352.

With this PR, we add the ability to add postprocessing after a stream closes.
This makes it easier for our utility functions to execute their logic that depends on chunks, e.g.:
- getToolCalls currently runs the entire mapping logic every single time instead of once after the stream
- getModuleResults (WIP) needs to merge the aggregated chunks at the end
- streamLock (WIP) needs to open the stream for the client again after the stream ends

Any other functionality we might want to add that depends on the closure of the stream also can be placed in this function.

### DIFF
--- a/packages/orchestration/src/orchestration-client.ts
+++ b/packages/orchestration/src/orchestration-client.ts
@@ -141,8 +141,8 @@ export class OrchestrationClient {
     });
 
     // Runs postProcessing of the response after the stream has been fully consumed.
-    // eslint-disable-next-line
     function postProcessing(
+      // eslint-disable-next-line
       response: OrchestrationStreamResponse<OrchestrationStreamChunkResponse>
     ): void {}
 


packages/orchestration/src/orchestration-client.ts
```diff
@@ -140,13 +140,30 @@ export class OrchestrationClient {
       }
     });
 
-    const stream = OrchestrationStream._create(streamResponse, controller);
-    response.stream = stream
+    // Runs postProcessing of the response after the stream has been fully consumed.
+    function postProcessing(
```
deekshas8, line 144: [q] Can you move this function to a separate private method? I imagine you would need to do something like 
`const postProcessWithContext = this.postProcessing.bind(this);` to use `this` inside the async generator

ZhongpinWang (changes requested):
[req] Can we not add another pipe function?

```ts
static async *_processFinal(
  stream: OrchestrationStream<OrchestrationStreamChunkResponse>
): AsyncGenerator<OrchestrationStreamChunkResponse> {
  for await (const chunk of stream) {
    yield chunk;
  }

  // Do some post-processing here if needed.
  // ...
}
```

tomfrenken:
After checking, post processing after the for await loop is the best approach, and I will instead put the post processing of each process function inside each of these, I'll open a follow-up to do this.

PR #845, opened 2025-07-07
feat: Support `.withStructuredOutput` in openai langchain client

deekshas8:
## Context

Closes SAP/ai-sdk-js-backlog#318.

Adds support for `withStructuredOutput` method


sample-code/src/langchain-azure-openai.ts
```diff
@@ -227,3 +227,111 @@ export async function streamChain(
...
+  const finalResponse = await structuredLlm.invoke('Tell me a joke about cats');
+  const parser = new StringOutputParser();
+  return parser.invoke(JSON.stringify(finalResponse));
+}
+
+/**
+ * With Structured Output error handling when refusal is expected.
+ * @returns The answer from GPT with structured output.
+ */
+export async function invokeWithStructuredOutputRefusal(): Promise<string> {
```
ZhongpinWang: [q] Is this function identical to the previous `invokeWithStructuredOutputJsonSchema()`? How do we trigger refusal?

packages/langchain/src/openai/chat.ts
```diff
@@ -105,6 +128,195 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
     } as Partial<AzureOpenAiChatCallOptions>);
   }
 
+  override withStructuredOutput<
```
ZhongpinWang, line 153: [req] There are other signatures for this method https://github.com/langchain-ai/langchainjs/blob/58db055f768546e6270e6b766651fcb770692a24/libs/langchain-openai/src/chat_models.ts#L3108, adding them helps improve type inference for the return value based on `includeRaw` property.

packages/langchain/src/openai/chat.ts
```diff
@@ -105,6 +128,195 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
...
+    const name = config?.name;
+    let method = config?.method;
+    const includeRaw = config?.includeRaw;
+
+    let llm: Runnable<BaseLanguageModelInput>;
+    let outputParser: Runnable<AIMessageChunk, RunOutput>;
+
+    if (config?.strict !== undefined && method === 'jsonMode') {
+      throw new Error(
+        "Argument `strict` is only supported for `method` = 'function_calling'"
```
ZhongpinWang: [pp] This is probably also a typo in official OpenAI client. The string literal indicates it should be `functionCalling` instead of the snake case. 

But I guess it doesn't really matter as from their implementation, everything else than `jsonMode` and `jsonSchema` go into the function calling branch. 😅

packages/langchain/src/openai/chat.ts
```diff
@@ -105,6 +128,195 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
...
+              return aiMessage.additional_kwargs.parsed as RunOutput;
+            }
+            return altParser;
+          }
+        );
+      } else {
+        outputParser = new JsonOutputParser<RunOutput>();
+      }
+    } else {
+      let functionName = name ?? 'extract';
```
ZhongpinWang, line 238: Who in OpenAI came up with this function name `extract` 🙃

sample-code/src/langchain-azure-openai.ts
```diff
@@ -227,3 +227,111 @@ export async function streamChain(
...
+
+  const finalResponse = await structuredLlm.invoke('Tell me a joke about cats');
+  const parser = new StringOutputParser();
+  return parser.invoke(JSON.stringify(finalResponse));
+}
+
+/**
+ * Invoke with structured output using tool calling.
+ * Tool calling is used when input is either a zod schema or an OpenAI Function Definition JSON schema (with name, description and parameters).
+ * Used for models that dont support `response_format` or `json_schema`.
```
ZhongpinWang:
```suggestion
 * Used for models that don't support `response_format` or `json_schema`.
```

sample-code/src/langchain-azure-openai.ts
```diff
@@ -227,3 +227,111 @@ export async function streamChain(
...
+  });
+
+  const finalResponse = await structuredLlm.invoke('Tell me a joke about cats');
+  const parser = new StringOutputParser();
+  return parser.invoke(JSON.stringify(finalResponse));
+}
+
+/**
+ * Invoke with structured output using tool calling.
+ * Tool calling is used when input is either a zod schema or an OpenAI Function Definition JSON schema (with name, description and parameters).
```
ZhongpinWang:
```suggestion
 * Tool calling is used when `mode` is neither set to `jsonMode` nor `jsonSchema`.
```
[req] I don't think the description of JSON mode and JSON schema is correct. In JSON mode, it is only ensured that output is any valid JSON where as JSON schema enforces additionally the schema compatibility. Maybe put a link here to the official OpenAI doc https://platform.openai.com/docs/guides/structured-outputs?api-mode=responses#json-mode or LangChain https://js.langchain.com/docs/concepts/structured_outputs/#json-mode would be helpful for our user.

sample-code/src/langchain-azure-openai.ts
```diff
@@ -227,3 +227,111 @@ export async function streamChain(
...
+ *   parameters: {
+ *     title: "Joke",
+ *     type: "object",
+ *     properties: {
+ *       setup: { type: "string", description: "The setup for the joke" },
+ *       punchline: { type: "string", description: "The joke's punchline" },
+ *     },
+ *     required: ["setup", "punchline"],
+ *   },
+ * }
```
ZhongpinWang, line 272: [pp] I would leave this part out. User should have knowledge about JSON schema (and Zod).

sample-code/src/langchain-azure-openai.ts
```diff
@@ -227,3 +227,111 @@ export async function streamChain(
...
+ * @returns The answer from GPT with structured output using tool calls.
+ */
+export async function invokeWithStructuredOutputToolCalling(): Promise<string> {
+  // initialize client with options
+  const llm = new AzureOpenAiChatClient({
+    modelName: 'gpt-35-turbo'
```
ZhongpinWang: [req] The usage of deprecated model should be avoided. Set `method` to `functionCalling` to force it going with the tool approach.

sample-code/src/server.ts
```diff
@@ -396,7 +397,17 @@ app.get(
 /* LangChain */
 app.get('/langchain/invoke', async (req, res) => {
   try {
-    res.header('Content-Type', 'text/plain').send(await invoke());
+    const response = await invoke();
+    res.send(JSON.stringify(response));
```
ZhongpinWang: [req] Is this change correct? `invoke()` returns just a plain text string due to `StringOutputParser()`.

sample-code/src/server.ts
```diff
@@ -396,7 +397,17 @@ app.get(
...
+  } catch (error: any) {
+    sendError(res, error);
+  }
+});
+
+app.get('/langchain/invoke-with-structured-output', async (req, res) => {
+  try {
+    const response = await invokeWithStructuredOutputJsonSchema();
```
ZhongpinWang, line 408: Whereas here, we will get a plain text JSON string?

sample-code/src/langchain-azure-openai.ts
```diff
@@ -227,3 +227,111 @@ export async function streamChain(
...
+    required: ['setup', 'punchline', 'rating']
+  };
+  const structuredLlm = llm.withStructuredOutput(jokeSchema, {
+    name: 'joke',
+    strict: true
+  });
+
+  const finalResponse = await structuredLlm.invoke('Tell me a joke about cats');
+  const parser = new StringOutputParser();
+  return parser.invoke(JSON.stringify(finalResponse));
```
ZhongpinWang: [req] I don't think it makes any sense to use a `StringOutputParser` for structured output. `finalResponse` is either an object conforms to the defined schema, or a `{ raw: ..., parsed: ...}` object. Just return `finalResponse` or JSON.stringify it would be the same.

sample-code/src/langchain-azure-openai.ts
ZhongpinWang: [q] `jsonMode` does not seem to work? 

If I call

```ts
llm.withStructuredOutput(jokeSchema, {
    name: 'joke',
    method: 'jsonMode',
    includeRaw: true
  });
```

I get

```json
{
  "error": {
    "code": null,
    "message": "'messages' must contain the word 'json' in some form, to use 'response_format' of type 'json_object'.",
    "param": "messages",
    "type": "invalid_request_error"
  }
}
```

Could you please double-check?

packages/langchain/src/openai/chat.ts
ZhongpinWang: We need tests for this method like https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-openai/src/tests/chat_models_structured_output.int.test.ts.

sample-code/src/langchain-azure-openai.ts
deekshas8: No, I was trying out a way to trigger refusal, but couldn't. I'll remove this.

packages/langchain/src/openai/chat.ts
deekshas8: There is an e2e test to check the returned response format with `json_schema` and `strict: true`. Do you want to have a test for every single case?

packages/langchain/src/openai/chat.ts
deekshas8: I will instead change this to `in not supported for jsonMode` since in other cases it works

packages/langchain/src/openai/chat.ts
ZhongpinWang: This function is quite complex. We should cover different cases like function calling, with or without strict:true, default behaviour, behaviour for models like gpt-4 etc. I would say we can write similar integration tests like what we have done for `bindTools()` that we check the actual mock request constructed by LangChain.

packages/langchain/src/openai/chat.ts
ZhongpinWang: Good point 👍

sample-code/src/langchain-azure-openai.ts
deekshas8: > Tool calling is used when `mode` is neither set to `jsonMode` nor `jsonSchema`.

I prefer explaining exactly the positive cases where tool calling is used. Instead of negated version. I will update the text for more clarity

>  don't think the description of JSON mode and JSON schema is correct.

But I'm not even explaining anything about jsonMode/jsonSchema options here 🤔

sample-code/src/server.ts
deekshas8: No, was a mistake. I changed it during testing and forgot to revert it.

sample-code/src/langchain-azure-openai.ts
ZhongpinWang: > I prefer explaining exactly the positive cases where tool calling is used. Instead of negated version. I will update the text for more clarity

True. I was wondering if we should treat tool calling as some kind of fallback method. Because user can potentially provide any string for the `method` property and tool calling will be used. But I guess this is an edge case which anyway should not happen.

> But I'm not even explaining anything about jsonMode/jsonSchema options here 🤔

Oh now I feel I misunderstood the meaning 😄 You were explaining JSON schema. But anyway, feels like we should mention JSON mode as LangChain supports it. Thanks for the clarification.

deekshas8: > Could you please double-check?

So `jsonMode` is even more weird. [Here](https://platform.openai.com/docs/guides/structured-outputs?api-mode=responses#json-mode) it explains that to even trigger json mode, the system prompt should explicitly say json, so thats the reason we get the error you posted here. If I say tell a joke and output json, then its fine. However since I passed a zod schema, it fails later since the output didn't match (also `strict` does. not work with `json_mode`). I now think maybe its not a good idea to even pass a schema, or if users really want to do it, they either use json_schema or handle the error. 
I am thinking of  maybe adding a sample for jsonMode and documenting these edge cases. WDYT?

ZhongpinWang: Thanks for investigating on this! I read through the text 🤯We should provide a separate example or doc section for this and link the OpenAI explanation..

I agree that passing schema to it is also somewhat not meaningful as user should do client-side schema verification afterwards.

ZhongpinWang: Also unfortunately, many some other providers other than OpenAI offers JSON mode. We probably need to support this. Otherwise I would even argue if we should just remove it. I found a pretty nice feature matrix table. https://js.langchain.com/docs/integrations/chat/#featured-providers

deekshas8: Yeah, I will leave in the support, maybe just document it then, its anyway not the default choice.

deekshas8:
```
  const structuredLlm = llm.withStructuredOutput({}, {
    name: 'joke',
    method: 'jsonMode'
  });

  const finalResponse = await structuredLlm.invoke('Tell me a joke about cats and output json');
```

The above works, I'll just document this and not add a sample. Doesnt feel worth the effort.

sample-code/src/langchain-azure-openai.ts
deekshas8: Done

sample-code/src/langchain-azure-openai.ts
deekshas8: Yeah, changed it to json.stringify

packages/langchain/src/openai/chat.ts
deekshas8: I have added unit tests for the different cases.

About the current `bindTools` integration tests — it seems they’re just running the code with mock responses without actually testing anything. Because of that, the four calls (all with the same response) don’t really add much value.

I think writing proper unit tests, similar to the ones for `withStructuredOutput` with real assertions and checks on what the function is actually doing under the hood, would make a lot more sense here.

deekshas8: Will do it in a follow-up PR

.changeset/solid-llamas-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/langchain': minor
+---
+
+[feat] Support `withStructuredOutput` method in OpenAI LangChain client.
```
ZhongpinWang:
```suggestion
[feat] Support `withStructuredOutput()` method in OpenAI LangChain client.
```

packages/langchain/src/openai/chat.test.ts
```diff
@@ -183,6 +187,139 @@ describe('Chat client', () => {
     });
   });
 
+  describe('withStructuredOutput', () => {
+    const joke = z.object({
```
ZhongpinWang: [pp] It would be nice if we put this into or reuse `test-util/tools.ts`.

packages/langchain/src/openai/chat.test.ts
```diff
@@ -183,6 +187,139 @@ describe('Chat client', () => {
...
+        client.withStructuredOutput(joke, {
+          name: 'joke',
+          strict: true,
+          method: 'jsonMode'
+        });
+      }).toThrow(
+        "Argument 'strict' is not supported for 'method' = 'jsonMode'."
+      );
+    });
+    it('should use `functionCalling` if method is specified', async () => {
```
ZhongpinWang:
```suggestion
    it('should use `functionCalling` method if specified', async () => {
```

packages/langchain/src/openai/chat.ts
```diff
@@ -1,14 +1,35 @@
...
+} from '@langchain/core/runnables';
+import {
+  getSchemaDescription,
+  isInteropZodSchema
+} from '@langchain/core/utils/types';
+import {
+  JsonOutputParser,
+  StructuredOutputParser
+} from '@langchain/core/output_parsers';
+import { toJsonSchema } from '@langchain/core/utils/json_schema';
```
ZhongpinWang, line 18: [pp] Using `toJsonSchema` feels better. Would you adjust also the places where zod v4 `z.toJSONSchema()` is used with LangChain? (Also docs).

packages/langchain/src/openai/chat.ts
```diff
@@ -105,6 +128,214 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
...
+      throw new Error(
+        "Argument 'strict' is not supported for 'method' = 'jsonMode'."
+      );
+    }
+
+    // TODO: Remove this once the deprecated models are removed
+    if (
+      !this.modelName.startsWith('gpt-3') &&
+      !this.modelName.startsWith('gpt-4-') &&
+      this.modelName !== 'gpt-4'
```
ZhongpinWang, line 182: Fyi, we can remove them after 2025-09-01.
Currently, `gpt-35-turbo`, `gpt-35-turbo-0125`, `gpt-4` are deprecated but still available until this date.

sample-code/src/langchain-azure-openai.ts
```diff
@@ -228,3 +228,66 @@ export async function streamChain(
...
+ * With Structured Output using `jsonSchema` option with `strict: true`.
+ * @returns The answer from GPT with exactly the structure defined in the schema.
+ */
+export async function invokeWithStructuredOutputJsonSchema(): Promise<string> {
+  // initialize client with options
+  const llm = new AzureOpenAiChatClient({
+    modelName: 'gpt-4o'
+  });
+
+  const joke = z.object({
```
ZhongpinWang, line 242: [pp] Same here, no strong preference. But it would be good if we can move reusable schema into `test-utils/tools.ts`.

packages/langchain/src/openai/chat.ts
deekshas8: Good to know, i'll make a note of this for the next model Update ticket

sample-code/src/langchain-azure-openai.ts
deekshas8: Didnt know there was this file. Will do

packages/langchain/src/openai/chat.ts
deekshas8: I removed it from everywhere except test-utils/tools.ts coz i would have to add the langchain package to the root package.json which feels like an overkill.

PR #852, opened 2025-07-10
feat: Add utility functionality to get the entire stream response

tomfrenken:
## Context

Part of SAP/ai-sdk-js-backlog#315.

With this PR we concatenate and validate the entire stream response, thus making it available like a regular completion response, once the stream ends.

This way, we can make the same utility functionality available for stream responses as we do already for "normal" completion responses.

Additionally, our validation layer is now always applied at the stream's end.

This functionality can then be used for our upcoming message history feature.


packages/orchestration/src/orchestration-stream-response.ts
```diff
@@ -1,39 +1,42 @@
...
-  /**
-   * Finish reasons for all choices.
-   */
-  private _finishReasons: Map<number, string> = new Map();
-  private _toolCallsAccumulators: Map<
-    number,
-    Map<number, ToolCallAccumulator>
-  > = new Map();
+  public _openStream = true;
+  public _data: Partial<CompletionPostResponse> = {};
```
ZhongpinWang, line 16: [q] if it is public, remove `_`?

packages/orchestration/src/orchestration-stream-response.ts
```diff
@@ -1,39 +1,42 @@
...
-  private _usage: TokenUsage | undefined;
-  /**
-   * Finish reasons for all choices.
-   */
-  private _finishReasons: Map<number, string> = new Map();
-  private _toolCallsAccumulators: Map<
-    number,
-    Map<number, ToolCallAccumulator>
-  > = new Map();
+  public _openStream = true;
```
ZhongpinWang, line 15: [req] Can we make this private and (if needed) only provide a getter for this?

packages/orchestration/src/orchestration-stream-response.ts
```diff
@@ -1,39 +1,42 @@
...
   TokenUsage
 } from './client/api/schema/index.js';
 import type { OrchestrationStream } from './orchestration-stream.js';
 
+const logger = createLogger({
+  package: 'orchestration',
+  messageContext: 'orchestration-stream-response'
+});
+
+function openStreamWarning(missingData: string): void {
```
ZhongpinWang:
```suggestion
function logOpenStreamWarning(missingData: string): void {
```

packages/orchestration/src/orchestration-stream-response.ts
```diff
@@ -1,39 +1,42 @@
...
-
-  /**
-   * @internal
-   */
-  _setTokenUsage(usage: TokenUsage): void {
-    this._usage = usage;
+    if (!this._openStream) {
+      return this._data.orchestration_result?.usage;
+    }
+    openStreamWarning('token usage');
```
ZhongpinWang: [pp] This `if` check is a precondition for all getters. So instead of adding a `if` branch and log everywhere, I would create a new function called `isStreamOpen()` and put logger in it. Then `if(isStreamOpen) return;`.

I feel we don't have to explicitly mention what information is missing as accessing before streaming ends is already a design time issue and should be avoided. It could be interesting though to indicate where it was logged by printing the stack trace (maybe?). See https://stackoverflow.com/a/57238353/18544589

packages/orchestration/src/orchestration-stream-response.ts
```diff
@@ -42,44 +45,87 @@ export class OrchestrationStreamResponse<T> {
...
+   * Parses the orchestration response and returns the tool calls generated by the model.
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns The message tool calls.
    */
-  _setFinishReasons(finishReasons: Map<number, string>): void {
-    this._finishReasons = finishReasons;
+  public getToolCalls(choiceIndex = 0): MessageToolCalls | undefined {
+    if (!this._openStream) {
+      const choice = this.findChoiceByIndex(choiceIndex);
+      return choice?.message?.tool_calls;
```
ZhongpinWang: [pp] Can we inline this? `this.findChoiceByIndex(choiceIndex)?.message?.tool_calls`

packages/orchestration/src/orchestration-stream.ts
```diff
@@ -53,122 +40,34 @@ export class OrchestrationStream<Item> extends SseStream<Item> {
...
-    }
-    for await (const chunk of stream) {
-      const usage = chunk.getTokenUsage();
-      if (usage) {
-        response._setTokenUsage(usage);
-        logger.debug(`Token usage: ${JSON.stringify(usage)}`);
-      }
-      yield chunk;
-    }
+    response._openStream = false;
```
ZhongpinWang, line 69: [req] I really feel we should put a try catch in `stream()` method or similar to also set this to false in case of any error thrown (could be by Axios or anything else). Otherwise in case of any error during parsing, this line is never reached.

packages/orchestration/src/util/stream.ts
```diff
@@ -0,0 +1,345 @@
...
+        );
+        break;
+      default:
+        mergedModuleResults[moduleName] = moduleResult;
+    }
+  }
+  return mergedModuleResults;
+}
+
+function mergeLlmModule(
```
ZhongpinWang, line 69:
```suggestion
function mergeOrchestrationResult(
```
[pp] I think name it after the property name makes it more clear.

packages/orchestration/src/orchestration-stream-response.ts
tomfrenken: it is public, but shouldn't be accessed by our users, similar to our "public " methods

packages/orchestration/src/util/stream.ts
tomfrenken: It is the llm module AND the orchestration result, they share the same type, hence the name

packages/orchestration/src/orchestration-stream-response.ts
tomfrenken: I prefer to split the logic a bit

packages/orchestration/src/orchestration-stream-response.ts
tomfrenken: Unfortunately there is no benefit in this case

packages/orchestration/src/util/stream.ts
```diff
@@ -0,0 +1,351 @@
...
+): ChoiceLogprobs | undefined {
+  if (!incoming) {
+    return existing;
+  }
+  if (!existing) {
+    return incoming;
+  }
+  return {
+    content: [...(existing.content ?? []), ...(incoming.content ?? [])],
+    refusal: [...(existing.refusal ?? []), ...(incoming.refusal ?? [])]
```
ZhongpinWang, line 190: [q] I am not sure how important this could be, but `ChoiceLogprobs` has `Record<string, any>` in it and we are not merging it.

tomfrenken: While true, merging "anything" is kinda impossible, we could at best do something like `...incoming` to at least set whatever is incoming, thoughts?

packages/orchestration/src/orchestration-client.ts
```diff
@@ -131,23 +141,38 @@ export class OrchestrationClient {
...
 
-    const stream = OrchestrationStream._create(streamResponse, controller);
-    response.stream = stream
-      ._pipe(OrchestrationStream._processChunk)
-      ._pipe(OrchestrationStream._processToolCalls, response)
-      ._pipe(OrchestrationStream._processFinishReason, response)
-      ._pipe(OrchestrationStream._processTokenUsage, response);
+      if (!controller.signal.aborted) {
+        controller.abort();
+      }
```
ZhongpinWang: [pp] I don't think it is necessary to check the `aborted` property. Abort again won't do anything or trigger event listener again if it is already aborted.

packages/orchestration/src/orchestration-stream-response.ts
```diff
@@ -89,6 +128,25 @@ export class OrchestrationStreamResponse<T> {
...
+  }
+
+  private findChoiceByIndex(index: number) {
+    return this.getChoices().find((c: { index: number }) => c.index === index);
+  }
+
+  private isStreamOpen(): boolean {
+    if (this._openStream) {
+      const stacktrace = new Error().stack;
+      logger.warn(
```
ZhongpinWang: [pp] I feel this is already an error since this is a design issue already which will most likely keep happening. And the code probably will never work.

packages/orchestration/src/orchestration-stream-response.ts
```diff
@@ -1,39 +1,36 @@
...
-  private _usage: TokenUsage | undefined;
-  /**
-   * Finish reasons for all choices.
-   */
-  private _finishReasons: Map<number, string> = new Map();
-  private _toolCallsAccumulators: Map<
-    number,
-    Map<number, ToolCallAccumulator>
-  > = new Map();
+  public _openStream = true;
```
ZhongpinWang, line 15: ~[req]/[q] I am really confused. I thought this is the stream lock and should be set on the client level not per response. Currently user can still call stream() if a stream is not finished?~

~And is the try catch block designed for stream lock? I thought we need that try catch because we want to set client level stream lock to false even if there is any underlying error. If not, I see no point of the try catch block by catching, logging and throwing again?~ 

Discussed. Stream lock not relevant to this PR.

ZhongpinWang (commented):
Perhaps I am lost with what we want to achieve with this PR. I thought we do stream lock. If `_openStream` is not meant for stream lock and this will happen in the other PR, I would remove try catch and add it when we implement stream lock. The stream response level check for `_openStream` when calling utilities is indeed helpful.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -131,23 +141,38 @@ export class OrchestrationClient {
...
 
-    const streamResponse = await this.executeRequest({
-      ...options,
-      requestConfig: {
-        ...options.requestConfig,
-        responseType: 'stream',
-        signal: controller.signal
-      }
-    });
+    try {
```
ZhongpinWang: [q] I don't think this try-catch is necessary as the `stream()` function one level above already has a try-catch?

packages/orchestration/src/util/stream.ts
```diff
@@ -0,0 +1,351 @@
...
+        logger.warn('LlmChoice is missing message information.');
+      } else {
+        validateMessage(choice.message);
+      }
+      if (!choice.finish_reason) {
+        logger.warn('LlmChoice is missing a finish reason.');
+      }
+      if (!choice.index && choice.index !== 0) {
+        logger.warn('LlmChoice must have a valid index.');
+      }
```
KavithaSiva: [pp] Also log, which `LLmChoice` we are logging, from code, it could be either from `llm` of `module_results`, `orchestration_result` or `output_unmasking`.

packages/orchestration/src/orchestration-stream-response.ts
tomfrenken: Yeah that's a fair assessment, I'll make it an error instead.

packages/orchestration/src/orchestration-client.ts
tomfrenken: I think the outer-loop is not catching errors that occur during the iteration of the for-await loop, I suppose I should add a test case for this.

packages/orchestration/src/util/stream.ts
```diff
@@ -0,0 +1,351 @@
...
+  }
+
+  try {
+    JSON.parse(toolCall.function?.arguments ?? '');
+  } catch {
+    logger.warn(
+      'ToolCall arguments are not valid JSON for tool: ' +
+        toolCall.function?.name ||
+        toolCall.id ||
+        'unknown'
```
KavithaSiva: [pp] Logs warning also for cases where a function has no arguments, only attempt parsing if arguments exist.

packages/orchestration/src/util/stream.ts
```diff
@@ -0,0 +1,351 @@
...
+
+  if (message.tool_calls) {
+    for (const toolCall of message.tool_calls) {
+      validateToolCall(toolCall);
+    }
+  }
+}
+
+function validateToolCall(toolCall: Partial<MessageToolCall>): void {
+  if (typeof toolCall.id !== 'string') {
```
KavithaSiva: [pp] Also check if the `id` is not empty.

packages/orchestration/src/util/stream.ts
```diff
@@ -0,0 +1,351 @@
...
+  }
+
+  validateLlmModuleResult(response._data.module_results?.llm);
+
+  validateLlmModuleResult(response._data.orchestration_result);
+
+  validateChoices(response._data.module_results?.output_unmasking);
+}
+
+function validateLlmModuleResult(
```
KavithaSiva: [pp] Not sure if this is always required for everything, but consider adding the source `module_results` or `orchestration_result` to logs to know the exact property being logged out.

packages/orchestration/src/util/stream.ts
```diff
@@ -0,0 +1,351 @@
...
+  incoming: ChatDelta | undefined
+): ResponseChatMessage {
+  if (!incoming) {
+    return existing;
+  }
+  return {
+    role: existing.role,
+    content: existing.content + (incoming.content ?? ''),
+    tool_calls: mergeToolCalls(existing.tool_calls, incoming.tool_calls),
+    refusal: incoming.refusal ?? existing.refusal
```
KavithaSiva, line 144: [q] Have you seen refusal in the orchestration service response? Is it easy to trigger with the the `input_filtering` ?

KavithaSiva (commented):
Thanks a lot for unifying the functionality for streaming and non-streaming variants.
Added few comments, review is still ongoing.

packages/orchestration/src/util/stream.ts
tomfrenken: Even if there is no arguments, it will be an empty object, example chunk:
```json
{
  "request_id": "18b6595f-28a0-4898-a76c-2ae8fa6b5d9f",
  "module_results": {
    "llm": {
      "id": "chatcmpl-BvjrapfelXZ3FgWXfbJLUGFI7ieXF",
      "object": "chat.completion.chunk",
      "created": 1753100694,
      "model": "gpt-4o-2024-08-06",
      "system_fingerprint": "fp_ee1d74bde0",
      "choices": [
        {
          "index": 0,
          "delta": {
            "role": "assistant",
            "content": "",
            "tool_calls": [
              {
                "index": 0,
                "function": {
                  "arguments": "{}"
                }
              }
            ]
          }
        }
      ]
    }
  },
  "orchestration_result": {
    "id": "chatcmpl-BvjrapfelXZ3FgWXfbJLUGFI7ieXF",
    "object": "chat.completion.chunk",
    "created": 1753100694,
    "model": "gpt-4o-2024-08-06",
    "system_fingerprint": "fp_ee1d74bde0",
    "choices": [
      {
        "index": 0,
        "delta": {
          "role": "assistant",
          "content": "",
          "tool_calls": [
            {
              "index": 0,
              "function": {
                "arguments": "{}"
              }
            }
          ]
        }
      }
    ]
  }
}

File: packages/orchestration/src/util/stream.ts (hunk 15 above)
Comment (tomfrenken) [reply packages/orchestration/src/util/stream.ts]: AFAIK the id is never an empty string, the first chunk of a tool_call will always contain the id with the complete string, would you still do this additional check?

File: packages/orchestration/src/util/stream.ts (hunk 16 above)
Comment (tomfrenken) [reply packages/orchestration/src/util/stream.ts]: Good idea

File: packages/orchestration/src/util/stream.ts (hunk 17 above)
Comment (tomfrenken) [reply packages/orchestration/src/util/stream.ts:144]: No I actually never managed to get it to refused (but also didn't test too much)

File: packages/orchestration/src/orchestration-client.ts (hunk 12 above)
Comment (ZhongpinWang) [reply packages/orchestration/src/orchestration-client.ts]: It would be a good learning if the outside try won't be able to catch everything. But yes, a test would help here.

File: packages/orchestration/src/orchestration-client.test.ts [hunk 18]
```diff
@@ -850,9 +851,86 @@ describe('orchestration service client', () => {
...
+
+      const controller = new AbortController();
+
+      // Abort immediately
+      controller.abort();
+
+      const client = new OrchestrationClient(config);
+
+      await expect(client.stream(undefined, controller)).rejects.toThrow();
+      expect(controller.signal.aborted).toBe(true);
```
ZhongpinWang: [pp] This is not necessary.

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -850,9 +851,86 @@ describe('orchestration service client', () => {
...
+      expect(errorSpy).toHaveBeenCalledWith(
+        'Error while creating the stream response:',
+        expect.any(Error)
+      );
+    });
+
+    it('should handle aborted requests gracefully', async () => {
```
ZhongpinWang: [q] I don't understand what you mean by `gracefully`? Don't we throw error? It doesn't feel graceful? I am a bit confused with the purpose of this test. Which part of the code would you like to test with this?

packages/orchestration/src/orchestration-client.ts
```diff
@@ -73,21 +73,27 @@ export class OrchestrationClient {
...
+        {
+          prompt,
+          requestConfig,
+          stream: true,
+          streamOptions: options
+        },
+        controller
       );
+    } catch (error) {
+      logger.error('Error while creating the stream response:', error);
```
ZhongpinWang: [pp] Feels unnecessary to log the error again as we throw it anyway later.

packages/orchestration/src/util/stream.ts
```diff
@@ -0,0 +1,381 @@
...
+  return {
+    content: [...(existing.content ?? []), ...(incoming.content ?? [])],
+    refusal: [...(existing.refusal ?? []), ...(incoming.refusal ?? [])]
+  };
+}
+
+function handleFinishReason(
+  existing: string | undefined,
+  incoming: string | undefined,
+  choiceIndex: number | undefined
```
ZhongpinWang: [pp] `choiceIndex` can never be undefined.

packages/orchestration/src/util/stream.ts
```diff
@@ -0,0 +1,381 @@
...
+ * @internal
+ */
+export function validateResponse(
+  response: OrchestrationStreamResponse<OrchestrationStreamChunkResponse>
+): void {
+  if (response._openStream) {
+    throw new Error(
+      "Stream wasn't closed properly. Please ensure the stream is closed after processing."
+    );
+  }
```
ZhongpinWang: [pp] I would remove this as the only time `validateResponse` is used is one line after setting `response._openStream` to `false`. I cannot think of a case that that line was not executed and validate was somehow called.

packages/orchestration/src/util/stream.ts
```diff
@@ -0,0 +1,381 @@
...
+        );
+      } else {
+        validateMessage(choice.message, sourceModule, choice.index);
+      }
+      if (!choice.finish_reason) {
+        logger.warn(
+          `${sourceModule}: LlmChoice ${choice.index} is missing a finish reason.`
+        );
+      }
+      if (!choice.index && choice.index !== 0) {
```
ZhongpinWang: [pp] Probably you want to test if `choice.index` is undefined?

packages/orchestration/src/util/stream.ts
```diff
@@ -0,0 +1,381 @@
...
+    }
+  }
+}
+
+function validateMessage(
+  message: Partial<ResponseChatMessage>,
+  sourceModule: string,
+  sourceChoice: number | undefined
+): void {
+  if (!message.role) {
```
ZhongpinWang: [pp] I don't mind having validations. But some of them are testing if a mandatory field is really there. This feels unnecessary since if we implement everything correctly, user should never see these warnings, at least not without seeing some error first (like stream ended unexpectedly). Other places like `choice.index`, `choice.message` are IMO also unnecessary. Can you walk through those cases and remove them if you also feel the same?

packages/orchestration/src/orchestration-client.test.ts
KavithaSiva: Have removed the abort signal assertion.

packages/orchestration/src/orchestration-client.test.ts
KavithaSiva: Renamed the test.

packages/orchestration/src/util/stream.ts
KavithaSiva: Removed the check as index is a mandatory field in choices.

packages/orchestration/src/util/stream.ts
KavithaSiva: I have removed some mandatory checks which feel redundant.

PR #860, opened 2025-07-14
fix: E2E document grounding get pipeline status

ZhongpinWang:
## Context

The service behaviour was changed. Now a pipeline will not stay in `INPROGRESS` status once created. Instead, it stays in `FINISHED` status.


tests/e2e-tests/src/document-grounding.test.ts
```diff
@@ -57,6 +57,6 @@ describe('document grounding', () => {
     const pipelineStatus = await getPipelineStatus(
       '7b17e2ab-4ecc-448e-837f-2c10c9359925'
     );
-    expect(pipelineStatus.status).toStrictEqual('INPROGRESS');
+    expect(pipelineStatus.status).toStrictEqual('FINISHED');
```
KavithaSiva, line 60: [q] Have they documented this behaviour change somewhere?

KavithaSiva (approved):
LGTM,  thanks for taking care!

tests/e2e-tests/src/document-grounding.test.ts
ZhongpinWang: I tried my best but found no docs for this... Also not from the real rage service. But since we are only testing if server responds with any status, I accepted this.

KavithaSiva: WDY mean by not from the real rage service here?

PR #865, opened 2025-07-16
chore: Use correct type cast in `findChoiceByIndex()`

shibeshduw:
## Context

Closes SAP/ai-sdk-js-backlog#306.

The bug listed on the ticket was resolved by Orchestration in 2506b update. 
Since `role` is now a required field in the response message, `getChoices()` can be safely typed as `LLMChoice[]`.

### DIFF
--- a/packages/orchestration/src/orchestration-response.ts
+++ b/packages/orchestration/src/orchestration-response.ts
@@ -5,7 +5,8 @@ import type {
   ChatMessage,
   ChatMessages,
   AssistantChatMessage,
-  MessageToolCalls
+  MessageToolCalls,
+  LlmChoice
 } from './client/api/schema/index.js';
 
 /**
@@ -100,10 +101,7 @@ export class OrchestrationResponse {
   }
 
   private findChoiceByIndex(index: number) {
-    // TODO: replace cast with LLMChoice[] after the bug in orchestration, where
-    // 'role' in ResponseChatMessage is optional when it should be mandatory, is fixed.
-    // https://github.com/SAP/ai-sdk-js-backlog/issues/306
-    return (this.getChoices() as any).find(
+    return (this.getChoices() as LlmChoice[]).find(
       (c: { index: number }) => c.index === index
     );
   }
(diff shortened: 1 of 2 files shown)


packages/orchestration/src/orchestration-response.ts
```diff
@@ -100,10 +101,7 @@ export class OrchestrationResponse {
   }
 
   private findChoiceByIndex(index: number) {
-    // TODO: replace cast with LLMChoice[] after the bug in orchestration, where
-    // 'role' in ResponseChatMessage is optional when it should be mandatory, is fixed.
-    // https://github.com/SAP/ai-sdk-js-backlog/issues/306
-    return (this.getChoices() as any).find(
+    return (this.getChoices() as LlmChoice[]).find(
```
deekshas8: [req] you dont need the assertion here. `getChoices()` returns the correct type. Add the return type to that function as a best practice maybe

shibeshduw: Yep. Completely went over my head. 🤦
Thanks!

packages/orchestration/src/orchestration-response.ts
```diff
@@ -95,16 +96,11 @@ export class OrchestrationResponse {
...
   }
 
-  private findChoiceByIndex(index: number) {
-    // TODO: replace cast with LLMChoice[] after the bug in orchestration, where
-    // 'role' in ResponseChatMessage is optional when it should be mandatory, is fixed.
-    // https://github.com/SAP/ai-sdk-js-backlog/issues/306
-    return (this.getChoices() as any).find(
-      (c: { index: number }) => c.index === index
-    );
+  private findChoiceByIndex(index: number): LlmChoice | undefined {
```
ZhongpinWang: [req] As we export this function in orchestration streaming, and foundation models, here we should also expose it publicly. The reason why we hide this before was because it was not really usable.

ZhongpinWang (changes requested):

PR #871, opened 2025-07-22
chore: Update model list

emincihangeri:
## Context

Closes SAP/ai-sdk-js-backlog#360.

The deprecated and retired models are removed from the available model list. New models are added such as `amazon--nova-premier`. The full correspondence with SAP notes is considered and applied.


packages/core/src/model-types.ts
```diff
@@ -27,10 +26,10 @@ export type AzureOpenAiEmbeddingModel = LiteralUnion<
  * GCP Vertex AI models for chat completion.
  */
 export type GcpVertexAiChatModel = LiteralUnion<
-  | 'gemini-1.5-pro'
-  | 'gemini-1.5-flash'
```
KavithaSiva: The retirement date of these models are set to 24th September. Why are we removing them sooner? It is not set to deprecated too.

But, I see that they are also in recently retired models list. @ZhongpinWang Have you seen something like this before?

ZhongpinWang: The retired models have version `001`. Version `002` of both models are not retired or deprecated. @emincihangeri Both of them should not be removed.

packages/core/src/model-types.ts
```diff
@@ -42,33 +41,24 @@ export type AwsBedrockChatModel = LiteralUnion<
   | 'anthropic--claude-3-sonnet'
   | 'anthropic--claude-3.5-sonnet'
   | 'anthropic--claude-3.7-sonnet'
+  | 'anthropic--claude-4-opus'
+  | 'anthropic--claude-4-sonnet'
   | 'amazon--titan-text-express'
   | 'amazon--titan-text-lite'
```
ZhongpinWang: [req] Both are deprecated. Please remove.

packages/core/src/model-types.ts
```diff
@@ -42,33 +41,24 @@ export type AwsBedrockChatModel = LiteralUnion<
...
-export type AwsBedrockEmbeddingModel = LiteralUnion<'amazon--titan-embed-text'>;
-
 /**
  * AI Core open source models for chat completion.
  */
 export type AiCoreOpenSourceChatModel = LiteralUnion<
-  | 'mistralai--mixtral-8x7b-instruct-v01'
   | 'mistralai--mistral-large-instruct'
   | 'mistralai--mistral-small-instruct'
   | 'meta--llama3.1-70b-instruct'
```
ZhongpinWang: [req] This model is deprecated. Please remove.

packages/core/src/model-types.ts
KavithaSiva: [req] `amazon--titan-text-express` and `amazon--titan-text-lite` are also marked as deprecated. Consider removing them as well.

ZhongpinWang (changes requested):
See the comments

packages/core/src/model-types.ts
KavithaSiva: [pp] `meta--llama3.1-70b-instruct` is also marked as deprecated but will only be retired by 30th September, so we can wait a bit longer here.

KavithaSiva (changes requested):
Also add `gemini-1.5-flash` and `gemini-1.5-pro` back in.

KavithaSiva:
> I am applying all the corrections and leave `meta--llama3.1-70b-instruct` for now on the list, right?

Also remember to change your base branch to `v1-main`

packages/core/src/model-types.ts
ZhongpinWang: No, we remove models in our SDK as soon as they are deprecated and announce the changes. This gives user some time to switch to newer model before the retirement. Otherwise, user may not check SAP notes and trust our SDK as we never announced anything. Then immediately at the retirement, their production code fails.

emincihangeri: I couldn't see a replacement for the deprecated model though. Is it okay to remove it or put it on the deprecated models list(in the documentation) without a replacement?

ZhongpinWang: In the SAP Notes, it says `mistralai--mistral-small-instruct` is the replacement in the "Recently Retired Models" table.

ZhongpinWang: Yes, for `meta--llama3.1-70b-instruct` if there is no replacement, we remove but also not suggest anything.

Issue #870, opened 2025-07-22
OrchestrationClient.bindTools() returns a non-functional RunnableBinding with no methods, breaking LangChain tool calling.


schardosin:
### Describe the Bug

It seems we have a bug in the @sap-ai-sdk/langchain package where the OrchestrationClient.bindTools() method returns a RunnableBinding object that is effectively empty and unusable.

When bindTools() is called on an OrchestrationClient instance, it successfully returns a RunnableBinding object without throwing an error. However, inspection of this returned object reveals that it has no methods of its own. The Object.getOwnPropertyNames() check on the returned runnable results in an empty array [].

This breaks the fundamental contract of the LangChain Expression Language (LCEL). While the object's prototype chain might still contain methods like invoke or stream, the core implementation logic, particularly the essential _generate method, is lost.

As a result, any attempt to use the bound runnable (e.g., calling .invoke()) fails. The call either hangs indefinitely or throws an error, because the underlying implementation required to make an API call with tools is missing. This makes the entire tool-calling feature incompatible with standard LangChain agent frameworks like createToolCallingAgent.

### Steps to Reproduce

Below is a code (generated with AI) to demonstrate the issue. Save to a bug.js file and run `node bug.js`

```
const { OrchestrationClient } = require('@sap-ai-sdk/langchain');
const { DynamicStructuredTool } = require('langchain/tools');
const { z } = require('zod');

// ============================================================================
// CONFIGURATION - UPDATE THESE VALUES WITH YOUR SAP AI CORE SETTINGS
// ============================================================================

const SAP_AI_CORE_CONFIG = {
	// Your SAP AI Core service URL (usually ends with /v2)
	serviceUrl: 'https://api.ai.prod.eu-central-1.aws.ml.hana.ondemand.com/v2',

	// OAuth credentials for authentication
	clientId: '<client_id>',
	clientSecret: '<client_secret>',

	// Your resource group
	resourceGroup: 'default',

	// Model to test with (any model you have deployed)
	modelName: 'gpt-4o-mini',

	// OAuth token URL (usually same domain as serviceUrl but with /oauth/token)
	tokenUrl: 'https://<subaccount>.authentication.sap.hana.ondemand.com/oauth/token',
};

// ============================================================================
// OAUTH AUTHENTICATION HELPER
// ============================================================================

async function getAccessToken() {
	const https = require('https');
	const querystring = require('querystring');

	const postData = querystring.stringify({
		grant_type: 'client_credentials',
[...]

PR #879, opened 2025-07-23
fix: Zod schema validation for both v3 and v4

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#371.
User issue: SAP/ai-sdk-js#870

Use `isInteropZodSchema` instead of `isZodSchemaV4` to support both Zod v3 and v4 schema input from user.


deekshas8 (approved):
LGTM

PR #882, opened 2025-07-24
fix: E2E with structured output

ZhongpinWang:
## Context

E2E was failing https://github.com/SAP/ai-sdk-js/actions/runs/16482805208

Returning object instead of JSON string to make it work.


deekshas8 (approved):
LGTM

PR #890, opened 2025-07-28
chore: Setup pipelines for `v1-main`

KavithaSiva:
## Context

Adjusting all workflows for v1 builds and releases.


.github/workflows/spec-update.yaml
```diff
@@ -16,7 +16,7 @@ on:
       file-ref:
         description: "Branch or tag to checkout the spec file from"
         required: false
-        default: main
+        default: v1-main
```
deekshas8: [q] This workflow should ideally run on both, or we decide to cherry pick changes over to main. Cherry pick might work for other purely generated packages, but not for orchestration.

.github/workflows/publish-canary.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: sap/ai-sdk-js/.github/actions/setup@main
+      - uses: sap/ai-sdk-js/.github/actions/setup@v1-main
```
deekshas8, line 19: [q] I imagine we will at some point release `next` for v2 for e2e testing. Will that be covered in a separate ticket?

.github/workflows/spec-update.yaml
ZhongpinWang: [req] `file-ref` is the branch name of the other service, where spec is defined, not our repo. It is default to `main`.

.github/workflows/spec-update.yaml
```diff
@@ -32,7 +32,7 @@ on:
...
```
ZhongpinWang: same

.github/workflows/spec-update.yaml
```diff
@@ -69,12 +69,12 @@ jobs:
         run: |
           git config --global user.email "cloudsdk@sap.com"
           git config --global user.name "SAP Cloud SDK Bot"
-          git fetch --unshallow origin main
+          git fetch --unshallow origin v1-main
```
ZhongpinWang: [req] We should only do spec update on main. Orchestration will also use this workflow to use the generator. And they should use our main branch.

.github/workflows/spec-update.yaml
```diff
@@ -69,12 +69,12 @@ jobs:
...
           
           BRANCH="spec-update/$CHOICE/$REF"
           git fetch origin $BRANCH || true
           git checkout -B $BRANCH origin/$BRANCH || git checkout -b $BRANCH
-          git rebase main
+          git rebase v1-main
```
ZhongpinWang: same

.github/workflows/spec-update.yaml
```diff
@@ -166,7 +166,7 @@ jobs:
             exit 0
           fi
           
-          PR_URL=$(gh pr create --base main --head $BRANCH --title "feat: Update $CHOICE specification" --body "
+          PR_URL=$(gh pr create --base v1-main --head $BRANCH --title "feat: Update $CHOICE specification" --body "
```
ZhongpinWang: same

.github/workflows/spec-update.yaml
```diff
@@ -191,7 +191,7 @@ jobs:
           BRANCH: ${{ steps.branch.outputs.branch }}
           PR_URL: ${{ steps.create-pr.outputs.pr_url }}
         run: |
-          DIFF_URL="https://github.com/SAP/ai-sdk-js/compare/main...$BRANCH"
+          DIFF_URL="https://github.com/SAP/ai-sdk-js/compare/v1-main...$BRANCH"
```
ZhongpinWang: same

.github/workflows/spec-update.yaml
ZhongpinWang, line 1: In general, this workflow from the `v1-main` branch should not be used anymore. 

I am just not sure if we remove it on the default branch, will we still see the `spec-update` workflow from `main` branch in the `Actions` tab. Can you double check? If we cannot see it anymore, maybe we don't remove it. But we should always use v2 `spec-update.yaml`

.github/workflows/spec-update-cron.yaml
```diff
@@ -11,17 +11,17 @@ on:
...
     secrets: inherit
   spec-update-document-grounding:
-    uses: SAP/ai-sdk-js/.github/workflows/spec-update.yaml@main
+    uses: SAP/ai-sdk-js/.github/workflows/spec-update.yaml@v1-main
     with:
       file: document-grounding
     secrets: inherit
   spec-update-prompt-registry:
-    uses: SAP/ai-sdk-js/.github/workflows/spec-update.yaml@main
+    uses: SAP/ai-sdk-js/.github/workflows/spec-update.yaml@v1-main
```
ZhongpinWang, line 24: (Read my other comments below first)

For this cron job specifically, since we should only run spec update for v2, and orchestration use `main` to run the generator, I would not use `v1-main` here but `main` directly, and only update that one in main.

`spec-update.yaml` on v1 should not be used.

ZhongpinWang (changes requested):
The spec update workflow is tricky. Need some changes.

.github/workflows/spec-update.yaml
KavithaSiva: Oh, okay, wasn't sure about this one, thanks for clarifying.

.github/workflows/spec-update.yaml
KavithaSiva: Added `ref` to the workflow, so that `main` branch is shown by default. If I remove the workflow from `v1-main`, it is no longer shown in the actions tab.

KavithaSiva: But, I saw two spec files maintained in the [llm spec folder](https://github.tools.sap/AI/llm-orchestration/tree/main/src/spec). Do you know if the development on `v1` is completely stopped or if we have to support it for sometime until `v2` ?

KavithaSiva: @ChristophMatthies Is development on `v1` spec completely stopped?

.github/workflows/spec-update.yaml
KavithaSiva: Clarity is still required here on whether the spec update has to be done for both v1 and v2 versions of orchestration.

.github/workflows/publish-canary.yml
KavithaSiva: Working on those changes in https://github.com/SAP/ai-sdk-js/pull/891 .
Is it right that when V2 is in development we will only have a job similar to this one to publish next and won't have the publish script until we choose to release v2 ?

.github/workflows/spec-update-cron.yaml
KavithaSiva: I would still let this be as I changed the workflows in this branch.

.github/workflows/spec-update.yaml
KavithaSiva: Discussed with Zhongpin, generations based on orchestration spec is only meant for `v2` versions now.

.github/workflows/spec-update.yaml
KavithaSiva: Clarified with Christoph, they are not planning to add anything in `v1`.

PR #891, opened 2025-07-28
chore: V2 pipeline setup

KavithaSiva:
## Context

Adjusting all workflows for v2 builds and releases.


.github/workflows/auto-fix-lint.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - main
+      - v1-main
```
deekshas8: [req] Keep this PR only for v2 main setup

.github/workflows/bump.yml
```diff
@@ -1,135 +1,136 @@
-name: bump
+# Disable workflow until releasing v2
```
deekshas8: [req] There is no need to disable it in this PR.

.github/workflows/codeql.yml
```diff
@@ -15,7 +15,7 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
-    branches: [ "main" ]
+    branches: [ "main", "v1-main" ]
```
deekshas8: [req] Remove references to v1-main

.github/workflows/documentation.yaml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - main
+      - v1-main
```
deekshas8: here

.github/workflows/publish-canary.yml
```diff
@@ -4,16 +4,16 @@ on:
   schedule:
     - cron: '0 1 * * *'
   workflow_dispatch:
-    inputs:
```
deekshas8: [q] Why?

.github/workflows/publish-canary.yml
```diff
@@ -4,16 +4,16 @@ on:
...
-      tag:
-        description: 'Dist tag for the release. If you chose something different than "canary", make sure to delete it once it is not needed anymore.'
-        type: string
-        required: false
-        default: 'canary'
 
 jobs:
+  e2e-tests:
```
deekshas8: [q] Isn't the general practice to first run e2e tests anyway before triggering a release?

KavithaSiva: I just thought it will be easier, if people forget to run the E2E tests from `main` branch.
Reverted it now

.github/workflows/publish-canary.yml
KavithaSiva: Felt we will only run the workflow for `next`. Reverted it now.

.github/workflows/bump.yml
KavithaSiva: The script can be still be run from `main`, so isn't it easier if we just comment it ?

KavithaSiva: Will do it in a separate PR.

.github/workflows/spec-update.yaml
```diff
@@ -1,10 +1,10 @@
-name: "Spec File Update Workflow"
+name: 'Spec File Update Workflow'
```
KavithaSiva, line 1: I will also comment this in a separate PR, as we don't want this to be triggered when we are working on the v2 changes in a separate package.

.github/workflows/publish-canary.yml
```diff
@@ -9,8 +9,8 @@ on:
         description: 'Dist tag for the release. If you chose something different than "canary", make sure to delete it once it is not needed anymore.'
         type: string
         required: false
-        default: 'canary'
-
+        default: 'next'
+
```
deekshas8: [req] Just setting tag next is not enough for a next release. The released version has to be 2.0.0. With the current setup 
```
cat <<EOT >> .changeset/canary-release-changeset.md
          ---
          '@sap-ai-sdk/core': patch
          ---
          Canary release
```
we'll end up publishing 1.0.0 as next which is misleading.

While we are in v2 development, we need to adjust is to use `major` instead of `patch`. **Once the release is done it has to be reverted to use `patch`**

deekshas8: I will also suggest to first run these commands (yarn changeset, etc) locally before to see how the `package.json` of all packages are updated. Only then run the `next` release workflow here.

deekshas8: > Just setting tag next is not enough for a next release. 

You can do this in the follow up PR. Since the `major` changeset would need to be reverted before release as well

PR #895, opened 2025-07-30
chore: Add weather MCP in agent tutorial

deekshas8:
## Context

Most weather mcp servers here https://github.com/modelcontextprotocol/servers are based on Accuweather/OpenWeathermap which require an API key with payment option. `Open-meteo` is a nice free API but the [mcp server](https://github.com/isdaniel/mcp_weather_server) is python based. To make setup as simple as possible (no additional unrelated prerequisites), I would prefer a TS based MCP server which also does not require a dedicated API key. 

https://github.com/microsoft/lets-learn-mcp-javascript provides that, but since this is a "lets-learn" project, and could potentially change, @ZhongpinWang suggested copying the code over. Its also a single file:   
https://github.com/microsoft/lets-learn-mcp-javascript/blob/main/main.ts.

I will update our tutorial docs.

### DIFF
--- a/sample-code/src/tutorials/agent-openai-langchain.ts
+++ b/sample-code/src/tutorials/agent-openai-langchain.ts
@@ -1,4 +1,4 @@
-/* eslint-disable no-console */
+/* eslint-disable no-console import/no-internal-modules */
 import {
   StateGraph,
   MessagesAnnotation,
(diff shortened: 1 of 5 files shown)


sample-code/src/tutorials/agent-openai-langchain.ts
```diff
@@ -8,48 +8,23 @@ import {
...
-  paris: { temperature: '22°C', condition: 'Mild and pleasant' },
-  reykjavik: { temperature: '3°C', condition: 'Cold and windy' }
-};
-
 const mockRestaurantData: Record<string, string[]> = {
   paris: ['Le Comptoir du Relais', "L'As du Fallafel", 'Breizh Café'],
   reykjavik: ['Dill Restaurant', 'Fish Market', 'Grillmarkaðurinn']
 };
 
-const getWeatherTool = tool(
```
MatKuhr, line 37: [pp] I'd prefer to leave examples for both MCP and non-MCP based tools in

sample-code/src/tutorials/mcp/weather-mcp-server.ts
```diff
@@ -0,0 +1,58 @@
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
+import { z } from "zod";
+
+const server = new McpServer({
+  name: "Weather Server",
+  version: "1.0.0"
+});
+
+server.tool(
```
MatKuhr, line 11: [pp] maybe point out that typically this part is not necessary, as one will often use provided / remote MCP servers, and only included here to have an independent example

MatKuhr (commented):
I'm not sure copying is worth it, because it requires us to include the [original MIT license](https://github.com/microsoft/lets-learn-mcp-javascript?tab=MIT-1-ov-file):

> The above copyright notice and this permission notice shall be included in all
    copies or substantial portions of the Software.

We should also link to the original code

sample-code/src/tutorials/mcp/weather-mcp-server.ts
```diff
@@ -0,0 +1,58 @@
...
+  'get-weather',
+  'Tool to get the weather of a city',
+  {
+    city: z.string().describe("The name of the city to get the weather for")
```
ZhongpinWang: [pp] As already replaced elsewhere, I would recommend promoting `meta()` method to replace `describe()`
```suggestion
    city: z.string().meta({ description: "The name of the city to get the weather for" })
```

sample-code/src/tutorials/agent-openai-langchain.ts
ZhongpinWang: The `getRestaurantsTool` is still non-MCP

sample-code/src/tutorials/agent-openai-langchain.ts
```diff
@@ -67,7 +42,7 @@ const getRestaurantsTool = tool(
 );
 
 // Define the tools for the agent to use
-const tools = [getWeatherTool, getRestaurantsTool];
+const tools = [...getWeatherTools, getRestaurantsTool];
```
ZhongpinWang: [pp] This feels weird though if we have a list of get weather tools. Can we rename this and call the actual used one `getWeatherTool`?

sample-code/src/tutorials/agent-openai-langchain.ts
deekshas8: Yes the other tool (mockRestraurant) is still used in the agent.

sample-code/src/tutorials/mcp/weather-mcp-server.ts
deekshas8: Yes, I was planning to do this when writing the tutorial and also give an example how to configure an `http` one for example.

PR #923, opened 2025-08-04
feat: Client-side history

tomfrenken:
## Context

Relates to SAP/ai-sdk-js-backlog#315.

This PR adds client-side chat history for both streaming and non-streaming.

The stream-lock will be in a dedicated PR.

### DIFF
--- a/packages/orchestration/src/orchestration-client.ts
+++ b/packages/orchestration/src/orchestration-client.ts
@@ -144,7 +144,7 @@ export class OrchestrationClient {
     return this.messagesHistory;
   }
 
-  private async *processStreamEnd(
+  private async *processMessageHistory(
     stream: OrchestrationStream<OrchestrationStreamChunkResponse>,
     response?: OrchestrationStreamResponse<OrchestrationStreamChunkResponse>
   ): AsyncGenerator<OrchestrationStreamChunkResponse> {
@@ -154,8 +154,6 @@ export class OrchestrationClient {
     for await (const chunk of stream) {
       yield chunk;
     }
-
-    response._openStream = false;
 
     if (this.useClientHistory) {
       this.messagesHistory = response.getAllMessages();
@@ -220,7 +218,8 @@ export class OrchestrationClient {
         OrchestrationStream._processOrchestrationStreamChunkResponse,
         response
       )
-      ._pipe(this.processStreamEnd.bind(this), response);
+      ._pipe(OrchestrationStream._processStreamEnd, response)
+      ._pipe(this.processMessageHistory.bind(this), response);
 
     return response;
   }
--- a/packages/orchestration/src/orchestration-stream.ts
+++ b/packages/orchestration/src/orchestration-stream.ts
@@ -55,6 +55,20 @@ export class OrchestrationStream<Item> extends SseStream<Item> {
     }
   }
 
+  static async *_processStreamEnd(
+    stream: OrchestrationStream<OrchestrationStreamChunkResponse>,
+    response?: OrchestrationStreamResponse<OrchestrationStreamChunkResponse>
+  ): AsyncGenerator<OrchestrationStreamChunkResponse> {
+    if (!response) {
+      throw new Error('Response is required to process stream end.');
+    }
+    for await (const chunk of stream) {
+      yield chunk;
+    }
+
+    response._openStream = false;
+  }
+
   /**
    * Transform a stream of chunks into a stream of content strings.
    * @param stream - Orchestration stream.


packages/orchestration/src/orchestration-client.ts
```diff
@@ -34,14 +38,18 @@ const logger = createLogger({
  * Get the orchestration client.
  */
 export class OrchestrationClient {
+  private useClientHistory = true;
```
ZhongpinWang: [q] I am not sure if we can enable client history by default due to the following reasons:

- This behaviour diverges from the official and our OpenAI clients. Most clients are stateless.
- The original BLI requires this only for LangChain Orchestration client.
- I am not sure if by default users want to have an in-memory storage and our only hope is garbage collector.
- For non-conversational case, user might not know they now need to initialize a new client for each and every call as the old result will be part of the message history.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -34,14 +38,18 @@ const logger = createLogger({
...
   /**
    * Creates an instance of the orchestration client.
    * @param config - Orchestration module configuration. This can either be an `OrchestrationModuleConfig` object or a JSON string obtained from AI Launchpad.
+   * @param clientConfig - Client configuration for the orchestration client.
    * @param deploymentConfig - Deployment configuration.
    * @param destination - The destination to use for the request.
    */
   constructor(
     private config: OrchestrationModuleConfig | string,
+    clientConfig?: ClientConfig,
```
ZhongpinWang, line 52: [pp] Adding one more params in the constructor is breaking and looks not really elegant. Since this feature is probably only used in our LangChain Orchestration client as an opt-in, maybe we can add a setter for this and then in LangChain Orchestration, we add a clientConfig option, and call the setter internally in LangChain Orchestration after the client initialization?

packages/orchestration/src/orchestration-client.ts
```diff
@@ -89,12 +129,40 @@ export class OrchestrationClient {
...
+    for await (const chunk of stream) {
+      yield chunk;
+    }
+
+    response._openStream = false;
+
+    if (this.useClientHistory) {
+      this.messagesHistory = response.getAllMessages();
+    }
+  }
```
ZhongpinWang: [req] I understand why you add this here. I would still reuse `OrchestrationStream._processStreamEnd` and add another async generator in this file to add message history if there is no better way 🧐

packages/orchestration/src/orchestration-client.ts
tomfrenken: This is actually not used in langchain at all.
I agree that adding another param is annoying, but I do not see a more elegant way, other than creating some consolidating object as an in-between layer.

packages/orchestration/src/orchestration-client.ts
tomfrenken: I would add it by default:
- Other clients don't have this, because OpenAI has this feature server-side, and LangChain has adapters, the requirement to have this type of feature is clearly there
- the original BLI was kind of clueless
- While I am also not 100% sure, my feeling is that it covers most use-cases, if this assumption turns out to be false we can still revert in v3, and users can simply turn it off already
- That's true, but with a major version, I would argue that this type of change is expected, as long as we document it

packages/orchestration/src/orchestration-client.ts
tomfrenken: So you would put all the logic, except the message history, back into `OrchestrationStream._processStreamEnd` and then only have something like `OrchestrationClient.processMessageHistory` on top of that?

packages/orchestration/src/orchestration-client.ts
tomfrenken: I've switched it to off by default 👍

packages/orchestration/src/orchestration-client.ts
ZhongpinWang: As discussed, when using LangChain's memory saver, since the message history saved comes from the request, which is not templated, not masked, not translated etc. In this case, we can only put them in the messages instead of `messagesHistory` in our client to let orchestration process all of them again. If we need processed content, we would have to read it from the `prompt_templating.template` property inside `intermediate_results`.

This is what we wanted to avoid. So instead of using LangChain's solution, we provide an opt-in feature in our orchestration client to have a sort-of in-memory list to save processed history.

I feel like calling a setter to set the message history and turn the client history on after initializing the orchestration client in LangChain could help.

packages/orchestration/src/orchestration-client.ts
ZhongpinWang: Yes, otherwise next time, we add something in the original `OrchestrationStream._processStreamEnd`, we probably will forget to add it to this one as well. In short, a separate async generator for

```
if (this.useClientHistory) {
      this.messagesHistory = response.getAllMessages();
    }
```

would be nice. I guess unfortunately, there is no other better way, at least not in my mind at this moment.

PR #930, opened 2025-08-07
chore: Disable workflows before v2 release

KavithaSiva:
## Context

- ~[ ] If my change will be merged into the `main` branch (for v2), I've updated (v2-Upgrade-Guide.md)[./v2-Upgrade-Guide.md] in case my change has any implications for users updating to SDK v2~

This PR disables some workflows and adapts the publish script to release a next version with major version changes in all packages.

**Note: These changes have to be reverted before v2 release**


.github/workflows/spec-update.yaml
```diff
@@ -1,212 +1,213 @@
...
-
-          if ${{ steps.spec_diff.outputs.spec_diff == 'true' }}; then
-            echo "| Client Generation | ${{ steps.generate.outcome == 'success' && '✅' || '❌' }} ${{ steps.generate.outcome }}" >> $GITHUB_STEP_SUMMARY
-            echo "| Client Compilation | ${{ steps.compile.outputs.compilation_result == 'success' && '✅' || '❌' }} ${{ steps.compile.outputs.compilation_result }}" >> $GITHUB_STEP_SUMMARY
-            echo "| Client Testing | ${{ steps.compile.outputs.test_result == 'success' && '✅' || steps.compile.outputs.test_result == 'skipped' && '⏩' || '❌' }} ${{ steps.compile.outputs.test_result }}" >> $GITHUB_STEP_SUMMARY
-            echo "| Branch Creation | ${{ steps.push.outcome == 'success' && '✅ [Branch Link]($DIFF_URL)' || '❌ failure' }}" >> $GITHUB_STEP_SUMMARY
-            echo "| Pull Request Creation | ${{ env.CREATE_PR == 'false' && '⏩ skipped' || '' }}${{ env.CREATE_PR == 'true' && steps.push.outcome == 'success' && '✅ [PR Link]($PR_URL)' || '' }}" >> $GITHUB_STEP_SUMMARY
-          fi
+# Enable workflow after orchestration v2 changes are merged
```
ZhongpinWang: [q] I am not sure if we should disable this as orchestration team will not be able to use the generator.

tomfrenken: [q] Why disable spec-file updates? I'd assume we would do v2 spec-file updates still?

.github/workflows/publish-canary.yml
```diff
@@ -22,13 +22,35 @@ jobs:
       - name: prepare canary changeset
         run: |
           rm -f .changeset/*.md || true
-          cat <<EOT >> .changeset/canary-release-changeset.md
+          project_version=$(cat package.json | jq '.version' | tr -d '"')
+          echo The project version is: $project_version
+
+          # If still on v1, do a major bump, otherwise do a patch
```
tomfrenken, line 28: [q] Why do a major bump on v1? I'd think v1 canary is intentionally v1?

tomfrenken (approved):
LGTM, but two open questions befor emerging

.github/workflows/bump.yml
```diff
@@ -1,135 +1,136 @@
-name: bump
+# Enable workflow while releasing v2
```
ZhongpinWang, line 1: [q] Why do we need to comment everything out? Are they affecting calling publish-canary?

KavithaSiva: As these jobs can be called from the Github UI with the `main` branch as input it's just a precautionary step.

KavithaSiva:
> I am not sure if I understand everything correctly. So you are trying to bump to the next major version everytime we do a canary (or more precisely next) release, is this correct? Is this what we did before for cloud sdk as I cannot recall this. Maybe @deekshas8 still remembers.
> 
Yes, this is correct, something similar was done for cloud sdk too: https://github.com/SAP/cloud-sdk-js/pull/5246/files
The idea is to release with version 2.x for `next` tags and we need to bump to next major version every time until we update the versions in `package.json`.

I think we only changed the version in `package.json` to 4.x.x during release

> Also why are we commenting out other files? Are they at this moment risky somehow?
Just commenting them out to prevent running them accidentally from the Github UI.

.github/workflows/spec-update.yaml
KavithaSiva: Do they use this workflow? I thought it was only for us to check.
I also wanted to disable it until your PR is done and we have a clean package structure for v2.

.github/workflows/publish-canary.yml
KavithaSiva: So, although the name is misleading, it is actually releasing a `next` instead of canary.
And we did something similar in the past for cloud sdk too: https://github.com/SAP/cloud-sdk-js/pull/5246/files
From what I understand, as we don't change the versions in `package.json` in the v2 branch, this major bump is required to release `next` versions with `2.x.x`.

.github/workflows/spec-update.yaml
KavithaSiva: I wanted to disable it until we have a clean package structure for orchestration v2.

KavithaSiva: Okay, you are right, looks like they run this from `main`, will re-enable it.

KavithaSiva: Enabled it back again, the orchestration team runs it from `main` branch.

PR #950, opened 2025-08-12
chore: Adjust v2 workflows

emincihangeri:
Co-authored-by: emincihangeri <emincihangeri@users.noreply.github.com>

## Context

Closes SAP/ai-sdk-js-backlog#379.

- [ ] I know which base branch I chose for this PR, as the default branch is `v1-main` now, which is not for v2 development.
- [ ] If my change will be merged into the `main` branch (for v2), I've updated (v2-Upgrade-Guide.md)[./v2-Upgrade-Guide.md] in case my change has any implications for users updating to SDK v2


.github/workflows/trigger-v1-tests.yml
```diff
@@ -0,0 +1,18 @@
...
+    - cron: 0 2 * * *
+
+jobs:
+  smoke-tests-v1:
+    uses: SAP/ai-sdk-js/.github/workflows/smoke-tests.yaml@v1-main
+    with:
+      branch: v1-main
+    secrets: inherit
+  e2e-tests-v1:
+    uses: SAP/ai-sdk-js/.github/workflows/e2e-tests.yaml@v1-main
```
KavithaSiva: [req] There are lint errors, the workflow should be `e2e-tests.yml`. 
~We also need to add the `workflow_call` to these workflows in this branch too, otherwise lint still complains.~

Looks like once the `v1-main` PR is merged, the `workflow_call key is not defined in the referenced workflow.` should go. We can test this out, just correct the file extension.

KavithaSiva: I have the Github Actions extension, this is the one showing the error, not lint. Maybe you haven't installed this and that's why you are not seeing the error.

emincihangeri: So I am just adding an empty workflow_call to the same files. But should I change the file names just how I call them?

KavithaSiva: Just change the file extension here, you don't need to add the `workflow_call` to the workflows in this branch.

.github/workflows/dependabot-pnpm-lock.yml
```diff
@@ -1,34 +0,0 @@
-name: Update pnpm-lock.yaml for Dependabot PRs
```
ZhongpinWang: Double check why this is skipped

.github/workflows/trigger-v1-tests.yml
```diff
@@ -0,0 +1,18 @@
+name: trigger v1 tests
+
+on:
+  schedule:
+    # At 2 AM UTC every day
+    - cron: 0 2 * * *
```
ZhongpinWang: Change the time maybe

emincihangeri: Do you mean the same frequency but different hour?

.github/workflows/dependabot-pnpm-lock.yml
KavithaSiva: This action is not skipped, it is still being run for dependabot PRs, so please put this workflow back in this PR.

.github/workflows/trigger-v1-tests.yml
KavithaSiva: Yes, the smoke tests of v2 also run at the same time so pick an hour later.

.github/workflows/trigger-v1-tests.yml
```diff
@@ -0,0 +1,18 @@
+name: trigger v1 tests
+
+on:
+  schedule:
+    # At 3 AM UTC every day
+    - cron: 0 3 * * *
+
+jobs:
```
KavithaSiva, line 8: On second thought, can you also trigger the canary release from v1, every night too?

.github/workflows/trigger-canary.yml
```diff
@@ -0,0 +1,13 @@
+name: trigger canary release
```
KavithaSiva: [pp] Adjust the name to reflect that it's triggering v1 release.

.github/workflows/trigger-v1-canary.yml
```diff
@@ -0,0 +1,13 @@
+name: trigger v1 canary release
+
+on:
+  schedule:
+    # At 4 AM UTC every day
+    - cron: 0 4 * * *
```
KavithaSiva: [pp] Also, the canary release should run before the tests as the smoke tests use the canary version for testing.

PR #953, opened 2025-08-12
chore!: Improve API facade

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#363.

Check the changeset and upgrade guide for all details.

ADR included in this PR. Review needed.
Mostly exporting strategy change. Also renaming some `data` to `_data` in the response object.


V2-Upgrade-Guide.md
```diff
@@ -39,8 +40,104 @@ Update all SAP Cloud SDK for AI packages to version 2.x in your `package.json` f
 
 ## Breaking Changes
 
+### `@sap-ai-sdk/foundation-models`
+
+#### Type Import Changes
+
+Generated types are no longer exported from `@sap-ai-sdk/foundation-models` and must be imported from `@sap-ai-sdk/foundation-models/internal.js`.
```
MatKuhr:
```suggestion
Generated types are no longer exported from `@sap-ai-sdk/foundation-models`.
For most use cases they replaced by stable types in the same package.
For edge cases where the underlying generated types are still required, they now must be imported from `@sap-ai-sdk/foundation-models/internal.js`.
```

[pP] Maybe to not overly promote the internal types

V2-Upgrade-Guide.md
```diff
@@ -39,8 +40,104 @@ Update all SAP Cloud SDK for AI packages to version 2.x in your `package.json` f
...
+} from '@sap-ai-sdk/foundation-models';
+```
+
+#### Chat Completion Parameter Type
+
+The `AzureOpenAiCreateChatCompletionRequest` type is no longer exported publicly.
+Use the new `AzureOpenAiChatCompletionParameters` type instead.
+
+**v1:**
+```typescript
```
MatKuhr, line 115: [pp] I feel the differences are hard to spot due to the lenghty names. Maybe try ```diff  instead?

packages/orchestration/src/orchestration-types.ts
```diff
@@ -428,3 +430,23 @@ export type AzureContentSafety = AzureContentSafetyOutput;
  */
 export type AzureContentSafetyFilterConfig =
   AzureContentSafetyOutputFilterConfig;
+
+/**
+ * Representation of the 'LlamaGuard38BFilterConfig' schema.
+ */
+export type LlamaGuardFilterConfig = LlamaGuard38BFilterConfig;
```
KavithaSiva, line 437: [suggestion] Use `Pick`, to only expose properties, we need, for e.g, I think `type` can be excluded from all module config types.
Current approach may introduce breaking changes.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -428,3 +430,23 @@ export type AzureContentSafety = AzureContentSafetyOutput;
...
+
+/**
+ * Representation of the 'DpiConfig' schema.
+ */
+export type DpiMaskingProviderConfig = DpiConfig;
+
+/**
+ * Representation of the 'SAPDocumentTranslation' schema.
+ */
+export type TranslationConfig = SAPDocumentTranslation;
```
KavithaSiva, line 452: [q] These types are only used for the `build..` convenience functions now. Why don't we also use them in  the `OrchestrationModuleConfig` ? Or use them in the interfaces defined for each module config?

packages/orchestration/src/util/filtering.ts
```diff
@@ -43,7 +43,7 @@ export function buildAzureContentSafetyFilter(
  */
 export function buildLlamaGuardFilter(
   ...categories: [LlamaGuardCategory, ...LlamaGuardCategory[]]
-): LlamaGuard38BFilterConfig {
+): LlamaGuardFilterConfig {
```
KavithaSiva, line 46: [q] I remember we discussed an API for other modules to differentiate between `input` and `output` config. Did you consider doing this for the convenience functions of the filtering module too?

packages/orchestration/src/orchestration-types.ts
```diff
@@ -428,3 +430,23 @@ export type AzureContentSafety = AzureContentSafetyOutput;
...
+
+/**
+ * Representation of the 'LlamaGuard38BFilterConfig' schema.
+ */
+export type LlamaGuardFilterConfig = LlamaGuard38BFilterConfig;
+
+/**
+ * Representation of the 'GroundingModuleConfig' schema.
+ */
+export type DocumentGroundingConfig = GroundingModuleConfig;
```
KavithaSiva, line 442: [q] Our `build...` convenience functions are used sometimes directly for a module and sometimes are used in nested structures. Should we consider unifying their behaviour in v2?

```ts
    filtering: {
      input: {
        filters: [buildLlamaGuardFilter('privacy')]
      }
    }
    grounding: buildDocumentGroundingConfig({
      placeholders: {
        input: ['groundingInput'],
        output: 'groundingOutput'
      },
      filters: [
        {
          data_repository_type: 'help.sap.com'
        }
      ]
    })
```

packages/orchestration/src/index.ts
```diff
@@ -36,3 +40,19 @@ export {
...
+ * Explicitly export the wrapper also in `internal.ts`.
+ */
+export type {
+  ChatMessage,
+  SystemChatMessage,
+  UserChatMessage,
+  AssistantChatMessage,
+  ToolChatMessage,
+  DeveloperChatMessage,
+  ChatCompletionTool,
```
KavithaSiva, line 56: [suggestion] `ChatCompletionTool` has a `type` that seems redundant, can we use a wrapper type instead in `Template` and not export this type?

packages/orchestration/src/orchestration-types.ts
ZhongpinWang: This is just an alias, in case in the future, only part of the type or any specific changes needed, we can just re-route/copy-paste it internally. User won't be affected.

packages/orchestration/src/orchestration-types.ts
ZhongpinWang: Because in between there are still one layer, which I don't feel necessary to expose. Also even using these wrapper/alias types there won't add on any more value. For our previous issue caused by splitting input and output, this won't help.

packages/orchestration/src/util/filtering.ts
ZhongpinWang: This will be done in a follow-up, also after azure content filter PR.

packages/orchestration/src/orchestration-types.ts
ZhongpinWang: This is due to the nature of orchestration config. We normally build something if there is a `type` property inside. But definitely open to ideas how we can improve our build functions. But maybe we do that in a follow-up?

packages/orchestration/src/index.ts
ZhongpinWang: I am not sure. Checking the sample-code, this is an example:

```ts
const addNumbersTool: ChatCompletionTool = {
  type: 'function',
  function: {
    name: 'add',
    description: 'Add two numbers',
    parameters: toJsonSchema(addNumbersSchema)
  }
};
```

User needs this type to create variable. I agree for most of the time, this seems redundant. Ideally, we can create similar convenience like LangChain's `tool()`, maybe that is a v2 feature 😄

V2-Upgrade-Guide.md
ZhongpinWang: I like this idea, but I just tried in docusaurus, they don't highlight diff. Eventually, this migration guide will end up in docs portal, so maybe I keep it as-is. I didn't find any plugins to do this.

packages/foundation-models/src/index.ts
```diff
@@ -15,4 +15,17 @@ export {
   AzureOpenAiChatCompletionStream
 } from './azure-openai/index.js';
 
-export type * from './azure-openai/client/inference/schema/index.js';
+/**
+ * Exporting frequently used types.
+ * In case of breaking changes, create wrappers.
+ * Explicitly export the wrapper also in `internal.ts`.
+ */
+export type {
```
deekshas8, line 23: [q] Why did we decide to introduce a facade for foundation model? 
We state in our official docs:
> We are continuously updating the client to match the latest API specification.

The user then has the option to use older version using `api-version` query param. This is how we communicated the update from `2024-06-01` to `2024-10-21` which also had breaking changes, which we accepted.

If we create wrappers, how long are we supposed to maintain this backward compatibility? Do we keep creating wrappers for older versions? We can say we'll deprecate with time, but then it again sets us up for releasing a major version just for the sake of it. We are increasing our workload to what end?

ZhongpinWang: I agree this is in general a problem. As for now, we are exporting 

```ts
  AzureOpenAiChatCompletionTool, // used in sample-code
  AzureOpenAiFunctionObject, // essential part of sample-code

  AzureOpenAiChatCompletionRequestMessage, // used in sample-code
  AzureOpenAiChatCompletionRequestSystemMessage,
  AzureOpenAiChatCompletionRequestUserMessage,
  AzureOpenAiChatCompletionRequestAssistantMessage,
  AzureOpenAiChatCompletionRequestToolMessage
```

I would say OpenAI probably won't introduce real breaking changes for these as they are extreme frequently used. Backward compatibility is probably done by the service side. I mean if their spec is not backward compatible, then user needs to migrate.

This is essentially the same problem as orchestration package, and next time when we update the spec, it might break. After discussing with @MatKuhr and @jjtang1985 we will try not introduce breaking changes within one major version or recommend using internal. Wrappers won't necessarily be able to protect us, but not adding them will definitely make it breaking. If really there is no way due to huge breaking change, then we can deprecate the old types and old functions.

Good thing is that azure openai spec does not update that frequently. The latest stable version has been stable for almost ten months and the last preview was April 2025. Finger crossed.

.changeset/azure-openai-chat-completion-response-improvements.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/foundation-models': minor
+---
+
+[New Functionality] Add `getTokenUsage()`, `getFinishReason()`, `getContent()`, `getToolCalls()`, `getRefusal()`, `getAssistantMessage()`, `findChoiceByIndex()` to Azure OpenAI chat completion response.
```
deekshas8:
```suggestion
[New Functionality] Add `getTokenUsage()`, `getFinishReason()`, `getContent()`, `getToolCalls()`, `getRefusal()`, `getAssistantMessage()`, `findChoiceByIndex()` methods to Azure OpenAI chat completion response.
```

.changeset/add-orchestration-intermediate-results.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[New Functionality] Add method to access intermediate processing results.
```
deekshas8: [pp] The second line is sufficient

adr/004-api-facade.md
```diff
@@ -0,0 +1,45 @@
...
+However, for packages like `foundation-models` and `orchestration`, only the generated types are used, and the client is implemented manually to provide more convenient features.
+
+Currently, all generated types in `foundation-models` and `orchestration` packages are exported publicly. 
+It means that if the schema in OpenAPI specification changes, we need to add wrappers or aliases, or deprecate the old types to avoid breaking changes.
+This can sometimes be tricky as those changed types might directly be used in the public client APIs and could lead to deprecation of the APIs in order to fully support the upcoming new features.
+
+## Decision
+
+For manual-written client, proper wrappers are needed and only necessary types are exported publicly.
+Generated types should only be exported via `internal.js`.
```
deekshas8: There are exceptions to this right? Maybe mention that.

packages/foundation-models/src/azure-openai/azure-openai-chat-client.ts
```diff
@@ -29,35 +29,37 @@ export class AzureOpenAiChatClient {
 
   /**
    * Creates a completion for the chat messages.
-   * @param data - The input parameters for the chat completion.
+   * @param request - Request containing chat completion input parameters.
```
deekshas8, line 32: [req] This is not really the whole request object right? For better readability, I would keep it like before. Also ultimately in the `executeRequest` method, the param is called data. 
Also the description is confusing.

packages/foundation-models/src/azure-openai/azure-openai-embedding-client.ts
```diff
@@ -25,12 +25,12 @@ export class AzureOpenAiEmbeddingClient {
...
   /**
    * Creates an embedding vector representing the given text.
-   * @param data - The text to embed.
+   * @param request - Request containing embedding input parameters.
    * @param requestConfig - The request configuration.
    * @returns The completion result.
    */
   async run(
-    data: AzureOpenAiEmbeddingParameters,
+    request: AzureOpenAiEmbeddingParameters,
```
deekshas8, line 33: [req] same here

packages/foundation-models/src/index.ts
deekshas8: > The latest stable version has been stable for almost ten months and the last preview was April 2025.

Most of there changes are generally backwards compatible, but in the last update, I remember many types were deleted. So going forward, we'll need to maintain those as well, marked as deprecated off course. Was it also discussed how long we maintain this for? Until the next major release I assume?

packages/orchestration/src/util/filtering.ts
```diff
@@ -43,7 +43,7 @@ export function buildAzureContentSafetyFilter(
  */
 export function buildLlamaGuardFilter(
   ...categories: [LlamaGuardCategory, ...LlamaGuardCategory[]]
-): LlamaGuard38BFilterConfig {
```
deekshas8, line 46: [q] Since they named it very specific to the model (also with a `type`), if in the future they have another llama config (4 12B), then we might have a naming problem.

packages/foundation-models/src/index.ts
ZhongpinWang: For internal types we don't care, if removed we remove.
For the aforementioned seven public exported types, the risk is rather low if they completely change or be removed IMO. 
We discussed, if breaking changes despite our wrappers cannot be avoided, we accept the breaking fact.

packages/orchestration/src/util/filtering.ts
ZhongpinWang: I agree, the only problem I see is that we call the build function `buildLlamaGuardFilter` which also doesn't say 3-8b anywhere. Maybe we should change our build function in the follow-up together with this type?

packages/foundation-models/src/azure-openai/azure-openai-chat-client.ts
ZhongpinWang: I wanted to align this parameter with orchestration client, since there was a renaming change of the previous `prompt` param to the current `request` in the last orchestration v2 PR.

Besides, in orchestration, the type for this property is called `ChatCompletionRequest`.
In Azure OpenAI, the type is called `AzureOpenAiCreateChatCompletionRequest` and we added this wrapper to call it `AzureOpenAiChatCompletionParameters` because we elsewhere like in the embedding client called the type parameters... In LangChain they call this input.

It is indeed confusing that `executeRequest` calls it data, but that is for us.

I don't have strong opinion on this. In the end user won't care too much what it is called. I reverted the change and open to discussion.

deekshas8: Orchestration probably just copied from OpenAI. Can't comment on why OpenAI named their types like that.

Even though the user will not directly interact with the variable itself, but looking at the API, I find it confusing, since a param named `request` I generally associate with the HTTP request object (using express/nestjs and other libs).

Also when we invoke `executeRequest(request, requestConfig)`, it again feels like we're passing a `req` object and req configuration.

`data` at least clearly denotes we're only talking about the payload.

packages/orchestration/src/util/filtering.ts
deekshas8: Yes. I was initially thinking if we could make the function extensible/generic to allow users to construct all types of Llama filters. But after azure filter types, can't say with confidence if that'll be possible.

PR #966, opened 2025-08-15
feat!: Update build azure content safety and support prompt shield

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#311.

- [ ] If my change will be merged into the `main` branch (for v2), I've updated (v2-Upgrade-Guide.md)[./v2-Upgrade-Guide.md] in case my change has any implications for users updating to SDK v2

Add `type: 'input' | 'output'` to the object in function parameter.

So now,

```ts
filtering: {
  input: {
    filters: [
      buildAzureContentSafetyFilter({
        type: 'input',
        hate: 'ALLOW_SAFE',
        violence: 'ALLOW_SAFE',
        prompt_shield: true
      })
    ]
  }
}
```

The only problem that I cannot solve is building a input with `prompt_shield` and then assign this return value to `fitlering.output.filters[]`. TS won't complain. TS does structural typing and do not care that the optional field `prompt_shield` does not exist in output filter. 

But I would argue if user provides `input` in the build function, then use it for output, this is clearly wrong.

For now, both llama guard and translation modules still receive same config in the spec. So no a problem for them.
In the future, this problem might also not happen in translation build function, because input and output config will contain distinct non-optional fields https://github.tools.sap/AI/llm-orchestration/pull/1490/files#diff-3032fe37036de2409b5d0094886d5873ae20cde65109642dcbb274a0f51f85faR50


deekshas8 (commented):
Shouldn't `type` be a separate parameter as its not related to the config options? Putting them together in one object is weird.

deekshas8:
> But I would argue if user provides input in the build function, then use it for output, this is clearly wrong.

Agreed

KavithaSiva:
> Shouldn't `type` be a separate parameter as its not related to the config options? Putting them together in one object is weird.

I also like this, either a separate parameter or we can assign member functions to a main function:
```TS
const inputFilters = buildAzureContentSafetyFilter.input({
  hate: 'ALLOW_SAFE_LOW_MEDIUM',
  self_harm: 'ALLOW_SAFE_LOW'
});

const outputFilters = buildAzureContentSafetyFilter.output({
  sexual: 'ALLOW_SAFE',
  violence: 'ALLOW_SAFE_LOW_MEDIUM'
});
```

ZhongpinWang:
> Shouldn't `type` be a separate parameter as its not related to the config options? Putting them together in one object is weird.

Thanks for the feedback. I also don't like this idea much. I can easily switch to other options.

So for now, two remaining options:

1. We put input output as params. It is not super elegant as these two build functions will have different signature than others, but I am also totally fine with it.
2. We assign functions to the build function object or create a const for it. Downside is we have a lot of other stuff in autocompletion like `apply()`, `call()` etc.

ZhongpinWang:
> > It is not super elegant as these two build functions will have different signature
> 
> You mean compared to llama ? I agree its different, but how they can be used is also diff (additional input param), so I guess its fine. I would vote for the 1st option.

Not to llama as it will also have input output (although the types are the same for now). Rather to masking or grounding build functions. But anyway I also agree, I will go for the 1st option then. Thanks for the feedback.

V2-Upgrade-Guide.md
```diff
@@ -369,15 +369,17 @@ const filter = buildAzureContentFilter({
...
   violence: 'ALLOW_SAFE_LOW_MEDIUM'
 });
 ```
 
 #### Azure Content Filter Changes
 
-The Azure content filter property names have been updated to use lowercase with underscores.
+The `buildAzureContentSafetyFilter()` function now requires a `type` parameter as the first argument to distinguish between input and output filter configurations.
+This change allows input and output filters to have different configuration options, such as the `prompt_shield` property that is only available for input filters.
+Additionally, the Azure content filter property names have been updated to use lowercase with underscores.
```
KavithaSiva, line 382: [q] But why? Isn't the types based on interfaces we define? Can't we still have camel case? 
Or is it because snake case is being used in many other places and this is easier?

packages/orchestration/src/orchestration-types.ts
```diff
@@ -361,9 +361,35 @@ export type DpiMaskingConfig = Omit<
...
+   */
+  self_harm?: AzureFilterThreshold;
+  /**
+   * The filter category for sexual content.
+   */
+  sexual?: AzureFilterThreshold;
+  /**
+   * The filter category for violence content.
+   */
+  violence?: AzureFilterThreshold;
```
KavithaSiva, line 382: [pp] Consider having a base interface for common properties shared between the two.

packages/orchestration/src/util/filtering.test.ts
```diff
@@ -21,16 +21,16 @@ describe('Content filter util', () => {
     });
 
     it('builds filter config with no config', async () => {
-      const filterConfig = buildAzureContentSafetyFilter();
+      const filterConfig = buildAzureContentSafetyFilter('input');
```
KavithaSiva, line 24: [pp] Consider documenting this usage in documentation portal. Do you also have a PR for documentation open ?

KavithaSiva (approved):
Mostly LGTM, thanks for proposing the new structure for the convenience API.
I have left a few minor comments.

V2-Upgrade-Guide.md
ZhongpinWang: It's just because it is easier to map. `buildAzureContentSafetyFilter()` implementation dynamically uses keys of the object and map the value to it. We can also hard code four lines for this and map the value.. But..

packages/orchestration/src/util/filtering.test.ts
ZhongpinWang: I created a documentation BLI for v2 release. We have a lot of changes but never documented anything yet, also because v2 isn't stable yet, at least not before the release. But yes we must update the doc before releasing.

packages/orchestration/src/orchestration-types.ts
ZhongpinWang: We can, if we are sure that the common part will stay common. For this special case, my feeling agrees with you this might never change in the future. I am not strongly opinionated on this one.

But since orchestration also created input and output types separately without having any common part, I followed them just in case. Adding it or not is the same to our users. And in general, I would avoid making tighter assumptions to any future orchestration behaviour.

PR #973, opened 2025-08-15
chore: Allow `bump` to be triggered from UI

KavithaSiva:
## Context

We recently shifted to `main` as default, but I noticed that the `bump` workflow cannot be triggered via UI if the default branch doesn't contain the `workflow_dispatch` trigger.

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -1,12 +1,24 @@
-# Enable workflow while releasing v2
-# name: bump
+#Fully Enable workflow while releasing v2
+name: bump
 
-# on:
-#   workflow_dispatch:
-#     inputs:
-#       majorVersion:
-#         description: Mandatory, when bumping a major version. Semver compatible version string (X.Y.Z). Must not be set for patch and minor version releases.
-#         required: false
+on:
+  workflow_dispatch:
+    inputs:
+      majorVersion:
+        description: Mandatory, when bumping a major version. Semver compatible version string (X.Y.Z). Must not be set for patch and minor version releases.
+        required: false
+
+jobs:
+  check-branch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if running on main branch
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/bump.yml
```diff
@@ -1,12 +1,24 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      majorVersion:
+        description: Mandatory, when bumping a major version. Semver compatible version string (X.Y.Z). Must not be set for patch and minor version releases.
+        required: false
+
+jobs:
+  check-branch:
```
deekshas8, line 12: In the release steps ticket, please add a note to remove this job before releasing

.github/workflows/trigger-v1-tests.yml
```diff
@@ -2,8 +2,8 @@ name: trigger v1 tests
 
 on:
   schedule:
-    # At 4 AM UTC every day
-    - cron: 0 4 * * *
+    # At 4 AM UTC every Monday (weekly)
```
ZhongpinWang, line 5: [pp] I don't mind to have the tests running daily though. Just in case something isn't working, e.g., orchestration service v1 is buggy or something.

.github/workflows/trigger-v1-canary.yml
```diff
@@ -2,8 +2,8 @@ name: trigger v1 canary release
 
 on:
   schedule:
-    # At 3 AM UTC every day
-    - cron: 0 3 * * *
+    # At 3 AM UTC every Monday (weekly)
+    - cron: 0 3 * * 1
 
 jobs:
   canary-release:
```
ZhongpinWang, line 9: [q] Maybe the workflow version should be v1-main instead of `v1.17.0`?

.github/workflows/bump.yml
KavithaSiva: I added a note to revert https://github.com/SAP/ai-sdk-js/pull/930/files to v2 release ticket.

PR #986, opened 2025-08-19
chore: Add deploymentId as optional

emincihangeri:
## Context

Closes SAP/ai-sdk-js-backlog#148.

Add deploymentId to allow user to provide in certain cases.

### DIFF
--- a/packages/orchestration/src/orchestration-client.ts
+++ b/packages/orchestration/src/orchestration-client.ts
@@ -15,7 +15,9 @@ import type {
   HttpResponse,
   CustomRequestConfig
 } from '@sap-cloud-sdk/http-client';
-import type { ResourceGroupConfig } from '@sap-ai-sdk/ai-api/internal.js';
+import type {
+  DeploymentIdConfig,
+  ResourceGroupConfig } from '@sap-ai-sdk/ai-api/internal.js';
 import type {
   OrchestrationModuleConfig,
   ChatCompletionRequest,
@@ -42,7 +44,7 @@ export class OrchestrationClient {
    */
   constructor(
     private config: OrchestrationModuleConfig | string,
-    private deploymentConfig?: ResourceGroupConfig,
+    private deploymentConfig?: Partial<ResourceGroupConfig & DeploymentIdConfig>,
     private destination?: HttpDestinationOrFetchOptions
   ) {
     if (typeof config === 'string') {


packages/ai-api/src/utils/deployment-resolver.ts
```diff
@@ -176,12 +176,14 @@ export async function getAllDeployments(
...
  * @param modelDeployment - The model deployment configuration.
  * @param executableId - The executable ID.
+ * @param scenarioId - The scenario ID.
  * @returns The ID of the deployment, if found.
  * @internal
  */
 export async function getDeploymentId(
   modelDeployment: ModelDeployment,
   executableId: string,
+  scenarioId: string,
```
deekshas8: [q] Why is this change needed? The idea was if ai-api is down, then we can still pass the deploymentId to the orchestration client. You will then not call this function to fetch it.

I would also wait for Zhongpin's PR, since he made some changes to this func.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -42,7 +45,9 @@ export class OrchestrationClient {
    */
   constructor(
     private config: OrchestrationModuleConfig | string,
-    private deploymentConfig?: ResourceGroupConfig,
+    private deploymentConfig?: Partial<
```
deekshas8: [q] Why `Partial` ? deploymentConfig is already optional. `ResourceGroupConfig | DeploymentIdConfig` or `Xor<ResourceGroupConfig, DeploymentIdConfig>` might make more sense.

sample-code/src/server.ts
```diff
@@ -723,3 +724,28 @@ app.get('/prompt-registry/template', async (req, res) => {
...
+    // Example: test with a model name (foundation-models)
+    const deploymentId1 = await getDeploymentId(
+      'gpt-4o',
+      'azure-openai',
+      'foundation-models'
+    );
+
+    // Example: test with a deploymentId config (orchestration)
+    const deploymentId2 = await getDeploymentId(
+      { deploymentId: 'd123456789', resourceGroup: 'my-group' },
```
deekshas8: [q] I dont understand this, we call `getDeploymentId` but also pass a deploymentId?

deekshas8 (changes requested):

packages/ai-api/src/utils/deployment-resolver.ts
ZhongpinWang: @emincihangeri 

https://github.com/SAP/ai-sdk-js/blob/cc0905b19270a70e6a9d148571e0d799dd61d2cc/packages/foundation-models/src/azure-openai/azure-openai-chat-client.ts#L80 Here we use `getDeploymentId()`.

What I meant by changing `getDeploymentId()` in the `ai-api` is that there, `foundation-models` is hard-coded and this is not indicated by the function name. You can try to move `getDeploymentId()` to `foundation-models` package.

In orchestration, https://github.com/SAP/ai-sdk-js/blob/cc0905b19270a70e6a9d148571e0d799dd61d2cc/packages/orchestration/src/orchestration-client.ts#L122 Here you need to check if `deploymentId` in `deploymentConfig` is already set. If so, do not call resolve.

packages/orchestration/src/orchestration-client.ts
ZhongpinWang: `ResourceGroupConfig | DeploymentIdConfig` should work as eventually both of them are optional. Xor maybe not as they are not mutual exclusive?

deekshas8: Yeah true, the header might still be needed.

sample-code/src/server.ts
emincihangeri: I don't know another way to test it. I couldn't find an example regarding this.

packages/orchestration/src/orchestration-client.ts
deekshas8: This still needs to be fixed. Change it to `private deploymentConfig?: ResourceGroupConfig | DeploymentIdConfig`

packages/ai-api/src/utils/deployment-resolver.ts
```diff
@@ -206,7 +206,7 @@ export async function getAllDeployments(
  * @returns The ID of the deployment, if found.
  * @internal
  */
-export async function getDeploymentId(
+export async function getFoundationModelDeploymentId(
```
deekshas8: [req] I think @ZhongpinWang 's suggestion was to move the respective `get...DeploymentId` methods to their respective packages. So `getFoundationModelDeploymentId` to the foundation-model package. And Orchestration to its own.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -119,11 +124,13 @@ export class OrchestrationClient {
...
           );
 
-    const deploymentId = await resolveDeploymentId({
-      scenarioId: 'orchestration',
-      ...(this.deploymentConfig ?? {}),
-      destination: this.destination
-    });
+    const deploymentId = this.deploymentConfig?.deploymentId
+      ? this.deploymentConfig.deploymentId
+      : await resolveDeploymentId({
```
deekshas8: [req] Here we should also use the new function `getOrchestrationDeploymentId` you just created above right

sample-code/src/server.ts
```diff
@@ -738,3 +739,19 @@ app.get('/prompt-registry/template', async (req, res) => {
     sendError(res, error);
   }
 });
+
+app.get('/test/deployment-id', async (req, res) => {
+  try {
+    // Example: test with a model name (foundation-models)
+    const deploymentId1 = await getFoundationModelDeploymentId(
```
deekshas8: [req] This test is not necessary. We already have a test at `/ai-api/deployments` which tests `getDeployments()` method.
What you need to test as part of this ticket is:
1. Add a new type test that user can pass a deploymentId to the orchestrationClient initialization. 
2. Add a unit test to check that if you set a `deploymentId`, then the resolve deployment method is not called. This should be possible with jest's `spyOn`.

packages/foundation-models/src/azure-openai/azure-openai-chat-client.test.ts
```diff
@@ -134,9 +134,9 @@ describe('Azure OpenAI chat client', () => {
 
     mockDeploymentsList(
       {
-        scenarioId: 'foundation-models',
+        scenarioId: 'azure-openai',
         resourceGroup: 'custom-resource-group',
-        executableId: 'azure-openai'
+        executableId: 'foundation-models'
```
deekshas8: [q] Isn't this the other way around? scenarioId is 'foundation-models' and executableId changes depending on the model: https://help.sap.com/docs/ai-launchpad/sap-ai-launchpad/models-and-scenarios-in-generative-ai-hub-fef463b24bff4f44a33e98bb1e4f3148
@ZhongpinWang

deekshas8 (changes requested):

packages/foundation-models/src/azure-openai/azure-openai-chat-client.test.ts
deekshas8: I suspect this is the reason why some of the tests are failing.

deekshas8 (dismissed):
LGTM 🚀

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -969,4 +971,21 @@ describe('orchestration service client', () => {
       ).rejects.toThrow();
     });
   });
+
+  describe('OrchestrationClient deploymentId behavior', () => {
+    it('does not call resolveDeployment when deploymentId is provided', async () => {
```
deekshas8, line 976: 👍🏽 
I think we should restructure this file a bit to improve readability, but that's out of scope for this ticket

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -969,4 +971,21 @@ describe('orchestration service client', () => {
...
+      const deploymentConfig = { deploymentId: 'test-deployment-id' };
+
+      // Spy on the resolveDeployment function
+      const spy = jest.spyOn({ resolveDeployment }, 'resolveDeployment');
```
deekshas8:
```suggestion
      const spy = jest.spyOn({ resolveDeploymentId }, 'resolveDeploymentId');
```

`resolveDeploymentId` is called directly within `getOrchestrationDeploymentId`, hence the spy should be on it, since it will also not be invoked.

[req] Please also move this test to `deployment-resolver.test.ts`

.changeset/curvy-pianos-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': patch
```
deekshas8:
```suggestion
'@sap-ai-sdk/orchestration': minor
```

PR #994, opened 2025-08-20
feat: Add resolve deployment url

ZhongpinWang:
## Context

Support resolve deployment url with the given criteria. Usage example:

```ts
const url = await resolveDeploymentUrl({
  scenarioId: 'foundation-models',
  model: { name: 'gpt-4o' }
});
```

### DIFF
--- a/.changeset/tiny-trains-tap.md
+++ b/.changeset/tiny-trains-tap.md
@@ -1,0 +1,5 @@
+---
+'@sap-ai-sdk/ai-api': minor
+---
+
+[New Functionality] Add `resolveDeploymentUrl()` function to get deployment URL matches the given criteria.
--- a/packages/ai-api/src/utils/deployment-resolver.ts
+++ b/packages/ai-api/src/utils/deployment-resolver.ts
@@ -102,28 +102,22 @@ export interface DeploymentResolutionOptions {
 }
 
 /**
- * Query the AI Core service for a deployment that matches the given criteria. If more than one deployment matches the criteria, the first one's ID is returned.
+ * Query the AI Core service for a deployment that matches the given criteria.
+ * If more than one deployment matches the criteria, the first one's ID is returned.
  * @param opts - The options for the deployment resolution.
  * @returns A promise of a deployment, if a deployment was found, fails otherwise.
  * @internal
  */
-export async function resolveDeploymentId(
+export async function resolveDeployment(
   opts: DeploymentResolutionOptions
-): Promise<string> {
+): Promise<AiDeployment> {
   const { model } = opts;
-
-  const cachedDeployment = deploymentCache.get(opts);
-  if (cachedDeployment?.id) {
-    return cachedDeployment.id;
-  }
-
   let deployments = await getAllDeployments(opts);
 
   if (model) {
     deployments = deployments.filter(
       deployment => extractModel(deployment)?.name === model.name
     );
-
     if (model.version) {
       deployments = deployments.filter(
         deployment => extractModel(deployment)?.version === model.version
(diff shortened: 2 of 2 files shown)


.changeset/tiny-trains-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/ai-api': minor
+---
+
+[New Functionality] Add `resolveDeploymentUrl()` function to get deployment URL matches the given criteria.
```
MatKuhr:
```suggestion
[New Functionality] Add `resolveDeploymentUrl()` function to dynamically resolve the deployment URL for a given model.
```

packages/ai-api/src/utils/deployment-resolver.ts
```diff
@@ -136,7 +130,36 @@ export async function resolveDeploymentId(
...
+/**
+ * Query the AI Core service for a deployment that matches the given criteria.
+ * If more than one deployment matches the criteria, the first one's URL is returned.
+ * @param opts - The options for the deployment resolution.
+ * @returns A promise of the deployment URL, if a deployment was found, fails otherwise.
+ */
+export async function resolveDeploymentUrl(
+  opts: DeploymentResolutionOptions
+): Promise<string | undefined> {
+  return (await resolveDeployment(opts)).deploymentUrl;
```
MatKuhr, line 166: Why skip the cache in this case?

sample-code/src/ai-api/deployment-api.ts
```diff
@@ -125,3 +126,23 @@ export async function deleteDeployments(
...
 }
+
+/**
+ * Get the deployment URL for a given model and scenario.
+ * @param model - Model name.
+ * @param scenarioId - Scenario ID.
+ * @param resourceGroup - AI-Resource-Group where the resources are available.
+ * @returns The deployment URL.
+ */
+export async function getDeploymentUrl(
```
MatKuhr: why export this again from here?

sample-code/src/ai-api/deployment-api.ts
```diff
@@ -125,3 +126,23 @@ export async function deleteDeployments(
...
+  model: string,
+  scenarioId: string,
+  resourceGroup: string
```
MatKuhr: should be optional and default to `default`. Compare with what is optional in `DeploymentResolutionOptions`

packages/ai-api/src/utils/deployment-resolver.ts
ZhongpinWang: Good catch. Updated our cache object.

sample-code/src/ai-api/deployment-api.ts
ZhongpinWang: This is sample code and this function is used in `server.ts`

sample-code/src/ai-api/deployment-api.ts
ZhongpinWang: This is just the sample code? No one is using it except us in `server.ts`?

.changeset/tiny-trains-tap.md
ZhongpinWang: Not necessarily a model. Model is optional and for things like orchestration, this function can still be used with just scenario id. But I agree, I changed the word `get` to `resolve` to make it more clear.

.changeset/tiny-trains-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/ai-api': minor
+---
+
+[New Functionality] Add `resolveDeploymentUrl()` function to resolve the deployment URL matches the given criteria.
```
deekshas8:
```suggestion
[New Functionality] Add `resolveDeploymentUrl()` function to resolve the deployment URL that matches the given criteria.
```

packages/ai-api/src/utils/deployment-resolver.ts
```diff
@@ -136,7 +130,40 @@ export async function resolveDeploymentId(
...
+
+/**
+ * Query the AI Core service for a deployment that matches the given criteria.
+ * If more than one deployment matches the criteria, the first one's URL is returned.
+ * @param opts - The options for the deployment resolution.
+ * @returns A promise of the deployment URL, if a deployment was found, fails otherwise.
+ */
+export async function resolveDeploymentUrl(
+  opts: DeploymentResolutionOptions
+): Promise<string | undefined> {
```
deekshas8, line 161: [q] Can the url be undefined ? I haven't checked the actual type that's returned.

sample-code/src/server.ts
```diff
@@ -158,6 +159,14 @@ app.get('/ai-api/models', async (req, res) => {
   }
 });
 
+app.get('/ai-api/deployment-url', async (req, res) => {
+  try {
+    res.send(await getDeploymentUrl('gpt-4o', 'foundation-models', 'default'));
```
deekshas8: [q] I also din't get it, since we're exporting the function now, why not directly call it from here? Why duplicate it?

deekshas8 (commented):
Left some questions

packages/ai-api/src/utils/deployment-resolver.ts
ZhongpinWang: Actually it can, at least from the type level because when creating a deployment, the url might need some time to appear.

sample-code/src/server.ts
ZhongpinWang: Now I see the confusion. True, I now directly use the resolve function.

ZhongpinWang:
> Do we merge this in v1? Since the user seems to need it now

I already applied to the user that we will do this in v2 and provided workaround for now. If they request this to be part of v1 asap, then I add this. I just want to avoid adding any more features to v1 and go full speed v2.

PR #1005, opened 2025-08-25
chore: Update model list

KavithaSiva:
## Context

Closes SAP/ai-sdk-js-backlog#382

- [x] I have created a PR for `v1-main`, if my changes need to be backported to v1.

Adds `gpt-5` models to the available model list.

### DIFF
--- a/.changeset/clean-guests-drum.md
+++ b/.changeset/clean-guests-drum.md
@@ -1,5 +1,5 @@
 ---
-'@sap-ai-sdk/core': major
+'@sap-ai-sdk/core': patch
 ---
 
 [Improvement] Add `gpt-5`,`gpt-5-mini` and `gpt-5-nano` to the available model list.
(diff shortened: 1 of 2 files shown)


.changeset/clean-guests-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/core': major
```
KavithaSiva: [q] I think this should be `major` as this will be part of v2 release notes. Am I right?
Some changesets I see in v2 are `patch` and `minor` though, should I change all of them to `major` ?

deekshas8: For v2, everything should be major, ideally. But even if not, I think changeset takes the highest version identifier. So in the end, shouldn't be a problem if some incorrect ones were committed.

ZhongpinWang: Technically speaking, this should follow semver based on this specific change. So here it should not be major. But when we do bump, changeset takes the highest from the changeset files.

PR #1007, opened 2025-08-25
feat!: Update llama guard and translation convenience function

KavithaSiva:
## Context

Closes SAP/ai-sdk-js-backlog#381.

Add `type: 'input' | 'output'` to the convenience functions for translation and llama guard filter configuration.


packages/orchestration/src/orchestration-types.ts
```diff
@@ -466,6 +487,13 @@ export type AzureContentSafetyFilterReturnType<T extends 'input' | 'output'> =
  */
 export type LlamaGuardFilterConfig = LlamaGuard38BFilterConfig;
 
+/**
+ * Filter return type for Llama Guard.
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export type LlamaGuardFilterReturnType<T extends 'input' | 'output'> =
```
KavithaSiva: Added this here, otherwise the type input would not be used anywhere for Llama guard.

ZhongpinWang: [pp] I am fine with 😄 But no strong opinion at all

```ts
export type LlamaGuardFilterReturnType<T extends 'input' | 'output'> =
  T extends 'input'
    ? LlamaGuardFilterConfig
    : LlamaGuardFilterConfig;
```

packages/orchestration/src/util/translation.ts
```diff
@@ -5,14 +5,16 @@ import type {
...
- * @example "buildTranslationConfig({ sourceLanguage: 'de-DE', targetLanguage: 'en-US' })"
+ * @returns SAP Document Translation configuration.
+ * @example "buildTranslationConfig('input', { sourceLanguage: 'de-DE', targetLanguage: 'en-US' })"
  */
-export function buildTranslationConfig(
-  config: TranslationConfigParams
+export function buildTranslationConfig<T extends 'input' | 'output'>(
+  type: T,
+  config: TranslationConfigParams<T>
 ): TranslationConfig {
```
ZhongpinWang: [req] This needs to be similar to `LlamaGuardFilterReturnType<T>`.

packages/orchestration/src/util/filtering.ts
```diff
@@ -66,13 +66,15 @@ export function buildAzureContentSafetyFilter<T extends 'input' | 'output'>(
...
  * @param categories - Categories to be enabled for filtering. Provide at least one category.
- * @returns Filter config object.
- * @example "buildLlamaGuardFilter('self_harm', 'hate')"
+ * @returns Llama Guard filter configuration.
+ * @example "buildLlamaGuardFilter('input', ['elections', 'hate'])"
  */
-export function buildLlamaGuardFilter(
-  ...categories: [LlamaGuardCategory, ...LlamaGuardCategory[]]
-): LlamaGuardFilterConfig {
+export function buildLlamaGuardFilter<T extends 'input' | 'output'>(
```
ZhongpinWang: [req] @deekshas8  and I were discussing we need to change this build function to `buildLlamaGuard38BFilter()` and also the return type to allow future llama guard filter types.

packages/orchestration/src/orchestration-types.ts
KavithaSiva: I will skip making this verbose and leave it as is.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -462,10 +490,17 @@ export type AzureContentSafetyFilterReturnType<T extends 'input' | 'output'> =
     : AzureContentSafetyOutputFilterConfig;
 
 /**
- * Representation of the 'LlamaGuard38BFilterConfig' schema.
+ * Union type representation of all Llama Guard filter types.
  */
 export type LlamaGuardFilterConfig = LlamaGuard38BFilterConfig;
```
KavithaSiva: Idea was to add other types to a union here.

.changeset/five-spoons-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': major
+---
+
+[Compatibility Note] `buildLlamaGuardFilter()` function is renamed to `buildLlamaGuard38BFilter()` function and now requires `type` parameter to distinguish between `input` and `output` filter configuration and accepts filter categories as an array.
```
deekshas8:
```suggestion
[Compatibility Note] `buildLlamaGuardFilter()` function has been renamed to `buildLlamaGuard38BFilter()`. It now requires a type parameter to distinguish between `input` and `output` filter configurations, and accepts filter categories as an array.
```

V2-Upgrade-Guide.md
```diff
@@ -409,6 +409,50 @@ buildAzureContentSafetyFilter('output', {
   violence: 'ALLOW_ALL'
 })
 ```
+#### Llama Guard Filter Changes
+
+The `buildLlamaGuardFilter()` function is now renamed to `buildLlamaGuard38BFilter()` function and now requires a `type` parameter as the first argument to distinguish between input and output filter configurations, and accepts categories as an array instead of individual parameters.
```
deekshas8:
```suggestion
The `buildLlamaGuardFilter()` function has been renamed to `buildLlamaGuard38BFilter()` function.
It now requires a `type` parameter as the first argument to distinguish between `input` and `output` filter configurations, and accepts categories as an array instead of individual parameters.
```

New line for each sentence.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -446,13 +445,52 @@ export interface TranslationConfigParams {
   targetLanguage: string;
 }
 
+/**
+ * Output parameters for translation output configuration.
+ */
+export interface TranslationOutputParameters {
```
deekshas8: [q] This is possible without code duplication, don't you think? Just a suggestion:
```
type TranslationConfigParams //(same as original)
export type TranslationInputParameters = TranslationConfigParams
export type TranslationOutputParameters = TranslationConfigParams
```

This still allows for future extension with `&`.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -446,13 +445,52 @@ export interface TranslationConfigParams {
...
+   * Language to which the text should be translated.
+   * @example targetLanguage: "en-US"
+   */
+  targetLanguage: string;
+}
+
+/**
+ * Parameters for translation configurations.
+ */
+export type TranslationConfigParams<T extends 'input' | 'output'> =
```
deekshas8: [q] Why not create a type for `'input' | 'output'` instead of duplicating it everywhere?

packages/orchestration/src/util/translation.ts
```diff
@@ -1,19 +1,21 @@
 import type {
-  TranslationConfig,
-  TranslationConfigParams
+  TranslationConfigParams,
+  TranslationReturnType
 } from '../orchestration-types.js';
 
 /**
  * Convenience function to build a document translation configuration for orchestration service.
+ * @param type - Type of the translation configuration, either 'input' or 'output'.
```
deekshas8:
```suggestion
 * @param type - Type of the translation configuration, either `input` or `output`.
```

packages/orchestration/src/util/filtering.ts
```diff
@@ -65,14 +65,16 @@ export function buildAzureContentSafetyFilter<T extends 'input' | 'output'>(
 }
 
 /**
- * Convenience function to build Llama guard filter.
+ * Convenience function to build Llama Guard 3 8B filter.
+ * @param type - Type of the filter, either 'input' or 'output'.
```
deekshas8:
```suggestion
 * @param type - Type of the filter, either `input` or `output`.
```

Looks better in the IDE

packages/orchestration/src/orchestration-types.ts
```diff
@@ -446,13 +445,52 @@ export interface TranslationConfigParams {
   targetLanguage: string;
 }
 
+/**
+ * Output parameters for translation output configuration.
+ */
+export interface TranslationOutputParameters {
+  /**
+   * Language of the text to be translated.
+   * @example sourceLanguage: "de-DE"
```
deekshas8:
```suggestion
   * @example sourceLanguage: 'de-DE'
```

Should not suggest using `"`

packages/orchestration/src/orchestration-types.ts
```diff
@@ -446,13 +445,52 @@ export interface TranslationConfigParams {
...
+   */
+  sourceLanguage?: string;
+  /**
+   * Language to which the text should be translated.
+   * @example targetLanguage: "en-US"
```
deekshas8:
```suggestion
   * @example targetLanguage: 'en-US'
```

packages/orchestration/src/util/filtering.ts
```diff
@@ -65,14 +65,16 @@ export function buildAzureContentSafetyFilter<T extends 'input' | 'output'>(
...
  * @param categories - Categories to be enabled for filtering. Provide at least one category.
- * @returns Filter config object.
- * @example "buildLlamaGuardFilter('self_harm', 'hate')"
+ * @returns Llama Guard 3 8B filter configuration.
+ * @example "buildLlamaGuard38BFilter('input', ['elections', 'hate'])"
```
deekshas8:
```suggestion
 * @example buildLlamaGuard38BFilter('input', ['elections', 'hate'])
```
`"` removes the code highlighting

packages/orchestration/src/util/translation.ts
```diff
@@ -1,19 +1,21 @@
...
  * @param config - Config for SAP Document Translation service.
  * The target language is mandatory, while source language will be auto-detected if not provided.
  * See https://help.sap.com/docs/translation-hub/sap-translation-hub/supported-languages-6854bbb1bd824ffebc3a097a7c0fd45d for list of supported languages.
- * @returns A `SAPDocumentTranslation` configuration object.
- * @example "buildTranslationConfig({ sourceLanguage: 'de-DE', targetLanguage: 'en-US' })"
+ * @returns SAP Document Translation configuration.
+ * @example "buildTranslationConfig('input', { sourceLanguage: 'de-DE', targetLanguage: 'en-US' })"
```
deekshas8:
```suggestion
 * @example buildTranslationConfig('input', { sourceLanguage: 'de-DE', targetLanguage: 'en-US' })
```

tests/type-tests/test/orchestration.test-d.ts
```diff
@@ -3,24 +3,26 @@ import {
   OrchestrationClient,
   buildAzureContentSafetyFilter,
   buildDocumentGroundingConfig,
-  buildLlamaGuardFilter,
-  buildDpiMaskingProvider
+  buildLlamaGuard38BFilter,
+  buildDpiMaskingProvider,
+  buildTranslationConfig
 } from '@sap-ai-sdk/orchestration';
+import type { TranslationReturnType } from '@sap-ai-sdk/orchestration/src/orchestration-types.js';
```
deekshas8: [req] This is part of public api no? Can be imported from `@sap-ai-sdk/orchestration`

tests/type-tests/test/orchestration.test-d.ts
```diff
@@ -3,24 +3,26 @@ import {
...
 } from '@sap-ai-sdk/orchestration';
 import type {
   CompletionPostResponse,
   TokenUsage,
   GroundingModuleConfig,
   ChatMessages,
   DpiConfig,
-  MessageToolCalls
+  MessageToolCalls,
+  LlamaGuard38BFilterReturnType
```
deekshas8: [req] I see this is also exported publicly

deekshas8 (dismissed):

packages/orchestration/src/orchestration-types.ts
ZhongpinWang: Good catch I agree, there are also other places I think we can just use aliases.

packages/orchestration/src/orchestration-types.ts
ZhongpinWang: One thing is that I found it bit hard to name this thing as it is used in many build functions. Calling it something like `type` also feels bad as we have too many "type"s already. If you have better idea, go ahead and change all.

tests/type-tests/test/orchestration.test-d.ts
ZhongpinWang: Oh this is auto imported wrong. Good catch.

tests/type-tests/test/orchestration.test-d.ts
ZhongpinWang: Yes this was initially not exported publicly, forgot to check them in this file.

packages/orchestration/src/orchestration-types.ts
deekshas8: I modified the Azure Config types, and made some other fixes.

packages/orchestration/src/orchestration-types.ts
deekshas8: I called it `ConfigType` for now, `ConfigMode` could also be an option. Don't have very good ideas either. I wasn't sure if we should export it publicly. Wdyt?

ZhongpinWang: True, anyway we don't export it. Good enough for me.

PR #1019, opened 2025-09-01
chore: Adjust related variables

emincihangeri:
## Context

Closes SAP/ai-sdk-js-backlog#386.

Change action variables to keep node versions up-to-date

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "@types/jest": "^30.0.0",
     "@types/jsonwebtoken": "^9.0.10",
     "@types/mock-fs": "^4.13.4",
-    "@types/node": "^20.19.13",
+    "@types/node": "^22.0.0",
     "depcheck": "^1.4.7",
     "eslint": "^9.34.0",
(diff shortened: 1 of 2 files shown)


.github/workflows/spec-update.yaml
```diff
@@ -57,10 +57,13 @@ jobs:
       CHOICE: ${{ inputs.file }}
       REF: ${{ inputs.file-ref }}
       CREATE_PR: ${{ inputs.create-pr }}
+    strategy:
```
ZhongpinWang: We do not need to use matrix for this kind of workflow. Spec update is only needed for one LTS version. Adding matrix will let the workflow run three times which does not make much sense.

emincihangeri: [q] So the matrix configuration is only relevant in build.yml?

ZhongpinWang: The purpose of having matrix is to see if our SDK can be built and work with different node versions.

PR #1057, opened 2025-09-08
fix: Update spec path

ZhongpinWang:
## Context

### DIFF
--- a/.github/workflows/spec-update.yaml
+++ b/.github/workflows/spec-update.yaml
@@ -98,7 +98,7 @@ jobs:
               FILE_PATH='packages/orchestration/src/spec/api.yaml'
               ;;
             prompt-registry)
-              API_URL="repos/AI/prompt-registry/contents/src/spec/prompt-registry.yaml?ref=$REF"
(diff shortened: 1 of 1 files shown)


.github/workflows/spec-update.yaml
```diff
@@ -98,7 +98,7 @@ jobs:
               FILE_PATH='packages/orchestration/src/spec/api.yaml'
               ;;
             prompt-registry)
-              API_URL="repos/AI/prompt-registry/contents/src/spec/prompt-registry.yaml?ref=$REF"
+              API_URL="repos/AI/prompt-registry/contents/src/spec/generated/prompt-registry-combined.yaml?ref=$REF"
               FILE_PATH='packages/prompt-registry/src/spec/prompt-registry.yaml'
```
KavithaSiva, line 102: [pp] Consider changing the name of the file in the SDK to `prompt-registry-combined.yaml` too.

ZhongpinWang: We didn't really care about the spec filenames in our repo. For example, the spec file in orchestration package is called `api.yaml`. Since the spec files are just for our internal use, I would leave them untouched.

PR #1058, opened 2025-09-08
fix: Read `vars` from input instead of `default`

ZhongpinWang:
## Context

[Reusable flow `setup` fails](https://github.com/SAP/ai-sdk-js/actions/runs/17546129492/job/49827913432) due to using `vars` in input's `default` property.

Fix by reading `vars` in individual workflow and provide the value over `node-version` input property to the reusable workflow.

### DIFF
--- a/.github/workflows/auto-fix-lint.yml
+++ b/.github/workflows/auto-fix-lint.yml
@@ -18,6 +18,7 @@ jobs:
         with:
           ref: ${{ github.head_ref }}
           token: ${{ secrets.GH_CLOUD_SDK_JS_ADMIN_WRITE_TOKEN }}
+          node-version: ${{ vars.DEFAULT_NODE_VERSION }}
       - run: pnpm lint:fix
       - name: Commit Changes if needed
         run: |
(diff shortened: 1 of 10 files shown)


deekshas8 (approved):

.github/workflows/auto-fix-lint.yml
```diff
@@ -18,6 +18,7 @@ jobs:
         with:
           ref: ${{ github.head_ref }}
           token: ${{ secrets.GH_CLOUD_SDK_JS_ADMIN_WRITE_TOKEN }}
+          node-version: ${{ vars.DEFAULT_NODE_VERSION }}
```
deekshas8, line 21: Seems to still fail, do we need to provide 22 value here?

deekshas8: Ignore my previous comment, it picks up the correct workflow file, but maybe not the updated actions file?

ZhongpinWang: I will change the action tag and revert to see if it works before merging. But you are right, it is due to the `@main` tag.

ZhongpinWang: I tried `build` workflow and worked. https://github.com/SAP/ai-sdk-js/actions/runs/17548628294?pr=1058

I would assume all other occurrences should work.

ZhongpinWang:
https://github.com/SAP/ai-sdk-js/actions/runs/17548710970/job/49836006753#step:3:3

It works in main 🥳

PR #1067, opened 2025-09-11
docs: Update V2 Upgrade Guide 

KavithaSiva:
## Context

Updating the v2 upgrade guide to add notes about prompt registry and langchain orchestration streaming.

### DIFF
--- a/V2-Upgrade-Guide.md
+++ b/V2-Upgrade-Guide.md
@@ -248,7 +248,6 @@ const config = {
 };
 ```
 
-
 #### Parameter Name Changes
 
 Several parameter names have been updated for consistency.


V2-Upgrade-Guide.md
```diff
@@ -286,6 +286,45 @@ promptTemplating: {
 }
 ```
 
+##### Prompt Registry Template Reference
+
+Due to the consolidation of `llm` and `templating` into `promptTemplating`, move Prompt Registry template reference from `templating.template_ref` to `promptTemplating.prompt.template_ref`.
```
ZhongpinWang: [req] `template_ref` is not relevant to previous `llm` module. Is this part not already covered by `Module Configuration Structure`? If people know `templating.template` is moved to `promptTemplation.prompt.template`, won't they know it is the same for `template_ref`? If we think it might be unclear, then just add one sentence there mentioning it is the same for `template_ref` would be enough.

V2-Upgrade-Guide.md
```diff
@@ -398,16 +437,17 @@ buildAzureContentSafetyFilter('input', {
   hate: 'ALLOW_SAFE',
   self_harm: 'ALLOW_SAFE_LOW',
   sexual: 'ALLOW_SAFE_LOW_MEDIUM',
-  violence: 'ALLOW_ALL'
-})
+  violence: 'ALLOW_ALL',
+  prompt_shield: true // Only available for input
```
ZhongpinWang: [req] This is a new feature in v2. It does not belong to Upgrade. Users were not using prompt shield, and do not need to read the upgrade guide for using this new feature.

V2-Upgrade-Guide.md
```diff
@@ -507,6 +547,57 @@ await orchestrationClient.invoke(messages, {
 });
 ```
 
+#### Stream Method Parameter Change
```
ZhongpinWang: [req] There was [no change to LangChain](https://github.com/SAP/ai-sdk-js/blob/cdb48f6c57fe570ae08c25e71f4cd57b7854ce9b/sample-code/src/langchain-orchestration.ts#L186) except our internal implementation. [The user interface is always signal and defined by LangChain.](https://github.com/langchain-ai/langchainjs/blob/b641daa8997ed3a6552432cfea4032b4fad1ad88/langchain-core/src/runnables/types.ts#L109)

Please double check AI generated content.

KavithaSiva: Ah okay, I do check the AI generated content, but I missed that this was already the existing state in Langchain and we didn't make these changes in v2.

V2-Upgrade-Guide.md
KavithaSiva: True, I hadn't thought about that.

V2-Upgrade-Guide.md
KavithaSiva: I added the prompt registry changes under `Parameter Name Changes`, but adding a line under `Module Configuration Structure` is better. Done.

V2-Upgrade-Guide.md
```diff
@@ -214,7 +214,7 @@ Prefer using the provided getter methods instead of accessing the data object di
 
 #### Module Configuration Structure
 
-The most significant change is the consolidation of `llm` and `templating` modules into a single `promptTemplating` module.
+The most significant change is the consolidation of `llm` and `templating` modules into a single `promptTemplating` module.In the new structure, the `templating` module is renamed to `prompt`.
```
ZhongpinWang:
```suggestion
The most significant change is the consolidation of `llm` and `templating` modules into a single `promptTemplating` module as `model` and `prompt` properties respectively.
```

V2-Upgrade-Guide.md
```diff
@@ -248,6 +248,8 @@ const config = {
 };
 ```
 
+As part of the configuration structure change, all properties previously in `templating`: `template`, `tools`, `template_ref`, `defaults`, and `response_format` are now located in `promptTemplating.prompt`.
```
ZhongpinWang:
```suggestion
```

[pp] I think this part is not necessary. When users use `template_ref` or `tools` etc and migrate to v2, it is clear enough when they read this migration guide and know the whole template module has been moved. It indicates naturally that all the underlying properties are moved as well.

KavithaSiva: Yeah it does, just wanted to reinforce the change again esp. since we don't have code samples with say for e.g `tools`.

PR #1071, opened 2025-09-15
chore: Test for an error

emincihangeri:
## Context

Closes SAP/ai-sdk-js-backlog#302.

Test for checking the orchestration error response structure.

### DIFF
--- a/packages/orchestration/src/orchestration-client.test.ts
+++ b/packages/orchestration/src/orchestration-client.test.ts
@@ -900,6 +900,26 @@ describe('orchestration service client', () => {
     `);
   });
   describe('OrchestrationClient Stream Error Handling', () => {
+    it('should throw an error when streaming with invalid JSON configuration', async () => {
+      const invalidJsonConfig = JSON.stringify({
+        config: {
+          stream: { enabled: true },
+          modules: {
+            prompt_templating: {
+              model: {
+                name: 'gpt-4o',
+                params: { version: 4.9 }
+              }
+            }
+          }
+        }
+      });
+
+      const client = new OrchestrationClient(invalidJsonConfig);
+
+      await expect(client.stream()).rejects.toThrow('Invalid JSON configuration');
+    });
+
     it('should abort controller and re-throw error when network request fails', async () => {
       const config: OrchestrationModuleConfig = {
         promptTemplating: {


packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -900,6 +900,28 @@ describe('orchestration service client', () => {
...
+          }
+        }
+      });
+
+      const client = new OrchestrationClient(invalidJsonConfig);
+
+      await expect(client.stream()).rejects.toThrow(
+        'Invalid JSON configuration'
+      );
+    });
```
ZhongpinWang: This is not the kind of error we are talking about in this ticket. The error comes from remote response. When streaming, after the initial http response, the http connection will remain alive. Then we keep receive chunks in the form of `data: ......`. The `...` part is normally a chunk payload JSON. In case of an error, the remote server will send `data: { error: { ... } }` instead and then stop sending any other chunks and close the connection. 

Have a look at https://html.spec.whatwg.org/multipage/server-sent-events.html and first understand SSE. In this BLI, you need to create (or reuse if already there) mocked streaming chunks in `test-util/data/...` and then let the SSE parser handle each and every chunk using a for await loop. 
Check `azure-openai-chat-completion-stream-chunks.txt` for a normal response.

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -900,6 +900,56 @@ describe('orchestration service client', () => {
...
+      // Call the streaming API
+      const response = await new OrchestrationClient(config).stream();
+
+      // Iterate over each chunk
+      for await (const chunk of response.stream) {
+        if (chunk._data.error) {
+          expect(chunk._data.error).toHaveProperty('message');
+          break;
+        }
+      }
```
KavithaSiva: [q] How are the chunks parsing successfully? Aren't we throwing an error when we encounter a stream error?

emincihangeri: I could throw an error here because I thought that it was supposed to pass the test if it correctly parses the error message

emincihangeri: or do you mean the error should be thrown from https://github.com/SAP/ai-sdk-js/blob/ae0257a05b7cb8050d4dd8976df0e497adbb5134/packages/core/src/stream/sse-stream.ts#L44

KavithaSiva: Correct, since we throw an error in `sse-stream.ts` it should have been thrown until here, not sure where we suppress it unintentionally now.

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -900,6 +900,57 @@ describe('orchestration service client', () => {
...
+      const response = await new OrchestrationClient(config).stream();
+
+      // Expect the iteration over the stream to throw an error
+      await expect(async () => {
+        for await (const chunk of response.stream) {
+          if (chunk._data.error) {
+            throw new Error(chunk._data.error.message);
+          }
+        }
+      }).rejects.toThrow();
```
ZhongpinWang: [pp] Do you want to check the throw message?

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -900,6 +900,57 @@ describe('orchestration service client', () => {
     `);
   });
   describe('OrchestrationClient Stream Error Handling', () => {
+    it('executes a streaming request and parses each SSE chunk correctly', async () => {
```
ZhongpinWang:
```suggestion
    it('throws error when getting an SSE error chunk ', async () => {
```
Test title should describe the behaviour clear. Avoid using ambiguous word like "correctly".

packages/orchestration/src/orchestration-client.test.ts
emincihangeri: should I just leave the loop empty?

ZhongpinWang: What I mean was check the error message with `toThrow('.....')` because there can be other error thrown. Or use `toThrowErrorMatchingSnapshot()` by running the test once, and then you should see the error message in a `__snapshots__` folder. Next time jest will check if the error message is the same as stored in the folder.

emincihangeri: when I tried toThrowErrorMatchingSnapshot(), the error message comes out as "[object Object]". is there a way to resolve it?

ZhongpinWang: Then you probably need to check the error object properties. Use JSON stringify to see the error.

ZhongpinWang (approved):
lgtm

packages/core/src/stream/sse-stream.ts
```diff
@@ -39,28 +39,35 @@ export class SseStream<Item> implements AsyncIterable<Item> {
...
             logger.error(`Could not parse message into JSON: ${sse.data}`);
             logger.error(`From chunk: ${sse.raw}`);
             throw e;
           }
+
+          if (data?.error) {
+            logger.error(
+              `Error received from server: ${JSON.stringify(data.error)}`
+            );
+            throw new Error(JSON.stringify(data.error));
```
ZhongpinWang, line 56: @emincihangeri I tried out and saw the problem. Here I now stringify the error message because a JS error normally consists of error.name and error.message, which both are suppose to be strings.

PR #1087, opened 2025-09-18
fix: Update outdated lockfile

ZhongpinWang:
The current lockfile is due to pnpm cache too outdated..


packages/document-grounding/src/tests/pipelines-api.test.ts
```diff
@@ -4,7 +4,7 @@ import {
   mockClientCredentialsGrantCall
 } from '../../../../test-util/mock-http.js';
 import { PipelinesApi } from '../client/api/index.js';
-import type { Pipelines } from '../client/api/index.js';
+import type { GetPipelines } from '../client/api/index.js';
```
deekshas8, line 7: Is this something that was missed before?

ZhongpinWang: yes, and the other type import. Somehow the tests passed before

PR #1090, opened 2025-09-18
chore: Revert "Disable workflows before v2 release (#930)"

KavithaSiva:
This reverts commit 4568eaeb946746717a7a14ebb3cdb9c59cda97f3.

## Context

-~[ ] I have created a PR for `v1-main`, if my changes need to be backported to v1.~

This PR enables all workflows for v2 release.

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: sap/ai-sdk-js/.github/actions/setup@main
+        with:
+          node-version: ${{ vars.DEFAULT_NODE_VERSION }}
       - name: Get Changelog
         uses: sap/cloud-sdk-js/.github/actions/get-changelog@main
         id: get-changelog
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -20,6 +20,7 @@ jobs:
         with:
           token: ${{ secrets.GH_CLOUD_SDK_JS_ADMIN_WRITE_TOKEN }}
           ref: 'main'
+          node-version: ${{ vars.DEFAULT_NODE_VERSION }}
       - name: bump version
         uses: sap/cloud-sdk-js/.github/actions/changesets-fixed-version-bump@main
         id: bump


.github/workflows/publish-canary.yml
```diff
@@ -23,42 +23,20 @@ jobs:
...
-          fi
 
       - name: publish
         run: |
           date=`date +%Y%m%d%H%M%S`
           pnpm changeset pre enter ${date}
           pnpm changeset version
           pnpm changeset pre exit
-          pnpm changeset publish --tag ${{ github.event_name == 'schedule' && 'next' || inputs.tag }}
+          pnpm changeset publish --tag ${{ github.event_name == 'schedule' && 'canary' || inputs.tag }}
```
ZhongpinWang, line 40: [q] What happens to v1 now? We probably stop any canary release for v1 right?

.github/workflows/draft-release.yml
```diff
@@ -1,23 +1,20 @@
...
-#         env:
-#           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-#         with:
-#           draft: true
-#           body: ${{ steps.get-changelog.outputs.changelog }}
+jobs:
+  draft-github-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: sap/ai-sdk-js/.github/actions/setup@main
```
ZhongpinWang, line 11: [q/req] Why are we removing the default node version? And also in many other places? We should in general avoid relying on the default value in the action and control versions with one single repository variable. (Unfortunately we cannot use such var in reusable action)

ZhongpinWang (commented):
As there are too many changes, maybe after resolving the minors I commented, we can just let it run and see.

.github/workflows/publish-canary.yml
KavithaSiva: Yes, I haven't opened the PR for adjusting v1 workflows yet, will do that shortly, which either removes the canary release job or publishes the canary version with a legacy tag.

.github/workflows/draft-release.yml
KavithaSiva: I did not remove them, I merely adapted to use  `vars.DEFAULT_NODE_VERSION` instead of `env.NODE_VERSION`

KavithaSiva: Hmm, but looks like I missed this particular occurence, will adapt. Thanks for noticing.

KavithaSiva: Done.

PR #1097, opened 2025-09-20
chore(v1): Adapt v1 pipelines

KavithaSiva:
## Context
- [x] I have created a PR for `v1-main`, if my changes need to be backported to v1.

I have adapted v1 pipelines to publish a version with `legacy` task and removed the job to publish canary versions.


tests/smoke-tests/package.json
```diff
@@ -19,12 +19,12 @@
     "langchain": "0.3.34",
     "@langchain/core": "0.3.75",
     "@langchain/textsplitters": "0.1.0",
-    "@sap-ai-sdk/ai-api": "canary",
-    "@sap-ai-sdk/document-grounding": "canary",
-    "@sap-ai-sdk/foundation-models": "canary",
-    "@sap-ai-sdk/langchain": "canary",
-    "@sap-ai-sdk/orchestration": "canary",
-    "@sap-ai-sdk/prompt-registry": "canary",
```
deekshas8, line 27: [q] Will we run smoke-tests on v1 anymore? Should we disable the trigger for this branch?

KavithaSiva: Don't we normally support a previous version for 6 months? So we should allow these pipelines to be run right?

KavithaSiva: We discussed to leave the smoke tests, but disable the automation for v1 smoke tests.

PR #1108, opened 2025-09-22
feat: Created github action for pnpm update (#394)

BrigittaK307:
Closes SAP/ai-sdk-js-backlog#394.

### DIFF
--- a/.github/workflows/pnpm-update.yml
+++ b/.github/workflows/pnpm-update.yml
@@ -18,6 +18,7 @@ jobs:
       - name: Update Transitive Dependencies
         run: |
           pnpm --no-save update
+          pnpm audit --audit-level moderate
           if [[ -n "$(git status --porcelain)" ]]; then
             git config --global user.email "cloud-sdk-js@github.com"
             git config --global user.name "cloud-sdk-js"


.github/workflows/nightly-dependency-update.yml
```diff
@@ -0,0 +1,93 @@
+name: Nightly Dependency Update
```
ZhongpinWang: [pp] I am not sure if nightly update is necessary. Maybe ~weekly~ monthly to avoid lock file merge conflicts? We need a somewhat stable lock file to better debug in the future in case some issue pops up after the lock file update.

.github/workflows/nightly-dependency-update.yml
```diff
@@ -0,0 +1,93 @@
...
+    steps:
+      - name: Setup Environment
+        uses: sap/ai-sdk-js/.github/actions/setup@main
+        with:
+          token: ${{ secrets.GH_CLOUD_SDK_JS_ADMIN_WRITE_TOKEN }}
+          node-version: ${{ vars.DEFAULT_NODE_VERSION }}
+
+      - name: Update Dependencies
+        run: |
+          pnpm update
```
ZhongpinWang: I tried this out and it looks like this line will update not only transitive dependencies, but also our SDK dependencies (i.e. our `package.json` and our direct deps in lock file are changed). This workflow will then supersedes the normal dependabot update (mostly for minor and patch updates). 

I wonder if there is a better way to do this?

BrigittaK307: I have found 3 commands which might do what is expected, but I am not sure which of the commands is the best suitable in our case:
pnpm audit fix
pnpm install --lockfile-only
pnpm update --depth Infinity

ZhongpinWang: Thanks for investigating on this. Perhaps we can use `pnpm --no-save update` to update the lock file only (instead of also changing the package.json, which is kind like the job of dependabot). 

With this, we (unfortunately?) still update the direct dependencies in the lock file based on the range defined in the `package.json`. But I would argue if this is really a bad thing as

1. We anyway try to auto-update dependency PRs if pipeline works. So not a big deal to already try out the latest version of direct dependencies, even if dependabot hasn't updated it in `package.json` yet.
2. If something failed, at least we know something is not working (and should either see the same issue again in the separate dependabot PR, or some of the transitive dependencies are not compatible with the direct dependency). And this is actually good to see, as user will try to install the latest version as well. We should discover the issue early.

Besides, we can run this workflow monthly, so this won't really create too many trouble for us.

@deekshas8 @KavithaSiva Wdyt?

KavithaSiva: Thanks for the detailed reply, I also agree with using ` pnpm --no-save update`.

deekshas8: I am wondering if we should run `pnpm audit --audit-level moderate` as another step, to be sure no vulnerabilities exist?

KavithaSiva: Good idea, I will just check for `high` severity instead as we also do that in the dependency-check.

PR #1109, opened 2025-09-22
chore: Adjust masking type

deekshas8:
## Context

This change avoids a future potential breaking change when orchestration changes the masking module config by adding a discriminator.


packages/orchestration/src/orchestration-types.ts
```diff
@@ -120,13 +120,14 @@ export interface FilteringModule {
 /**
  * Representation of the `MaskingModuleConfig` schema.
  */
-export interface MaskingModule {
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
```
deekshas8, line 123: [info] This rule can be removed later when we introduce the discriminator

PR #1118, opened 2025-09-25
fix: JSDoc example of `buildAzureContentSafetyFilter()` function

ZhongpinWang:
## Context

Unfortunately, the JSDoc of this function was overlooked and was not changed to the correct usage.

### DIFF
--- a/.changeset/empty-wasps-lick.md
+++ b/.changeset/empty-wasps-lick.md
@@ -1,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': patch
+---
+
+[Fixed Issue] Fix JSDoc example of `buildAzureContentSafetyFilter()` function.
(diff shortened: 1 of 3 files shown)


.changeset/empty-wasps-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': patch
+---
+
+[Fixed Issue] Fix JSDoc example of `buildAzureContentSafetyFilter()` function.
```
KavithaSiva, line 5: [q] Do we really have to add a release note for this?

ZhongpinWang: I was also thinking but no strong opinion. I guess it is also kinda useful to tell people the previous JSDoc was wrong.

PR #1135, opened 2025-09-29
feat: Embedding client for orchestration

KavithaSiva:
## Context

### DIFF
--- a/packages/orchestration/src/orchestration-embedding-client.ts
+++ b/packages/orchestration/src/orchestration-embedding-client.ts
@@ -1,0 +1,141 @@
+import { executeRequest } from '@sap-ai-sdk/core';
+import { createLogger } from '@sap-cloud-sdk/util';
+import { getOrchestrationDeploymentId } from './deployment-resolver.js';
+import { OrchestrationEmbeddingResponse } from './orchestration-embedding-response.js';
+import type {
+  HttpResponse,
+  CustomRequestConfig
+} from '@sap-cloud-sdk/http-client';
+import type {
+  DeploymentIdConfig,
+  ResourceGroupConfig
+} from '@sap-ai-sdk/ai-api/internal.js';
+import type {
+  OrchestrationEmbeddingConfig,
+  EmbeddingRequest
+} from './orchestration-types.js';
+import type {
+  EmbeddingsPostRequest,
+  EmbeddingsOrchestrationConfig,
+  EmbeddingsModuleConfigs
+} from './client/api/schema/index.js';
+import type { HttpDestinationOrFetchOptions } from '@sap-cloud-sdk/connectivity';
+
+const logger = createLogger({
+  package: 'orchestration',
+  messageContext: 'orchestration-embedding-client'
+});
+
+/**
+ * Orchestration embedding client for generating embeddings with optional orchestration modules.
+ */
+export class OrchestrationEmbeddingClient {
+  /**
+   * Creates an instance of the orchestration embedding client.
+   * @param config - Orchestration embedding configuration.
+   * @param deploymentConfig - Deployment configuration.
+   * @param destination - The destination to use for the request.
+   */
+  constructor(
+    private config: OrchestrationEmbeddingConfig,
+    private deploymentConfig?: ResourceGroupConfig | DeploymentIdConfig,
+    private destination?: HttpDestinationOrFetchOptions
+  ) {}
+
+  /**
+   * Generate embeddings for the given input.
+   * @param request - Embedding request configuration.
+   * @param requestConfig - Custom request configuration.
+   * @returns Promise resolving to embedding response.
+   */
+  async embed(
+    request: EmbeddingRequest,
+    requestConfig?: CustomRequestConfig
+  ): Promise<OrchestrationEmbeddingResponse> {
+    logger.debug('Generating embeddings', {
+      inputType: typeof request.input,
+      inputLength: Array.isArray(request.input) ? request.input.length : 1,
+      type: request.type
+    });
+
+    const response = await this.executeRequest(request, requestConfig);
+    return new OrchestrationEmbeddingResponse(response);
+  }
+
+  private async executeRequest(
+    request: EmbeddingRequest,
+    requestConfig?: CustomRequestConfig
+  ): Promise<HttpResponse> {
+    const body = this.constructEmbeddingPostRequest(request);
+
+    const deploymentId = await getOrchestrationDeploymentId(
+      this.deploymentConfig ?? {},
+      this.destination
+    );
+
+    if (!deploymentId) {
+      throw new Error('Failed to resolve deployment ID');
+    }
+
+    logger.debug('Executing request', {
+      deploymentId,
+      endpoint: '/embeddings'
+    });
+
+    return executeRequest(
+      {
+        url: `/inference/deployments/${deploymentId}/v2/embeddings`,
+        ...(this.deploymentConfig ?? {})
+      },
+      body,
+      requestConfig,
+      this.destination
+    );
+  }
+
+  private constructEmbeddingPostRequest(
+    request: EmbeddingRequest
+  ): EmbeddingsPostRequest {
+    const orchestrationConfig: EmbeddingsOrchestrationConfig = {
+      modules: this.buildModulesConfig()
+    };
+
+    const embeddingRequest: EmbeddingsPostRequest = {
+      config: orchestrationConfig,
+      input: {
+        text: request.input,
+        ...(request.type && { type: request.type })
+      }
+    };
+
+    logger.debug('Constructed embedding request', {
+      hasEmbeddingsConfig: !!orchestrationConfig.modules.embeddings,
+      hasMaskingConfig: !!orchestrationConfig.modules.masking,
+      inputType: request.type
+    });
+
+    return embeddingRequest;
+  }
+
+  private buildModulesConfig(): EmbeddingsModuleConfigs {
+    const modules: EmbeddingsModuleConfigs = {
+      embeddings: {
+        model: {
+          name: this.config.modules.embeddings.model.name,
+          version: this.config.modules.embeddings.model.version ?? 'latest',
+          ...(this.config.modules.embeddings.model.params && {
+            params: this.config.modules.embeddings.model.params
+          })
+        }
+      }
+    };
+
+    if (this.config.modules.masking) {
+      modules.masking = {
+        masking_providers: this.config.modules.masking.masking_providers
+      };
+    }
+
+    return modules;
+  }
+}
--- a/packages/orchestration/src/orchestration-embedding-response.ts
+++ b/packages/orchestration/src/orchestration-embedding-response.ts
@@ -1,0 +1,91 @@
+
+import type { HttpResponse } from '@sap-cloud-sdk/http-client';
+import type {
+  EmbeddingsPostResponse,
+  EmbeddingResult,
+  EmbeddingsUsage,
+  ModuleResultsBase
+} from './client/api/schema/index.js';
+import type { EmbeddingData, EmbeddingUsageInfo } from './orchestration-types.js';
+
+/**
+ * Response wrapper for orchestration embedding requests.
+ */
+export class OrchestrationEmbeddingResponse {
+  private _data: EmbeddingsPostResponse;
+
+  constructor(private response: HttpResponse) {
+    this._data = response.data;
+  }
+
+  /**
+   * Request ID for the embedding operation.
+   * @returns The request ID string.
+   */
+  get requestId(): string {
+    return this._data.request_id;
+  }
+
+  /**
+   * Final embedding results.
+   * @returns Array of embedding data or undefined if no results.
+   */
+  get embeddings(): EmbeddingData[] | undefined {
+    if (!this._data.final_result?.data) {
+      return undefined;
+    }
+
+    return this._data.final_result.data.map((result: EmbeddingResult) => ({
+      embedding: result.embedding,
+      index: result.index
+    }));
+  }
+
+  /**
+   * Model used for generating embeddings.
+   * @returns The model name or undefined.
+   */
+  get model(): string | undefined {
+    return this._data.final_result?.model;
+  }
+
+  /**
+   * Usage information.
+   * @returns Usage information or undefined.
+   */
+  get usage(): EmbeddingUsageInfo | undefined {
+    if (!this._data.final_result?.usage) {
+      return undefined;
+    }
+
+    const usage: EmbeddingsUsage = this._data.final_result.usage;
+    return {
+      promptTokens: usage.prompt_tokens,
+      totalTokens: usage.total_tokens
+    };
+  }
+
+  /**
+   * Intermediate results from orchestration modules.
+   * @returns Intermediate results or undefined.
+   */
+  get intermediateResults(): ModuleResultsBase | undefined {
+    return this._data.intermediate_results;
+  }
+
+  /**
+   * Raw response data.
+   * @returns The raw embeddings post response data.
+   */
+  get data(): EmbeddingsPostResponse {
+    return this._data;
+  }
+
+  /**
+   * Raw HTTP response.
+   * @returns The raw HTTP response object.
+   */
+  get httpResponse(): HttpResponse {
+    return this.response;
+  }
+}
--- a/packages/orchestration/src/orchestration-types.ts
+++ b/packages/orchestration/src/orchestration-types.ts
@@ -23,7 +23,9 @@ import type {
   ErrorResponse,
   AzureContentSafetyInputFilterConfig,
   AzureContentSafetyOutputFilterConfig,
-  LlamaGuard38BFilterConfig
+  LlamaGuard38BFilterConfig,
+  EmbeddingsModelDetails as OriginalEmbeddingsModelDetails,
+  EmbeddingsModelParams as OriginalEmbeddingsModelParams
 } from './client/api/schema/index.js';
 
 /**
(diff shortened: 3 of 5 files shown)


packages/orchestration/src/orchestration-embedding-client.ts
```diff
@@ -0,0 +1,91 @@
...
+   * Generate embeddings for the given input.
+   * @param request - Embedding request configuration.
+   * @param requestConfig - Custom request configuration.
+   * @returns Promise resolving to embedding response.
+   */
+  async embed(
+    request: EmbeddingRequest,
+    requestConfig?: CustomRequestConfig
+  ): Promise<OrchestrationEmbeddingResponse> {
+    logger.debug('Generating embeddings', {
```
ZhongpinWang: Perhaps the debug is only for dev purpose? We can remove this kind of debug logs before merge as they are not adding too much value once the SDK is released. We should guarantee code correctness with tests.

packages/orchestration/src/orchestration-embedding-response.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  constructor(public readonly response: HttpResponse) {
+    this._data = response.data;
+  }
+
+  /**
+   * Final embedding results.
+   * @returns Array of embedding vectors, where each element is either a number array or a base-64 encoded string of the embedding.
+   */
+  getEmbeddingVectors(): (number[] | string)[] {
```
KavithaSiva: For now, I have allowed both response types, I have opened a discussion with the Orchestration team.

Assumption here is that the service will always only either have a list of float array or a list of encoded strings and not a mix of both, although the return types allow it.

KavithaSiva: On second thought, I don't think this is a good idea as there is also `binary` return type which can only be used while using `amazon--titan-embed-text`.

I think it's better if we either expose a new type for the embedding result and have dedicated convenience method for returning `number[]` and users are expected to access the raw response object if they need other response types like `binary` or `base64` encoded strings like Java.

WDY think @deekshas8 @ZhongpinWang

KavithaSiva: I have now added this suggestion instead in the PR.

packages/orchestration/src/util/module-config.ts
```diff
@@ -197,3 +202,58 @@ function isTemplate(
...
+    }
+  };
+
+  logger.debug('Constructed embedding request', {
+    hasEmbeddingsConfig: !!orchestrationConfig.modules.embeddings,
+    hasMaskingConfig: !!orchestrationConfig.modules.masking,
+    inputType: request.type
+  });
+
+  return embeddingRequest;
```
ZhongpinWang, line 232: [pp] Inline this and remove debug

packages/orchestration/src/util/module-config.ts
```diff
@@ -197,3 +202,58 @@ function isTemplate(
...
+        })
+      }
+    }
+  };
+
+  if (config.masking) {
+    modules.masking = {
+      masking_providers: config.masking.masking_providers
+    };
+  }
```
ZhongpinWang: [pp/req] Align the style with completion. Avoid having long property access like `config.embeddings.model.xyz`

```ts
const  { embeddings, masking } = config;
```

[req] And we need more than just `masking_providers` property or? Also we need to check length of keys.
```ts
{
  embeddings: { ... },
  ...(masking && Object.keys(masking).length && { masking }
}
```

packages/orchestration/src/util/module-config.ts
```diff
@@ -197,3 +202,58 @@ function isTemplate(
...
+  logger.debug('Constructed embedding request', {
+    hasEmbeddingsConfig: !!orchestrationConfig.modules.embeddings,
+    hasMaskingConfig: !!orchestrationConfig.modules.masking,
+    inputType: request.type
+  });
+
+  return embeddingRequest;
+}
+
+function buildEmbeddingModulesConfig(
```
ZhongpinWang, line 235: [pp] I like the idea of separating this part of logic out of construct request. Can we maybe do this for the existing chat completion construct request as well? So that we have the same structure for easier future maintenance purpose.

packages/orchestration/src/orchestration-embedding-client.test.ts
```diff
@@ -0,0 +1,304 @@
...
+    const clientWithResourceGroup = new OrchestrationEmbeddingClient(config, {
+      resourceGroup: 'custom-resource-group'
+    });
+
+    const response = await clientWithResourceGroup.embed(request);
+    expect(response).toBeInstanceOf(OrchestrationEmbeddingResponse);
+    expect(response._data.request_id).toBe('random-request-id');
+  });
+
+  it('handles HTTP error responses', async () => {
```
ZhongpinWang: [q] What is the point of this test? It is basically testing if Axios throws when response is 400. And since 400 is mocked, it is not relevant to the given `config`, `request` or `constructEmbeddingPostRequest()`. Can you clarify what is under test or we just remove this?

packages/orchestration/src/orchestration-embedding-client.ts
```diff
@@ -0,0 +1,91 @@
...
+    );
+
+    if (!deploymentId) {
+      throw new Error('Failed to resolve deployment ID');
+    }
+
+    logger.debug('Executing request', {
+      deploymentId,
+      endpoint: '/embeddings'
+    });
```
ZhongpinWang: [pp] This I personally don't mind to log. Perhaps we log the full `url` like line 83? But I am more in favour of removing it, as the deployment id resolve logic has been used by chat completion client for quite long. 
We should in general only log valuable, critical and necessary steps, which could be helpful when bugs come.

packages/orchestration/src/orchestration-embedding-response.test.ts
```diff
@@ -0,0 +1,304 @@
...
+
+    const response = new OrchestrationEmbeddingResponse(httpResponse);
+    const embeddings = response.getEmbeddingVectors();
+    expect(embeddings).toEqual([
+      [0.1, 0.2, 0.3, 0.4, 0.5],
+      [0.6, 0.7, 0.8, 0.9, 1.0]
+    ]);
+  });
+
+  it('should throw error for string embeddings for embedding vectors', () => {
```
ZhongpinWang:
```suggestion
  it('should throw error when using `getEmbeddingVectors()` for string embeddings', () => {
```

packages/orchestration/src/orchestration-embedding-response.ts
```diff
@@ -0,0 +1,62 @@
...
+  constructor(public readonly response: HttpResponse) {
+    this._data = response.data;
+  }
+
+  /**
+   * Final embedding results with index information.
+   * @returns Array of embedding data objects containing both vectors and indices.
+   */
+  getEmbeddings(): EmbeddingData[] {
+    return this._data.final_result!.data.map((result: EmbeddingResult) => ({
```
ZhongpinWang: [q] I guess the `!` will be removed later once orchestration releases their type fix right?

packages/orchestration/src/orchestration-embedding-response.ts
```diff
@@ -0,0 +1,62 @@
...
+   */
+  getEmbeddingVectors(): number[][] {
+    return this._data.final_result!.data.map((result: EmbeddingResult) => {
+      if (typeof result.embedding === 'string') {
+        throw new Error(
+          'String embeddings are not supported in getEmbeddingVectors(). Use getEmbeddings() instead.'
+        );
+      }
+      return result.embedding;
+    });
```
ZhongpinWang: [q] Are we sure that the `n-th` element in the array always has index `n`?

ZhongpinWang: If this is the case, then I guess for string response, we can also return a string[] instead of EmbeddingData[] with `embedding` and `index` as the property names of its element?

I would double check this and align the return type of both methods. If possible merge the two methods

- either all use `EmbeddingData[]`
- or use `(string | number[])[]`

packages/orchestration/src/orchestration-types.ts
```diff
@@ -506,3 +508,78 @@ export type LlamaGuard38BFilterReturnType<T extends ConfigType> =
...
+   * input: "This is a text to embed" or
+   * input: ["Text 1", "Text 2", "Text 3"]
+   */
+  input: string | string[];
+
+  /**
+   * Type of embedding input.
+   * @default 'text'
+   */
+  type?: 'text' | 'document' | 'query';
```
ZhongpinWang, line 545: [q] Do we know how this `type` is making a difference in orchestration service?
We should probably document the usage later for its usage.

ZhongpinWang (dismissed):
First round. Besides the comments listed below, I would also add type tests for this client.

packages/orchestration/src/orchestration-embedding-client.ts
KavithaSiva: Done.

packages/orchestration/src/util/module-config.ts
KavithaSiva: Done.

packages/orchestration/src/util/module-config.ts
KavithaSiva: Done.

packages/orchestration/src/orchestration-embedding-client.ts
KavithaSiva: Done.

packages/orchestration/src/orchestration-embedding-response.ts
KavithaSiva: Yes, they already merged it, will probably part of the next release, added Todos as reminders.

packages/orchestration/src/orchestration-embedding-response.ts
KavithaSiva: >[q] Are we sure that the n-th element in the array always has index n?

Yes, that's always the case, see: https://platform.openai.com/docs/api-reference/embeddings/object.

>I would double check this and align the return type of both methods. If possible merge the two methods
>either all use EmbeddingData[] or use (string | number[])[]

I discussed about this in the [thread above](https://github.com/SAP/ai-sdk-js/pull/1135#discussion_r2455432891), the response formats for now can be base64 encoded `string` or `float array` or `binary` which results in an array of 0 and 1. My idea of not using  `(string | number[])[]` was to not introduce any breaking changes if in the future the service adds a new return type.

IMO, having an endpoint to return just `number[]` (as this is probably the most frequently used method) and for all other use cases asking users to use `EmbeddingData[]` instead is cleaner and non-breaking. WDY think?

ZhongpinWang: I thought about this again and checked the OpenAI SDK [embeddings](https://github.com/openai/openai-node/blob/1b6c789b5faba94f310bc72ec5436f9e8419cc83/src/resources/embeddings.ts#L21C77-L21C100). They don't even provide any convenience for the embedding response.

I am not even sure if having a `number[][]` is a convenience in reality for the user considering the mapping logic is fairly simple and people might have already done that many times before with other SDKs. 

I know Java made the consumption that people might most frequently use `number[][]`. I doubt it as I have seen a lot of other users using base64 array with OpenAI official SDK. And nobody complained about not having a helper to get embedding objects mapped to array.. Having two methods with one of them only support number case and throw in the other case feels cumbersome and not necessary.

My suggestion would be removing the helper `getEmbeddingVectors()`. Also in the `EmbeddingData`, add the hardcoded `object: 'embedding'` back to [fully align with OpenAI official return value](https://github.com/openai/openai-node/blob/1b6c789b5faba94f310bc72ec5436f9e8419cc83/src/resources/embeddings.ts#L122).

packages/orchestration/src/orchestration-types.ts
KavithaSiva: No, I tried with different `type` values and the same `text` input and still got the same embeddings as result.
Will ask colleagues.

packages/orchestration/src/orchestration-embedding-response.ts
KavithaSiva: Done.

packages/orchestration/src/orchestration-types.ts
KavithaSiva: It refers to https://docs.cloud.google.com/vertex-ai/generative-ai/docs/embeddings/task-types. Basically controls what kind of search we do to obtain the embeddings.

KavithaSiva: Added some information in JSDoc, would be nice to map to the orchestration documentation once they document  more details.

packages/orchestration/src/orchestration-embedding-client.test.ts
KavithaSiva: The test asserts the error message thrown by the service instead now.

packages/orchestration/src/orchestration-embedding-response.ts
davidkna-sap: With the recent orchestration update, the `!` can be removed now.

packages/orchestration/src/orchestration-embedding-response.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  constructor(public readonly response: HttpResponse) {
+    this._data = response.data;
+  }
+
+  /**
+   * Final embedding results with index and object type (which is always `embedding`) information.
+   * @returns Array of embedding data objects containing both vectors, indices, and object types.
+   */
+  getEmbeddings(): EmbeddingData[] {
```
davidkna-sap, line 24: [np] `getEmbeddingsFloat` with throw-on-different-type could be useful here since most users may opt use the default `float` embedding

packages/orchestration/src/orchestration-embedding-response.ts
```diff
@@ -0,0 +1,47 @@
...
+      index: result.index,
+      object: 'embedding'
+    }));
+  }
+  /**
+   * Usage information.
+   * @returns Usage information or undefined.
+   */
+  getTokenUsage(): EmbeddingsUsage {
+    return this._data.final_result!.usage;
```
ZhongpinWang: Here same remove !

packages/orchestration/src/orchestration-embedding-client.test.ts
```diff
@@ -0,0 +1,317 @@
...
+      }
+    );
+
+    const response = await new OrchestrationEmbeddingClient(config).embed(
+      request
+    );
+
+    expect(response).toBeInstanceOf(OrchestrationEmbeddingResponse);
+    expect(response._data.request_id).toBe('random-request-id');
+    expect(response._data.final_result?.model).toBe('text-embedding-3-small');
```
ZhongpinWang: Here remove `?`

packages/orchestration/src/orchestration-embedding-client.test.ts
```diff
@@ -0,0 +1,317 @@
...
+      {
+        url: 'inference/deployments/1234/v2/embeddings'
+      }
+    );
+
+    try {
+      await new OrchestrationEmbeddingClient(config).embed(request);
+      fail('Expected an error to be thrown');
+    } catch (error: any) {
+      expect(error.cause?.response?.data?.error?.message).toBe(
```
ZhongpinWang: [pp] I would prefer using a snapshot.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -522,3 +524,82 @@ export type LlamaGuard38BFilterReturnType<T extends ConfigType> =
...
+ * Embedding module configuration.
+ */
+export interface EmbeddingModuleConfig {
+  /**
+   * Embeddings model configuration.
+   */
+  embeddings: EmbeddingModelConfig;
+
+  /**
+   * Optional masking module configuration.
```
ZhongpinWang:
```suggestion
   * Masking module configuration.
```

packages/orchestration/src/orchestration-types.ts
```diff
@@ -522,3 +524,82 @@ export type LlamaGuard38BFilterReturnType<T extends ConfigType> =
...
+ */
+export interface EmbeddingRequest {
+  /**
+   * Text input for which embeddings need to be generated.
+   * Can be a single string or array of strings.
+   * @example
+   * input: "This is a text to embed" or
+   * input: ["Text 1", "Text 2", "Text 3"]
+   */
+  input: string | string[];
```
davidkna-sap, line 539: [pp] empty lists and strings cause errors, so it could be useful to have users guard against empty lists.

```suggestion
  input: string | [string, ...string[]];
```

sample-code/src/orchestration.ts
```diff
@@ -763,3 +765,37 @@ export async function chatCompletionStreamWithTools(
     streamOptions
   );
 }
+
+/**
+ * Generate embeddings with masking for personal information.
+ * @returns The orchestration embedding response with masking applied.
+ */
+export async function orchestrationEmbeddingWithMasking(): Promise<OrchestrationEmbeddingResponse> {
```
davidkna-sap, line 773: [q] Should these functions be added to the e2e tests?

packages/orchestration/src/orchestration-embedding-response.ts
KavithaSiva: I initially implemented this, but after discussions, decided against it, refer [here](https://github.com/SAP/ai-sdk-js/pull/1135#discussion_r2469038147)

We can anyway add this later if needed, so I wouldn't consider this a blocker for merging this PR.

sample-code/src/orchestration.ts
KavithaSiva: Good catch.
I missed adding this in, done now.

packages/orchestration/src/orchestration-types.ts
KavithaSiva: I checked and orchestration service already has proper validation for empty lists and empty strings. So I wouldn't want to restrict this in the client's side.

ZhongpinWang (approved):

PR #1136, opened 2025-09-29
fix: Modified auto merge condition

BrigittaK307:
Fixed auto merge condition and changed settimeout to https://www.npmjs.com/package/async-retry
library.


.github/workflows/pnpm-update.yml
```diff
@@ -19,76 +19,102 @@ jobs:
...
+            git push --set-upstream origin "$BRANCH_NAME"
           else
             echo "No changes to commit"
-            exit 0
           fi
 
       - name: Create Pull Request
         if: success()
-        uses: actions/github-script@v8
+        uses: actions/github-script@v6
```
ZhongpinWang: [req] Why downgrading?

.github/workflows/pnpm-update.yml
```diff
@@ -19,76 +19,102 @@ jobs:
         run: |
           pnpm --no-save update
           pnpm audit --audit-level moderate
+          
           if [[ -n "$(git status --porcelain)" ]]; then
             git config --global user.email "cloud-sdk-js@github.com"
             git config --global user.name "cloud-sdk-js"
-            git checkout -b "bot/monthly-transitive-dependency-update-$(date +%Y-%m-%d)"
+            BRANCH_NAME="bot/monthly-transitive-dependency-update-$(date -u +%Y-%m-%d)"
```
ZhongpinWang: [pp] I would add hhmmss as well to avoid conflicts when triggering multiple in a day.

.github/workflows/pnpm-update.yml
```diff
@@ -19,76 +19,102 @@ jobs:
...
-        uses: actions/github-script@v8
+        uses: actions/github-script@v6
         with:
           github-token: ${{ secrets.GH_CLOUD_SDK_JS_ADMIN_WRITE_TOKEN }}
           script: |
             const { repo, owner } = context.repo;
+            const branch = `bot/monthly-transitive-dependency-update-${new Date().toISOString().split('T')[0]}`;
             const result = await github.rest.pulls.create({
-              title: 'chore: Monthly Transitive Dependency Update',
+              title: 'Monthly Transitive Dependency Update',
```
ZhongpinWang: [req] We need `chore: ` in front of the title to avoid pr lint workflow to fail. I think some old changes were lost when doing the merge.

.github/workflows/pnpm-update.yml
```diff
@@ -19,76 +19,102 @@ jobs:
...
+              head: branch,
               base: 'main',
               body: 'This is an automated PR to update pnpm transitive dependencies.'
             });
             console.log(`Pull Request created: ${result.data.html_url}`);
 
       - name: Auto-merge if checks pass
         if: success()
-        uses: actions/github-script@v8
+        uses: actions/github-script@v6
```
ZhongpinWang: Same

.github/workflows/pnpm-update.yml
```diff
@@ -19,76 +19,102 @@ jobs:
...
         with:
           github-token: ${{ secrets.GH_CLOUD_SDK_JS_ADMIN_WRITE_TOKEN }}
           script: |
+            const retry = require('async-retry');
```
ZhongpinWang, line 66: Nice 👍

.github/workflows/pnpm-update.yml
```diff
@@ -19,76 +19,102 @@ jobs:
...
             const { repo, owner } = context.repo;
+            const branch = `bot/monthly-transitive-dependency-update-${new Date().toISOString().split('T')[0]}`;
+
             const pulls = await github.rest.pulls.list({
               owner,
               repo,
-              head: `bot/monthly-transitive-dependency-update-${new Date().toISOString().split('T')[0]}`,
+              head: `${owner}:${branch}`,
```
ZhongpinWang: [q] Why do we need the owner?

.github/workflows/pnpm-update.yml
```diff
@@ -19,76 +19,102 @@ jobs:
...
+                const allAcceptable = checks.check_runs.every(run =>
+                  ['success', 'skipped', 'neutral'].includes(run.conclusion)
+                );
+
+                if (!allCompleted) {
+                  throw new Error(`PR #${pr.number} still has pending checks. Retrying...`);
+                }
+
+                if (!allAcceptable) {
+                  throw new Error(`PR #${pr.number} has failed checks that prevent merging.`);
```
ZhongpinWang: [req] In such case, retry won't change anything right? We can already return something like 'failed' here and let the workflow finish.

.github/workflows/pnpm-update.yml
```diff
@@ -19,76 +19,102 @@ jobs:
...
+                }
+
+                return 'success';
+              },
+              {
+                retries: 30,
+                factor: 1.5,
+                minTimeout: 10000,
+                maxTimeout: 60000,
```
ZhongpinWang:
```suggestion
                retries: 5,
                minTimeout: 2 * 60 * 1000, // Wait 2 min for the first time
                maxTimeout: 2 * 60 * 1000, // Wait max. 2 min between two retries
```

Retry 30 times is a bit too much 😄 I feel this workflow will run ~25 min or something.

With the suggested changes, we are sending request at

```
0s 2min 4min 6min 8min 10min
```

If we waited for more than 10 min, something must be wrong.

.github/workflows/pnpm-update.yml
```diff
@@ -19,76 +19,102 @@ jobs:
...
+                onRetry: (error, attempt) => {
+                  console.log(`Attempt ${attempt} failed: ${error.message}. Retrying...`);
```
ZhongpinWang: [pp] Would be nice if we print out the full time so that we see when it was executed

.github/workflows/pnpm-update.yml
```diff
@@ -19,76 +19,102 @@ jobs:
...
                 owner,
                 repo,
                 pull_number: pr.number,
                 merge_method: 'squash'
               });
-              console.log(`Pull Request #${pr.number} merged successfully`);
-            } else {
-              console.log(`Not all checks passed for PR #${pr.number}`);
+              console.log(`PR #${pr.number} merged successfully.`);
             }
```
ZhongpinWang, line 134: [pp] If `result` not equal to `success`? Any hint?

.github/workflows/pnpm-update.yml
ZhongpinWang: Log an error here. Don't throw. Throw will immediately jump out of the function and it will never reach the return.

.github/workflows/pnpm-update.yml
```diff
@@ -19,16 +19,17 @@ jobs:
         run: |
           pnpm --no-save update
           pnpm audit --audit-level moderate
+          
           if [[ -n "$(git status --porcelain)" ]]; then
             git config --global user.email "cloud-sdk-js@github.com"
             git config --global user.name "cloud-sdk-js"
-            git checkout -b "bot/monthly-transitive-dependency-update-$(date +%Y-%m-%d)"
+            BRANCH_NAME="bot/monthly-transitive-dependency-update-$(date -u +%Y-%m-%d-%H%M%S)"
```
ZhongpinWang:
```suggestion
            BRANCH_NAME="bot/monthly-transitive-dependency-update-$(date -u +%Y%m%d-%H%M%S)"
```

.github/workflows/pnpm-update.yml
```diff
@@ -38,11 +39,12 @@ jobs:
           github-token: ${{ secrets.GH_CLOUD_SDK_JS_ADMIN_WRITE_TOKEN }}
           script: |
             const { repo, owner } = context.repo;
+            const branch = `bot/monthly-transitive-dependency-update-${new Date().toISOString().split('T')[0]}`;
```
ZhongpinWang: Now we changed the branch name and included H M S, we should probably also change this line or?

.github/workflows/pnpm-update.yml
```diff
@@ -39,7 +39,14 @@ jobs:
...
           script: |
             const { repo, owner } = context.repo;
-            const branch = `bot/monthly-transitive-dependency-update-${new Date().toISOString().split('T')[0]}`;
+            const now = new Date();
+            const pad = num => num.toString().padStart(2, '0');
+            const formattedDate = `${now.getUTCFullYear()}${pad(
+              now.getUTCMonth() + 1
+            )}${pad(now.getUTCDate())}-${pad(now.getUTCHours())}${pad(
+              now.getUTCMinutes()
+            )}${pad(now.getUTCSeconds())}`;
```
ZhongpinWang, line 48: Oh true, plain JS does not have the date formatter. We used `moment.js` in our SDK but I guess as long as this works correctly we use it. Thanks for taking care of this.

BrigittaK307: I have tested it in plain JS compiler, and the formattedDate looks something like this: 20251002-102936

ZhongpinWang (commented):
One comment still not resolved.

ZhongpinWang (approved):
Thanks! Look good to me.

PR #1151, opened 2025-10-01
fix: Support AI core service key changes at runtime

### LINKED ISSUE #1150: changes of ai core service key are not supported at runtime
## Description
The `aiCoreServiceBinding` variable in **`packages/core/src/context.ts`** currently caches the first service binding resolution (whether from environment variables or service bindings) and never updates it. This means that if environment variables change at runtime, those changes are not reflected.

## Side Note Question
`aiCoreServiceBinding` is defined as a global variable, which raises the question: what is the reasoning behind making it global in the first place?

The main issue here is handling environment variable changes at runtime. A potential fix could be to improve the caching mechanism for this global variable - for example, by adding an option to clear/reset it.  

**Alternatively**, it may be possible to remove the cache/global variable entirely. This would simplify the implementation and avoid the complexity of a cache mechanism, since the performance cost is low and there are no API calls involved that would add noticeable overhead.

### Steps to Reproduce

1. update process.env["AICORE_SERVICE_KEY"] with service key: "service-key-1"
2. create orchestration client
3. call orchestrationClient.stream
4. update process.env["AICORE_SERVICE_KEY"] with different service key: "service-key-2"
5. call orchestrationClient.stream
6. it will fail because aiCoreServiceBinding is still using the old service key

### Expected Behavior

Changing service keys in runtime will be supported

### Screenshots

_No response_

### Used Versions

- Node version via `node -v`: v22.17.0
- NPM version via `npm -v: 10.9.2
- SAP Cloud SDK for AI version: 2.0.0
  - ...
  - ...

### Code Examples

_No response_

### Log File

_No response_

### Region

EU10

Production

### Impact

Blocked

### Timeline

_No response_

### Additional Context

_No response_

yuvalman:
## Context
Closes #1150

This PR removes the global variable aiCoreServiceBinding (previously used as a cache) in order to support runtime changes of the service key.  

The cache added unnecessary complexity and prevented updates when environment variables changed at runtime. Since there are no costly API calls involved, the performance impact of removing the cache is negligible, while improving flexibility and correctness.

### DIFF
--- a/.changeset/loose-banks-itch.md
+++ b/.changeset/loose-banks-itch.md
@@ -1,0 +1,5 @@
+---
+'@sap-ai-sdk/core': patch
+---
+
+Support AI core service key changes at runtime
--- a/packages/core/src/context.test.ts
+++ b/packages/core/src/context.test.ts
@@ -1,13 +1,36 @@
 import nock from 'nock';
 import { mockClientCredentialsGrantCall } from '../../../test-util/mock-http.js';
+import { dummyToken } from '../../../test-util/mock-jwt.js';
 import { getAiCoreDestination } from './context.js';
 
 describe('context', () => {
+  const originalEnv = process.env;
+
+  beforeEach(() => {
+    // Reset environment variables before each test
+    process.env = { ...originalEnv };
+    delete process.env.AICORE_SERVICE_KEY;
+
+    // Clean all nock interceptors
+    nock.cleanAll();
+  });
+
+  afterEach(() => {
+    process.env = originalEnv;
+  });
+
   afterAll(() => {
     nock.cleanAll();
   });
 
   it('should throw if client credentials are not fetched', async () => {
+    // Clean up any existing mocks first
+    nock.cleanAll();
+
+    // Don't set environment variable so it falls back to service binding
+    delete process.env.AICORE_SERVICE_KEY;
+
+    // Mock the authentication to return an error
     mockClientCredentialsGrantCall(
       {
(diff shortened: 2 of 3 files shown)


ZhongpinWang:
Hi @yuvalman ,

First of all, thanks for contributing. I am not entirely sure about the purpose of this behaviour change, like in which case will an env variable change at runtime. Environment variables normally doesn't change.

In case different AI Core instances are needed, it is always possible to [create a destination programmatically](https://sap.github.io/ai-sdk/docs/js/connecting-to-ai-core#creating-a-custom-destination) and use it when [initializing a client](https://sap.github.io/ai-sdk/docs/js/orchestration/chat-completion#custom-destination). Will this meet your needs?

yuvalman:
Thanks @ZhongpinWang,

Actually, passing the destination does meet my needs!

However, I encountered an issue when trying to use `registerDestination` as you suggested - I received a failure with status code 400 and the message "Missing Tenant Id" from the `executeHttpRequest` function in `packages/core/src/http-client.ts`.

Instead, I successfully built the destination using `import { transformServiceBindingToDestination } from "@sap-cloud-sdk/connectivity"`, which works as expected.

Is this approach considered a good practice, or would you recommend still using `registerDestination`? If `registerDestination` is preferred, do you have an example of how to register a destination that can be passed to the orchestration client and doesn't encounter the error I mentioned?

ZhongpinWang:
Then I will close this PR and the related issue. Feel free to reach out again for any further issues! 😊

PR #1161, opened 2025-10-06
chore: Update model list

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#399.

Update available model list to latest version:
Add `cohere--command-a-reasoning` and `mistralai--mistral-medium-instruct`, and remove `deepseek-ai--deepseek-r1`  from available model list.

### DIFF
--- a/packages/core/src/model-types.ts
+++ b/packages/core/src/model-types.ts
@@ -46,7 +46,6 @@ export type AwsBedrockChatModel = LiteralUnion<
   | 'anthropic--claude-3.7-sonnet'
   | 'anthropic--claude-4-opus'
   | 'anthropic--claude-4-sonnet'
-  | 'cohere--command-a-reasoning'
   | 'amazon--nova-pro'
   | 'amazon--nova-lite'
   | 'amazon--nova-micro'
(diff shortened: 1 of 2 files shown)


packages/core/src/model-types.ts
```diff
@@ -46,6 +46,7 @@ export type AwsBedrockChatModel = LiteralUnion<
   | 'anthropic--claude-3.7-sonnet'
   | 'anthropic--claude-4-opus'
   | 'anthropic--claude-4-sonnet'
+  | 'cohere--command-a-reasoning'
```
ZhongpinWang: [req] This one is not available in orchestration service and `AwsBedrockChatModel` is imported there. And since it cannot be used anywhere else either in our SDK, we should leave it out

packages/core/src/model-types.ts
```diff
@@ -58,6 +59,6 @@ export type AwsBedrockChatModel = LiteralUnion<
 export type AiCoreOpenSourceChatModel = LiteralUnion<
   | 'mistralai--mistral-large-instruct'
   | 'mistralai--mistral-small-instruct'
+  | 'mistralai--mistral-medium-instruct'
```
ZhongpinWang: [req] This is not available in orchestration

Issue #1145, opened 2025-09-30
Support Langchain MultiServerMCPClient


lemaiwo:
### Describe the Problem

Dear, Would it be possible to support LangChain MultiServerMCPClient in combination with SAP AI SDK? Langchain integration is supported but not the mcp server client, see documentation here: https://sap.github.io/ai-sdk/docs/js/langchain/orchestration

### Propose a Solution

Support MCP Server client for langchain in SAP AI SDK

### Affected Development Phase

Production

### Impact

Blocked

lemaiwo:
This is the response we always get: 

<img width="526" height="639" alt="Image" src="https://github.com/user-attachments/assets/9b0f42cd-95db-45ba-97b6-8b223fdaf2bf" />

It doesn't seem to work with langchain in general

lemaiwo:
The same mcp with langchain and anthropic llm does work not using SAP AI SDK

lemaiwo:
It does seem to work when using AzureOpenAiChatClient

lemaiwo:
Thank you for the clarification! It makes more sense now!

One more question, would it be possible to combine SAP AI SDK with langgraph function createReactAgent? This isn't working for me and confused me.

ZhongpinWang:
@lemaiwo You are welcome. 😄Regarding `createReactAgent` could you please provide us some logs or analysis showing which part is not working? I will add a ticket in our backlog to try it out.

ZhongpinWang:
In the meanwhile, I see that in the next 1.0 version of LangChain, `createReactAgent()` will be replaced by `createAgent()` https://docs.langchain.com/oss/javascript/releases/langchain-v1#3-agent-migration. Have you tried this one out?

PR #1162, opened 2025-10-07
chore: Use `zod/v3` in Agent tutorial and add plain LangChain MCP sample code

ZhongpinWang:
## Context

Relates to SAP/ai-sdk-js#1145.

Relates to SAP/ai-sdk-js-backlog#374

The first part is about switching from using `zod/v4` to `zod/v3` with `zod@3.x` package in `sample-code/src/tutorials/mcp/weather-mcp-server.ts` since `@modelcontextprotocol/sdk` does not even support `zod/v4` from `zod@3.x` due to the outdated zod dependency.

`modelcontextprotocol` typescript sdk is working on supporting `zod@4.x` package directly. See https://github.com/modelcontextprotocol/typescript-sdk/pull/869.

The second part is to add a new sample-code show-casing how to use LangChain's MCP adapter with pure LangChain `bindTools()` without LangGraph. (as requested in SAP/ai-sdk-js#1145).

### DIFF
--- a/sample-code/src/langchain-orchestration.ts
+++ b/sample-code/src/langchain-orchestration.ts
@@ -1,5 +1,5 @@
 import { StringOutputParser } from '@langchain/core/output_parsers';
-import { AzureOpenAiChatClient, OrchestrationClient } from '@sap-ai-sdk/langchain';
+import { OrchestrationClient } from '@sap-ai-sdk/langchain';
 import {
   buildAzureContentSafetyFilter,
   buildDpiMaskingProvider,
@@ -21,6 +21,7 @@ import {
 } from '@langchain/core/messages';
 // eslint-disable-next-line import/no-internal-modules
 import * as z from 'zod/v4';
+// eslint-disable-next-line import/no-internal-modules
 import { getMcpTools } from './tutorials/mcp/mcp-adapter.js';
 import type { BaseMessage, AIMessageChunk } from '@langchain/core/messages';
 import type { LangChainOrchestrationModuleConfig } from '@sap-ai-sdk/langchain';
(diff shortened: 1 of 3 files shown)


sample-code/src/tutorials/mcp/weather-mcp-server.ts
```diff
@@ -3,7 +3,7 @@
 /* eslint-disable import/no-internal-modules */
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
 import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
-import * as z from 'zod/v4';
+import * as z from 'zod/v3';
```
ZhongpinWang, line 6: https://github.com/modelcontextprotocol/typescript-sdk/blob/e0de0829019a4eab7af29c05f9a7ec13364f121e/package.json#L75 This zod version is outdated and does not support `zod/v4` feature and is not compatible with our latest `zod@3.x` package.

sample-code/src/langchain-orchestration.ts
```diff
@@ -336,3 +338,55 @@ export async function invokeToolChain(): Promise<string> {
...
+    { maxRetries: 0 }
+  );
+
+  const tools = [...getMcpTools];
+
+  const messages: BaseMessage[] = [
+    new HumanMessage('What is the weather like in Berlin?')
+  ];
+
+  const response = await client.bindTools(tools).invoke(messages);
```
deekshas8, line 364: [pp] We could just use `getMcpTools` here since it returns an array anyway, no point creating a copy of the same array

ZhongpinWang: True, initially I was doing this, but I need to access `tools[0]` later again to invoke the actual function, thus, I extracted this out, otherwise we call `getMcpTools()` twice.

PR #1188, opened 2025-10-15
chore: Update publish.yml

deekshas8:
## Context

Relates https://github.com/SAP/ai-sdk-js-backlog/issues/410

Make necessary updates t support Trusted Publishing. Remove the npmjs token.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -7,9 +7,6 @@ on:
 env:
   DOCS_REPO: SAP/ai-sdk
 
-permissions:
-  id-token: write
-
 jobs:
   check-release-notes-pr:
     name: Check Release Notes PR Status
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -7,6 +7,9 @@ on:
 env:
   DOCS_REPO: SAP/ai-sdk
 
+permissions:
+  id-token: write
+
```
davidkna-sap: Perhaps move this into `publish-npm:`

deekshas8: From what I understood, the workflow needs to have the permission, not just the job

davidkna-sap: Okay, that may be safer. That said, [quite a few projects](https://github.com/search?q=id-token%3A%20write%20path%3A.github%2Fworkflows%2F*.yaml&type=code) configure this for OIDC at the job level.

deekshas8: I moved it. If this works, then it's good enough to have it in the job.

PR #1193, opened 2025-10-17
chore: Add repository.url in `package.json`

deekshas8:
## Context

NPM Publish fails with:
```
PUT https://registry.npmjs.org/@sap-ai-sdk%2flangchain - Error verifying sigstore provenance bundle: Failed to validate repository information: package.json: "repository.url" is "", expected to match "https://github.com/SAP/ai-sdk-js" from provenance
```

Adding relevant information to all packages.

### DIFF
--- a/packages/ai-api/package.json
+++ b/packages/ai-api/package.json
@@ -7,7 +7,7 @@
     "sap-ai-sdk",
     "ai-api"
   ],
-"repository": {
+  "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


ZhongpinWang (approved):
[q] Why do we need this?

deekshas8:
> [q] Why do we need this?

Basically additional checks for publishers. Trusted publishing has automatic provenance attestation (where the code was published, if authorized user published a package, etc). For it to work, this config should be part of package.json

PR #1218, opened 2025-10-22
fix: Github action pnpm update

BrigittaK307:
Because of some yet unsolved vulnerabilities in the `validator` package version currently in use, the command `pnpm audit --audit-level moderate` was failing and because of this, the whole pipeline as well. This is why the change into `pnpm audit --audit-level high` was neccessary.


.github/workflows/pnpm-update.yml
```diff
@@ -57,22 +54,19 @@ jobs:
             });
             console.log(`Pull Request created: ${result.data.html_url}`);
 
+      - name: Install async-retry for auto-merge
+        if: steps.update-deps.outputs.branch-created == 'true'
+        run: pnpm install async-retry -w
```
deekshas8, line 56: I didn't get why we need this for auto-merge?

BrigittaK307: If used without pnpm install, the running of the github action gave me an error.
The usage of the `async-retry` package was not neccessary, but was recommended by Zhongpin for me.

ZhongpinWang: This was used for checking the PR status with retry before merging in the script below.

PR #1230, opened 2025-10-24
chore: Map the parameters for reasoning models

emincihangeri:
## Context

Closes SAP/ai-sdk-js-backlog#415.


packages/langchain/src/openai/util.ts
```diff
@@ -297,7 +315,17 @@ export function mapLangChainToAiClient(
...
+
+  if (isReasoningModel(client.modelName)) {
+    params.max_completion_tokens =
+      client.max_tokens === -1 ? undefined : client.max_tokens;
+  } else {
+    params.max_tokens =
+      client.max_tokens === -1 ? undefined : client.max_tokens;
+  }
+
+  return removeUndefinedProperties<AzureOpenAiChatCompletionParameters>(params);
```
davidkna-sap, line 337: [pp] I would usually prefer a test for something like this. Though if we decided to map `max_tokens` to `max_completion_tokens` in every case, existing tests should handle this. (orchestration `gpt-4o` etc. supports it)

sample-code/src/langchain-azure-openai.ts
```diff
@@ -42,6 +42,33 @@ export async function invoke(): Promise<string> {
...
 }
 
+/**
+ * Ask GPT about the capital of France.
+ * @returns The answer from GPT-o3.
+ */
+export async function invokeReasoning(): Promise<string> {
+  // initialize client with options
+  const client = new AzureOpenAiChatClient({
+    modelName: 'o3',
```
davidkna-sap: [pp] Let's use one of the faster models for this test to avoid slowing things down:

```suggestion
    modelName: 'gpt-5-mini',
```

sample-code/src/langchain-azure-openai.ts
```diff
@@ -42,6 +42,33 @@ export async function invoke(): Promise<string> {
   return parser.invoke(response);
 }
 
+/**
+ * Ask GPT about the capital of France.
+ * @returns The answer from GPT-o3.
+ */
+export async function invokeReasoning(): Promise<string> {
```
davidkna-sap: [pp] To make purpose of the test more clear:
```suggestion
export async function invokeReasoningWithMaxTokens(): Promise<string> {
```

davidkna-sap:
[E2E](https://github.com/SAP/ai-sdk-js/actions/runs/19891908224/job/57013107733) for 97db0d6097662ac2e798969744f7d4c1e86a6960 is green

Issue #1249, opened 2025-10-30
Huge error messages get pushed up the stack in orchestration mode


brambar5:
### Describe the Bug

The SAP AI SDK doesn't handle large error payloads gracefully in *orchestration mode*. When API calls fail, the entire conversation history, request payload, response data, and error context get serialized into a single error message string.

In many case I get errors that are 8MB big. So in error.message I see a json string which I could parse to get the real error but this is slow with such amount of data. Is it really useful that the API, or even the AI Core service sends back the complete conversation back?

### Steps to Reproduce

Just setup orchestration mode and send a message big enough to overflow the context of the LLM

### Expected Behavior

I don't need the complete conversation history in the response. I first want to see the error message like:
Context too long for the LLM

### Used Versions

- Node version via `node -v`: ...
- NPM version via `npm -v: ...
- SAP Cloud SDK for AI version:
  - ...
  - ...

### Region

EU10 Canary

### Affected Development Phase

Development

### Impact

Impaired

ZhongpinWang:
We merged the fix #1250 and it should be available in our next patch release. Thanks again for pointing this out.

brambar5:
Please still reconsider if the server should really send back the complete just received context as part of the error response or there are other ways to indicate what was received. Let me know if I should report that somewhere else.

Thank you for improving the logging!

PR #1250, opened 2025-10-30
fix: Remove active logging during SSE streaming by throwing errors

ZhongpinWang:
## Context

Closes SAP/ai-sdk-js-backlog#419.

Relates to user issue #1249.

We should not log error payload during SSE streaming. Instead, we should throw the error and let user handle it.

### DIFF
--- a/.changeset/fluffy-apples-make.md
+++ b/.changeset/fluffy-apples-make.md
@@ -1,0 +1,5 @@
+---
+'@sap-ai-sdk/core': patch
+---
+
+[Fixed Issue] Replace active logging during streaming with error throwing to avoid logging the response payload.


packages/core/src/stream/sse-stream.ts
```diff
@@ -44,16 +44,11 @@ export class SseStream<Item> implements AsyncIterable<Item> {
...
-            throw e;
+            throw new ErrorWithCause('Could not parse message into JSON', e);
           }
 
           if (data?.error) {
-            logger.error(
-              `Error received from server: ${JSON.stringify(data.error)}`
-            );
-            throw new Error(JSON.stringify(data.error));
+            throw new Error(`Error received from the server.\n${JSON.stringify(data.error)}`);
```
ZhongpinWang: I can only throw error here, because there is no error thrown by anything else, different from the case when AxiosError is thrown by Axios, we can directly pass the Axios error to ErrorWithCause.

deekshas8 (approved):
LGTM

PR #1262, opened 2025-11-05
chore: Update to new orchestration release (2510a)

BrigittaK307:
Updated orchestration release to [https://github.tools.sap/AI/llm-orchestration/releases/tag/v0.105.0](url)


BrigittaK307:
Had to modify the orchestration package.json `postgenerate` script in order to run it on Windows. The command I used on Windows is: `node -e "import { readdirSync, unlinkSync } from 'fs'; import { join } from 'path'; const dir = './src/client/api'; readdirSync(dir).filter(f => f.endsWith('.ts')).forEach(f => unlinkSync(join(dir, f)));"
`

packages/orchestration/src/util/filtering.ts
ZhongpinWang, line 1: I changed this just to align the code with translation module.

packages/orchestration/src/util/translation.ts
```diff
@@ -17,13 +17,29 @@ export function buildTranslationConfig<T extends ConfigType>(
...
+  const { sourceLanguage, targetLanguage } = config;
+  const baseConfig = {
+    ...(sourceLanguage && {
+      source_language: sourceLanguage
+    }),
+    target_language: targetLanguage
+  };
+
+  if (type === 'input') {
+    const { applyTo } = config as TranslationConfigParams<'input'>;
```
ZhongpinWang: [req] If you already want to support `applyTo`, please add a changeset and create a documentation PR, write exactly how this should be used. 

Otherwise, revert the changes + create follow-up BLI.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -432,17 +434,37 @@ interface TranslationConfigParameters {
...
+  /**
+   * Language of the text to be translated.
+   * @example sourceLanguage: 'de-DE'
+   */
+  sourceLanguage?: string;
+  /**
+   * Language to which the text should be translated.
+   * @example targetLanguage: 'en-US'
+   */
+  targetLanguage: SAPDocumentTranslationOutputTargetLanguage;
```
ZhongpinWang: [req] Same, if you would like to support the selector in this PR already, changeset + documentation.

Otherwise, change it back to string + create follow-up BLI.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -418,10 +420,31 @@ export type LlamaGuard38BCategory = keyof LlamaGuard38B;
...
+  sourceLanguage?: string;
+  /**
+   * Language to which the text should be translated.
+   * @example targetLanguage: 'en-US'
+   */
+  targetLanguage: string;
+  /**
+   * List of selectors to apply translation to.
+   */
+  applyTo?: SAPDocumentTranslationApplyToSelector[];
```
ZhongpinWang: [req] If you decided not to support the new feature, then this needs to be removed as well. Otherwise, user can still define `applyTo` when using the build function.

packages/orchestration/src/util/translation.ts
BrigittaK307: Reverted changes, will apply the new ones in another PR

packages/orchestration/src/orchestration-types.ts
```diff
@@ -418,10 +419,27 @@ export type LlamaGuard38BCategory = keyof LlamaGuard38B;
 export type ConfigType = 'input' | 'output';
 
 /**
- * Parameters for translation configuration.
+ * Input parameters for translation configuration.
  * @internal
```
ZhongpinWang:
```suggestion
```
`@internal` is only for exported things.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -418,10 +419,26 @@ export type LlamaGuard38BCategory = keyof LlamaGuard38B;
...
+  /**
+   * Language to which the text should be translated.
+   * @example targetLanguage: 'en-US'
+   */
+  targetLanguage: string;
+}
+
+/**
+ * Output parameters for translation configuration.
  * @internal
```
ZhongpinWang:
```suggestion
```

PR #1269, opened 2025-11-10
feat: Update document-grounding specification

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#408
Closes SAP/ai-sdk-js#951

Based on SAP/ai-sdk-js#951, only adding a changeset.


.changeset/orange-shrimps-poke.md
```diff
@@ -0,0 +1,24 @@
+---
```
ZhongpinWang, line 1: Due to the limitation of the release workflow, one changeset should only contain one change, and do not include empty lines (this I am not sure if I remembered correctly, but ...)

ZhongpinWang (commented):
One concern about the the changeset files. Otherwise lgtm

ZhongpinWang (approved):
Thanks for providing a list of all compatibility issues.

PR #1271, opened 2025-11-11
feat: Support auto-streaming with langchain

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#344.

This PR adds basic support for passing `streaming: true` to the constructors and saving it as an attribute. Results are collected via `.concat()`. In addition this PR updates the unit tests, sample code and e2e tests.


packages/langchain/src/orchestration/client.ts
```diff
@@ -54,6 +56,10 @@ export class OrchestrationClient extends BaseChatModel<
     };
 
     super(langchainOptions);
+
+    // Todo: Extend BaseChatModelParams?
+    this.streaming =
+      (langchainOptions as { streaming?: boolean })?.streaming ?? false;
```
davidkna-sap: The `streaming`-option is supported here, but I felt like extending or replacing `BaseChatModelParams` might be considered a breaking change which leads to any use needing to avoid type-checks.

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -513,4 +711,63 @@ describe('orchestration service client', () => {
...
+      // langgraph will only enable streaming in a granular streaming mode
+      { streamMode: 'messages' as const }
+    );
+
+    let finalOutput;
+    for await (const chunk of stream) {
+      finalOutput =
+        finalOutput !== undefined ? finalOutput.concat(chunk) : chunk;
+    }
+
```
davidkna-sap, line 693: A snapshot test directly on the final output is not possible here, because langgraph will add uuids to the response object, but other tests should already be verifying the general execution with auto-streaming enabled.

packages/langchain/src/openai/chat.ts
```diff
@@ -55,6 +55,8 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
   max_tokens?: number;
   supportsStrictToolCalling?: boolean;
   modelName: string;
+  streaming: boolean;
+  disableStreaming: boolean;
```
ZhongpinWang: [req] Make `disableStreaming` optional https://github.com/langchain-ai/langchainjs/blob/cc502e1b67dbadcd123a7ea2964c791c2bbad581/libs/langchain-core/src/language_models/chat_models.ts#L101

[info] LangChain default it to false if not set https://github.com/langchain-ai/langchainjs/blob/cc502e1b67dbadcd123a7ea2964c791c2bbad581/libs/langchain-core/src/language_models/chat_models.ts#L220

packages/langchain/src/openai/chat.ts
```diff
@@ -55,6 +55,8 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
   max_tokens?: number;
   supportsStrictToolCalling?: boolean;
   modelName: string;
+  streaming: boolean;
```
ZhongpinWang: [req] Default `streaming` to `false` https://github.com/langchain-ai/langchainjs/blob/cc502e1b67dbadcd123a7ea2964c791c2bbad581/libs/providers/langchain-openai/src/llms.ts#L127C3-L127C21

packages/langchain/src/openai/chat.ts
```diff
@@ -72,6 +74,13 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
     this.presence_penalty = fields.presence_penalty;
     this.frequency_penalty = fields.frequency_penalty;
     this.max_tokens = fields.max_tokens;
+    this.disableStreaming = fields.disableStreaming ?? false;
+    // If both streaming and disableStreaming are set, disable streaming
+    this.streaming = (fields.streaming ?? false) && !this.disableStreaming;
+    // if streaming is false, streaming is disabled
+    if (fields?.streaming === false) {
+      this.disableStreaming = true;
+    }
```
ZhongpinWang: [req] Consider `disableStreaming` and `streaming` both optional.

packages/langchain/src/openai/chat.ts
```diff
@@ -72,6 +74,13 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
     this.presence_penalty = fields.presence_penalty;
     this.frequency_penalty = fields.frequency_penalty;
     this.max_tokens = fields.max_tokens;
+    this.disableStreaming = fields.disableStreaming ?? false;
+    // If both streaming and disableStreaming are set, disable streaming
```
ZhongpinWang:
```suggestion
    // Enable streaming only when `streaming` is `true` (default `false`) and `disableStreaming` is `false` (default `false`).
```
[pp] Perhaps this is more clear than saying `are set`.

packages/langchain/src/openai/chat.ts
```diff
@@ -86,6 +95,20 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
     options: typeof this.ParsedCallOptions,
     runManager?: CallbackManagerForLLMRun
   ): Promise<ChatResult> {
+    // If both streaming and disableStreaming are set, disable streaming
+    if ((options?.stream ?? this.streaming) && !this.disableStreaming) {
```
ZhongpinWang: [req] I would follow the LangChain OpenAI implementation and at this position only check `streaming`

Check `params.stream`
https://github.com/langchain-ai/langchainjs/blob/cc502e1b67dbadcd123a7ea2964c791c2bbad581/libs/providers/langchain-openai/src/llms.ts#L291

Get `params` from `invocationParams()`
https://github.com/langchain-ai/langchainjs/blob/cc502e1b67dbadcd123a7ea2964c791c2bbad581/libs/providers/langchain-openai/src/llms.ts#L275

In `invocationParams()`, `stream` is equal to `this.streaming`
https://github.com/langchain-ai/langchainjs/blob/cc502e1b67dbadcd123a7ea2964c791c2bbad581/libs/providers/langchain-openai/src/llms.ts#L225

ZhongpinWang: I am a bit confused, why are we allowing passing `stream` in the option? LangChain official OpenAI doesn't seem to allow this? This should only be set when instantiating the client?

packages/langchain/src/openai/types.ts
```diff
@@ -63,6 +70,7 @@ export type AzureOpenAiChatCallOptions = BaseChatModelCallOptions &
     | 'function_call'
   > & {
     strict?: boolean;
+    stream?: boolean;
```
ZhongpinWang: [q/req] I might be wrong, but please double check if passing `stream` to a invoke call is possible. I feel it is only possible when setting `streaming` in the constructor. 

So if client has `streaming` to `true`, then `invoke()` / `_generate()` redirects to streaming. And vice versa.

sample-code/src/langchain-orchestration.ts
```diff
@@ -50,6 +50,33 @@ export async function invokeChain(): Promise<string> {
...
+  } as any)
+    .pipe(new StringOutputParser())
+    .invoke([
+      {
+        role: 'user',
+        content: 'Tell me about SAP Cloud SDK'
+      }
+    ]);
+}
+
```
ZhongpinWang: [pp] Honestly, since this is LangChain feature, as long as we have tests, I would not show a sample code, also because setting `streaming` explicity to `true` and call `invoke()` instead of `streaming()` is a bit confusing. This property is more meant to be used in auto-streaming.

sample-code/src/langchain-orchestration.ts
```diff
@@ -144,7 +167,15 @@ export async function invokeLangGraphChain(): Promise<string> {
 
   // Add memory
   const memory = new MemorySaver();
-  const app = workflow.compile({ checkpointer: memory });
+  return workflow.compile({ checkpointer: memory });
+}
+
+/**
+ * Invoke the model with memory.
+ * @returns The answer from ChatGPT.
```
ZhongpinWang: [pp] ChatGPT is nowadays either a chatbot or the specific model announced in 2022. Here it is actually answer from a LangGraph workflow.
```suggestion
 * @returns The answer from LangGraph workflow.
```

sample-code/src/langchain-orchestration.ts
```diff
@@ -169,6 +200,54 @@ export async function invokeLangGraphChain(): Promise<string> {
   return `${firstResponse}\n\n${secondResponse}`;
 }
 
+/**
+ * Stream responses using LangGraph Orchestration client.
+ * @returns The answer from ChatGPT.
```
ZhongpinWang: Same

sample-code/src/langchain-orchestration.ts
```diff
@@ -144,7 +167,15 @@ export async function invokeLangGraphChain(): Promise<string> {
...
-  const app = workflow.compile({ checkpointer: memory });
+  return workflow.compile({ checkpointer: memory });
+}
+
+/**
+ * Invoke the model with memory.
+ * @returns The answer from ChatGPT.
+ */
+export async function invokeLangGraphChain(): Promise<string> {
+  const app = createLangGraphApp();
```
ZhongpinWang, line 151: 👍Good that we have this part extracted.

packages/langchain/src/openai/chat.ts
davidkna-sap: Removed `disableStreaming`, because base class provides it.

packages/langchain/src/openai/chat.ts
davidkna-sap: Added, though i amended the comment a bit.

packages/langchain/src/openai/chat.ts
davidkna-sap: Not completely sure what you had in mind, I hope this works for you.

packages/langchain/src/openai/chat.ts
davidkna-sap: Adding `stream` was a mistake, I've removed it. I also handled `disableStreaming` to preemptively avoid any issues, because `disableStreaming` disable streaming completely (`stream` is redirected to `invoke`).

packages/langchain/src/openai/types.ts
davidkna-sap: I re-investigated this and decided to remove the `stream` option handling.

packages/langchain/src/openai/types.ts
```diff
@@ -35,6 +35,18 @@ export type AzureOpenAiChatModelParams = Pick<
...
+   * If {@link disableStreaming} is set to `true`, this option will be ignored.
+   * If {@link streaming} is explicitly set to `false`, {@link disableStreaming} will be set to `true`.
+   */
+  streaming?: boolean;
+  /**
+   * Whether streaming should be disabled.
+   * If `true`, `stream()` calls will be redirected to `invoke()`.
+   * Setting `streaming: false` will automatically set this to `true`.
+   */
+  disableStreaming?: boolean;
```
ZhongpinWang: [req] Do you still need this since it is part of `BaseChatModelParams` which is unioned in the next line?

https://github.com/langchain-ai/langchainjs/blob/cc502e1b67dbadcd123a7ea2964c791c2bbad581/libs/langchain-core/src/language_models/chat_models.ts#L101

packages/langchain/src/openai/chat.ts
```diff
@@ -72,6 +73,19 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
     this.presence_penalty = fields.presence_penalty;
     this.frequency_penalty = fields.frequency_penalty;
     this.max_tokens = fields.max_tokens;
+    // Initialize streaming flags with LangChain-compatible behavior:
+    // - `streaming`: true enables auto-streaming in `invoke()` calls
+    // - `disableStreaming`: true overrides streaming flag
+    // - `streaming`: `false` causes `disableStreaming` to be set to `true` for framework compatibility
+    this.disableStreaming = fields.disableStreaming ?? this.disableStreaming;
```
ZhongpinWang: [req] The difference between your implementation and https://github.com/langchain-ai/langchainjs/blob/cc502e1b67dbadcd123a7ea2964c791c2bbad581/libs/providers/langchain-openai/src/chat_models/base.ts#L479 is that openai **always set** the value to `false` if not **strictly** equal `true`. Your solution using `??` might be problematic if user writes plain JS and anything such as `''`, `0` other than `undefined` is assigned to the left.

Also your left can potentially be `undefined` if it was previously `undefined`. And later in line 82, if the `if` condition does not apply, `this.disableStreaming` will be `undefined` forever, which is different from the OpenAI behaviour. (Although https://github.com/langchain-ai/langchainjs/blob/cc502e1b67dbadcd123a7ea2964c791c2bbad581/libs/langchain-core/src/language_models/chat_models.ts#L220 here it is set to `false` by default by LangChain, so not a real issue but hope you see my point about yours and OpenAI's implementation technically being different. Be careful about `??` in TS and I personally find `bar = foo === true` a smart way for making sure `bar` will either be `true` or `false` and also excluding all other truthy values).

packages/langchain/src/openai/chat.ts
```diff
@@ -72,6 +73,19 @@ export class AzureOpenAiChatClient extends BaseChatModel<AzureOpenAiChatCallOpti
...
+    // if streaming is explicitly false, streaming is disabled
+    if (fields?.streaming === false) {
+      this.disableStreaming = true;
+    }
+    // Enable streaming only when `streaming` is `true` (default `false`) and `disableStreaming` is not `true` (default `undefined`).
+    this.streaming =
+      (fields?.streaming ?? this.streaming) && this.disableStreaming !== true;
```
ZhongpinWang: [pp] Here similarly, truthy value of `fields?.streaming` might be a problem in JS

sample-code/src/langchain-azure-openai.ts
```diff
@@ -71,6 +71,29 @@ export async function invokeReasoningWithMaxTokens(): Promise<string> {
...
+ * Ask GPT about the capital of France, using streaming internally.
+ * @returns The answer from GPT.
+ */
+export async function invokeWithStreaming(): Promise<string> {
+  // initialize client with options
+  const client = new AzureOpenAiChatClient({
+    modelName: 'gpt-4o',
+    max_tokens: 1000,
+    temperature: 0.7,
+    streaming: true
```
ZhongpinWang: [pp] Just still feel maybe we don't need this sample code for show-casing this feature. We shouldn't create sample code for just turning on one property and the rest remains the same and internal remains invisible. Not helpful for users IMO.

packages/langchain/src/openai/chat.test.ts
```diff
@@ -332,6 +338,160 @@ describe('Chat client', () => {
...
+            }
+          }
+        },
+        {
+          data: mockResponseStream,
+          status: 200
+        },
+        endpoint
+      );
+      jest.spyOn(AzureOpenAiChatClient.prototype, '_streamResponseChunks');
```
ZhongpinWang, line 363: Relevant to my removing sample-code comment, this unit test is good enough. For user, we maybe need to add documentation, or link it to LangChain's documentation since we did the same for other LangChain default behavioursl.

packages/langchain/src/orchestration/client.ts
```diff
@@ -54,6 +57,24 @@ export class OrchestrationClient extends BaseChatModel<
...
+    // If streaming is explicitly false, streaming is disabled
+    if ((langchainOptions as { streaming?: boolean })?.streaming === false) {
+      this.disableStreaming = true;
+    }
+
+    // Enable streaming only when `streaming` is `true` (default `false`) and `disableStreaming` is not `true` (default `undefined`).
+    this.streaming =
+      ((langchainOptions as { streaming?: boolean })?.streaming ??
+        this.streaming) &&
+      this.disableStreaming !== true;
```
ZhongpinWang: See my comment for the LangChain OpenAI client above.

ZhongpinWang (commented):
Thanks for the changes. 👍 I added some more comments.

packages/langchain/src/openai/chat.ts
ZhongpinWang: See my explanation in the latest review round.

packages/langchain/src/openai/chat.ts
davidkna-sap: I replaced all `??` instances in the relevant code with boolean equality checks.

packages/langchain/src/openai/chat.test.ts
```diff
@@ -332,6 +338,160 @@ describe('Chat client', () => {
...
+
+      client.streaming = false;
+      client.disableStreaming = true;
+
+      const finalOutput = await client.invoke('What is the capital of France?');
+
+      expect(finalOutput).toMatchSnapshot();
```
ZhongpinWang, line 409: [pp] Is this `expect` important for this non-streaming test? If not we can remove it and remove the snapshot. The snapshots are too many.

packages/langchain/src/openai/chat.test.ts
```diff
@@ -332,6 +338,160 @@ describe('Chat client', () => {
...
+
+      client.disableStreaming = true;
+      client.streaming = false;
+
+      const stream = await client.stream('What is the capital of France?');
+
+      let finalOutput: AIMessageChunk | undefined;
+      for await (const chunk of stream) {
+        finalOutput = finalOutput ? finalOutput.concat(chunk) : chunk;
+      }
```
ZhongpinWang: [q] Just a question, this loop will only run once right? One chunk generated with the full content?

packages/langchain/src/openai/chat.test.ts
```diff
@@ -406,5 +566,58 @@ describe('Chat client', () => {
...
+      const workflow = new StateGraph(MessagesAnnotation)
+        // Define the (single) node in the graph
+        .addNode('model', callModel)
+        .addEdge(START, 'model')
+        .addEdge('model', END);
+
+      const app = workflow.compile();
+      const stream = await app.stream(
+        { messages: [{ role: 'user', content: 'Hello!' }] },
+        // langgraph will only enable streaming in a granular streaming mode
```
ZhongpinWang: [pp] Perhaps put the LangChain [doc](https://docs.langchain.com/oss/javascript/langgraph/streaming#supported-stream-modes) here? I find this comment, especially granular` hard to understand if not reading the LangChain doc and knowing other modes.

> `messages`: Streams 2-tuples (LLM token, metadata) from any graph nodes where an LLM is invoked.

packages/langchain/src/openai/chat.test.ts
```diff
@@ -406,5 +566,58 @@ describe('Chat client', () => {
...
+        { streamMode: 'messages' as const }
+      );
+
+      let finalOutput;
+      for await (const chunk of stream) {
+        finalOutput =
+          finalOutput !== undefined ? finalOutput.concat(chunk) : chunk;
+      }
```
ZhongpinWang:
```suggestion
      for await (const _ of stream) {
        // Empty
      }
```

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -389,6 +395,160 @@ describe('orchestration service client', () => {
...
+        streaming: true,
+        disableStreaming: true
+      } as any);
+      expect(client.streaming).toBe(false);
+      expect(client.disableStreaming).toBe(true);
+
+      const finalOutput = await client.invoke([
+        { role: 'user', content: 'Hello!' }
+      ]);
+      expect(finalOutput).toMatchSnapshot();
```
ZhongpinWang: [pp] Similar, do we need to check this for non-streaming case since this is not different from a normal non-streaming case?

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -513,4 +673,63 @@ describe('orchestration service client', () => {
...
+      .addNode('model', callModel)
+      .addEdge(START, 'model')
+      .addEdge('model', END);
+
+    const app = workflow.compile();
+    const stream = await app.stream(
+      {
+        messages
+      },
+      // langgraph will only enable streaming in a granular streaming mode
```
ZhongpinWang: [pp] Same

packages/langchain/src/orchestration/client.ts
```diff
@@ -33,11 +33,14 @@ function isInputFilteringError(error: any): boolean {
 
 /**
  * The Orchestration client.
+ * Remark: `langchainOptions` supports passing `streaming` flag for auto-streaming behavior, despite not being part of `BaseChatModelParams`.
```
ZhongpinWang: [pp/req] Maybe not putting this remark here in the JS Doc since it is a bit too technical. Put it in just a line comment maybe afterwards.

packages/langchain/src/orchestration/client.ts
```diff
@@ -33,11 +33,14 @@ function isInputFilteringError(error: any): boolean {
...
  */
 export class OrchestrationClient extends BaseChatModel<
   OrchestrationCallOptions,
   OrchestrationMessageChunk
 > {
+  streaming: boolean = false;
+
   constructor(
     public orchestrationConfig: LangChainOrchestrationModuleConfig,
     public langchainOptions: BaseChatModelParams = {},
```
ZhongpinWang: [req] If `streaming` is not part of the `BaseChatModelParams`, we can create a new type for LangChain orchestration similar to LangChain OpenAI and union the `streaming` property.

packages/langchain/src/orchestration/client.ts
```diff
@@ -54,6 +57,22 @@ export class OrchestrationClient extends BaseChatModel<
...
     super(langchainOptions);
+
+    // Initialize streaming flags with LangChain-compatible behavior:
+    // - `streaming`: true enables auto-streaming in `invoke()` calls
+    // - `disableStreaming`: true overrides streaming flag
+    // - `streaming`: `false` causes `disableStreaming` to be set to `true` for framework compatibility
+    this.disableStreaming = langchainOptions?.disableStreaming === true;
+
+    // If streaming is explicitly false, streaming is disabled
+    if ((langchainOptions as { streaming?: boolean })?.streaming === false) {
```
ZhongpinWang: [req] Once the new type is created, this can be cleaned up.

packages/langchain/src/orchestration/client.ts
```diff
@@ -54,6 +57,22 @@ export class OrchestrationClient extends BaseChatModel<
...
+      this.disableStreaming = true;
+    }
+
+    // Enable streaming only when `streaming` is `true` (default `false`) and `disableStreaming` is not `true` (default `undefined`).
+    this.streaming =
```
ZhongpinWang, line 72: same

ZhongpinWang (commented):
Minor suggestions

packages/langchain/src/orchestration/client.ts
davidkna-sap: Completely removed comment due to `langchainOptions` type changes.

packages/langchain/src/orchestration/client.ts
davidkna-sap: Updated, I previously avoided this due to backward-compatibility concerns.

packages/langchain/src/orchestration/client.test.ts
davidkna-sap: Removed, previously this test was more useful in that it tested the extra `disableStreaming` condition.

packages/langchain/src/openai/chat.test.ts
davidkna-sap: This will run only once, I have updated the test accordingly.

packages/langchain/src/openai/chat.test.ts
davidkna-sap: I've reduced the amount of snapshot tests a bit.

packages/langchain/src/orchestration/client.ts
ZhongpinWang: Thanks for changing this. I believe this shouldn't be backward incompatible since the new property `streaming` is optional and TS does structural typing.

ZhongpinWang (changes requested):
Oh before we merge, can you add a changeset for the changes?

ZhongpinWang (approved):
Thanks.

PR #1272, opened 2025-11-11
fix: Use trusted publishing for the canary workflow

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#416.

This PR changes `deploy.yml` to be a proxy for any release-triggering events, which are forwarded to `publish_canary.yml` or `publish_release.yml` (renamed from `publish.yml`) via a `workflow_call`:

>  Some GitHub Actions workflows use `workflow_call` to invoke other workflows that run npm publish, or use `workflow_dispatch` for manual publishing. When this happens, validation checks the calling workflow's name instead of the workflow that actually contains the publish command, which can cause configuration mismatches.

(via. [npm/documentation#1755](https://www.github.com/npm/documentation/issues/1755) and https://docs.npmjs.com/trusted-publishers#troubleshooting)

If completely combining all jobs in a single file as proposed in SAP/ai-sdk-js-backlog#416 is preferred for reduced GitHub-actions complexity or variation between canary and releases, I would also be glad to combine all workflows in a single file.

### DIFF
--- a/.github/workflows/publish-canary.yml
+++ b/.github/workflows/publish-canary.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - uses: sap/ai-sdk-js/.github/actions/setup@main
         with:
-          node-version: ${{ vars.DEFAULT_NODE_VERSION }}
+          node-version: 24 # Will install npm 11 needed for trusted publishing
           pnpm-install-args: ''
 
       - name: prepare canary changeset


.github/workflows/publish-canary.yml
```diff
@@ -31,12 +29,22 @@ jobs:
...
         run: |
           date=`date +%Y%m%d%H%M%S`
           pnpm changeset pre enter ${date}
           pnpm changeset version
           pnpm changeset pre exit
-          pnpm changeset publish --tag ${{ github.event_name == 'schedule' && 'canary' || inputs.tag }}
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPMJS_ACCESS_TOKEN }}
+
+  publish-npm:
```
deekshas8: [q] Why do we need to split this?  Also, why remove the `canary` tag?

davidkna-sap: I split this to align the file with the handling in the regular release job, and I mistakenly removed the `--tag` because I copied the job definition from there. I've added the `--tag` back now.

deekshas8: I would honestly just keep it as it was. Its simpler and more straightforward.

davidkna-sap: I've reverted it, it may not even have work as expected due to not operating on the same checkout.

PR #1277, opened 2025-11-14
chore(lint): Be more permissive with/ignore generated cjs files

davidkna-sap:
## Context

Alongside the `dist` folder, disable additional rules for `dist-cjs` which will fail after compiling for cjs.


ZhongpinWang (approved):
Thx for taking care of this. Bothered me everytime when I generate cjs files. 😄

PR #1278, opened 2025-11-17
fix: Improve reliability of Spec File Update Workflow

davidkna-sap:
## Context

The workflow will fail if there is a conflict (https://github.com/SAP/ai-sdk-js/actions/runs/19399804740).

This PR gives `git rebase` a resolution strategy  (`theirs`) to allow resolving conflicts and continuing with the workflow.

### DIFF
--- a/.github/workflows/spec-update.yaml
+++ b/.github/workflows/spec-update.yaml
@@ -74,7 +74,7 @@ jobs:
           BRANCH="spec-update/$CHOICE/$REF"
           git fetch origin $BRANCH || true
           git checkout -B $BRANCH origin/$BRANCH || git checkout -b $BRANCH
-          git rebase origin/main
+          git rebase -X theirs origin/main
           git push --force --set-upstream origin $BRANCH
           echo "branch=$BRANCH" >> "$GITHUB_OUTPUT"
 


.github/workflows/spec-update.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
           BRANCH="spec-update/$CHOICE/$REF"
           git fetch origin $BRANCH || true
           git checkout -B $BRANCH origin/$BRANCH || git checkout -b $BRANCH
-          git rebase origin/main
+          git rebase -X theirs origin/main
```
ZhongpinWang, line 77: [q] Have you checked why it failed before? Normally, the auto-created `spec-update/grounding/main` should not contain conflicting changes to main.
And if this flow is triggered by orchestration colleagues in their PRs, then they anyway should never changed the auto created branch.

davidkna-sap: For grounding I think it started failing because we closed #951 without deleting its branch. Prompt registry updates are also failing, likely alongside the orval bump which IIRC changed some of the relevant files.

ZhongpinWang (approved):
In general, I would even prefer deleting the old branch except for orchestration runs which are triggered by orchestration PRs. And for all case, if rebasing failed, we anyway just delete the old branch to make everything work. Maybe helpful for orchestration colleagues to see why it failed. But accepting theirs from our main can hopefully also solve the issue.

PR #1285, opened 2025-11-18
feat: Support protected_material_code

BrigittaK307:
The PR's purpose is to support protected_material_code in output content filtering module.

### DIFF
--- a/sample-code/src/orchestration.ts
+++ b/sample-code/src/orchestration.ts
@@ -206,6 +206,39 @@ export async function orchestrationPromptRegistry(): Promise<OrchestrationRespon
 }
 
 /**
+ * Use a template stored in the prompt registry store at
+ * the resource group scope.
+ * @returns The orchestration service response.
+ */
+export async function orchestrationCompletionPromptRegistryScoped(): Promise<OrchestrationResponse> {
+  const orchestrationClient = new OrchestrationClient(
+    {
+      promptTemplating: {
+        prompt: {
+          // refer to a prompt template stored at resource group scope
+          template_ref: {
+            name: 'e2e-test-scoped',
+            scenario: 'e2e-test-scoped',
+            version: '0.0.1',
+            scope: 'resource_group'
+          }
+        },
+        // define the language model to be used
+        model: {
+          name: 'gpt-4o'
+        }
+      }
+    },
+    // provide resource group where the template is stored
+    { resourceGroup: 'ai-sdk-js-e2e' }
+  );
+
+  return orchestrationClient.chatCompletion({
+    placeholderValues: { country: 'France' }
+  });
+}
+
+/**
  * Apply multiple content filters to the input.
  * @returns The orchestration service error response.
  */
--- a/sample-code/src/index.ts
+++ b/sample-code/src/index.ts
@@ -10,6 +10,7 @@ export {
   orchestrationChatCompletion,
   orchestrationTemplating,
   orchestrationPromptRegistry,
+  orchestrationCompletionPromptRegistryScoped,
   orchestrationInputFiltering,
   orchestrationOutputFiltering,
   orchestrationRequestConfig,
--- a/tests/e2e-tests/src/orchestration.test.ts
+++ b/tests/e2e-tests/src/orchestration.test.ts
@@ -2,6 +2,7 @@ import {
   orchestrationChatCompletion,
   orchestrationTemplating,
   orchestrationPromptRegistry,
+  orchestrationCompletionPromptRegistryScoped,
   orchestrationInputFiltering,
   orchestrationOutputFiltering,
   orchestrationRequestConfig,
@@ -48,6 +49,12 @@ describe('orchestration', () => {
 
   it('should complete a chat with a template reference', async () => {
     const response = await orchestrationPromptRegistry();
+
+    assertContent(response);
+  });
+
+  it('should complete a chat with a scoped template reference', async () => {
+    const response = await orchestrationCompletionPromptRegistryScoped();
 
     assertContent(response);
   });


packages/orchestration/src/util/filtering.ts
```diff
@@ -31,31 +31,41 @@ export function buildAzureContentSafetyFilter<T extends ConfigType>(
...
           ])
         ),
-        ...(prompt_shield !== undefined && { prompt_shield })
+        ...(prompt_shield !== undefined && { prompt_shield }),
+        ...(protected_material_code !== undefined && {
+          protected_material_code
+        })
       }
     } as AzureContentSafetyFilterReturnType<T>;
   }
```
KavithaSiva: `protected_material_code` is only supported with Output filters, please remove the code for adding config to input.
Refer [here](https://github.tools.sap/AI/llm-orchestration/blob/main/src/spec/v2.yaml#L1289)

packages/orchestration/src/util/filtering.test.ts
```diff
@@ -33,6 +33,36 @@ describe('Content filter util', () => {
...
+      });
+      const expectedFilterConfig = {
+        type: 'azure_content_safety',
+        config: {
+          hate: 0,
+          protected_material_code: true
+        }
+      };
+      expect(filterConfig).toEqual(expectedFilterConfig);
+    });
```
KavithaSiva: As pointed out above, feature is only valid for output filter configuration.

packages/orchestration/src/orchestration-types.ts
KavithaSiva: [req] Define the input parameters separately don't use output type to build the input types as the types have diverged now.

KavithaSiva (changes requested):
In addition to changing the input type definition also add type tests for this feature.
Refer to `tests/type-tests/test/orchestration.test-d.ts` with `prompt_shield` for examples.

sample-code/src/orchestration.ts
```diff
@@ -223,7 +223,8 @@ export async function orchestrationInputFiltering(): Promise<OrchestrationErrorR
           // Build Azure content filter with only safe content allowed for hate and violence
           buildAzureContentSafetyFilter('input', {
             hate: 'ALLOW_SAFE',
-            violence: 'ALLOW_SAFE'
+            violence: 'ALLOW_SAFE',
+            protected_material_code: true
```
KavithaSiva: [req] Remove this, will not work here once you adjust the types.

sample-code/src/orchestration.ts
```diff
@@ -268,7 +269,8 @@ export async function orchestrationOutputFiltering(): Promise<OrchestrationRespo
           // Build Azure content filter with only safe content allowed for hate and violence
           buildAzureContentSafetyFilter('output', {
             hate: 'ALLOW_SAFE',
-            violence: 'ALLOW_SAFE'
+            violence: 'ALLOW_SAFE',
+            protected_material_code: true
```
KavithaSiva: [req] I think we can remove `protected_material_code` from this example as the prompt will not generate code samples from llm. 
Documenting it in the portal should be enough, doesn't require sample code IMO.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -381,17 +380,36 @@ export interface AzureContentSafetyFilterOutputParameters {
...
+  sexual?: AzureFilterThreshold;
+  /**
+   * The filter category for violence content.
+   */
+  violence?: AzureFilterThreshold;
+  /**
+   * A flag to use prompt shield.
+   */
+  prompt_shield?: boolean;
+}
```
KavithaSiva: [req] Please move the common properties in input and output to an internal base e.g. `AzureContentSafetyFilterBaseParameters` interface which can be extended by both input and output interfaces.
Then add only the additional properties in the input and output interfaces.

.changeset/smooth-showers-throw.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ai-sdk/orchestration': minor
+'@sap-ai-sdk/type-tests': minor
+---
+
+[Feat] Added support for `protected_material_code` to `AzureContentSafetyFilterOutputParameters`
```
KavithaSiva:
```suggestion
[New Functionality] Added support for `protected_material_code` to `AzureContentSafetyFilterOutputParameters`
```

packages/orchestration/src/orchestration-types.ts
```diff
@@ -362,9 +361,10 @@ export type DpiMaskingConfig = Omit<
 };
 
 /**
- * Output Parameters for Azure content safety output filter.
+ * Base parameters for Azure content safety filter.
+ * @internal
  */
-export interface AzureContentSafetyFilterOutputParameters {
+export interface AzureContentSafetyFilterBaseParameters {
```
KavithaSiva: [req] You don't have to export it. Remove `@internal` and `export`.

KavithaSiva (changes requested):
Mostly good to go, just two more changes.

.changeset/smooth-showers-throw.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ai-sdk/orchestration': minor
+'@sap-ai-sdk/type-tests': minor
+---
+
+[New Functionality] Added support for `protected_material_code` to `AzureContentSafetyFilterOutputParameters`
```
KavithaSiva:
```suggestion
[New Functionality] Added support for `protected_material_code` property to `buildAzureContentSafetyFilter()` function for output filter configuration to allow detecting protected code content from known github repositories.
```

PR #1287, opened 2025-11-19
feat: Support for applyTo and new translation input/output types

BrigittaK307:


packages/orchestration/src/orchestration-types.ts
```diff
@@ -437,30 +438,36 @@ interface TranslationConfigParametersInput {
...
+   * Language of the text to be translated.
+   * @example sourceLanguage: 'de-DE'
+   */
+  sourceLanguage?: string;
+  /**
+   * Configuration for applying translation to specific placeholders or message roles, or a target language string.
+   * @example targetLanguage: { category: 'placeholders', items: ['assistant_response'] }
+   */
+  targetLanguage?: string | SAPDocumentTranslationApplyToSelector;
+}
```
KavithaSiva: [req] Previously, we had an internal interface, `TranslationConfigParametersOutput` which was exposed directly via type `TranslationOutputParameters`. 
Please revert to the old state, and also make `targetLanguage` a required field as making this optional is a **breaking change.** 

Please always review your changes, they should never include breaking changes in minor version releases, exceptions are possible in generated clients, but even then we include compatibility notes.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -437,30 +438,36 @@ interface TranslationConfigParametersInput {
...
+  /**
+   * Language of the text to be translated.
+   * @example sourceLanguage: 'de-DE'
+   */
+  sourceLanguage?: string;
+  /**
+   * Configuration for applying translation to specific placeholders or message roles, or a target language string.
+   * @example targetLanguage: { category: 'placeholders', items: ['assistant_response'] }
+   */
+  targetLanguage?: string | SAPDocumentTranslationApplyToSelector;
```
KavithaSiva: [req] As, `targetLanguage` should also additionally allow another type in addition to string in our defined types, consider adding a new type for `SAPDocumentTranslationApplyToSelector` instead of using the generated type directly.

Reason is that we define our own types to act as wrappers for generated orchestration types. This ensures that even if orchestration types introduce a breaking change, we can try to handle it in a non-breaking manner.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -437,30 +438,36 @@ interface TranslationConfigParametersInput {
...
+  translateMessagesHistory?: boolean;
   /**
-   * Language to which the text should be translated.
-   * @example targetLanguage: 'en-US'
+   * Configuration for applying translation to specific placeholders or message roles.
+   * @example applyTo: [{ category: 'placeholders', items: ['user_input'], source_language: 'de-DE' }]
    */
-  targetLanguage: string;
-}
+  applyTo?: SAPDocumentTranslationApplyToSelector[];
```
KavithaSiva: [req] These fields can be added to `TranslationConfigParametersInput` instead of adding it as intersection type. I don't see an advantage of the intersection here.

BrigittaK307:
> Please rework on the types and interfaces used for translation based on my comments.
> 
> Current state in `main` is:
> 
> For configuring both input and output translation Module we have internal Interfaces: `TranslationConfigParametersInput`, `TranslationConfigParametersOutput`
> 
> And these are exposed via types:`TranslationInputParameters`,`TranslationOutputParameters` -> `TranslationConfigParams<T extends ConfigType>`, this is the input for our build convenience function.
> 
> `TranslationInputConfig`, `TranslationOutputConfig` -> `TranslationReturnType<T extends ConfigType>`, this is the output of build function, so they will be always adhere to the Orchestration generated types.
> 
> Hope this helps you to understand the current state

Thank you for the detailed explanation, it helps a lot in understanding how everything works!

packages/orchestration/src/orchestration-types.ts
```diff
@@ -438,7 +452,17 @@ interface TranslationConfigParametersInput {
    * Language to which the text should be translated.
    * @example targetLanguage: 'en-US'
    */
-  targetLanguage: string;
+  targetLanguage: TranslationTargetLanguage;
```
davidkna-sap: [req] I tried using an `DocumentTranslationApplyToSelector` as `target_language` in Bruno and it doesn't seem to work. I even got an error that only `string` is supported so please clarify if this is a spec mistake. `DocumentTranslationApplyToSelector` does not have a `target_language` field so I am not sure how It could work in practice.

davidkna-sap: Nevermind, I managed to make it work.

packages/orchestration/src/util/translation.ts
```diff
@@ -12,18 +13,38 @@ import type {
...
+      ...(inputConfig.translateMessagesHistory !== undefined && {
+        translate_messages_history: inputConfig.translateMessagesHistory
+      }),
+      config: {
+        ...baseConfig,
+        ...(inputConfig.applyTo && {
+          apply_to: inputConfig.applyTo
+        })
+      }
+    } as TranslationReturnType<T>;
```
davidkna-sap, line 53: [pp] I think this may require removing `baseConfig`, but I would prefer using `satisfies`  for the checks in this function (followed by `as` if necessary).

This helped me find https://github.com/SAP/ai-sdk-js/pull/1287/files#r2581014800

packages/orchestration/src/orchestration-types.ts
davidkna-sap: [req] For `SAPDocumentTranslationInput` this is string only:
```suggestion
  targetLanguage: string;
```

packages/orchestration/src/orchestration-types.ts
```diff
@@ -425,6 +426,19 @@ export type LlamaGuard38BCategory = keyof LlamaGuard38B;
  */
 export type ConfigType = 'input' | 'output';
 
+/**
+ * Configuration selector for applying translation to specific placeholders or message roles.
+ */
+export type DocumentTranslationApplyToSelector =
+  SAPDocumentTranslationApplyToSelector;
```
davidkna-sap: [req] This uses `source_language` but `buildTranslationConfig` uses camelCase `sourceLanguage`. This should probably use a convenience type and have name translations in `buildTranslationConfig`.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -438,7 +452,17 @@ interface TranslationConfigParametersInput {
    * Language to which the text should be translated.
    * @example targetLanguage: 'en-US'
    */
-  targetLanguage: string;
+  targetLanguage: TranslationTargetLanguage;
+  /**
+   * If true, the messages history will be translated as well.
+   * @example translateMessagesHistory: true
+   */
+  translateMessagesHistory?: boolean;
```
davidkna-sap, line 481: [req] Per the docs this defaults to `true`
```suggestion
   * If false, the messages history will not be translated.
   * @example translateMessagesHistory: false
   */
  translateMessagesHistory?: boolean;
```

packages/orchestration/src/util/translation.ts
```diff
@@ -12,18 +13,38 @@ import type {
  * See https://help.sap.com/docs/translation-hub/sap-translation-hub/supported-languages-6854bbb1bd824ffebc3a097a7c0fd45d for list of supported languages.
  * @returns SAP Document Translation configuration.
  * @example buildTranslationConfig('input', { sourceLanguage: 'de-DE', targetLanguage: 'en-US' })
+ * @example buildTranslationConfig('input', { sourceLanguage: 'de-DE', targetLanguage: 'en-US', translateMessagesHistory: true })
```
davidkna-sap:
```suggestion
 * @example buildTranslationConfig('input', { sourceLanguage: 'de-DE', targetLanguage: 'en-US', translateMessagesHistory: false })
```

packages/orchestration/src/orchestration-types.ts
BrigittaK307: Yes, the default is true but in the description in the api/schema file looks like this:
```
 /**
   * If true, the messages history will be translated as well.
   * Default: true.
   */
  translate_messages_history?: boolean;
```
Do you think that in the description I should mention what should happen if not the default value is set?

davidkna-sap: Maybe you can use `@default true` instead of an `@example`

tests/type-tests/test/orchestration.test-d.ts
davidkna-sap, line 1: The removed tests here should be re-added

davidkna-sap (approved):

sample-code/src/orchestration.ts
```diff
@@ -744,11 +744,22 @@ export async function orchestrationTranslation(): Promise<OrchestrationResponse>
...
+          }
+        ]
       }),
       output: buildTranslationConfig('output', {
         sourceLanguage: 'de-DE',
-        targetLanguage: 'fr-FR'
+        targetLanguage: {
+          category: 'placeholders',
+          items: ['assistant_response']
+        }
```
KavithaSiva: [req] I tested the sample code and in it's current state it does not work. The following changes had to be made to make it work:

1.  `sourceLanguage` had to be removed from input config, as `applyTo` cannot be combined with a `sourceLanguage` property at the root.
2. items in output config cannot contain `assistant_response` as it is not a placeholder value. Change it to `user_input` instead.

BrigittaK307: I made the changes, and currently the Bruno request is working and looks like this:
Screenshot 2025-12-08 112302

BrigittaK307: Removed it, now the Bruno response for that api call looks like this:
image

PR #1292, opened 2025-11-20
chore: Add sample-code/e2e-test for scoped prompt-template refs

davidkna-sap:
## Context

Related SAP/ai-sdk-js-backlog#429.

Types for this were added in the recent orchestration spec upgrade, but sample-code and an e2e tests were still missing.

### DIFF
--- a/sample-code/src/orchestration.ts
+++ b/sample-code/src/orchestration.ts
@@ -206,6 +206,39 @@ export async function orchestrationPromptRegistry(): Promise<OrchestrationRespon
 }
 
 /**
+ * Use a template stored in the prompt registry store at
+ * the resource group scope.
+ * @returns The orchestration service response.
+ */
+export async function orchestrationCompletionPromptRegistryScoped(): Promise<OrchestrationResponse> {
+  const orchestrationClient = new OrchestrationClient(
+    {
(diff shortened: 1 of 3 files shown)


sample-code/src/orchestration.ts
```diff
@@ -205,6 +205,39 @@ export async function orchestrationPromptRegistry(): Promise<OrchestrationRespon
   });
 }
 
+/**
+ * Use a template stored in the prompt registry store at
+ * the resource group scope.
+ * @returns The orchestration service response.
+ */
+export async function orchestrationCompletionPromptRegistryScoped(): Promise<OrchestrationResponse> {
```
BrigittaK307, line 213: Consider exporting this function in the `server.ts` file as well, so that it can be tested out more easily in a `curl` or `Bruno` request.

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -52,6 +53,12 @@ describe('orchestration', () => {
     assertContent(response);
   });
 
+  it('should complete a chat with a scoped template reference', async () => {
```
BrigittaK307, line 56: This test case is failing for me, it returns this error message:

Image

PR #1302, opened 2025-11-26
fix: Adapt RAG example to use `createAgent `

KavithaSiva:
## Context
This PR adapts the current RAG example using the latest [RAG agent](https://docs.langchain.com/oss/javascript/langchain/rag) tutorial.

It also fixes smoke-test failures by adding the missing `@langchain/classic` package.

Successful test run: https://github.com/SAP/ai-sdk-js/actions/runs/19854088195

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -287,7 +287,7 @@ importers:
         version: 5.2.1
       langchain:
         specifier: ^1.0.0
-        version: 1.1.3(@langchain/core@1.1.2(openai@6.9.1(ws@8.18.3)(zod@4.1.13)))(openai@6.9.1(ws@8.18.3)(zod@4.1.13))(zod-to-json-schema@3.25.0(zod@4.1.13))
+        version: 1.1.5(@langchain/core@1.1.4(openai@6.9.1(ws@8.18.3)(zod@4.1.13)))(openai@6.9.1(ws@8.18.3)(zod@4.1.13))(zod-to-json-schema@3.25.0(zod@4.1.13))
       uuid:
         specifier: ^13.0.0
         version: 13.0.0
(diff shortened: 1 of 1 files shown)


tests/smoke-tests/package.json
```diff
@@ -15,6 +15,7 @@
     "lint:fix": "eslint . --fix && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -w --log-level error"
   },
   "dependencies": {
+    "@langchain/classic": "^1.0.5",
```
deekshas8, line 18: [q] Is it a requirement to have it always installed?

KavithaSiva: yes, it is required for our RAG sample code.

sample-code/src/langchain-azure-openai.ts
```diff
@@ -96,49 +95,59 @@ export async function invokeRagChain(): Promise<string> {
...
+      Use the tool to help answer user queries.
       If you don't know the answer, just say that you don't know.
       Use ten sentences maximum and keep the answer concise.
-      Always include code snippets in your answer including class instantiation.
-
-      Question: {question}
-      Context: {context}
-      Answer:`
+      Always include code snippets in your answer including class instantiation.`
   );
```
davidkna-sap: [pp/np] With LC v1 this can just be a string:

```suggestion
  const systemPrompt = `You are an assistant for question-answering tasks.
      You have access to a tool that retrieves information related to a query.
      Use the tool to help answer user queries.
     If you don't know the answer, just say that you don't know.
     Use ten sentences maximum and keep the answer concise.
      Always include code snippets in your answer including class instantiation.`;
```

sample-code/src/langchain-azure-openai.ts
```diff
@@ -96,49 +95,59 @@ export async function invokeRagChain(): Promise<string> {
...
-    llm,
-    prompt: promptTemplate,
-    outputParser: new StringOutputParser()
+  const agent = createAgent({
+    model: chatClient,
+    tools: [retrieve],
+    systemPrompt
   });
 
-  // Create a retriever for the vector store
```
davidkna-sap, line 158: [q] I feel like the [dynamic system prompt variant](https://docs.langchain.com/oss/javascript/langchain/agents#dynamic-system-prompt) is closer to the old `ragChain` example and show-cases best practices a bit more strongly. That said the current approach is a bit simpler/easier to maintain, so either way would be fine with me.

KavithaSiva: I took a look at this again now. Originally, we were passing the question and the retrieved context.
But, now since we use the agent, it automatically calls the tool, to retrieve context and the question is still injected dynamically using `agentInputs`. 
So, I will keep this as is.

PR #1293, opened 2025-11-20
feat: Update prompt-registry specification

davidkna-sap:
## Context

Related SAP/ai-sdk-js-backlog#428.

## What this PR does and why it is needed

So far this is #1149 with a newer spec and an eslint exception for `no-useless-escape` in the generated files.

## Definition of Done

- [ ] Unit / type tests cover new classes
- [x] Release notes / Changeset updated

PR #1339, opened 2025-12-04
fix: Update orval configuration for handling JSON examples in prompt registry specification

KavithaSiva:
## Context

Updates orval config to ignore linting issues with JSON examples in spec.

With recent updates to prompt registry spec: https://github.tools.sap/AI/prompt-registry/pull/310 the SDK generation step was failing.

Reason is we use `orval` in the post generation step to generate zod schemas, and one of the descriptions in the spec file:
```
This selector allows you to define the scope of translation, such as specific placeholders or messages with specific roles. For example, `{"category": "placeholders", "items": ["user_input"], "source_language": "de-DE"}` targets the value of "user_input" in `placeholder_values` specified in the request payload; and considers the value to be in German.
```
uses double quotes in the JSON example which is escaped using \  by orval leading to linting errors.
As this lint rule severity is low, I decided to ignore the rule instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "eslint": "^9.39.1",
     "glob": "^13.0.0",
     "jest": "^30.2.0",
-    "jsonwebtoken": "^9.0.2",
+    "jsonwebtoken": "^9.0.3",
     "mock-fs": "^5.5.0",
     "nock": "^14.0.10",
     "orval": "^7.17.0",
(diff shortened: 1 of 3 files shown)


ZhongpinWang:
I would even want to disable eslint for `src/zod` folder. The zod file is generated and makes almost no sense to lint / prettify it. 
If you want @davidkna-sap , you can do this in your #1293. Check the package json in prompt registry package and modify both `lint` and `lint:fix` (perhaps also add this folder to a local prettierignore).

PR #1393, opened 2026-01-05
feat: Support for Orchestration config persistence in Orchestration client

KavithaSiva:
## Context

Closes SAP/ai-sdk-js-backlog#428.

This PR introduces initialising vanilla orchestration client using orchestration configuration references.

```TS
    scenario: 'sdk-test-scenario',
    name: 'test-config-for-OrchestrationTest',
    version: '0.0.1'
  });

//or

    id: 'c1a4f2e2-BETA-4d2b-9f7a-1234567890ab'
  })
```


.changeset/rich-onions-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[New Functionality] Add support for orchestration configuration references in the vanilla orchestration client.
```
davidkna-sap: [np] Should be clear from context:

```suggestion
[New Functionality] Add support for orchestration configuration references in the orchestration client.
```

[q] Have you checked if this can be accessed from the langchain client?

packages/orchestration/src/util/module-config.ts
```diff
@@ -52,6 +55,42 @@ export function constructCompletionPostRequestFromJsonModuleConfig(
...
+ * @internal
+ */
+export function constructCompletionPostRequestFromConfigReference(
+  configRef: OrchestrationConfigReference,
+  request?: ChatCompletionRequest
+):
+  | CompletionRequestConfigurationReferenceById
+  | CompletionRequestConfigurationReferenceByNameScenarioVersion {
+  // Determine if it's by-id or by-name reference
+  if ('id' in configRef && configRef.id !== undefined) {
```
davidkna-sap: [pp] I think that `OrchestrationConfigReference` can be forwarded as-is. If you would rather keep it separated, I feel like `configRef !== undefined` should be skipped, because `{ id: undefined }` leading to a `{ scenario: undefined, name: undefined, version: undefined }` configuration reference payload could be unexpected.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -62,6 +72,11 @@ export class OrchestrationClient {
     request?: ChatCompletionRequest,
     requestConfig?: CustomRequestConfig
   ): Promise<OrchestrationResponse> {
+    if (this.isConfigReference(this.config) && request?.messages?.length) {
+      logger.warn(
+        'The messages field in request is not supported when using an orchestration config reference. Messages should be part of the referenced configuration or provided via messagesHistory. The messages field will be ignored.'
```
davidkna-sap, line 73: [req] `request` should be in backticks to highlight that it's an argument.

```suggestion
        'The messages field in `request` is not supported when using an orchestration config reference. Messages should be part of the referenced configuration or provided via messagesHistory. The messages field will be ignored.'
```

packages/orchestration/src/orchestration-client.ts
```diff
@@ -89,6 +104,18 @@ export class OrchestrationClient {
           'Stream options are not supported when using a JSON module config.'
         );
       }
+      if (this.isConfigReference(this.config)) {
+        if (options) {
+          logger.warn(
+            'Stream options are not supported when using an orchestration config reference. Streaming is only supported if the referenced config has streaming configured.'
```
davidkna-sap, line 106: likewise.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -115,12 +158,17 @@ export class OrchestrationClient {
             request,
             stream
           )
-        : constructCompletionPostRequest(
-            this.config,
-            request,
-            stream,
-            streamOptions
-          );
+        : this.isConfigReference(this.config)
```
davidkna-sap: [pp]: I feel like it would be nice to print a warning if `"config_ref" in this.config` and `this.isConfigReference(this.config)` is `false`.

.changeset/rich-onions-sort.md
KavithaSiva: Yes, in the langchain client, the orchestration configuration is modified, we add some additional configs  `tools` and `stop`. 
As this feature is about maintaining a static configuration, I would not extend this feature to the langchain orchestration client.

packages/orchestration/src/orchestration-client.ts
KavithaSiva: Done.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -220,6 +220,39 @@ export interface OrchestrationModuleConfig {
...
+ * created via the Prompt Registry API before it can be referenced.
+ * Reference by ID.
+ * @example
+ * const configRef: OrchestrationConfigReference = {
+ *   id: 'f47ac10b-58cc-4372-a567-0e02b2c3d479'
+ * };
+ * Reference by scenario, name and version.
+ * @example
+ * const configRef: OrchestrationConfigReference = {
+ *   scenario: 'foundation-models',
```
marikaner: [q] Isn't the scenario always "orchestration" for Orchestration?

packages/orchestration/src/util/module-config.ts
```diff
@@ -3,11 +3,14 @@ import type {
   ChatCompletionRequest,
   StreamOptions,
   OrchestrationModuleConfig,
+  OrchestrationConfigReference,
   EmbeddingModuleConfig,
   EmbeddingRequest
 } from '../orchestration-types.js';
 import type {
   CompletionPostRequest,
+  CompletionRequestConfigurationReferenceById,
```
marikaner, line 12: [pp] For consistency, I think `Config` in type names is better than `Configuration`.
[pp] For brevity, I would prefer `Ref` instead of `Reference`.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -39,17 +41,25 @@ const logger = createLogger({
 export class OrchestrationClient {
   /**
    * Creates an instance of the orchestration client.
-   * @param config - Orchestration module configuration. This can either be an `OrchestrationModuleConfig` object or a JSON string obtained from AI Launchpad.
+   * @param config - Orchestration configuration. Can be:
+   * - An `OrchestrationModuleConfig` object for inline configuration
+   * - A JSON string obtained from AI Launchpad
+   * - An `OrchestrationConfigReference` to reference a stored configuration by ID or name.
```
marikaner: [req] => https://github.com/SAP/cloud-sdk/blob/main/STYLEGUIDE.md#include-a-descriptive-noun-before-or-after-code
```suggestion
   * - An object of type`OrchestrationConfigReference` to reference a stored configuration by ID or name.
```

packages/orchestration/src/orchestration-client.ts
```diff
@@ -39,17 +41,25 @@ const logger = createLogger({
 export class OrchestrationClient {
   /**
    * Creates an instance of the orchestration client.
-   * @param config - Orchestration module configuration. This can either be an `OrchestrationModuleConfig` object or a JSON string obtained from AI Launchpad.
+   * @param config - Orchestration configuration. Can be:
```
marikaner, line 45: [q] Does doc generation work with multiline descriptions?

packages/orchestration/src/orchestration-client.ts
```diff
@@ -105,6 +132,22 @@ export class OrchestrationClient {
     }
   }
 
+  /**
+   * Type guard to check if config is a config reference.
+   * @param config - The config to check.
+   * @returns True if config is a config reference.
```
marikaner: [pp] In JS this returns true, but it TS it is a little more complex than that. Just pp, though.

```suggestion
   * @returns Whether the config is a config reference.
```

packages/orchestration/src/orchestration-client.ts
KavithaSiva: Done.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -105,6 +132,22 @@ export class OrchestrationClient {
...
+   */
+  private isConfigReference(
+    config: OrchestrationModuleConfig | string | OrchestrationConfigReference
+  ): config is OrchestrationConfigReference {
+    return (
+      typeof config === 'object' &&
+      config !== null &&
```
marikaner: [q] Can it be `null` here?

packages/orchestration/src/orchestration-client.ts
```diff
@@ -62,6 +72,11 @@ export class OrchestrationClient {
     request?: ChatCompletionRequest,
     requestConfig?: CustomRequestConfig
   ): Promise<OrchestrationResponse> {
+    if (this.isConfigReference(this.config) && request?.messages?.length) {
+      logger.warn(
+        'The `messages` field in `request` is not supported when using an orchestration config reference. Messages should be part of the referenced configuration or provided via `messagesHistory`. The `messages` field will be ignored.'
```
marikaner: [pp] I think formatting in logs is not a good idea, because it will not be formatted, will it?

packages/orchestration/src/orchestration-client.ts
```diff
@@ -39,17 +41,25 @@ const logger = createLogger({
...
+      | OrchestrationModuleConfig
+      | string
+      | OrchestrationConfigReference,
     private deploymentConfig?: ResourceGroupConfig | DeploymentIdConfig,
     private destination?: HttpDestinationOrFetchOptions
   ) {
     if (typeof config === 'string') {
       this.validateJsonConfig(config);
+    } else if (this.isConfigReference(config)) {
+      this.config = config;
```
marikaner: [req] This line effectively does the same as `this.config = this.config`. Maybe instead you can turn this around and just do this:

```
} else if (!this.isConfigReference(config)) {
  this.config =
    typeof config.promptTemplating.prompt === 'string'
    // ...
```

packages/orchestration/src/util/module-config.ts
KavithaSiva: True, forwarded it as-is with a type assertion instead now.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -105,6 +132,22 @@ export class OrchestrationClient {
     }
   }
 
+  /**
+   * Type guard to check if config is a config reference.
+   * @param config - The config to check.
+   * @returns True if config is a config reference.
+   */
+  private isConfigReference(
```
marikaner: [pp] I find it a little odd, that this is a private method of the orchestration client, as this has semantically nothing to do with the client itself. It is then always called by passing the instance variable `this.config` (except in the constructor, where it could be changed), which is odd as well, because methods have access to instance variables, so you wouldn't need to pass it.
In my opinion typeguards belong to the types and should be located there.

packages/orchestration/src/util/module-config.ts
```diff
@@ -52,6 +55,42 @@ export function constructCompletionPostRequestFromJsonModuleConfig(
...
+      name: configRef.name,
+      version: configRef.version
+    },
+    ...(request?.placeholderValues && {
+      placeholder_values: request.placeholderValues
+    }),
+    ...(request?.messagesHistory && {
+      messages_history: request.messagesHistory
+    })
+  };
```
marikaner: +1 on @davidkna-sap's comment. Also I see quite a bit of duplicate code here, that could be avoided.

packages/orchestration/src/orchestration-client.ts
KavithaSiva: Do you have a specific edge case in mind? Because, as we have an XOR type, at least via the API, we would definitely guarantee that the config reference input has the required properties.

packages/orchestration/src/util/module-config.ts
KavithaSiva: Done.

packages/orchestration/src/orchestration-types.ts
KavithaSiva: This scenario is different from that one.
The `orchestration` scenario is used while creating an orchestration deployment.
The scenario here is to differentiate the multiple orchestration configuration you create for your use cases.
Refer [here](https://help.sap.com/docs/sap-ai-core/generative-ai/create-orchestration-config?version=CLOUD)

KavithaSiva: I adjusted the scenario values to reflect this better.

packages/orchestration/src/util/module-config.ts
KavithaSiva: But, these are types generated by the generator. I have already adjusted the type that I created.

packages/orchestration/src/orchestration-client.ts
KavithaSiva: Moved the typeguard to the type declaration file.

packages/orchestration/src/orchestration-client.ts
KavithaSiva: Screenshot 2026-01-08 at 07 13 53

yes, works.

packages/orchestration/src/orchestration-client.ts
KavithaSiva: No, removed it.

packages/orchestration/src/orchestration-client.ts
KavithaSiva: True, making them plaintext instead.

packages/orchestration/src/orchestration-types.ts
marikaner: Understood. 😅

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -25,6 +25,7 @@ import { getOrchestrationDeploymentId } from './deployment-resolver.js';
 import type { CompletionPostResponse } from './client/api/schema/index.js';
 import type {
   OrchestrationModuleConfig,
+  OrchestrationConfigReference,
```
marikaner: This is outdated right?

packages/orchestration/src/util/module-config.ts
marikaner: are you aware of this file though: scripts/postgenerate-orchestration.ts ?

packages/orchestration/src/orchestration-client.test.ts
KavithaSiva: Yeah, missed it. Done now.

marikaner (approved):
lgtm 😊 I will still reach out to the orchestration team to adjust the spec with consistent naming if possible

PR #1415, opened 2026-01-13
fix: Pin @langchain/core to 1.1.8 due to regression

davidkna-sap:
## Context

Should be reverted once this is merged: https://github.com/langchain-ai/langchainjs/pull/9781

The lockfile update (ade22f63c42b596a80989809294fc449f4adae04) lead to it being updated.

### DIFF
--- a/packages/langchain/package.json
+++ b/packages/langchain/package.json
@@ -43,6 +43,6 @@
     "@langchain/langgraph": "^1.0.7"
   },
   "peerDependencies": {
-    "@langchain/core": "^1.1.8"
+    "@langchain/core": "=1.1.8"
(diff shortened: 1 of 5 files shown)


packages/langchain/package.json
```diff
@@ -43,6 +43,6 @@
     "@langchain/langgraph": "^1.0.7"
   },
   "peerDependencies": {
-    "@langchain/core": "^1.1.8"
+    "@langchain/core": "=1.1.8"
```
marikaner, line 46: is this correct? shouldn't it be
```suggestion
    "@langchain/core": "1.1.8"
```

PR #1419, opened 2026-01-13
feat: RPT-1 beta

marikaner:
## Context

Closes SAP/ai-sdk-js-backlog#452.

This adds a client for RPT.

Currently this is a separate package, however it relies on the foundation model package. We could consider moving the utility function to ai-core instead. Please leave your comments.

Leftovers: `predictParquet` API + gzipping. I would like to do this on separate PRs, if possible.

### DIFF
--- a/packages/rpt/package.json
+++ b/packages/rpt/package.json
@@ -39,6 +39,7 @@
   "dependencies": {
     "@sap-ai-sdk/ai-api": "workspace:^",
     "@sap-ai-sdk/core": "workspace:^",
-    "@sap-ai-sdk/foundation-models": "workspace:^"
+    "@sap-cloud-sdk/connectivity": "^4.3.1",
+    "@sap-cloud-sdk/util": "^4.3.1"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -225,9 +225,12 @@ importers:
       '@sap-ai-sdk/core':
         specifier: workspace:^
         version: link:../core
-      '@sap-ai-sdk/foundation-models':
-        specifier: workspace:^
-        version: link:../foundation-models
+      '@sap-cloud-sdk/connectivity':
+        specifier: ^4.3.1
+        version: 4.3.1
+      '@sap-cloud-sdk/util':
+        specifier: ^4.3.1
+        version: 4.3.1
 
   sample-cap:
     dependencies:


sample-code/src/rpt.ts
```diff
@@ -0,0 +1,120 @@
...
+          PRICE: 89.99,
+          CUSTOMER: 'Tech Forward',
+          COUNTRY: 'India',
+          __row_idx__: '204',
+          SALESGROUP: 'Networking Devices'
+        }
+      ]
+    })
+      // TODO: I assume this should be done differently
+      .setBasePath('/inference/deployments/d1dad0c7fd1b1ce9')
```
ZhongpinWang: [pp] If providing a complete handwritten client is a bit of overkill, then we can maybe provide an [`executeRequest()`](https://github.com/SAP/ai-sdk-js/blob/96b75ccc9c0f01e3192a1621c04288bb236fe9e8/packages/foundation-models/src/azure-openai/azure-openai-chat-client.ts#L74) function, which takes the payload, get the deployment id automatically and build the request.

sample-code/src/rpt.ts
```diff
@@ -0,0 +1,120 @@
...
+        // this is how it is in the type and docs, but doesn't work
+        // target_columns: [{ name: 'SALESGROUP', prediction_placeholder: '[PREDICT]' }]
+
+        // this works, but breaks the type
+        target_columns: [
+          { name: 'SALESGROUP', placeholder_value: '[PREDICT]' } as any
+        ]
+      },
+      index_column: '__row_idx__',
+      data_schema: {
```
ZhongpinWang: [pp] I wonder if we can provide some convenience for building the `data_schema`, but hard to think of the reality, how this part and also the rows are going to be transformed from some table into what is required by RPT...

package.json
```diff
@@ -19,12 +19,13 @@
...
     "ai-api": "pnpm -F=@sap-ai-sdk/ai-api",
-    "foundation-models": "pnpm -F=@sap-ai-sdk/foundation-models",
-    "orchestration": "pnpm -F=@sap-ai-sdk/orchestration",
     "core": "pnpm -F=@sap-ai-sdk/core",
     "document-grounding": "pnpm -F=@sap-ai-sdk/document-grounding",
+    "foundation-models": "pnpm -F=@sap-ai-sdk/foundation-models",
     "langchain": "pnpm -F=@sap-ai-sdk/langchain",
+    "orchestration": "pnpm -F=@sap-ai-sdk/orchestration",
     "prompt-registry": "pnpm -F=@sap-ai-sdk/prompt-registry",
+    "rpt": "pnpm -F=@sap-ai-sdk/rpt",
```
ZhongpinWang, line 28: [q] RPT is technically a foundation model. I would put it in the `foundation-models` package next to openai. But... how can we publish the package with partial beta content, and potentially going to be unstable for a while. Or we can publish it separately, then deprecate the beta package and move the content back to foundation models package. Not sure.

marikaner: I think if we decide to put it into foundation models, we can do it right from the beginning and note in the readme and in the docs that this is beta and might change in the future. However, even though it is technically foundation-models it feels wrong to put this together into one package, because they are not related at all. Users who use one of the packages always carry the second one along, whether they need it or not.
Therefore I vote for separate packages.

sample-code/src/rpt.ts
marikaner: Will do 😊 I was a little uninspired yesterday, but that sounds legit.

package.json
ZhongpinWang: I agree to the package size point. I would also rather separate them. But I feel like it is then in general a problem naming this package foundation models as usually people would probably only use one client from this package and everything else is redundant. Maybe we can consider if we can improve this in the next major version.

marikaner: Agreed, I now also continued with the deployment resolution and I see that even for us, technically we benefit from putting it together. But really just minor ;)

packages/rpt/README.md
```diff
@@ -0,0 +1,31 @@
+# @sap-ai-sdk/rpt <!-- omit in toc -->
+
+SAP Cloud SDK for AI is the official Software Development Kit (SDK) for **SAP AI Core**, **SAP Generative AI Hub**, and **Orchestration Service**.
+
+This package incorporates: TODO!
```
davidkna-sap: [req] Please handle the TODO.

packages/rpt/src/client.ts
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Creates an instance of the RPT client.
+   * @param modelDeployment - This configuration is used to retrieve a deployment. Depending on the configuration use either the given deployment ID or the model name to retrieve matching deployments. If model and deployment ID are given, the model is verified against the deployment.
+   * @param destination - The destination to use for the request.
+   */
+  constructor(
+    private modelDeployment: ModelDeployment<SAPRptModel> = 'sap-rpt-1-small',
+    private destination?: HttpDestinationOrFetchOptions
+  ) {}
+
```
davidkna-sap, line 29: [q] Does the parquet endpoint work yet, should it be exposed on `RptClient`?

marikaner: It does not work yet, but should be soon released. I will try to find out when.

marikaner: I got message it was released in the mean time 😊

packages/rpt/src/client.test.ts
marikaner, line 1: Should I add more tests? I tested the new logic that I added, but everything else was already there, more or less. There are still E2E and type tests.

packages/rpt/src/client.ts
```diff
@@ -0,0 +1,87 @@
...
+  ): Promise<PredictResponsePayload> {
+    return this.executePrediction(predictionData, dataSchema);
+  }
+
+  /**
+   * Predict based on prediction data. Uses automatic data type parsing.
+   * @param predictionData - Data to base prediction on.
+   * @returns Prediction response.
+   */
+  async predictWithAutomaticSchemaInference(
```
marikaner: What do you think of the name? A bit long but I had no better ideas.

packages/rpt/src/index.ts
```diff
@@ -0,0 +1,7 @@
+export * from './client.js';
+export type { PredictionData } from './types.js';
+export type {
+  PredictResponseMetadata,
+  PredictResponsePayload,
+  PredictResponseStatus
```
marikaner, line 6: Those are the only generated types that are publicly exported.

packages/foundation-models/src/azure-openai/index.ts
```diff
@@ -7,4 +7,5 @@ export * from './azure-openai-chat-completion-types.js';
 export * from './azure-openai-embedding-client.js';
 export * from './azure-openai-embedding-response.js';
 export * from './azure-openai-embedding-types.js';
+export * from './deployment-resolver.js';
```
marikaner: This is only internal

sample-code/src/rpt.ts
```diff
@@ -0,0 +1,83 @@
+import { RptClient } from '@sap-ai-sdk/rpt';
+import type { PredictResponsePayload, PredictionData } from '@sap-ai-sdk/rpt';
+
+//
+// NOTES:
```
marikaner: TODO

packages/rpt/src/types.ts
```diff
@@ -0,0 +1,111 @@
...
+ * It is an object that maps the known column names to a list of values.
+ * If no data schema is given, it maps string to a list of values.
+ * @template T - Type of the data schema.
+ */
+type ColType<T extends DataSchema> = {
+  [P in keyof RowType<T>]: RowType<T>[P][];
+};
+
+//
+// This could be simplified, if `TargetColumnConfig` in the spec would set `additionalProperties` to `false`.
```
marikaner, line 54: TODO

packages/rpt/src/client.ts
davidkna-sap: Maybe `predcitWithInferredSchema`, `predictAuto`, or `predictUnsafe`? Another option would be to rename this to `predict` and call the schema-version `predictWithSchema`.

packages/rpt/src/types.ts
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * How many predictions to output per classification column.If not provided, only a single prediction is returned. Only relevant for classification.
+     */
+    top_k?: number | any;
+  }[];
+}
+
+/**
+ * Optional schema defining the data types of each column. If provided, this will override automatic data type parsing.
+ * @internal
```
davidkna-sap, line 89: [q] This is used in some types, should this be public?

```suggestion
```

packages/rpt/README.md
```diff
@@ -0,0 +1,34 @@
+# @sap-ai-sdk/rpt
+
+This package is currently experimental and subject to change.
+Do not use in production.
```
davidkna-sap: [pp]
```suggestion
> [!CAUTION]
> This package is currently experimental and subject to change.
> Do not use in production.
```

packages/core/src/model-types.ts
```diff
@@ -77,3 +77,8 @@ export type AiCoreOpenSourceChatModel = LiteralUnion<
  */
 export type AiCoreOpenSourceEmbeddingModel =
   LiteralUnion<'nvidia--llama-3.2-nv-embedqa-1b'>;
+
+/**
+ * SAP RPT models.
+ */
+export type SAPRptModel = LiteralUnion<'sap-rpt-1-small' | 'sap-rpt-1-large'>;
```
davidkna-sap: [req] Styleguide (unless there is an exception for this case):
```suggestion
export type SapRptModel = LiteralUnion<'sap-rpt-1-small' | 'sap-rpt-1-large'>;
```

packages/foundation-models/src/azure-openai/index.ts
davidkna-sap: [q] Should this be moved to the core package instead?

packages/rpt/src/types.ts
```diff
@@ -0,0 +1,111 @@
...
+ */
+type ColNames<T extends DataSchema> = T extends readonly any[]
+  ? T[number]['name']
+  : string;
+
+/**
+ * Maps the type from the spec ('numeric', 'string', 'date') to a TypeScript type.
+ * @template T - Type of the data schema.
+ */
+type TsType<T extends ColumnType> = T extends 'numeric' ? number : string;
```
davidkna-sap: [pp] I think per the API-docs dates can be typed:

```suggestion
type Digit = 0|1|2|3|4|5|6|7|8|9;
type Month = `0${1|2|3|4|5|6|7|8|9}` | `1${0|1|2}`;
type Day = `0${Digit}` | `1${Digit}` | `2${Digit}` | `30` | `31`;
type YYYY = `${Digit}${Digit}${Digit}${Digit}`;

export type DateString = `${YYYY}-${Month}-${Day}`;


type TsType<T extends ColumnType> =
  T extends 'numeric' ? number :
  T extends 'date' ? DateString : string;
```

davidkna-sap (changes requested):

packages/rpt/src/client.ts
marikaner: The service colleagues voted for keeping the name with schema simpler, so it will be more likely used. I'll use one of your suggestions. Thanks 😊

marikaner: How about `predictWithoutSchema()`?

packages/rpt/src/types.ts
marikaner: It should rather stay internal. This is only used in generic types and users should not use it directly. If they were to assert this type on their schema, they will loose all automatic type inference in our client (aka. on TS level).
I adjusted the comment a bit, it was taken from the original generated type.

packages/core/src/model-types.ts
marikaner: No there isn't, I just forgot that.

packages/rpt/src/types.ts
marikaner: nice, I didn't think of that. 👍

marikaner: Unfortunately this doesn't work because the type becomes "too complex to represent". => https://github.com/microsoft/TypeScript/issues/59015

marikaner: But this less restrictive version works: type DateString = `${number}${number}${number}${number}-${number}${number}-${number}${number}`;

packages/foundation-models/src/azure-openai/index.ts
marikaner: I was thinking the same thing. I see no blockers to put it there.

marikaner: Well, I just checked and it is a problem, this would introduce a circular dependency, because the deployment resolution functionality is in ai-api. So by moving to core I would get core > ai-api > core.
I could move it to ai-api or duplicate this to rpt.
We also have a deployment resolver in orchestration. If I move the foundation one to ai-api, I am thinking to also move the other one.

ZhongpinWang: I would not move the deployment resolver to core as the logic of the `getXYZDeploymentId()` from foundation models and orchestration are a bit different and more package specific.

For example, in foundation model package, deployment id can be resolved by model name + scenarioId, whereas in orchestration, scenario id is fixed.

marikaner: @ZhongpinWang the plan is only to move this specific function and not to generalize it. I created #1473 to do this. Alternatively, I could just copy the code to the rpt client package. @davidkna-sap and I agreed, that it doesn't really matter and decided to move the code. If you have a strong opinion let me know.

packages/rpt/src/types.ts
marikaner: Still waiting for a spec update, it should happen this week.

packages/foundation-models/src/azure-openai/index.ts
ZhongpinWang: I totally see the point of moving this function for enabling the reuse in rpt client. Do we then have a plan to move orchestration resolver as well? I only feel a bit uncomfortable to make ai-api less general and introduce foundation model or orchestration concept in it. But I am okay if we keep the change consistent.

.github/workflows/spec-update-cron.yaml
```diff
@@ -25,3 +25,21 @@ jobs:
...
+  # see comments in spec-update.yaml
+  # get-rpt-tag:
+  #   runs-on: ubuntu-latest
+  #   outputs:
+  #     tag: ${{ steps.get-tag.outputs.tag }}
+  #   steps:
+  #     - name: Get latest tag
+  #       id: get-tag
+  #       run: |
+  #         TAG=$(git -c 'versionsort.suffix=-' ls-remote --tags --sort='v:refname' https://github.tools.sap/DL-COE/sap-rpt-1-public-content | tail --lines=1 | cut -f 2)
```
ZhongpinWang, line 37: [q] I hope the tag naming will follow the same pattern in the future and there won't be any strange tag starting with an alphabet or something like this?

packages/rpt/src/client.ts
ZhongpinWang: I vote for `predict()` without schema, and `predictWithSchema()`. Did the service colleagues suggest any default behaviour? Like in most cases user needs / doesn't need to provide a schema?

In general, I feel it might be too complex for user to understand what inference / auto / unsafe means. Or when they use pure JS, it would be strange.

cbiehlsap: Ideally, users who know the data types should communicate them to the API by passing the `data_schema` parameter in the JSON (since type inference is never 100% perfect in all cases). Is there a good way to make this clear in the code or SDK docs?

From the usability point of view, you would likely know better than me what is best here. I just feared that users would import `predict` and just ignore the other function (even if they know the data types). If you want to give it another name than `predict`, I would probably try to keep the name consistent between both functions, like `predictWithSchema` and `predictWithoutSchema`. One could also consider logging a message that recommends passing the schema and shows a little example for what the parameter should look like.

packages/foundation-models/src/azure-openai/index.ts
marikaner: I had the same concerns, but I did not plan to move it. I don't have a strong opinion about this inconsistency, that is why I left it. I can move it though, if you think that's better (which I would not disagree with ;) ).

packages/rpt/src/client.ts
marikaner: I agree with the take that `predict()`  will be the goto method and if it shouldn't be the one without schema. Default case is with schema and the deviation is, you don't wanna say. I could live with naming both functions consistently as @cbiehlsap suggested.

.github/workflows/spec-update-cron.yaml
marikaner: The team is open to meet our needs here. They also suggested to keep the latest released spec on main instead, but then we don't have a channel to try the latest changes if needed.

packages/rpt/src/client.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    const body = {
+      data_schema:
+        dataSchema?.reduce(
+          (merged, { name, ...schemaFieldConfig }) => ({
+            ...merged,
+            [name]: schemaFieldConfig
+          }),
+          {}
+        ) || null,
```
davidkna-sap: [pp]
```suggestion
    dataSchema
        ? Object.fromEntries(
            dataSchema.map(({ name, ...schemaFieldConfig }) => [name, schemaFieldConfig])
        )
        : null,
```

packages/rpt/src/client.test.ts
davidkna-sap: For completeness sake, I would prefer having a unit test for `predictWithoutSchema` here, but handling that in a follow-up would be fine.

packages/rpt/src/client.ts
ZhongpinWang: I will go with @cbiehlsap's idea and call them `predictWithSchema` and `predictWithoutSchema`.
Also added hints in the jsdoc to prefer using the one with schema when it is known.

packages/foundation-models/src/azure-openai/index.ts
ZhongpinWang: I moved it to keep it clean. 🧹

packages/rpt/src/client.ts
ZhongpinWang: Nice improvement

davidkna-sap:
This needs to be published manually, or have a preliminary stub release because the package can't have trusted publishing configured until it is published.

packages/rpt/README.md
```diff
@@ -0,0 +1,35 @@
+# @sap-ai-sdk/rpt
+
+> [!CAUTION]
+> This package is currently experimental and subject to change.
+> Do not use in production.
```
davidkna-sap, line 5:
```suggestion

<!-- vale off -->
> [!CAUTION]
> This package is currently experimental and subject to change.
> Do not use in production.
<!-- vale on -->
```

ZhongpinWang: Any specific reason to vale this part off?

davidkna-sap: There was a vale warning about !C (`[SAP.Spacing] '!C' should have one space.`)

ZhongpinWang: Okay, this always happens and perhaps we can write a vale rule using regex. I still won't completely disable it just in case we change some text in the future inside this block and still have vale turned off. Warning is fine. Thanks for checking!

PR #1424, opened 2026-01-14
chore: Add a dynamic model example

emincihangeri:
## Context

Closes SAP/ai-sdk-js-backlog#405.


sample-code/src/langchain-orchestration.ts
```diff
@@ -391,6 +392,55 @@ export async function invokeToolChain(): Promise<string> {
...
+    }
+  });
+
+  const agent = createAgent({
+    model: basicModel,
+    tools: [],
+    middleware: [dynamicModelSelection]
+  });
+
+  const inputMessage = 'What is the SAP?';
```
davidkna-sap: [np]

```suggestion
  const inputMessage = 'What is SAP?';
```

```suggestion
  const inputMessage = 'What is the SAP Cloud SDK?';
```

PR #1432, opened 2026-01-15
feat: Support `.withStructuredOutput` in orchestration langchain client

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#355.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "@types/jest": "^30.0.0",
     "@types/jsonwebtoken": "^9.0.10",
     "@types/mock-fs": "^4.13.4",
-    "@types/node": "^22.19.9",
+    "@types/node": "^22.19.10",
     "depcheck": "^1.4.7",
     "eslint": "^9.39.2",
     "glob": "^13.0.1",
--- a/packages/langchain/package.json
+++ b/packages/langchain/package.json
@@ -40,8 +40,8 @@
     "uuid": "^13.0.0"
   },
   "devDependencies": {
-    "@langchain/langgraph": "^1.1.3",
-    "zod": "^4.3.5"
+    "@langchain/langgraph": "^1.1.4",
+    "zod": "^4.3.6"
   },
   "peerDependencies": {
     "@langchain/core": "^1.1.16"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -12,7 +12,7 @@ importers:
     devDependencies:
       '@changesets/cli':
         specifier: ^2.29.8
-        version: 2.29.8(@types/node@22.19.9)
+        version: 2.29.8(@types/node@22.19.10)
       '@jest/globals':
         specifier: ^30.2.0
         version: 30.2.0
@@ -56,8 +56,8 @@ importers:
         specifier: ^4.13.4
         version: 4.13.4
       '@types/node':
-        specifier: ^22.19.9
-        version: 22.19.9
+        specifier: ^22.19.10
+        version: 22.19.10
       depcheck:
         specifier: ^1.4.7
         version: 1.4.7
@@ -69,7 +69,7 @@ importers:
         version: 13.0.1
       jest:
         specifier: ^30.2.0
-        version: 30.2.0(@types/node@22.19.9)(ts-node@10.9.2(@types/node@22.19.9)(typescript@5.9.3))
+        version: 30.2.0(@types/node@22.19.10)(ts-node@10.9.2(@types/node@22.19.10)(typescript@5.9.3))
       jsonwebtoken:
         specifier: ^9.0.3
         version: 9.0.3
@@ -87,10 +87,10 @@ importers:
         version: 3.8.1
       ts-jest:
         specifier: ^29.4.6
-        version: 29.4.6(@babel/core@7.29.0)(@jest/transform@30.2.0)(@jest/types@30.2.0)(babel-jest@30.2.0(@babel/core@7.29.0))(jest-util@30.2.0)(jest@30.2.0(@types/node@22.19.9)(ts-node@10.9.2(@types/node@22.19.9)(typescript@5.9.3)))(typescript@5.9.3)
+        version: 29.4.6(@babel/core@7.29.0)(@jest/transform@30.2.0)(@jest/types@30.2.0)(babel-jest@30.2.0(@babel/core@7.29.0))(jest-util@30.2.0)(jest@30.2.0(@types/node@22.19.10)(ts-node@10.9.2(@types/node@22.19.10)(typescript@5.9.3)))(typescript@5.9.3)
       ts-node:
         specifier: ^10.9.2
-        version: 10.9.2(@types/node@22.19.9)(typescript@5.9.3)
+        version: 10.9.2(@types/node@22.19.10)(typescript@5.9.3)
       tsx:
         specifier: ^4.21.0
         version: 4.21.0
@@ -180,10 +180,10 @@ importers:
         version: 13.0.0
     devDependencies:
       '@langchain/langgraph':
-        specifier: ^1.1.3
-        version: 1.1.3(@langchain/core@1.1.19(openai@6.18.0(ws@8.19.0)(zod@4.3.6)))(zod-to-json-schema@3.25.1(zod@4.3.6))(zod@4.3.6)
+        specifier: ^1.1.4
+        version: 1.1.4(@langchain/core@1.1.19(openai@6.18.0(ws@8.19.0)(zod@4.3.6)))(zod-to-json-schema@3.25.1(zod@4.3.6))(zod@4.3.6)
       zod:
-        specifier: ^4.3.5
+        specifier: ^4.3.6
         version: 4.3.6
 
   packages/orchestration:
@@ -247,8 +247,8 @@ importers:
         specifier: workspace:^
         version: link:../packages/orchestration
       '@sap/cds':
-        specifier: ^9.7.0
-        version: 9.7.0(@eslint/js@9.39.2)
+        specifier: ^9.7.1
+        version: 9.7.1(@eslint/js@9.39.2)
       '@sap/xssec':
         specifier: ^4.12.2
         version: 4.12.2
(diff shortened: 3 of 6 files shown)


.changeset/clever-queens-bet.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/langchain': minor
+---
+
+[feat] Support `withStructuredOutput()` method in the Orchestration LangChain client.
```
marikaner: [pp] I prefer full sentences, but I see other people not doing it, so up to you ;)

```suggestion
[feat] Support the `withStructuredOutput()` method in the Orchestration LangChain client.
```

packages/langchain/src/orchestration/client.ts
```diff
@@ -171,6 +189,138 @@ export class OrchestrationClient extends BaseChatModel<
...
+      | 'jsonMode';
+    const name = config?.name ?? 'extract';
+    const description = getSchemaDescription(outputSchema);
+    const strict = config?.strict;
+    const includeRaw = config?.includeRaw;
+
+    let llm: Runnable<BaseLanguageModelInput>;
+    let outputParser: Runnable<AIMessageChunk, RunOutput>;
+
+    // jsonSchema method: Use orchestration's native responseFormat
```
ZhongpinWang: [pp/req] Can we move this comment to the below line 279? Was a bit confusing.
```
else if (method === 'jsonSchema') {
```

packages/langchain/src/orchestration/client.ts
```diff
@@ -171,6 +189,138 @@ export class OrchestrationClient extends BaseChatModel<
...
+        })
+      });
+      llm = this.withConfig({
+        tools: [
+          {
+            type: 'function' as const,
+            function: {
+              name,
+              description,
+              parameters: jsonSchema,
```
ZhongpinWang, line 265: [pp] I would put `toJsonSchema(outputSchema)` here inline. Because putting this before the whole big `if else` block with `jsonMode` not using it feels unclean.

packages/langchain/src/orchestration/client.ts
```diff
@@ -171,6 +189,138 @@ export class OrchestrationClient extends BaseChatModel<
...
+      outputParser = isInteropZodSchema(outputSchema)
+        ? StructuredOutputParser.fromZodSchema(outputSchema)
+        : new JsonOutputParser<RunOutput>();
+      llm = this.withConfig({
+        responseFormat: {
+          type: 'json_schema' as const,
+          json_schema: {
+            name,
+            description,
+            schema: jsonSchema,
```
ZhongpinWang, line 300: here as well

packages/langchain/src/orchestration/client.ts
```diff
@@ -171,6 +189,138 @@ export class OrchestrationClient extends BaseChatModel<
...
+      if (strict !== undefined) {
+        throw new Error(
+          'The "strict" option is not supported with the "jsonMode" structured output method. Please use "jsonSchema" or "functionCalling" methods for strict output instead.'
+        );
+      }
+      outputParser = isInteropZodSchema(outputSchema)
+        ? StructuredOutputParser.fromZodSchema(outputSchema)
+        : new JsonOutputParser<RunOutput>();
+      llm = this.withConfig({
+        responseFormat: {
```
ZhongpinWang, line 284: [q/req] We might missed `ls_structured_output_format` here

See https://github.com/SAP/ai-sdk-js/blob/e4c15d569d04f53a70ecb2d921896ad4b9fcddb9/packages/langchain/src/openai/chat.ts#L221

ZhongpinWang: Please check other occurrences with our existing openai langchain client or official client.

packages/langchain/src/orchestration/client.ts
```diff
@@ -171,6 +189,138 @@ export class OrchestrationClient extends BaseChatModel<
...
+          type: 'json_schema' as const,
+          json_schema: {
+            name,
+            description,
+            schema: jsonSchema,
+            ...(strict !== undefined && { strict })
+          }
```
ZhongpinWang, line 302: [q/req] Same here, missing `ls_structured_output_format`, please double check 

https://github.com/SAP/ai-sdk-js/blob/e4c15d569d04f53a70ecb2d921896ad4b9fcddb9/packages/langchain/src/openai/chat.ts#L238C9-L238C36

packages/langchain/src/orchestration/client.ts
```diff
@@ -171,6 +189,138 @@ export class OrchestrationClient extends BaseChatModel<
...
+        }
+      } satisfies Partial<OrchestrationCallOptions>);
+    } else {
+      method satisfies never;
+      throw new Error(
+        `Unsupported structured output method: ${method}. Supported methods are 'jsonSchema', 'functionCalling', and 'jsonMode'.`
```
ZhongpinWang, line 309: [q] The langchain official openai client uses function calling as the last check and "fallback"/default if not the other two methods are defined.

Logically our implementation is also not wrong, just checking if this is aligned with other langchain client behaviour.

packages/langchain/src/orchestration/client.ts
```diff
@@ -171,6 +189,138 @@ export class OrchestrationClient extends BaseChatModel<
...
+          zodSchema: outputSchema as InteropZodType<RunOutput>
+        })
+      });
+      llm = this.withConfig({
+        tools: [
+          {
+            type: 'function' as const,
+            function: {
+              name,
+              description,
```
ZhongpinWang, line 264: [q] I guess this `getSchemaDescription()` should return the same as `toJsonSchema(..).description`?

https://github.com/SAP/ai-sdk-js/blob/e4c15d569d04f53a70ecb2d921896ad4b9fcddb9/packages/langchain/src/openai/chat.ts#L268

packages/langchain/src/orchestration/client.ts
```diff
@@ -171,6 +189,138 @@ export class OrchestrationClient extends BaseChatModel<
...
+              parameters: jsonSchema,
+              ...(strict !== undefined && { strict })
+            }
```
ZhongpinWang, line 267: [pp/req] Similar here missing `ls_structured_output_format`

https://github.com/SAP/ai-sdk-js/blob/e4c15d569d04f53a70ecb2d921896ad4b9fcddb9/packages/langchain/src/openai/chat.ts#L279

packages/langchain/src/orchestration/client.ts
```diff
@@ -171,6 +189,138 @@ export class OrchestrationClient extends BaseChatModel<
...
+          }
+        ]
+        // TODO: Set `tool_choice` if it becomes supported in Orchestration
```
ZhongpinWang: [q] Do you have a bruno collection or something saying this is not supported? But I guess not a big deal as there is anyway only one tool..

packages/langchain/src/orchestration/client.ts
davidkna-sap: @ZhongpinWang I should have added a comment for this - IIRC `ls_structured_output_format` was causing some type-issues, and I considered it less important (langsmith), but I will investigate again.

packages/langchain/src/orchestration/client.ts
davidkna-sap: `tool_choice` can be used via `params`, but that will always be model-dependent.

packages/langchain/src/orchestration/client.ts
davidkna-sap: I mostly used it for alignment with upstream, `getSchemaDescription()` has specific handling for reading description directly from zod v4 and v3 and verifies that the description is a string if a JSON-schema is provided instead of a zod schema.

packages/langchain/src/orchestration/client.ts
davidkna-sap: I wouldn't be opposed to changing this, but I saw that for some of the non-openai in-langchain-repo clients, they would also throw if an unsupported method was provided.

I considered this to be the safer choice in case they decide to expand this in the future. The general default is also supposed to be  `functionCalling` if no method is provided, but our OpenAI client is also redirecting to `jsonSchema` and it's a popular option with other clients.

I haven't checked how this interacts with [`providerStrategy`](https://docs.langchain.com/oss/javascript/langchain/structured-output#provider-strategy) yet ([see also](https://github.com/SAP/ai-sdk-js-backlog/issues/455)) - without `tool_choice` it might be more brittle than necessary if they default us to `functionCalling`.

packages/langchain/src/orchestration/client.ts
davidkna-sap: I did not move it inline because all paths (apart from invalid `method`) now require `jsonSchema`

packages/langchain/src/orchestration/client.ts
davidkna-sap: I added `ls_structured_output_format`-handling.

packages/langchain/src/orchestration/client.ts
davidkna-sap: During testing some of the smaller models were not using the function call.

packages/langchain/src/orchestration/client.ts
ZhongpinWang: Does `jsonMode` case need this?

davidkna-sap: It needs it for `ls_structured_output_format`

ZhongpinWang: I also found this out seconds ago 😄

packages/langchain/src/orchestration/client.ts
ZhongpinWang: Then we keep the current implementation as-is. Also not a big deal imo.

packages/langchain/src/orchestration/client.ts
ZhongpinWang: Can you maybe create a list of models which does not support the `tool_choice` and provide this feedback to llm-orchestration? Users will ask question why they used function call or provided `tool_choice` but it did not work. Thanks.

sample-code/src/langchain-orchestration.ts
```diff
@@ -492,3 +492,57 @@ export async function invokeMcpToolChain(): Promise<string> {
...
+  includeRaw = false as T
+): Promise<
+  T extends true
+    ? {
+        raw: BaseMessage;
+        parsed: {
+          setup: string;
+          punchline: string;
+          rating: number;
+        };
```
ZhongpinWang: [pp] Tiny suggestion, maybe create a small type for `{ setup: ..., punchline: ..., rating: ... }`?

I am kinda surprised as well that LangChain hasn't typed `{ raw: ..., parsed: ...}` 😄

ZhongpinWang:
Out of curiosity, was this missing feature reported by langchain standard test?

packages/langchain/src/orchestration/client.ts
davidkna-sap: @ZhongpinWang I mostly meant this in the sense that the option is also handled differently for e.g. Antrophic and Google models. (Smaller models are more likely not to call the function or not output json without `tool_choice`)

sample-code/src/langchain-orchestration.ts
davidkna-sap: IIRC it is typed, but TS is not smart enough to properly resolve it in this case.

PR #1454, opened 2026-01-21
feat: Support orchestration prompt module fallback

davidkna-sap:
## Summary
- Add support for orchestration prompt module fallback by allowing a list of module configurations when constructing an `OrchestrationClient`

Handles non-streaming part of https://github.com/SAP/ai-sdk-js-backlog/issues/461

The main complexity is related to handling streaming options.

## Test plan
- Added comprehensive test coverage for module fallback scenarios
- Tests verify fallback behavior with multiple module configurations
- Tests confirm proper error handling when all modules in the list fail

> [!WARNING]
> Streaming with module fallbacks is handled in #1470 (targeting this PR)


sample-code/src/server.ts
```diff
@@ -323,6 +325,14 @@ app.get('/orchestration/:sampleCase', async (req, res) => {
         .send(
           `Embedding with masking applied successfully:${JSON.stringify(embeddingResult.getIntermediateResults()?.input_masking?.data, null, 2)}\nEmbeddings: ${embedding}\nUsage - Prompt tokens: ${embeddingResult.getTokenUsage()?.prompt_tokens}\nUsage - Total tokens: ${embeddingResult.getTokenUsage()?.total_tokens}`
         );
+    } else if (sampleCase === 'fallbackModules') {
+      const intermediateFailures = (
+        result as OrchestrationResponse
+      ).getIntermediateFailures();
+      const content = (result as OrchestrationResponse).getContent();
+      res.send(
```
KavithaSiva: [pp] Add the `text/plain` header to make the output readable.

packages/orchestration/src/util/module-config.test.ts
```diff
@@ -265,3 +268,103 @@ describe('constructCompletionPostRequestFromConfigReference', () => {
...
+
+describe('constructCompletionPostRequest with module fallback configs', () => {
+  const primaryConfig: OrchestrationModuleConfig = {
+    promptTemplating: {
+      prompt: {
+        template: [{ role: 'user', content: 'Hello {{?name}}' }]
+      },
+      model: {
+        name: 'gpt-4o',
+        params: { max_tokens: 100, timeout: 5 }
```
KavithaSiva: [q] `timeout` property can be set outside `params`.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -220,6 +220,28 @@ export interface OrchestrationModuleConfig {
...
 }
 
+/**
+ * Non-empty list of orchestration module configurations for module fallback.
+ * The orchestration service will try each configuration in order until one succeeds.
+ * @example
+ * const fallbackConfig: OrchestrationModuleConfigList = [
+ *   {
+ *     promptTemplating: {
+ *       model: { name: 'gpt-4o', params: { timeout: 5 } }
```
KavithaSiva:
```suggestion
 *       model: { name: 'gpt-4o', timeout: 5  }
```

packages/orchestration/src/orchestration-client.ts
```diff
@@ -44,23 +48,45 @@ export class OrchestrationClient {
...
+          `Fallback configurations contain duplicate models: [${models.join(', ')}]. ` +
+            'Consider using different models for meaningful fallback behavior.'
+        );
+      }
     } else if (!isConfigReference(config)) {
-      this.config =
-        typeof config.promptTemplating.prompt === 'string'
-          ? this.parseAndMergeTemplating(config)
-          : config;
+      this.config = parseOrchestrationModuleConfig(config);
```
KavithaSiva: [req] The orchestration config handling logic is not easily readable, consider splitting based on the config type. 

Consider moving `parseOrchestrationModuleConfig` to a dedicated method, renaming it, `parseTemplatingConfig` or `parseTemplatingModule` is clearer.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -259,15 +281,41 @@ export type OrchestrationConfigRef = Xor<
...
+    | string
+    | OrchestrationConfigRef
+): config is OrchestrationModuleConfigList {
+  const isArray = Array.isArray(config);
+  if (isArray && config.length === 0) {
+    throw new Error('Configuration array must not be empty.');
+  }
+  return isArray;
+}
+
```
KavithaSiva, line 355: [req] Please also check if the elements of the array contains properties that make it a `OrchestrationModuleConfig`.

Please also remove throwing the error, as type guards usually don't throw errors but are just type validators.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -168,6 +200,15 @@ export class OrchestrationClient {
       requestConfig,
       this.destination
     );
+
+    // Log summary when fallbacks were used
+    if (!stream && response.data?.intermediate_failures?.length > 0) {
+      logger.info(
+        `Orchestration used ${response.data.intermediate_failures.length} fallback(s) before success`
```
KavithaSiva: [suggestion] Consider logging which configuration/model succeeded.

packages/orchestration/src/util/module-config.test.ts
davidkna-sap: Sorry, I had already fixed this somewhere else but failed to clean it up everywhere.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -257,4 +283,51 @@ export class OrchestrationClient {
...
+   * @returns The validated and parsed configuration list.
+   * @throws {Error} If the array is empty or contains invalid elements.
+   */
+  private parseModuleConfigList(
+    config:
+      | OrchestrationModuleConfig
+      | OrchestrationModuleConfigList
+      | string
+      | OrchestrationConfigRef
+  ): OrchestrationModuleConfigList {
```
KavithaSiva, line 310: [pp] Won't the array check above already rule out the other config types?
```suggestion
  private parseModuleConfigList(
    config:
      OrchestrationModuleConfigList
  ): OrchestrationModuleConfigList {
```

packages/orchestration/src/orchestration-types.ts
```diff
@@ -259,15 +281,79 @@ export type OrchestrationConfigRef = Xor<
...
+    | string
+    | OrchestrationConfigRef
+): config is OrchestrationModuleConfigList {
+  try {
+    assertIsOrchestrationModuleConfigList(config);
+    return true;
+  } catch {
+    return false;
+  }
+}
```
KavithaSiva, line 354: [pp] I would have personally preferred the other way around, the type guard being used in the assert function with the additional error throwing.

davidkna-sap: I feel like doing it the other way around in this case is better, because it allows for more differentiated error types and messages.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -259,15 +281,79 @@ export type OrchestrationConfigRef = Xor<
...
+
+/**
+ * Type guard to check if config is a valid list of orchestration module configs.
+ * @internal
+ * @param config - The config to check.
+ * @returns True if config is a non-empty array with valid OrchestrationModuleConfig elements.
+ */
+export function isOrchestrationModuleConfigList(
```
KavithaSiva, line 341: [q] Should we consider exporting this publicly? Considering we export `isConfigReference` ?

packages/orchestration/src/orchestration-types.ts
KavithaSiva: True.

PR #1464, opened 2026-01-22
feat: Add `rawResponse` property to `OrchestrationStreamResponse`& `getRequestId`-helpers

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#457.

### DIFF
--- a/.changeset/brave-lions-dance.md
+++ b/.changeset/brave-lions-dance.md
@@ -4,4 +4,3 @@
 
 [feat] Add `rawResponse` property to `OrchestrationStreamResponse`.
 The new property exposes the raw HTTP response from the orchestration service for advanced use cases.
-SSE data is not part of the immediate HTTP response.


packages/orchestration/src/orchestration-stream-response.ts
```diff
@@ -18,6 +19,12 @@ export class OrchestrationStreamResponse<T> {
   public _data: Partial<CompletionPostResponse> = {};
   private _stream: OrchestrationStream<T> | undefined;
 
+  /**
+   * Creates an orchestration stream response.
+   * @param rawHttpResponse - The raw HTTP response from the orchestration service. This does not containy any of the SSE stream payload data.
+   */
+  constructor(public readonly rawHttpResponse: HttpResponse) {}
```
davidkna-sap: I called it `rawHttpResponse` instead of `rawResponse` to stress that it doesn't contain the actual data being streamed.

packages/orchestration/src/orchestration-stream-response.test.ts
davidkna-sap, line 1: This file did not have tests yet, the AI decided to generate a bunch for things that were not covered.
 I kept these to increase overall code coverage, but I would be glad to remove additional tests unrelated to the primary PR objective to ease reviewing.

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -123,6 +123,9 @@ describe('orchestration', () => {
 
   it('should return stream of orchestration responses', async () => {
     const response = await chatCompletionStream(new AbortController());
+    expect(response.rawHttpResponse.config.headers['X-Request-ID']).toBe(
+      'custom-request-id-123'
+    );
```
davidkna-sap: The Orchestration HTTP response does not set or forward `'X-Request-ID'`, but it can be accessed via `response.rawHttpResponse.config.headers` if it is included in the request.

packages/orchestration/src/orchestration-stream-response.ts
jerome-benoit: It breaks namespace consistency across the API.
The stream use case is orthogonal, so the collision is not a concern here.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -186,6 +183,11 @@ export class OrchestrationClient {
       }
     });
 
+    const response =
+      new OrchestrationStreamResponse<OrchestrationStreamChunkResponse>(
+        streamResponse
```
ZhongpinWang, line 188: [pp/req] Instead of putting it in the constructor, how about calling a setter afterwards. The workaround below `OrchestrationStreamResponse({} as any)` is not elegant.

packages/orchestration/src/orchestration-stream-response.ts
ZhongpinWang: This `HttpResponse` contains `data` property, which is a Node.js `Readable` stream object set by Axios.

See how we are parsing the SSE.

https://github.com/SAP/ai-sdk-js/blob/4c5c6b6ea629f7f5c163cf32bd6d1cac97705020/packages/orchestration/src/orchestration-stream.ts#L23

https://github.com/SAP/ai-sdk-js/blob/4c5c6b6ea629f7f5c163cf32bd6d1cac97705020/packages/core/src/stream/sse-stream.ts#L101

The idea is to expose this `HttpResponse` object in `OrchestrationStreamResponse` so that users can parse the `data` SSE by themselves. And if they do parse by themselves, then `OrchestrationStreamResponse` cannot be used but then it is user's choice.

But please double check, if `data` is consumable. My understanding is that if user consumes the stream by running a for await loop, then the stream is not consumable any more. So please check if it is not consumed by a for await, then can the `Readable` object still be consumed directly from the raw response.

packages/orchestration/src/orchestration-stream.test.ts
```diff
@@ -65,7 +65,7 @@ describe('Orchestration chat completion stream', () => {
           () => asyncGeneratorChunk,
           new AbortController()
         ),
-        new OrchestrationStreamResponse()
+        new OrchestrationStreamResponse({} as any)
       );
```
ZhongpinWang, line 78: [pp/req] Unfortunately, we have exposed `OrchestrationStreamResponse` not just as a type. Adding an additional required property in the constructor is a breaking change. Please consider using setter later.

Or perhaps we accept this as a compatibility note. WDYT?

sample-code/src/orchestration.ts
```diff
@@ -88,7 +88,10 @@ export async function chatCompletionStream(
       ]
     },
     controller.signal,
-    streamOptions
+    streamOptions,
+    {
+      headers: { 'X-Request-ID': 'custom-request-id-123' }
+    }
```
ZhongpinWang: [req] I don't think adding this in sample code is necessary. Sample code should be simple and the usage should be straightforward. Consider a new user looking at this header, they would wonder do I also need to set this header when using our SDK.

tests/e2e-tests/src/orchestration.test.ts
ZhongpinWang: [req] Once we expose raw response `HttpResponse`, users can read headers, data, etc out themselves. Check the type `HttpResponse`, it contains `data`, `status`, `headers`, `request` and other any string fields. 

Image

ZhongpinWang: And I don't fully understand where does the `config` property come from.

packages/orchestration/src/orchestration-stream-response.test.ts
ZhongpinWang: [pp] I prefer adding this test file in a separate PR. This file is a bit too much to be considered as a small improvement. But thanks for considering the test coverage and please do create a follow-up for this. 👍

packages/orchestration/src/orchestration-stream-response.ts
ZhongpinWang: Also please call this property `rawResponse` same as in the non-streaming response. (I see that we have some inconsistency between embedding and chat client... embedding response calls this `response` instead of `rawResponse`... But let's be at least consistent within one client.)

ZhongpinWang: [pp] If you want, consider deprecate `response` in embedding client and rename it to `rawResponse` and check docs. Cosmetic stuff cc @KavithaSiva WDYT?

packages/orchestration/src/orchestration-stream.test.ts
davidkna-sap: I did add a compatibility note. 

I decided on doing it this way because it means that the new property does not have to be optional, and `OrchestrationResponse` also add the raw response via a constructor argument.
And while this is a breaking change at the type-level, it doesn't break anything at runtime.

ZhongpinWang: Make sense. Would of course be nicer if we have exported this as a type only https://github.com/SAP/ai-sdk-js/blob/4c5c6b6ea629f7f5c163cf32bd6d1cac97705020/packages/orchestration/src/index.ts#L46

Perhaps something you can consider doing, also for other response types.

packages/orchestration/src/orchestration-stream-response.ts
davidkna-sap: @jerome-benoit Please clarify what you mean with "namespace consistency across the API" in this context. Would you prefer it being called `rawResponse`? Given the other review comments I will be renaming it to `rawResponse`.

tests/e2e-tests/src/orchestration.test.ts
davidkna-sap: I found `config` after investigating the response object, I didn't realise it wasn't an official API.

packages/orchestration/src/orchestration-stream-response.ts
jerome-benoit: > @jerome-benoit Please clarify what you mean with "namespace consistency across the API" in this context. Would you prefer it being called `rawResponse`? Given the other review comments I will be renaming it to `rawResponse`.

Two different transport layers - slightly -, same overall semantic expected on both content => the namespace should match.

davidkna-sap: I have confirmed `data` can be streamed, but decided against adding an e2e-test for this feature for now. Edit: I added a unit test for this.

packages/orchestration/src/orchestration-stream.test.ts
davidkna-sap: I've changed the code to handle this gracefully by deprecating the constructor without the `OrchestrationResponse`, and throwing if `OrchestrationResponse` is `undefined` when `rawResponse` is accessed.

.changeset/purple-dolphins-swim.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[compat] Support for constructing an `OrchestrationStreamResponse` without an `HttpResponse` has been deprecated, and will be removed in the next major release.
+Code directly instantiating this class should be updated to provide an `HttpResponse` object as the first parameter.
```
ZhongpinWang:
```suggestion
Code directly instantiating this class should be updated to provide an `HttpResponse` object as the first parameter to allow reading from raw HTTP response.
```

.changeset/brave-lions-dance.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[feat] Add `rawResponse` property to `OrchestrationStreamResponse`.
+The new property exposes the raw HTTP response from the orchestration service for advanced use cases.
+SSE data is not part of the immediate HTTP response.
```
ZhongpinWang:
```suggestion
```
I would remove this just because if people don't understand what "immediate" means, they maybe confused if they can iterate `data` later by themselves.

PR #1470, opened 2026-01-23
feat: Support orchestration prompt module fallback (with stream)

davidkna-sap:
## Context

Closes https://github.com/SAP/ai-sdk-js-backlog/issues/461

Streaming-related part of #1454

### DIFF
--- a/packages/orchestration/src/orchestration-client.test.ts
+++ b/packages/orchestration/src/orchestration-client.test.ts
@@ -1572,7 +1572,7 @@ describe('orchestration service client', () => {
         overrides: [
           { promptTemplating: { include_usage: true } },
           { promptTemplating: { include_usage: false } }
-        ] as any
+        ]
       };
 
       mockInference(
--- a/packages/orchestration/src/orchestration-types.ts
+++ b/packages/orchestration/src/orchestration-types.ts
@@ -399,7 +399,9 @@ export interface BaseStreamOptions {
 
 /**
  * Stream options with support for per-config overrides in module fallback scenarios.
- * The `overrides` property allows you to specify stream options that apply to specific module configurations by their index in the fallback list. If an override is not provided for a config, it will use the shared options defined in `BaseStreamOptions`.
+ * The `overrides` property allows you to specify stream options that apply to specific module configurations by their index in the fallback list.
+ * If an override is not provided for a config, it will use the shared options defined in `BaseStreamOptions`.
+ * Overrides do not merge with shared options.
  */
 export type StreamOptionsWithOverrides = Omit<
   BaseStreamOptions,
@@ -416,9 +418,9 @@ export type StreamOptionsWithOverrides = Omit<
    *   global: { chunk_size: 100 },
    *   promptTemplating: { include_usage: false },
    *   overrides: {
-   *     0: { promptTemplating: { include_usage: true } },  // config 0
-   *     // 1: uses base configuration
-   *     2: { outputFiltering: { overlap: 50 } }            // config 2
+   *     0: { promptTemplating: { include_usage: true } },  // config 0, replaces base options fully
+   *     // 1: resolves base options because no override is provided (`{ promptTemplating: { include_usage: false } }`)
+   *     2: { outputFiltering: { overlap: 50 } }            // config 2, replaces base options fully
    *   }
    * };
    * ```
--- a/packages/orchestration/src/util/module-config.ts
+++ b/packages/orchestration/src/util/module-config.ts
@@ -277,7 +277,7 @@ export function addStreamOptions(
     : [moduleConfigs];
 
   if (!Array.isArray(moduleConfigs) && streamOptions?.overrides) {
-    logger.warn(
+    throw new Error(
       'Overrides in stream options are not supported when a single module configuration is provided.'
     );
   }
--- a/packages/orchestration/src/util/stream.ts
+++ b/packages/orchestration/src/util/stream.ts
@@ -64,10 +64,10 @@ function mergeIntermediateFailures(
 
   // Deduplicate messages with structural equality checks
   const existingSet = new Set(
-    existing.map(f => JSON.stringify(Object.entries(f).sort()))
+    existing.map(f => JSON.stringify([f.request_id, f.code, f.message]))
   );
   const newFailures = transformedIncoming.filter(
-    f => !existingSet.has(JSON.stringify(Object.entries(f).sort()))
+    f => !existingSet.has(JSON.stringify([f.request_id, f.code, f.message]))
   );
   return [...existing, ...newFailures];
 }
--- a/packages/orchestration/src/util/module-config.test.ts
+++ b/packages/orchestration/src/util/module-config.test.ts
@@ -423,6 +423,20 @@ describe('addStreamOptions with module fallback configs', () => {
     expect(result.stream).toEqual({ enabled: true, chunk_size: 100 });
   });
 
+  it('should throw when overrides are provided for a single module config', () => {
+    const config = createModuleConfig('gpt-4o');
+    const streamOptions: StreamOptions = {
+      overrides: {
+        0: { promptTemplating: { include_usage: true } }
+      }
+    };
+
+    // @ts-expect-error overrides are not supported for single configs
+    expect(() => addStreamOptions(config, streamOptions)).toThrow(
+      'Overrides in stream options are not supported when a single module configuration is provided.'
+    );
+  });
+
   it('should warn only once when output filtering options are set but no config has filtering', () => {
     const logger = createLogger({
       package: 'orchestration',


packages/orchestration/src/orchestration-client.ts
```diff
@@ -158,12 +153,19 @@ export class OrchestrationClient {
...
+                request,
+                stream,
+                streamOptions
+              )
+            : constructCompletionPostRequest(
+                this.config,
+                request,
+                stream,
+                streamOptions as StreamOptions | undefined
+              );
```
KavithaSiva, line 170: [req] Why is this conditional required? Don't you have overloads for the function, can't the type be automatically inferred?

packages/orchestration/src/util/module-config.ts
```diff
@@ -126,61 +126,181 @@ export function addStreamOptionsToOutputFilteringConfig(
...
-  streamOptions?: StreamOptions
+  moduleConfigs: ModuleConfigs | ModuleConfigs[],
+  streamOptions?: StreamOptions | StreamOptionsArray
 ): OrchestrationConfig {
-  const { prompt_templating, filtering } = moduleConfigs;
-  const outputFiltering = streamOptions?.outputFiltering;
+  // Validate stream options compatibility
+  validateStreamOptions(moduleConfigs, streamOptions);
+
+  // Handle array of stream options
```
KavithaSiva: [pp] Please remove comments for self explanatory code snippets.

packages/orchestration/src/util/module-config.ts
```diff
@@ -126,61 +126,181 @@ export function addStreamOptionsToOutputFilteringConfig(
...
+/**
+ * @internal
+ * Implementation
  */
 export function addStreamOptions(
-  moduleConfigs: ModuleConfigs,
-  streamOptions?: StreamOptions
+  moduleConfigs: ModuleConfigs | ModuleConfigs[],
+  streamOptions?: StreamOptions | StreamOptionsArray
 ): OrchestrationConfig {
```
KavithaSiva: [req] This function is currently difficult to read with multiple cases tackled together for single/multiple module config and stream options.

Could you please refactor the function to have cleaner separations for each of these different cases?

KavithaSiva (changes requested):
Initial review done, with few refactoring requests.

packages/orchestration/src/orchestration-client.ts
davidkna-sap: The overload with module fallbacks supports a list of `streamOptions` whereas the overload for a single configuration doesn't.
By handling it separately I tried to increase type-safety, reducing the use of `as`. If you prefer, I can replace it with a single forward that uses `streamOptions as StreamOptions | undefined`.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -392,6 +395,20 @@ export interface StreamOptions {
...
+ *
+ * Global options (chunk_size, etc.) apply to the entire stream and can only be
+ * specified in the first element. This is enforced at the type level using `Omit<StreamOptions, 'global'>[]`
+ * for subsequent elements, ensuring type-safe configuration.
+ */
+export type StreamOptionsArray = [
+  StreamOptions,
+  ...Omit<StreamOptions, 'global'>[]
+];
+
```
KavithaSiva: [suggestion] I was brainstorming. Instead of providing an array of stream options what about having a type like this:
```TS
  interface StreamOptionsWithOverrides {                                                                                                                     
    global?: GlobalStreamOptions;                         // Top level only                                                                                 
    default?: Omit<StreamOptions, 'global'>;             // No global here                                                                                   
    overrides?: Record<number, Omit<StreamOptions, 'global'>>;  // No global here                                                                            
  }                                                                                                                                                          
```     
Which makes it simpler to provide global config and default config that has to fit all orchestration configs?

So, the usage looks cleaner:

```TS
// Array config - structured                                                                                                                               
  const arrayClient = new OrchestrationClient([config1, config2, config3]);                                                                                  
  arrayClient.stream(req, sig, {                                                                                                                             
    global: { chunk_size: 100 },                                                                                                                             
    default: { promptTemplating: { include_usage: false } },                                                                                                 
    overrides: {                                                                                                                                             
      0: { promptTemplating: { include_usage: true } }                                                                                                       
    }                                                                                                                                                        
  });                                                                                                                                                        
                                                                                                                                                             
  // Array config - minimal (just global)                                                                                                                    
  arrayClient.stream(req, sig, {                                                                                                                             
    global: { chunk_size: 100 }                                                                                                                              
  });                                                                                                                                                        
                                                                                                                                                             
  // Array config - just overrides                                                                                                                           
  arrayClient.stream(req, sig, {                                                                                                                             
    overrides: {                                                                                                                                             
      1: { outputFiltering: { overlap: 50 } }                                                                                                                
    }                                                                                                                                                        
  });           
```

Honestly, it would have been much easier if the stream options in module configuration had also been directly initialised in the constructor instead, a consideration for v3 maybe.

Another option would be to only allow `global` and `default` options for now and only allow `overrides` if there is enough demand.

What do you think of these suggestions?

davidkna-sap: I do think this is cleaner, though as far as ease-of-use goes, simple cases can still use a single non-list `StreamOptions` as usual anyway. I mostly only implemented this more complicated way of doing this because the documentation/demos stressed the importance of each module being independent and very customised for each the module configuration.

That said, I will try implementing your suggestion to see how it looks.

marikaner: @davidkna-sap I would like to throw in a slight change to @KavithaSiva's idea: You could also change the API to be something like: 

```ts
type StreamOptionsWithOverrides  = StreamOptions & {                                                                                                                                                                                                      
  overrides?: Record<number, Omit<StreamOptions, 'global'>>;  // No global here                                                                            
}
```

That would keep the stream options as is for the simple case and allow overrides for fallbacks.

KavithaSiva: Wait a bit before implementing my suggestion. I discussed the API with @marikaner today and she also has some suggestions for the API. Let's discuss before we implement more.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -391,7 +391,44 @@ export interface StreamOptions {
...
+   *   promptTemplating: { include_usage: false },
+   *   overrides: [
+   *      { promptTemplating: { include_usage: true }, outputFiltering: { overlap: 50 } }, // config 0
+   *      undefined, // config 1 uses base configuration
+   *      { promptTemplating: { include_usage: false }, outputFiltering: { overlap: 0 } }  // config 2
+   *   ]
+   * };
+   * ```
+   */
+  overrides?: Partial<Record<number, ModuleStreamOptions>>;
```
KavithaSiva: [req] With this change, you also permit the user to provide overrides for single configuration. 

I prefer creating a dedicated type for stream options that correspond to multiple module configs:

```TS
type StreamOptionsWithOverrides  = StreamOptions & {                                                                                                                                                                                                      
  overrides?: Record<number, Omit<StreamOptions, 'global'>>;  // No global here                                                                            
}
```
What was your reasoning behind adding overrides to the same type over the union with a new dedicated type?

packages/orchestration/src/orchestration-types.ts
```diff
@@ -391,7 +391,44 @@ export interface StreamOptions {
...
+   *   global: { chunk_size: 128 },
+   *   promptTemplating: { include_usage: false },
+   *   overrides: [
+   *      { promptTemplating: { include_usage: true }, outputFiltering: { overlap: 50 } }, // config 0
+   *      undefined, // config 1 uses base configuration
+   *      { promptTemplating: { include_usage: false }, outputFiltering: { overlap: 0 } }  // config 2
+   *   ]
+   * };
+   * ```
+   */
```
KavithaSiva, line 427: [req] Consider simplifying the JSDoc, it is atm confusing. 
Suggestion from claude:

```
   * Per-config stream options that override shared settings.                                                                                                                    
   * Supports both object and array notation.                                                                                                                                    
   *                                                                                                                                                                             
   * **Object notation (recommended for sparse overrides):**                                                                                                                     
   * Use numeric keys for config indices you want to override.                                                                                                                   
   * Omit keys for configs that should use shared options.                                                                                                                       
   *                                                                                                                                                                             
   * **Array notation:**                                                                                                                                                         
   * Use array indices matching config positions.                                                                                                                                
   * Use `undefined` for configs that should use shared options.                                                                                                                 
   * Array length should match or be less than config count. 
   (Examples follow)     
```

[suggestion] It would be nice we can skip allowing the array notation altogether and only permit object notations. With the index, it is much clearer which config belongs to which fallback.
Or if this is not easy, at least not adevertise the array notation usage in JSDoc.

packages/orchestration/src/util/module-config.ts
```diff
@@ -124,45 +124,190 @@ export function addStreamOptionsToOutputFilteringConfig(
...
+    return;
+  }
+  const arrayLength = streamOptions.overrides.length;
+  logger.warn(
+    `Override array has ${arrayLength} element(s) but there are ${configurationCount} module configuration(s). ` +
+      `Configs at indices ${arrayLength}-${configurationCount - 1} will use shared options. ` +
+      'If this is intentional, use object input to silence this warning: `{...streamOptionsArray}`'
+  );
+}
+
```
KavithaSiva, line 166: [suggestion] Consider silent skip for invalid override indices and short overrides(adds log noise) and keep errors for semantic misuse, like overrides on single config.

davidkna-sap: I feel like this could be nice-to-have for debugging? What do you think about moving this to a different log-level like info or debug instead?

packages/orchestration/src/orchestration-types.ts
davidkna-sap: @KavithaSiva

1. The code actually does support single options without module fallback now.
2. We can't restrict the type in a user-facing anyway because we don't have any generic argument to tell us if module fallbacks are in use, the differentiation I previously added was mostly restricted to non-user-facing functions.

KavithaSiva: I agree we can't get compile-time checking without generics, but the combination of clearer types and runtime validation (for overrides in single config) still improves the API IMO.

packages/orchestration/src/util/module-config.ts
KavithaSiva: Both these warnings lean toward wrong API usage, if you like to keep them, debug is better.

packages/orchestration/src/util/stream.ts
```diff
@@ -34,9 +36,42 @@ export function mergeStreamResponse(
...
+  }
+
+  // Deduplicate messages with structural equality checks
+  const existingSet = new Set(
+    existing.map(f => JSON.stringify(Object.entries(f).sort()))
+  );
+  const newFailures = transformedIncoming.filter(
+    f => !existingSet.has(JSON.stringify(Object.entries(f).sort()))
+  );
+  return [...existing, ...newFailures];
```
KavithaSiva, line 72: [suggestion] Do we have to consider all properties in  `OrchestrationError` during deduplication. Sorting and stringifying all properties looks like an overkill.

Wouldn't directly comparing the necessary fields like `request_id`, `code`, `message` and `location` be enough?

packages/orchestration/src/orchestration-client.test.ts
```diff
@@ -1318,4 +1320,356 @@ describe('orchestration service client', () => {
...
+          }
+        }
+      };
+
+      const streamOptions: StreamOptions = {
+        global: { chunk_size: 100 },
+        overrides: [
+          { promptTemplating: { include_usage: true } },
+          { promptTemplating: { include_usage: false } }
+        ] as any
```
KavithaSiva:
```suggestion
          { promptTemplating: { include_usage: true } },
          { promptTemplating: { include_usage: false } }
        ] 
```
Is the cast to `any` required?

packages/orchestration/src/util/module-config.ts
```diff
@@ -124,74 +125,253 @@ export function addStreamOptionsToOutputFilteringConfig(
...
+  if (configsWithoutFilter.length === configs.length) {
+    logger.warn(
+      'Output filter stream options are not applied because no module configuration has output filtering enabled.'
+    );
+  } else if (configsWithoutFilter.length > 0) {
+    const configWord =
+      configsWithoutFilter.length > 1 ? 'configurations' : 'configuration';
+    const positions = configsWithoutFilter.map(i => `#${i + 1}`).join(', ');
+    logger.warn(
+      `Output filter stream options will not be applied to ${configWord} ${positions} because output filtering is not configured for those modules.`
```
KavithaSiva, line 230: [pp] I personally prefer these also to be set to `debug` level instead of `warn`.

packages/orchestration/src/util/module-config.ts
```diff
@@ -124,74 +125,253 @@ export function addStreamOptionsToOutputFilteringConfig(
...
 ): OrchestrationConfig {
-  const { prompt_templating, filtering } = moduleConfigs;
-  const outputFiltering = streamOptions?.outputFiltering;
-  const globalStreamOptions = streamOptions?.global;
+  const configs = Array.isArray(moduleConfigs)
+    ? moduleConfigs
+    : [moduleConfigs];
 
-  if (!moduleConfigs?.filtering?.output && outputFiltering) {
+  if (!Array.isArray(moduleConfigs) && streamOptions?.overrides) {
```
KavithaSiva, line 279: [suggestion] Should we consider throwing an error here as it is a clear API misuse?

packages/orchestration/src/orchestration-types.ts
```diff
@@ -391,7 +391,55 @@ export interface StreamOptions {
...
+   * @example
+   * ```typescript
+   *
+   * const optionsWithOverrides: StreamOptions = {
+   *   global: { chunk_size: 100 },
+   *   promptTemplating: { include_usage: false },
+   *   overrides: {
+   *     0: { promptTemplating: { include_usage: true } },  // config 0
+   *     // 1: uses base configuration
+   *     2: { outputFiltering: { overlap: 50 } }            // config 2
```
KavithaSiva: [suggestion] Config 2 would resolve to `{ outputFiltering: { overlap: 50 },  promptTemplating: { include_usage: false }}` correct?

Consider documenting the final value the configuration resolves to.

packages/orchestration/src/util/module-config.ts
davidkna-sap: I am leaning towards keeping it because this was also `warn` pre-module-fallback (https://github.com/SAP/ai-sdk-js/blob/v2.6.0/packages/orchestration/src/util/module-config.ts#L136-L140).

packages/orchestration/src/orchestration-types.ts
davidkna-sap: it evaluates to `2: { outputFiltering: { overlap: 50 } }`, I have updated the comments to make this more obvious.

PR #1480, opened 2026-01-27
feat: Advertise AbortSignal support for HTTP request cancellation

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#458.

### DIFF
--- a/packages/langchain/src/orchestration/client.ts
+++ b/packages/langchain/src/orchestration/client.ts
@@ -102,6 +102,7 @@ export class OrchestrationClient extends BaseChatModel<
     options: typeof this.ParsedCallOptions,
     runManager?: CallbackManagerForLLMRun
   ): Promise<ChatResult> {
+    options.signal?.throwIfAborted();
     // Auto-streaming: transparently stream and concatenate when enabled
     if (this.streaming) {
       let generation;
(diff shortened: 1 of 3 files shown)


packages/core/src/http-client.ts
```diff
@@ -28,7 +28,8 @@ export type CustomRequestConfig = Pick<
   | 'httpsAgent'
   | 'parameterEncoder'
 > &
-  Record<string, any>;
+  // TODO: Move to `Pick` after updating SAP Cloud SDK dependency
```
ZhongpinWang, line 31: [q/req] Is this TODO still relevant? Can we now pick it?

packages/langchain/src/orchestration/client.ts
```diff
@@ -183,6 +184,7 @@ export class OrchestrationClient extends BaseChatModel<
     options: typeof this.ParsedCallOptions,
     runManager?: CallbackManagerForLLMRun
   ): AsyncGenerator<ChatGenerationChunk> {
+    options.signal?.throwIfAborted();
```
ZhongpinWang, line 187: [q] I think this `throwIfAborted` is just a one time check? So it is just meant for an early abort?

davidkna-sap: It's an early abort, langchain recently added this for all official client, so I decided to follow suit.

packages/core/src/http-client.ts
marikaner: +1

marikaner (approved):

PR #1484, opened 2026-01-29
feat: Support Parquet input in RPT client

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#469.

It would be nicer if this was using https://github.com/SAP/cloud-sdk-js/pull/6276, but this PR also works without these additional changes while preserving type-safety at the expense of additional boilerplate.

### DIFF
--- a/packages/rpt/src/client.test.ts
+++ b/packages/rpt/src/client.test.ts
@@ -112,6 +112,7 @@ describe('rpt', () => {
     expect(result.predictions).toEqual([{ SALESGROUP: 'test' }]);
   });
 
+  // Only uses the required parameters, the Parquet file and the prediction configuration, only giving the mandatory first two paramters to the method.
   it('should upload Parquet file without optional parameters', async () => {
     mockDeploymentsList(
       {
--- a/sample-code/src/rpt.ts
+++ b/sample-code/src/rpt.ts
@@ -81,29 +81,37 @@ const __dirname = dirname(__filename);
 const parquetFilePath = join(__dirname, 'product_data.parquet');
 
 /**
- * Predict the sales group of products by passing a Parquet file.
- * @param type - The type of the file to be sent, either 'Blob' or 'File' (a `Blob` with a filename).
+ * Predict the sales group of products by passing a Parquet file with filename (`File`).
  * @returns The prediction results.
  */
-export async function predictParquet(
-  type: 'Blob' | 'File' = 'Blob'
-): Promise<PredictResponsePayload> {
-  // You can choose to pass either a Blob or a File.
+export async function predictParquetFile(): Promise<PredictResponsePayload> {
   const parquetFileBlob = await openAsBlob(parquetFilePath, {
     type: 'application/vnd.apache.parquet'
   });
-  // If type is 'File', the filename will be forwarded to the RPT service instead of a generic name.
+  // Create a File with a filename that will be forwarded to the RPT service
   const parquetFile = new File([parquetFileBlob], 'product_data.parquet', {
     type: 'application/vnd.apache.parquet'
   });
   // Send the Parquet file to the RPT service for predictions
   const client = new RptClient();
-  return client.predictParquet(
-    type === 'File' ? parquetFile : parquetFileBlob,
-    data.prediction_config,
-    {
-      index_column: data.index_column,
-      parse_data_types: false
-    }
-  );
+  return client.predictParquet(parquetFile, data.prediction_config, {
+    index_column: data.index_column,
+    parse_data_types: false
+  });
+}
+
+/**
+ * Predict the sales group of products by passing a Parquet file without a filename (`Blob`).
+ * @returns The prediction results.
+ */
+export async function predictParquetBlob(): Promise<PredictResponsePayload> {
+  const parquetFileBlob = await openAsBlob(parquetFilePath, {
+    type: 'application/vnd.apache.parquet'
+  });
+  // Send the Parquet blob to the RPT service for predictions
+  const client = new RptClient();
+  return client.predictParquet(parquetFileBlob, data.prediction_config, {
+    index_column: data.index_column,
+    parse_data_types: false
+  });
 }
--- a/sample-code/src/index.ts
+++ b/sample-code/src/index.ts
@@ -70,5 +70,6 @@ export {
 export {
   predictWithSchema,
   predictAutomaticParsing,
-  predictParquet
+  predictParquetFile,
+  predictParquetBlob
 } from './rpt.js';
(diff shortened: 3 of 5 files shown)


packages/rpt/src/client.ts
```diff
@@ -53,6 +58,61 @@ export class RptClient {
...
+
+    // Construct intermediate body object to handle type-safety
+    const bodyObject: BodyPredictParquet = {
+      // The schema says bytes, but SAP Cloud SDK generates it as string
+      file: parquetData as unknown as string,
+      prediction_config: JSON.stringify(predictionConfig),
+      ...(options || {})
+    };
+
+    const body = new FormData();
```
davidkna-sap, line 96: Instead of vanilla node `FormData` the [`form-data`-package](https://www.npmjs.com/package/form-data) would allow more freedom towards supporting different binary or stream data sources for the parquet file. But at least on the local file-systems `fs.openAsBlob` will do some streaming internally.

packages/rpt/src/client.ts
```diff
@@ -53,6 +58,61 @@ export class RptClient {
...
+    const { resourceGroup, deploymentId } =
+      await this.getResourceGroupAndDeploymentId();
+
+    // Construct intermediate body object to handle type-safety
+    const bodyObject: BodyPredictParquet = {
+      // The schema says bytes, but SAP Cloud SDK generates it as string
+      file: parquetData as unknown as string,
+      prediction_config: JSON.stringify(predictionConfig),
+      ...(options || {})
+    };
```
BrigittaK307, line 94: I believe that the `bodyObject` is not used at all (besides the prediction_config parameter).
Could be easier to remove the `bodyObject` entirely, and append `JSON.stringify(predictionConfig)` instead.

davidkna-sap: I've handled it this way to ensure that the assembled body matches the expected `BodyPredictParquet`-format despite the lack of a typed API-endpoint.

BrigittaK307: Okay, I understand.
My only concern now is the `file` property's casting as unknown.
Couldn't there be a workaround to do this?

davidkna-sap: The issue is that the OpenAPI generator maps this to `string` instead of `Blob` - this is the workaround for that:
```yaml
"type": "string",
"format": "binary",
```
I am using `as unknown as string` instead of `as any` to make sure it gets caught if we update the generator to handle this case differently.

BrigittaK307: I think that in this case I don't have any other issues, so I'll approve it.

sample-code/resources/generate_parquet.py
marikaner, line 1: [q] Why Python?

packages/rpt/src/client.ts
```diff
@@ -53,6 +58,72 @@ export class RptClient {
...
+
+    body.append('prediction_config', bodyObject.prediction_config);
+
+    for (const [key, value] of Object.entries(options || {})) {
+      if (value !== undefined) {
+        body.append(key, String(value));
+      }
+    }
+
+    return RptApi.predictParquet(body)
```
marikaner, line 117: [q] Why did you not use the form data generation from the Cloud SDK? => https://github.com/SAP/cloud-sdk-js/pull/6276
I did this to avoid duplicate implementations in this (and future) clients.

packages/rpt/src/client.test.ts
```diff
@@ -79,4 +79,63 @@ describe('rpt', () => {
...
+        ]
+      },
+      { index_column: '__row_idx__', parse_data_types: false }
+    );
+
+    expect(requestSpy.isDone()).toBe(true);
+    expect(result.predictions).toEqual([{ SALESGROUP: 'test' }]);
+  });
+
+  it('should upload Parquet file without optional parameters', async () => {
```
marikaner, line 116: I don't get this test case. What are the optional parameters that are not uploaded? Maybe add a comment somewhere?

packages/rpt/src/types.ts
```diff
@@ -121,3 +121,18 @@ export type PredictionData<T extends DataSchema> = {
...
+/**
+ * Options for Parquet-based predictions.
+ */
+export interface PredictionOptionsParquet {
+  /**
+   * The name of the index column. If provided, the service will return this column's value in each prediction object to facilitate aligning the output predictions with the input rows on the client side. If not provided, the column will not be included in the output.
+   */
+  index_column?: string;
+  /**
+   * Whether to parse the data types of the columns. If set to True, numeric columns will be parsed to float or integer and dates in ISO format YYYY-MM-DD will be parsed.
```
marikaner:
```suggestion
   * Whether to parse the data types of the columns. If set to true, numeric columns will be parsed to float or integer and dates in ISO format YYYY-MM-DD will be parsed.
```

sample-code/src/rpt.ts
```diff
@@ -72,3 +75,35 @@ export async function predictAutomaticParsing(): Promise<PredictResponsePayload>
...
+  const client = new RptClient();
+  return client.predictParquet(
+    type === 'File' ? parquetFile : parquetFileBlob,
+    data.prediction_config,
+    {
+      index_column: data.index_column,
+      parse_data_types: false
+    }
+  );
+}
```
marikaner: [pp] For better readability, I would split this into two samples. I found it quite confusing why we load the file twice + you have a `type` parameter. Without that the sample is so simple, that it will be much easier and quicker to grasp.

packages/rpt/src/client.ts
davidkna-sap: I avoided https://github.com/SAP/cloud-sdk-js/pull/6276 to have control over the filename, but https://github.com/SAP/cloud-sdk-js/pull/6276 may have worked if I had discovered `File` for this use-case earlier.

I created https://github.com/SAP/cloud-sdk-js/pull/6301 as a follow-up to https://github.com/SAP/cloud-sdk-js/pull/6276 with additional support for content-types, how stringifications is performed (`String` or `JSON.stringify`, `charset`), recognising `string` with `binary` format. https://github.com/SAP/ai-sdk-js/pull/1520 is based on that.

sample-code/resources/generate_parquet.py
davidkna-sap: Because it's an utility script, I didn't want to add additional npm-dependencies and python has an inline-syntax for script dependencies.

packages/rpt/src/client.test.ts
davidkna-sap: In this case it's just not passing the third argument with `index_column` and `parse_data_types`, I will add a comment.

PR #1488, opened 2026-01-30
feat: Support compressed requests in the RPT client

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#470.
Needs https://github.com/SAP/cloud-sdk-js/pull/6291


packages/rpt/src/vendor/compress-request-middleware.ts
davidkna-sap, line 1: > [!CAUTION]
> Vendored from https://github.com/SAP/cloud-sdk-js/pull/6291

marikaner: Don't forget to remove this.

davidkna-sap:
> I only managed to take a short look today, will continue tomorrow, but I already have a question: If this is implemented in a SAP Cloud SDK middleware, why do we even need any code here? Shouldn't we only add docs on how to use the middleware? Might be a brain fart though ;)

The idea with adding it here is to automatically compress any larger payloads by default.

packages/rpt/src/types.ts
```diff
@@ -117,3 +128,35 @@ export type PredictionData<T extends DataSchema> = {
...
   }
 >;
+
+/**
+ * Compression algorithm to use for request Body.
+ */
+export type RptRequestCompressionAlgorithm = Extract<
+  'gzip',
+  RequestCompressionAlgorithm
+>;
```
marikaner: [q] Isn't this the same as:
```suggestion
export type RptRequestCompressionAlgorithm = 'gzip';
```

packages/rpt/src/types.ts
```diff
@@ -117,3 +128,35 @@ export type PredictionData<T extends DataSchema> = {
...
+
+/**
+ * Compression middleware options for requests to the RPT service endpoint.
+ * @template C - The compression algorithm type.
+ */
+export type RptRequestCompressionMiddlewareOptions<
+  C extends RptRequestCompressionAlgorithm = 'gzip'
+> = RequestCompressionMiddlewareOptions<C>;
```
marikaner: [req] This seems quite unnecessary, given that we only support gzip here. Shouldn't it just be:
```suggestion
export type RptRequestCompressionMiddlewareOptions = RequestCompressionMiddlewareOptions<'gzip>;
```

That way you also lose the generic.

packages/rpt/src/client.ts
```diff
@@ -32,36 +33,42 @@ export class RptClient {
...
    * @param dataSchema - Prediction data follows this schema. When using TypeScript, the data schema type is used to infer the types of the prediction data. In that case, the data schema must be provided as a constant (`as const`).
    * @param predictionData - Data to base prediction on.
+   * @param customRequest - Custom request options.
    * @returns Prediction response.
    */
   async predictWithSchema<const T extends DataSchema>(
     dataSchema: T,
-    predictionData: PredictionData<T>
+    predictionData: PredictionData<T>,
+    customRequest: RptRequestOptions = {}
```
marikaner: [pp/req] In the other clients we call this `requestConfig` which I find more speaking.

packages/rpt/src/client.ts
```diff
@@ -83,9 +90,28 @@ export class RptClient {
...
-      .addCustomHeaders({ 'ai-resource-group': resourceGroup || 'default' })
-      .execute(this.destination);
+    const { requestCompression, ...customRequest } = customRequestAll;
+
+    if (requestCompression?.mode !== 'never') {
+      customRequest.middleware = [
+        compressRequest(requestCompression),
+        ...(customRequest.middleware || [])
+      ];
+    }
```
marikaner: [pp] I like this for convenience. Reconsider the naming though. I would like to suggest:
```ts
const { requestCompression, ...customRequestConfig } = requestConfig;
```

packages/rpt/src/client.ts
```diff
@@ -83,9 +90,28 @@ export class RptClient {
...
+      },
+      body,
+      customRequest,
+      this.destination
+    );
+
+    if (response.data) {
+      return response.data;
+    }
+    throw new Error('No data received from RPT prediction request');
```
marikaner: [req] Why not keep the old code and add `.addCustomRequestConfiguration(customRequest)` instead?

sample-code/src/rpt.ts
```diff
@@ -64,6 +68,23 @@ export async function predictWithSchema(): Promise<PredictResponsePayload> {
   return client.predictWithSchema(schema, data);
 }
 
+/**
+ * Predict the sales group of products with gzip compression.
+ * @param algorithm - The compression algorithm to use.
+ * @returns The prediction results.
+ */
+export async function predictWithSchemaCompressed(
+  algorithm: RptRequestCompressionAlgorithm = 'gzip'
```
marikaner: [req] This seems unnecessary as we only support 'gzip'.
```suggestion
```

sample-code/src/rpt.ts
```diff
@@ -64,6 +68,23 @@ export async function predictWithSchema(): Promise<PredictResponsePayload> {
...
+): Promise<PredictResponsePayload> {
+  const client = new RptClient();
+  return client.predictWithSchema(schema, data, {
+    requestCompression: {
+      mode: 'always', // force-enable compression
+      algorithm
+    }
+  });
```
marikaner, line 77: Why does the user have to set the algorithm, if we only support 'gzip'. Also, I think it would be more convenient, if the default compression mode was "always". My reasoning is: when a user sets requestcompression on a request and omits the mode, they might expect that the request is compressed. Otherwise they have to "force" the compression by setting the mode.

What to you think about the following:
1. remove algorithm
2. make the mode required

Then users would always have to set `requestCompression: { mode: 'XY' }` making it more obvious what mode is used.

@KavithaSiva maybe you also have an opinion 😊

marikaner (changes requested):
I think we should use gzip without making the user set it. Everything else is in the comments 😊 

Also, we should consider adding the request configuration to the new parquet endpoint as well.

packages/rpt/src/types.ts
davidkna-sap: I wanted to have generics, in order to have the option to extend this in the future without breaking changes. In the future more algorithms could become supported by the service. WDYT?

packages/rpt/src/client.ts
davidkna-sap: Will change this, not very familiar with the APIs there yet, so I copied the approach the other modules have taken.

packages/rpt/src/types.ts
marikaner: Let's ask the colleagues about their plans. If there are no plans for this, I would treat it as YAGNI.

sample-code/src/rpt.ts
davidkna-sap: I feel like the current default `auto` works well because it avoids additional complexity for smaller payloads, and `algorithm` leaves the door open for future extensions.

marikaner: I generally agree from the point of view of the Cloud SDK, but not the AI SDK.

Currently we already compress by default using "auto". When I set request compression explicitly, I need to set at least `{requestCompression: {algorithm: 'gzip'}}`. I find this quite wonky, given that I have no other choice, but let's assume we keep it at that for now.

Then I need to *know* that it will only be compressed, when it is "big" enough, aka. that I didn't change the default behavior. The configuration does not reflect that.
Otherwise I have to set `{requestCompression: {algorithm: 'gzip', mode: 'always'}}`.

When using a middleware in the SDK it is a different story. Then you pass a middleware function, that is opinionated, which makes sense...

marikaner: About the algorithm, I asked the colleagues there are no plans to allow other algorithms in the foreseeable future. That is why I would not include it in our config. It can still later be changed, so that gzipping is the default but adjustable without breaking changes.

Those two discussions cannot be separated completely, because when we remove the `algorithm` parameter `{requestCompression: {algorithm: 'gzip'}}` changes to `{requestCompression: {}}`. Which is not useful. This is why I would make the mode required in this case. That would solve all other issues: by default auto is there anyways and when you want to configure compression, you have to do it explicitly.

davidkna-sap: I think the type should resolve correctly without explicit `algorithm` (missing algorithm is treated like `gzip`), but I can remove the generics.

marikaner: Another thought: currently I can set `header-only`, however, I cannot pass a gzipped body. I think we should either have an API to pass the compressed body or not allow the `header-only` mode.

KavithaSiva: I agree with your idea @marikaner, of removing `alogrithm` and explicitly setting `mode`.

PR #1520, opened 2026-02-06
feat: Support Parquet input in RPT client

davidkna-sap:
## Context

This PR is https://github.com/SAP/ai-sdk-js/pull/1484 but powered by the [multi-part SAP Cloud SDK PR](https://github.com/SAP/cloud-sdk-js/pull/6301).

Related https://github.com/SAP/ai-sdk-js-backlog/issues/469.
Cloud SDK PR/branch: https://github.com/SAP/cloud-sdk-js/pull/6301

### DIFF
--- a/packages/rpt/src/client.ts
+++ b/packages/rpt/src/client.ts
@@ -2,11 +2,13 @@ import {
   getFoundationModelDeploymentId,
   getResourceGroup
 } from '@sap-ai-sdk/ai-api/internal.js';
+import { compress as compressMiddleware } from '@sap-cloud-sdk/http-client';
 import { type PredictionConfig, RptApi } from './internal.js';
-import {
-  type DataSchema,
-  type PredictionData,
-  type PredictionOptionsParquet
+import type {
+  DataSchema,
+  PredictionData,
+  RptRequestOptions,
+  PredictionOptionsParquet
 } from './types.js';
 import type {
   PredictRequestPayload,
@@ -37,25 +39,29 @@ export class RptClient {
    * Prefer using this method when the data schema is known.
    * @param dataSchema - Prediction data follows this schema. When using TypeScript, the data schema type is used to infer the types of the prediction data. In that case, the data schema must be provided as a constant (`as const`).
    * @param predictionData - Data to base prediction on.
+   * @param requestConfig - Custom request configuration.
    * @returns Prediction response.
    */
   async predictWithSchema<const T extends DataSchema>(
     dataSchema: T,
-    predictionData: PredictionData<T>
+    predictionData: PredictionData<T>,
+    requestConfig: RptRequestOptions = {}
   ): Promise<PredictResponsePayload> {
-    return this.executePrediction(predictionData, dataSchema);
+    return this.executePrediction(predictionData, dataSchema, requestConfig);
   }
 
   /**
    * Predict based on prediction data with data schema inferred.
    * Prefer using `predictWithSchema` when the data schema is known.
    * @param predictionData - Data to base prediction on.
+   * @param requestConfig - Custom request configuration.
    * @returns Prediction response.
    */
   async predictWithoutSchema(
-    predictionData: PredictionData<DataSchema>
+    predictionData: PredictionData<DataSchema>,
+    requestConfig: RptRequestOptions = {}
   ): Promise<PredictResponsePayload> {
-    return this.executePrediction(predictionData);
+    return this.executePrediction(predictionData, undefined, requestConfig);
   }
 
   /**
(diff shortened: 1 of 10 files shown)


packages/rpt/src/spec/rpt.json
davidkna-sap, line 1: These changes are not in the upstream spec:

I've requested that `encoding` is added and AFAIK `prediction_config` types were already requested.

package.json
davidkna-sap, line 1: To be reverted

sample-code/src/rpt.ts
```diff
@@ -72,3 +75,35 @@ export async function predictAutomaticParsing(): Promise<PredictResponsePayload>
...
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+const parquetFilePath = join(__dirname, 'product_data.parquet');
+
+/**
+ * Predict the sales group of products by passing a Parquet file.
+ * @param type - The type of the file to be sent, either 'Blob' or 'File' (a `Blob` with a filename).
+ * @returns The prediction results.
+ */
+export async function predictParquet(
```
marikaner: [req] I think it would much better serve as an example, if this was separated into two parts. `predictParquetFile()` and `predictParquetBlob()`

sample-code/src/rpt.ts
```diff
@@ -72,3 +75,35 @@ export async function predictAutomaticParsing(): Promise<PredictResponsePayload>
   const client = new RptClient();
   return client.predictWithoutSchema(data);
 }
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+const parquetFilePath = join(__dirname, 'product_data.parquet');
```
marikaner: [q] Does this not work?

```suggestion
const __filename = fileURLToPath(import.meta.url);
const __dirname = dirname(__filename);
const parquetFilePath = join(import.meta.dirname, 'product_data.parquet');
```

packages/rpt/src/client.ts
```diff
@@ -53,6 +58,56 @@ export class RptClient {
...
+    // JavaScript has a few Blob-like types for binary data (e.g., Buffer, ArrayBuffer, etc.) which
+    // users might try to use here.
+    // Note: This check also covers File
+    if (!(parquetData instanceof Blob)) {
+      throw new Error(
+        `parquetData must be of type Blob or File. Received: ${typeof parquetData}`
+      );
+    }
+
+    // Workaround: Endpoint requires a filename that ends with .parquet
```
marikaner, line 94: [q] Does it make sense to ask the service owners to ignore the filename and just look at the format?

packages/rpt/src/client.ts
```diff
@@ -53,6 +58,56 @@ export class RptClient {
     return this.executePrediction(predictionData);
   }
 
+  /**
+   * Predict based on Parquet file data.
+   * @param parquetData - Parquet file data as Blob. Can also be a File to forward the filename.
```
marikaner, line 70: [nth] Maybe add a few words an what is in the file. I think prediction data and data schema.

packages/rpt/src/client.ts
```diff
@@ -53,6 +58,56 @@ export class RptClient {
...
+      prediction_config: predictionConfig,
+      ...(options || {})
+    };
+
+    return RptApi.predictParquet(body)
+      .setBasePath(`/inference/deployments/${deploymentId}`)
+      .addCustomHeaders({
+        'ai-resource-group': resourceGroup || 'default',
+        // SAP Cloud SDK does not (yet) set the Content-Type header automatically for multipart/form-data
+        'content-type': 'multipart/form-data'
```
marikaner: [req]

```suggestion
        'ai-resource-group': resourceGroup || 'default'
```

packages/rpt/src/client.test.ts
```diff
@@ -79,4 +79,63 @@ describe('rpt', () => {
...
+  it('should upload Parquet file with all parameters', async () => {
+    mockDeploymentsList(
+      {
+        scenarioId: 'foundation-models',
+        executableId: 'aicore-sap'
+      },
+      { id: '1234', model: { name: 'sap-rpt-1-small', version: 'latest' } }
+    );
+
+    const requestSpy = nock('https://api.ai.ml.hana.ondemand.com', {
```
marikaner: [nth] A request spy is exactly what it is, however it might be mistaken for a jest spy instead of a nock scope. Consider only calling actual spies "spy". Not a must, though.

packages/rpt/src/client.test.ts
```diff
@@ -79,4 +79,63 @@ describe('rpt', () => {
...
+      reqheaders: {
+        'ai-resource-group': 'default'
+      }
+    })
+      .post('/v2/inference/deployments/1234/predict_parquet', () => true)
```
marikaner: [req] I don't fully understand the idea of this test. It will always pass, I assume even if I would send optional parameters like `{ peter } as any`. The only thing that I see being tested is that you can pass optional parameters and leave them out. That shouldn't be tested here, but in the type tests instead.

packages/rpt/src/client.ts
davidkna-sap: I [already asked](https://github.tools.sap/DL-COE/sap-rpt-1-public-content/issues/10).

sample-code/src/rpt.ts
davidkna-sap: It look like I already updated this in the other PR, I will check the diff between both to see if anything else is missing.

packages/rpt/src/spec/rpt.json
marikaner: awesome 😊

packages/rpt/src/client.ts
marikaner: awesome!

packages/rpt/src/client.test.ts
```diff
@@ -79,4 +79,36 @@ describe('rpt', () => {
     await new RptClient().predictWithoutSchema({} as any);
     expect(requestSpy.isDone()).toBe(true);
   });
+
+  it('should upload Parquet file with all parameters', async () => {
```
marikaner: [pp] 
```suggestion
  it('should upload Parquet file', async () => {
```

PR #1533, opened 2026-02-11
chore: Fix rpt

marikaner:
I noticed that the top_k parameter is still there, although it should have been removed. I added it by accident. Also, the `task_type` is `any`, which it shouldn't.

### DIFF
--- a/.changeset/many-crabs-stick.md
+++ b/.changeset/many-crabs-stick.md
@@ -1,0 +1,5 @@
+---
+'@sap-ai-sdk/rpt': patch
+---
+
+[Improvement] Restrict the `task_type` property in the `PredictionConfig` type.


packages/rpt/src/types.ts
```diff
@@ -75,11 +75,7 @@ interface PredictionConfig<T extends DataSchema> {
     /**
      * The type of prediction task for this column. If not provided, the model will infer the task type from the data.
      */
-    task_type?: 'classification' | 'regression' | any;
-    /**
-     * How many predictions to output per classification column.If not provided, only a single prediction is returned. Only relevant for classification.
-     */
-    top_k?: number | any;
```
KavithaSiva, line 82: [q] Why should this be removed? Did they change something in their spec?

marikaner: It is not released yet, and I removed it from the spec. But I forgot that this is a manually created type here.

KavithaSiva (approved):

PR #1545, opened 2026-02-13
chore: Combine ai client type headers

marikaner:
## Context

Closes SAP/ai-sdk-js-backlog#477.

Combine custom client type with ours

### DIFF
--- a/packages/core/src/http-client.test.ts
+++ b/packages/core/src/http-client.test.ts
@@ -30,7 +30,7 @@ describe('http-client', () => {
   it('adds ai-client-type header, when there is a custom ai-client-type header', async () => {
     const scope = nock(aiCoreDestination.url, {
       reqheaders: {
-        'ai-client-type': /.*AI SDK JavaScript.*/
+        'ai-client-type': 'AI SDK JavaScript, custom client'
       }
     })
       .post('/v2/some/endpoint')


packages/core/src/http-client.ts
```diff
@@ -105,15 +106,24 @@ function mergeWithDefaultRequestConfig(
...
   };
+
   return {
     ...defaultConfig,
     ...requestConfig,
-    headers: mergeIgnoreCase(defaultConfig.headers, requestConfig?.headers),
+    headers: {
+      ...mergeIgnoreCase(defaultConfig.headers, requestConfig?.headers),
+      // add 'ai-client-type' header and merge with custom client type if needed
+      'ai-client-type': [
```
davidkna-sap: I am not sure what this does in practice, but I would prefer if this also used `mergeIgnoreCase`.

The spread above with `requestConfig?.headers` might add `'ai-client-type'`-headers with different casing(s).

marikaner: > I am not sure what this does in practice, but I would prefer if this also used `mergeIgnoreCase`.

I changed the code a bit to hopefully improve readability. What it does is add `'AI SDK JavaScript'` and either the custom client type or undefined to a list, then filter this list to remove undefined, if it is there and then join with `', '`. So the result is either `'AI SDK JavaScript'` or `'AI SDK JavaScript', 'CUSTOM CLIENT TYPE'`

> The spread above with `requestConfig?.headers` might add `'ai-client-type'`-headers with different casing(s).

You are right. I missed that, it is fixed now. I had a hard time adding a sensible test for this and decided that this is trivial enough to not bother testing it specifically. Let me know if you disagree or if you have a good idea how to test it. My approach was to mock `executeHttpRequest()`, however with ESM and jest, I didn't manage in a sensible time frame.

packages/core/src/http-client.test.ts
```diff
@@ -15,6 +14,36 @@ describe('http-client', () => {
...
+      .reply(200);
+
+    await executeRequest({ url: '/some/endpoint' }, { prompt: 'test' });
+    expect(scope.isDone()).toBe(true);
+  });
+
+  it('adds ai-client-type header, when there is a custom ai-client-type header', async () => {
+    const scope = nock(aiCoreDestination.url, {
+      reqheaders: {
+        'ai-client-type': /.*AI SDK JavaScript.*/
```
davidkna-sap: [pp]

```suggestion
        'ai-client-type': 'AI SDK JavaScript, custom client'
```

PR #1546, opened 2026-02-13
chore: Generate API docs

emincihangeri:
## Context

Closes SAP/ai-sdk-js-backlog#453.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Check dependencies
         run: pnpm check:deps
       - name: License Check
-        uses: sap/cloud-sdk-js/.github/actions/check-license@main
+        uses: sap/cloud-sdk-js/.github/actions/check-license@license-checker
       - run: pnpm doc
         name: API Doc Check
 


package.json
```diff
@@ -52,6 +54,7 @@
     "@types/node": "^22.19.11",
     "depcheck": "^1.4.7",
     "eslint": "^9.39.2",
+    "execa": "^5.0.0",
```
marikaner: [q] Why this old version? In the AI SDK, we can use ESM.

scripts/generate-docs.ts
```diff
@@ -0,0 +1,199 @@
+/* eslint-disable no-console */
+import { readFileSync, renameSync, readdirSync, lstatSync } from 'fs';
```
marikaner: [pp] Mark node imports with `node:fs`. That makes it easier to differentiate from other imports + it is more secure.
[pp] I would prefer if we could apply the async functions instead of the sync ones.

package.json
emincihangeri: I was following cloud-sdk but I forgot to check this. Corrected it now

scripts/generate-docs.ts
emincihangeri: I changed the implementation accordingly. Should I also do the similar changes in other script files in this PR?

scripts/generate-docs.ts
```diff
@@ -0,0 +1,197 @@
...
+
+    const encodedAdjustedData = await compressJson(navigationItems);
+    return `window.navigationData = "${encodedAdjustedData}"`;
+  });
+}
+
+function removeUnderlinePrefix(str: string) {
+  const i = str.indexOf('_');
+  // Remove the first `_`
+  return str.substring(0, i) + str.substring(i + 1);
```
marikaner: [q] I believe this is taken from the cloud SDK, however it surprises me. This could be replaces with `str.replace('_', '')`, right?

[q] Can it happen that a filename does not start with _? Then it will remove the first _ in the name, right? Should we also check that it really starts with that?

scripts/generate-docs.ts
```diff
@@ -0,0 +1,197 @@
...
+  await Promise.all(
+    htmlPaths.map((filePath: string) =>
+      transformFile(filePath, (file: string) =>
+        file.replace(/<a href="[^>]*_[^>]*.html[^>]*>/gi, removeUnderlinePrefix)
+      )
+    )
+  );
+
+  await Promise.all(
+    htmlPaths.map((filePath: string) => removeUnderlinePrefixFromFileName(filePath))
```
marikaner: [req] Can't we do something like this:
```suggestion
    htmlPaths.map((filePath: string) => {
      const newPath = removeUnderlinePrefix(filePath);
      return rename(filePath, newPath)
    })
```

emincihangeri: I think it has a little different functioning here. Because we don't remove the first underscore from the filePath, rather the leading underscores from the fileName. (i.e. path/_fileName -> path/fileName). But I can still reuse the function removeUnderlinePrefix there as you suggested.

davidkna-sap (changes requested):
Please add a documentation building step to `bump.yml`, [like for the SAP Cloud SDK](https://github.com/SAP/cloud-sdk-js/blob/main/.github/workflows/bump.yml#L32-L38).

I would also prefer to keep documentation artifacts outside this repository, but let’s discuss it with the team.

.github/workflows/build.yml
```diff
@@ -53,6 +53,8 @@ jobs:
         run: pnpm check:deps
       - name: License Check
         uses: sap/cloud-sdk-js/.github/actions/check-license@main
+      - run: pnpm doc
```
KavithaSiva, line 56: [q] I would like to skip this, and only generate docs during releases.

KavithaSiva (dismissed):

emincihangeri:
> I would also prefer the same, to avoid the bloat, can the workflow be adapted to generate docs on-the-fly during releases without committing them?

Done! Should we apply the same logic to cloud-sdk-js? @KavithaSiva

scripts/generate-docs.ts
```diff
@@ -0,0 +1,196 @@
...
+  const [, invalidLinks] = generationLogs.split(invalidLinksMessage);
+  if (invalidLinks) {
+    throw Error(`Error: ${invalidLinksMessage}\n${invalidLinks}`);
+  }
+}
+
+async function generateDocs() {
+  const generationLogs = await execa('typedoc', ['--tsconfig', 'tsconfig.typedoc.json'], {
+    cwd: resolve()
+  });
```
davidkna-sap: [pp] Consider removing `execa` again and using:
```ts
import { execFile } from 'node:child_process';
const execFileP = promisify(execFile);
```
and here:

```suggestion
  const generationLogs = await execFileP('typedoc', ['--tsconfig', 'tsconfig.typedoc.json'], {
    cwd: resolve()
  });
```

.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,65 @@
...
+          git config --global user.name "ai-sdk-js"
+          
+          # Use the version input parameter
+          VERSION="${{ inputs.version }}"
+          
+          cd ..
+          git clone --depth 1 git@github.com:SAP/ai-sdk.git
+          
+          # Copy generated docs to versioned folder
+          rsync -avz ai-sdk-js/knowledge-base/api-reference/ ai-sdk/static/api/${VERSION}/
```
davidkna-sap: [req] To clean-up any removed files:
```suggestion
          rsync -avz --delete ai-sdk-js/knowledge-base/api-reference/ ai-sdk/static/api/${VERSION}/
```

davidkna-sap (changes requested):
(wrong review type on previous review)

package.json
```diff
@@ -52,6 +54,7 @@
     "@types/node": "^22.19.13",
     "depcheck": "^1.4.7",
     "eslint": "^9.39.2",
+    "execa": "^9.6.1",
```
davidkna-sap:
```suggestion
```

.github/workflows/bump.yml
```diff
@@ -37,6 +37,14 @@ jobs:
           version: ${{ steps.bump.outputs.version }}
           user-name: sap-ai-sdk
 
+  generate-api-docs:
```
davidkna-sap, line 40: [nth] I don't wish to drag this PR out any further, but we should consider pushing the updated documentation alongside the release notes PR in a follow-up.

scripts/generate-docs.ts
marikaner: Generally, yes, please do minor clean ups alongside your implementation, but be careful to not do too much. Then rather create a separate PR.

PR #1568, opened 2026-02-23
chore: Add ADR for file input convenience

davidkna-sap:
## Context

ADR for https://github.com/SAP/ai-sdk-js-backlog/issues/485

### DIFF
--- a/adr/005-file-input-convenience.md
+++ b/adr/005-file-input-convenience.md
@@ -2,7 +2,7 @@
 
 ## Status
 
-proposed
+accepted
 
 ## Context
 


adr/005-file-input-convenience.md
```diff
@@ -0,0 +1,278 @@
...
+
+proposed
+
+## Context
+
+The orchestration package supports multi-modal messages containing files (PDF, CSV, DOCX, MP3) alongside text.
+
+The OpenAPI specification describes `file.file_data` as "Base64 encoded file content or file URL".
+
+In practice, the orchestration service only accepts RFC 2397 data URIs with a non-empty MIME type for inline file content.
```
BrigittaK307, line 13: [suggestion] I am wondering if it would be helpful to include some examples showing error cases?

adr/005-file-input-convenience.md
```diff
@@ -0,0 +1,278 @@
...
+<!-- vale off -->
+- [OpenAI Responses SDK-API for PDF](https://developers.openai.com/api/docs/guides/pdf-files/#uploading-files)
+- [Langchain PDF input](https://docs.langchain.com/oss/javascript/langchain/messages#multimodal)
+- [Vercel AI SDK PDF input](https://ai-sdk.dev/providers/ai-sdk-providers/openai#pdf-support)
+- Other providers mostly split base64 encoding and MIME type management into separate properties, without requiring users to manually construct data URIs.
+<!-- vale on -->
+
+## Proposed Decision
+
+Accept a structured `{ data: string; mimeType: string }` shape for `file_data` in the SDK (Option D) and serialize it to the required RFC 2397 data URI string at request-construction time.
```
BrigittaK307: [suggestion] Maybe we should include a type definition for this and the file_data?

BrigittaK307 (commented):
Generally looks really good to me, it is a very detailed ADR.

adr/005-file-input-convenience.md
```diff
@@ -0,0 +1,370 @@
...
+| nova-lite        | ❌  | ❌  |  ❌  | ❌  |
+| gpt-5-mini       | ❌  | ❌  |  ❌  | ❌  |
+| sonar            | ✅  | ✅  |  ✅  | ❌  |
+| mistral-medium   | ❌  | ❌  |  ❌  | ❌  |
+
+Helpful links:
+
+<!-- vale off -->
+
+- [OpenAI Responses SDK-API for PDF](https://developers.openai.com/api/docs/guides/pdf-files/#uploading-files)
```
KavithaSiva:
```suggestion
- [OpenAI Responses SDK-API for PDF](https://developers.openai.com/api/docs/guides/file-inputs#uploading-files)
```
Typo in link.

adr/005-file-input-convenience.md
```diff
@@ -0,0 +1,370 @@
...
+
+client.chatCompletion({
+  messages: [
+    {
+      role: 'user',
+      content: [
+        { type: 'text', text: 'Summarise this document.' },
+        {
+          type: 'file',
+          file: { file_data: fileData, filename: 'document.pdf' }
```
KavithaSiva, line 106: Looking at the [orchestration spec,](https://github.tools.sap/AI/llm-orchestration/blob/main/src/spec/v2.yaml#L589) I see that users can either give the encoded string (`data:<mimeType>;base64,<base64EncodedData>`) or a url pointing to the file.

adr/005-file-input-convenience.md
```diff
@@ -0,0 +1,370 @@
...
+          type: 'file',
+          file: {
+            file_data: { data: base64, mimeType: 'application/pdf' },
+            filename: 'document.pdf'
+          }
+        }
+      ]
+    }
+  ]
+});
```
KavithaSiva, line 338: I would like to suggest a slightly different option with discriminated union instead as orchestration service allows both URL and encoded file data.

```
 const base64 = await readFile('./document.pdf', 'base64');                                                                                           
                                                                                                                                                         
  client.chatCompletion({                                                                                                                                
    messages: [
      {
        role: 'user',
        content: [
          { type: 'text', text: 'Summarise this document.' },
          {
            type: 'file',
            file: {
              type: 'base64',  // discriminator
              data: base64,
              mimeType: 'application/pdf',
              filename: 'document.pdf'
            }
          }
        ]
      }
    ]
  });
```

KavithaSiva (commented):
Really detailed ADR, loved the depth and the various considerations.

I also lean towards D, but have a different suggestion for the types introduced, inspired by Langchain.

```TS
type FileContentInput =
    | {
        type: 'base64';
        data: string | Buffer;
        mimeType: string;
        filename?: string;
      }
    | {
        type: 'url';
        url: string;
        mimeType?: string;
        filename?: string;
      };
```
We would also have to override the generated `UserChatMessageContentItem`.

```TS
  type ChatMessageContentItem =
    | { type: 'text'; text: string; }
    | { type: 'image_url'; image_url: ImageContentUrl }
    | { type: 'file'; file: FileContentInput };  // Nested like image_url
```

KavithaSiva:
> @KavithaSiva Sounds good, but if a `url` is provided I would disallow `mimeType` (mimeType?: never). Arguably `mimeType` could be useful for HTTPS URLs too (server returning unexpected MIME types in response), but currently the service does not support this.

Agreed.
 @marikaner Can you also please look at the proposed option/suggestions and give your opinion?

PR #1589, opened 2026-03-03
chore: Responses API ADR

marikaner:
## Context

Relates to SAP/ai-sdk-js-backlog#482.

This introduces a propsal with options on how to implement the Respones API and thereby change the Chat API.
Other ideas are welcome.

### DIFF
--- a/adr/005-responses-api.md
+++ b/adr/005-responses-api.md
@@ -130,7 +130,7 @@ const response = await new AzureOpenAiChatClient('gpt-5')
 
 #### Cons
 
-- maybe atypical for JS
+- not idiomatic for JS
 
 ## Option B.1
 


adr/005-responses-api.md
```diff
@@ -0,0 +1,199 @@
...
+Often a list of options with pros and cons including the selection implementation. -->
+
+## Option A.1 - Keep everything as is with minimal renaming
+
+We currently have one client for the Chat API and one for Embeddings.
+We could keep them and add another for the Responses API:
+
+```ts
+// Responses API
+const response = await new AzureOpenAiResponsesClient('gpt-5').createResponse({
```
ZhongpinWang:
```suggestion
const response = await new AzureOpenAiResponsesClient('gpt-5').create({
```
[pp] Since the client name already indicate which type it is, I would keep the create method name shorter.

adr/005-responses-api.md
```diff
@@ -0,0 +1,199 @@
...
+The current Chat and Embedding APIs each have one method `run()` and `stream()`.
+The Responses API has more than one endpoint, so this approach doesn't work for this API.
+With renamed method names (e.g. `createChatCompletion()`), naming becomes more difficult (e.g. `streamCreateChatCompletion()`);
+Instead, the actual method should be the same for streaming and non streaming, and streaming should be toggled per option.
+
+There are multiple places to put this option:
+
+### Option A.2.1 - In the data
+
+The `stream` option already exists in the data.
```
ZhongpinWang, line 69: This I like. Similar to the official OpenAI SDK approach by using `stream` property to decide the actual return type.

adr/005-responses-api.md
```diff
@@ -0,0 +1,199 @@
...
+We could use it same as OpenAI does in their clients:
+
+```ts
+// Chat API
+const response = await new AzureOpenAiChatClient('gpt-5').createChatCompletion({
```
ZhongpinWang, line 74: Similar to my suggestion above, I would name the method just as `create()` instead of `createChatCompletion()`.

https://github.com/openai/openai-node?tab=readme-ov-file#streaming-responses

adr/005-responses-api.md
```diff
@@ -0,0 +1,199 @@
...
+});
+```
+
+#### Pros
+
+- similar API as OpenAI
+
+#### Cons
+
+- differs from our other APIs, where data and config are separated
```
ZhongpinWang, line 86: Well technically, it is possible to put everything in one `create()` method.

Image

Looking at the `createStream()` method we currently use for streaming case, the diff between streaming and non-streaming is mostly

`request`
```ts
        stream: true,
        stream_options: {
          include_usage: true
        }
```

and `requestConfig`

```ts
        responseType: 'stream',
        signal: controller.signal // I think anyway we support signal in both streaming and non-streaming cases
```

So we can for sure just use `stream` property in the `request` property to decide if we go down to the streaming or non-streaming path.

I also just checked, even now we allow providing `stream` property in `run()` method 🥲 . Perfect extension point.

adr/005-responses-api.md
```diff
@@ -0,0 +1,199 @@
...
+We could put them into one common options object:
+
+```ts
+// Chat API
+const response = await new AzureOpenAiChatClient('gpt-5').createChatCompletion(
+  {
+    messages: [...]
+  },
+  {
+    stream: true,
```
ZhongpinWang, line 100: IMO no this option is not great. `stream` is already part of `request`. Duplicating it into `requestConfig` seems to complicate things.

adr/005-responses-api.md
```diff
@@ -0,0 +1,199 @@
...
+  });
+```
+
+#### Pros
+
+- no changes to the API needed
+
+#### Cons
+
+- maybe atypical for JS
```
ZhongpinWang: Also not up voting this. The request diff between streaming and non-streaming is rather small. And true, it feels so Java...

adr/005-responses-api.md
```diff
@@ -0,0 +1,199 @@
...
+
+The naming currently includes quite a long prefix, making the client classes a bit wonky, especially if there are more than 2 now.
+We could consider something like a namespaces to organize the clients.
+Namespaces are not recommended in modern TS as many build tools cannot handle them, therefore I would bring in structure through exports:
+
+```ts
+import { AzureOpenAi } from '@sap-ai-sdk/foundation-models';
+
+// Chat API
+const response = await new AzureOpenAi.ChatClient('gpt-5').createChatCompletion({
```
ZhongpinWang:
```suggestion
const response = await new AzureOpenAi.ChatClient('gpt-5').create({
```

ZhongpinWang: I somewhat like the official OpenAI approach to have a "base" client  maybe call it `AzureOpenAi`. Then `AzureOpenAi.chat` or `AzureOpenAi.response` or `AzureOpenAi.embed`, which can also return the concrete client object if we anyway want to group different client types.

I feel we have too many clients spread separately.

adr/005-responses-api.md
```diff
@@ -0,0 +1,199 @@
...
+});
+```
+
+alternative usage:
+
+```ts
+import { Chat } from '@sap-ai-sdk/foundation-models/azure-openai';
+
+// Chat API
+const response = await new ChatClient('gpt-5').createChatCompletion({
```
ZhongpinWang: Feels no big improvement but only shorten the name.

adr/005-responses-api.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Option B.2
+
+We could even get rid of the classes on API level, further shortening the naming, by exporting builder functions instead:
+
+```ts
+import { AzureOpenAi } from '@sap-ai-sdk/foundation-models';
+
+// Chat API
+const response = await AzureOpenAi.chat('gpt-5').createChatCompletion({
```
ZhongpinWang, line 177: I prefer this one over the B.1. We can still make our changes non-breaking, by still initializing an actual client object and return it when calling `chat()`. In this way, users still have the chance to use the actual `AzureOpenAIChatClient` directly. 

We can keep the `AzureOpenAi` in the current major version just as a convenience for better autocompletion, then later step by step  extract maybe common parts out of each client.

ZhongpinWang (commented):
Overall, I believe there is a way to introduce a client grouping without having breaking changes. We can still keep the old client, the old `stream()` method as is and extend the `run()` (or rather create a new one called `create()`) and using `stream` property inside of it. In the next major version, we will do some clean up and remove / hide the old ones.

marikaner:
> Overall, I believe there is a way to introduce a client grouping without having breaking changes. We can still keep the old client, the old `stream()` method as is and extend the `run()` (or rather create a new one called `create()`) and using `stream` property inside of it. In the next major version, we will do some clean up and remove / hide the old ones.

Agreed, no breaking changes will be necessary, most likely deprecation.

adr/005-responses-api.md
marikaner: I saw that too, I guess that was not intended...

adr/005-responses-api.md
marikaner: I would not duplicate, rather move. The fact that the `stream` property is currently exposed is a mistake, I think.

adr/005-responses-api.md
marikaner: I agree. I didn't think of better arguments for cons (maybe implementation effort), but it just "feels" wrong ;)

adr/005-responses-api.md
marikaner: I agree. And I was thinking about something like that (+ David also suggested something in that direction). The reason I didn't put it here explicitly is that OpenAI has something to initialize this client with, we don't have anything, right?
Than there is no point in having a constructor just to expose functions. That can be done through modules + users can benefit from treeshaking.

adr/005-responses-api.md
```diff
@@ -0,0 +1,199 @@
...
+// Chat API
+const response = await new AzureOpenAi.ChatClient('gpt-5').createChatCompletion({
+  messages: [...]
+});
+```
+
+alternative usage:
+
+```ts
+import { Chat } from '@sap-ai-sdk/foundation-models/azure-openai';
```
marikaner:
```suggestion
import { ChatClient } from '@sap-ai-sdk/foundation-models/azure-openai';
```

marikaner:
> ## **A**
> Let's also consider a transparent Azure OpenAI client that can route to Chat vs Responses based on model/flags (IIRC LangChain does something like this for their openai client, and other SDKs often do something like that). That said, I'd still prefer to also expose per-API clients/classes.
> 
> Cons:
> 
> * May make it harder to expose API-specific features, and to debug issues.
> * New models don't support chat completion API anymore -> better to avoid features for older API?
> 

Given that there are only cons and you prefer other options I would leave it at that.

> ## **Option A.2 (stream toggled)**
> I think it would be hard to express semantically in types, and internally implementations of the sync and streaming versions are handled very differently. But I do l feel like it could be convenient for users to handle it this way. My feeling is that streaming should be a separate method, because toggling streaming tends to make APIs harder to read and type.
> 
> I'm most partial to A.2.1 if we go this route, because it matches OpenAI’s own clients and keeps the call site simple. That said the official SDK is just auto-generated from the spec.
> 

Typing is a good point. I would like to take a look at this and make a small PoC before we dismiss this option as I think it is among the more promising options.

> ## **Option B.1 (grouped exports)**
> LGTM for organizing / shortening names, but I’m concerned about naming collisions and debuggability. If people import plain `ChatClient` from a subpath, issue reports often omit imports, so we lose context. This ambiguity already exists with langchain/non-langchain classes sharing names.
> 

I will add this to the cons. In my opinion this isn't a big issue though, because in issue reports we can ask where it was imported from. What are your concerns about debuggability?

> ## **Option B.2 (function-driven)**
> I would consider this for a v3. I can see the appeal of a more function-driven surface (e.g., `chat(model, ...)`, `stream(model, ...)`, where `model` is a string or a model config). But I’d treat B.2 as a future major breaking change.
> 

As @ZhongpinWang pointed out, I don't think this necessarily needs to be breaking. We could still keep the old client under the hood and expose it as deprecated. 

> ## **Future**
> OpenAI recently introduced websocket APIs. We may need a first-class `Session` concept at some point, not just as a convenience layer.
> 
> Conceptually something like:
> 
> * `const session = client.session()`
> * `session.chat('What is SAP')`
> * `session.chat('Thank you')`

Interesting. In the interest of time, I will look at this optionally and mention it in the ADR. I feel that this might be a little out of scope, but it might also influence our design to be future proof.

adr/005-responses-api.md
```diff
@@ -0,0 +1,278 @@
...
+```
+
+#### Pros
+
+- no maintenance, updating and documentation effort for a whole client, only it's configuration
+- users might already be used to the API, no new abstractions
+
+#### Cons
+
+- AzureOpenAI requires the `model` to be set in the payload (but apparently the value doesn't matter) => potentially forces users to pass the model twice
```
ZhongpinWang, line 234: Yes, this is unfortunately the downside..

adr/005-responses-api.md
```diff
@@ -0,0 +1,278 @@
...
+    destination
+  );
+  const resourceGroup = getResourceGroup(config) || 'default';
+  const aiCoreDestination = await getAiCoreDestination(destination);
+  const { baseURL, params } = await buildHttpRequest(aiCoreDestination);
+
+  return {
+    apiVersion: config.apiVersion || '2025-08-07',
+    azureADTokenProvider: async () =>
+      (await buildHttpRequest(aiCoreDestination)).headers.authorization.split(
```
ZhongpinWang: [req] I would put `await getAiCoreDestination()` inside the `azureADTokenProvider` function for token refetching if it is no longer valid.

And I am not sure why we need to call `buildHttpRequest()`? I thought `getAiCoreDestination()` already contains the auth token?

So maybe something like 

```ts
export async function tokenProvider() {
  const destination = await getAiCoreDestination();
  return destination.authTokens![0].value;
}
```

And use this as the `azureADTokenProvider`.

ZhongpinWang: Okay I see `buildHttpRequest()` will fetch the destination with token. But `getAiCoreDestination()` does it already.

adr/005-responses-api.md
davidkna-sap: I think `model` in the payload may be required for the v1 spec either way: https://github.com/openai/openai-java?tab=readme-ov-file#optional-url-path-mode-configuration

adr/005-responses-api.md
```diff
@@ -0,0 +1,278 @@
...
+- shorter naming
+
+#### Cons
+
+-
+
+## Option C
+
+Actually? Why do we have our own clients, when there are clients from OpenAI that are maintained by OpenAI?
+We could provide a helper function to retrieve a config based on destination, model deployment configuration and API version.
```
marikaner, line 206:
```suggestion
We could provide a helper function to retrieve a config based on destination, model deployment configuration and API version.

This is a hands-on PoC: https://github.tools.sap/I824643/responses-api-poc.
```

adr/005-responses-api.md
marikaner: This is just a naive implementation to make it work. The token provider is a function and I believe it will be called when a new token is needed again. It is not 100% correct right now, I'll leave this to the real implementation though.

adr/005-responses-api.md
```diff
@@ -0,0 +1,280 @@
...
+  const aiCoreDestination = await getAiCoreDestination(destination);
+  const { baseURL, params } = await buildHttpRequest(aiCoreDestination);
+
+  return {
+    apiVersion: config.apiVersion || '2025-08-07',
+    azureADTokenProvider: async () =>
+      (await buildHttpRequest(aiCoreDestination)).headers.authorization.split(
+        ' '
+      )[1],
+    defaultHeaders: {
```
marikaner:
```suggestion
```

adr/005-responses-api.md
```diff
@@ -0,0 +1,280 @@
...
+      'ai-resource-group': resourceGroup
+    },
+    defaultQuery: params,
+    baseURL: `${baseURL}/v2/inference/deployments/${deploymentId}`
+  };
+}
+```
+
```
marikaner:
```suggestion
```

adr/005-responses-api.md
```diff
@@ -0,0 +1,237 @@
...
+#### Pros
+
+- no maintenance, updating and documentation effort for a whole client, only it's configuration
+- users might already be used to the API, no new abstractions
+
+#### Cons
+
+- AzureOpenAI requires the `model` to be set in the payload (but apparently the value doesn't matter) => potentially forces users to pass the model twice
```
marikaner, line 234:
```suggestion
- AzureOpenAI requires the `model` to be set in the payload (but apparently the value doesn't matter) => potentially forces users to pass the model twice
- SAP Cloud SDK features like resilience could not be offered OOB anymore.
```

adr/005-responses-api.md
marikaner: Weirdly, the client requires it, but I tried it in bruno and it works without. ¯\\_(ツ)_/¯

marikaner:
> For langchain specifically, I would still consider providing a client that subclasses `ChatOpenAICompletions` (or `ChatOpenAIResponses` which many of the community integrations do) for the improved UX and reduced maintenance effort. One way to reduce/share support burden in that case would also be to upstream the integration.

Could you elaborate? What UX would specifically be improved and what maintenance effort would be reduced?

adr/005-responses-api.md
```diff
@@ -0,0 +1,238 @@
...
+- SAP Cloud SDK features like resilience could not be offered OOB anymore.
+
+## Outlook
```
marikaner, line 243:
```suggestion
#### Open Concerns

- How does OpenAI SDK handle errors?
- How can we ensure a smooth migration path, when OpenAI SDK has a major version update with breaking changes?
- Is it a problem that users could simply overwrite the "ai-client-type" header?

## Outlook
```

adr/005-responses-api.md
```diff
@@ -0,0 +1,238 @@
+# Responses, Chat and Embeddings API
+
+## Status
+
+<!-- What is the status, such as decided, proposed, outdated -> one sentence reason, superseded -> link to followup ADR. -->
+
+proposed
```
marikaner:
```suggestion
tentatively decided - reevaluate when implementing this
```

adr/005-responses-api.md
ZhongpinWang: > Weirdly, the client requires it, but I tried it in bruno and it works without. ¯\_(ツ)_/¯

Same for me. Could it be that llm-access add this to the payload? For official vendor API I believe `model` is required.

adr/005-responses-api.md
```diff
@@ -0,0 +1,282 @@
...
+## Outlook
+
+We should consider how OpenAI's WebSocket API (`Session`) plays into this.
+Option C is likely already future proof in this sense.
+
+## Option D
+
+Option C requires (and if it was optional: allows) users to set the model.
+However, the model is already defined through the deployment.
+In order to prevent confusion we could take back more ownership and wrap the official client.
```
ZhongpinWang, line 252:
```suggestion
Option C requires (and if it was optional: allows) users to set the model.
However, the model is already defined through the deployment.
In order to prevent confusion we could take back more ownership and wrap the official client.

SDK header is also important for us to track the usage, thus, we want to enforce this other than letting user add it.

From the perspective of extensibility to other model providers, the wrapper option can help us hide adoption complexity from users such as url construction, SDK header and auth token setting, deployment URL resolvement and many more.
Check the [PoC examples](https://github.tools.sap/I563080/sdk-js-wrapper-poc/tree/main/examples) for understanding the complexity of supporting other providers.
```

adr/005-responses-api.md
```diff
@@ -0,0 +1,282 @@
...
+  instructions: 'You are a coding assistant that talks like a pirate',
+  input: 'Are semicolons optional in JavaScript?'
+});
+```
+
+#### Pros
+
+- less maintenance, updating and documentation effort than for a whole client
+- users might already be used to the API, no new abstractions
+- more ownership
```
ZhongpinWang, line 272:
```suggestion
- more ownership, allowing us to control SDK header plus setting auth token, adding deployment URL resolvement and lower adoption complexity with out-of-the-box client.
```

adr/005-responses-api.md
```diff
@@ -0,0 +1,282 @@
...
+- SAP Cloud SDK features like resilience could be offered OOB
```
ZhongpinWang, line 273: [q] I wonder how we can add this 🤔

adr/005-responses-api.md
```diff
@@ -0,0 +1,282 @@
...
+
+#### Cons
+
+- tighter coupling between SDK and OpenAI amplifies: support responsibility lines more blurred
```
ZhongpinWang, line 277: We need to define support strategy / scope here if we go down this path. It can be a danger zone, e.g., when openai introduces breaking changes. Then we might be asked to make it non-breaking by adding an adapter...

adr/005-responses-api.md
davidkna-sap: I think it's possible even with Option C (for OpenAI):
```ts
import OpenAI from 'openai';
import fetch from 'my-fetch';

const client = new OpenAI({ fetch });
```

[reference](https://developers.openai.com/api/reference/typescript#customizing-the-fetch-client)

`fetchOptions` is also powerful.

ZhongpinWang: Thanks and it looks at least doable. I see the response type is from undici type. So my feeling is that our axios client cannot be directly used, not without adding some mapping (might be hard considering streaming). Worth investigation.

adr/005-responses-api.md
marikaner: but all of this is already possible with option C. auth token refresh might not work with the OpenAIClient, however I was under the impression that we would provide the AzureOpenAI client.
Why "lower adoption complexity"? Because of the removed model?

adr/005-responses-api.md
marikaner: > SDK header is also important for us to track the usage, thus, we want to enforce this other than letting user add it.

I don't see a big difference in option C and D with this. Using `withOptions()` or the the request options (second parameter in e.g. `responses.create()`. This is equally easy to overwrite.

marikaner: > From the perspective of extensibility to other model providers, the wrapper option can help us hide adoption complexity from users such as url construction, SDK header and auth token setting, deployment URL resolvement and many more.
Check the [PoC examples](https://github.tools.sap/I563080/sdk-js-wrapper-poc/tree/main/examples) for understanding the complexity of supporting other providers.

This works with the assumption, that we want the same solution for all providers, right? This should be added here, otherwise I find it confusing how this is related to this ADR.

adr/005-responses-api.md
marikaner: What I meant by this point is that resilience could be offered with the well known API from Cloud SDK if we have a wrapper, although I didn't think it through and it will likely require more effort than just mere wrapping. This wasn't part of my investigation.

adr/005-responses-api.md
marikaner: should we persist this in the ADR?

adr/005-responses-api.md
ZhongpinWang: Because of url construction. Also I have heard that there might be service side breaking changes both from the original service provider and llm-access that would affect the url. Having a wrapper would help us help our user not face this complexity.

adr/005-responses-api.md
ZhongpinWang: I agree we should not mix responses api with how we want to implement our next client. I suggest we leave now merge this ADR as-is and create a new one for client implementation + a new PoC package for that. Regarding different options, I will discuss with you in details. I have some preference in mind.

PR #1641, opened 2026-03-16
chore: Revise decision on file input API

davidkna-sap:
# Revise Decision on File Input API (ADR Update)

### Documentation

📝 Updated the Architecture Decision Record (ADR) `005-file-input-convenience.md` to reflect a revised decision on the file input API design.

### Changes

The ADR previously proposed **Option D** (a custom discriminated union type with `FileContentInput`), but this decision has been **withdrawn in favor of Option A** — exposing the API as-is, aligned with the OpenAI spec.

* `adr/005-file-input-convenience.md`:
  - Changed section heading from `## Proposed Decision` to `## Decision`.
  - Replaced the earlier decision text (Option D with custom type definitions) with a rationale for withdrawing it in favor of **Option A** (expose as-is).
  - Removed custom TypeScript type definitions (`FileContentBase`, `FileUrlContent`, `FileBase64Content`, `FileContentInput`) from the decision section.
  - Added reasoning for rejecting API-surface-changing options: non-bijective message rewriting complexity, deviation from spec and other SDK providers, misalignment with image input API, insufficient complexity reduction, and inability to prevent incorrect MIME types.
  - Added a "To discuss" section proposing basic validation that string inputs conform to the `data:` URI format, with options ranging from no validation to validating messages and message history.
  - Added a note to create a follow-up BLI to investigate helper utilities (Options E/F) for MIME type support.
  - Updated the example code to show the simplified `as-is` approach using a manually constructed data URI string passed directly as `file_data`, replacing the earlier structured `FileContentInput` object examples.

- [ ] 🔄 Regenerate and Update Summary

---
📬 [Subscribe to the Hyperspace PR Bot DL](https://url.sap/451kgs) to get the latest announcements and pilot features!

<details>
<summary>PR Bot Information</summary>

**Version:** `1.18.5` | 📖 [Documentation](https://url.sap/dy9ocn) | 🚨 [Create Incident](https://url.sap/budnv9) | 💬 [Feedback](https://url.sap/my4dn3)

- Correlation ID: `dc599660-21d7-11f1-84bd-b49dc531367a`
- Event Trigger: `issue_comment.edited`
</details>

### DIFF
--- a/adr/005-file-input-convenience.md
+++ b/adr/005-file-input-convenience.md
@@ -161,7 +161,7 @@ Options that change the API surface:
 - Does not sufficiently reduce user complexity.
 - Does not prevent issues around incorrect MIME types.
 
-To discuss:
+Regarding runtime warnings:
 
 Add basic validation that any string input starts with `data:` and contains `;base64,` and a non-empty MIME type, with helpful error messages for common formatting issues.
 
@@ -170,6 +170,8 @@ Options for validation:
 - No validation, rely on service errors and suggest improvements to error messages in the orchestration service as needed.
 - Validate for messages
 - Validate for messages and message history.
+
+Choice: **No validation** as existing error messages from the service are decent (see collapsed samples above).
 
 For support regarding help with mime types (`application/vnd.openxmlformats-officedocument.wordprocessingml.document`), create a follow-up BLI to investigate helpers (Options E/F) and consider also improving the experience of sending images.
 
--- a/adr/005-file-input-convenience.md
+++ b/adr/005-file-input-convenience.md
@@ -149,92 +149,27 @@ Helpful links:
 - [MDN Web Docs - MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)
 <!-- vale on -->
 
-## Proposed Decision
+## Decision
 
-Based on Option D, accept a discriminated union of either a base64 string or `Buffer` with an explicit MIME type, or a URL (either a data URI or a public HTTPS URL).
-The SDK will handle assembling the data URI for base64 inputs, and require users to make a conscious choice about whether to provide a URL or base64 content with a MIME type.
+Earlier Decision (Option D) is _withdrawn_ in favor of **Option A** (expose as-is).
 
-### Type Definitions
+Options that change the API surface:
 
-```ts
-interface FileContentBase {
-  filename?: string; // optional filename for both URL and base64 inputs
-}
+- Rewriting all messages both input and output has significant additional complexity and is not bijective.
+- Deviating from the spec and other SDK providers may be confusing.
+- Not aligned with the image input API.
+- Does not sufficiently reduce user complexity.
+- Does not prevent issues around incorrect MIME types.
 
-interface FileUrlContent extends FileContentBase {
-  type: 'url';
-  url: string; // RFC 2397 data URI or public HTTPS URL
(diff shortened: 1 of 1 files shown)


davidkna-sap (commented):
Draft implementation is available at https://github.com/SAP/ai-sdk-js/pull/1645

adr/005-file-input-convenience.md
```diff
@@ -149,92 +149,27 @@ Helpful links:
...
+To discuss:
 
-// Raw readFile with base64 FileContentInput
-const fileContent: FileContentInput = {
-  type: 'base64',
-  data: await readFile('./document.pdf'),
-  mimeType: 'application/pdf',
-  filename: 'document.pdf'
-};
+Add basic validation that any string input starts with `data:` and contains `;base64,` and a non-empty MIME type, with helpful error messages for common formatting issues.
```
ZhongpinWang, line 166: 👍, but I would consider log a warning instead of throwing as the official OpenAI SDK does no validation and the server-side behaviour is not documented (at least I didn't see any?) and such data URL requirement was only figured out with manual testing. But no strong opinion and I doubt OpenAI or Orchestration will ever change this server-side behaviour.

adr/005-file-input-convenience.md
```diff
@@ -149,92 +149,27 @@ Helpful links:
...
-      role: 'user',
-      content: [
-        { type: 'text', text: 'Summarise this document.' },
-        { type: 'file', file: fileContent }
-      ]
-    }
-  ]
-```
+- No validation, rely on service errors and suggest improvements to error messages in the orchestration service as needed.
+- Validate for messages
```
ZhongpinWang, line 171: I would check the server-side error. If it is good enough, I personally tend to go with the no validation option (+ optionally a warning log).

ZhongpinWang (commented):
Thanks and overall lgtm. Suggestions:

I would suggest writing a minimal example for option A.
Also consider adding some convenience for "Long MIME types" in decision section.

adr/005-file-input-convenience.md
davidkna-sap: My main issue with server-side errors is that they are not consistent between models/providers, and depending on the model they are verbatim upstream responses.

There a few sample errors in the context section already, which are decent enough, though e.g. plain base64 with gemini just leads to `Internal Server Error`.

That said, printing the warnings as implemented in https://github.com/SAP/ai-sdk-js/pull/1645 is also somewhat involved and O(n), even if it much simpler than rewriting messages.

ZhongpinWang: Can you report the 500 to orchestration?
They probably just forward the content as-is to the actual provider. So my understanding is that unless orchestration decides to restrict the content to be data URL, we can't be sure that data URL other than plain b64 is always required. I would provide this feedback to orchestration and ask how this should behave or have they tested all models.

davidkna-sap: When creating the ticket I noticed that my local bruno request body was broken, I tried it again with the correct body, and now it shows the error from the ADR.

Given that, I think leaving validation to the service should be fine.

marikaner (commented):
A bit late to the party, I agree with the comments and decisions that were made. Nothing to add from my side 😊

ZhongpinWang (commented):
~Can you check why the whole "Decision" section is underscored in preview?~ Weird, if I view the complete file I don't see underscore anymore. Whatever.

ZhongpinWang (approved):
LGTM

PR #1699, opened 2026-04-01
chore: Use action to detect changes

marikaner:
We currently use git to detect changes, I find it much easier to read when this action is used.


davidkna-sap (commented):
I would generally want to avoid depending on additional external actions, does something like `git diff --quiet origin/main -- SPEC_PATH` work? IIRC cloud sdk is also already using https://github.com/step-security/changed-files which may also work here.

PR #1703, opened 2026-04-02
chore: Sync model types from SAP Notes and tooling

KavithaSiva:
Automated sync of model types from [SAP Notes 3437766](https://me.sap.com/notes/3437766).

## Changes

Updated LiteralUnion type blocks in `packages/core/src/model-types.ts` to reflect current model availability.

- Removed `anthropic--claude-4.5-opus` (not available in orchestration)
- Removed `amazon--titan-embed-image` (not available in orchestration)

Also includes fixes to `scripts/sync-model-types.ts`:
- Fixed regex to handle multi-line type declarations (`=\nLiteralUnion<...>`)
- Retired models are now filtered before the executableId mapping lookup, so they no longer appear in the skipped-models warning
- Skipped models (unknown executableId) are now summarised at the end of the run with actionable guidance

## Definition of Done
- [x] Model names are correct
- [x] Compilation passes
- [x] Release notes / Changeset updated if needed


scripts/sync-model-types.ts
```diff
@@ -0,0 +1,314 @@
...
+  return null;
+}
+
+function hasConcreteRetirementDate(retirementDate: string): boolean {
+  // Only treat as retired if the value is a concrete date (YYYY-MM-DD).
+  // "not earlier than ...", "No current plans for retirement", "-", etc. are NOT concrete.
+  return /^\d{4}-\d{2}-\d{2}$/.test(retirementDate.trim());
+}
+
+function isRetiredSoon(retirementDate: string): boolean {
```
davidkna-sap, line 86: [req] This function should use `hasConcreteRetirementDate()`

.claude/commands/update-models.md
```diff
@@ -0,0 +1,83 @@
...
+   git commit -m "chore: sync model types from SAP Notes 3437766"
+   gh pr create --base main --title "chore: Sync model types from SAP Notes" --body "Automated sync of model types from [SAP Notes 3437766](https://me.sap.com/notes/3437766).\n\n## Changes\n\nUpdated LiteralUnion type blocks in \`packages/core/src/model-types.ts\` to reflect current model availability.\n\n## Definition of Done\n- [ ] Model names are correct\n- [ ] Compilation passes\n- [ ] Release notes / Changeset updated if needed"
+   ```
+
+## Prerequisites
+
+- **Google Chrome (or Chromium) must be installed.**
+  The `chrome-devtools-mcp` plugin connects to Chrome via the DevTools Protocol — it cannot run without a live Chrome process.
+  See step 1 for the launch command on your OS.
+- `chrome-devtools-mcp` plugin must be installed and connected in Claude Code
```
davidkna-sap: [nth] As-is, `playwright-mcp` might be easier to use.

.claude/skills/update-models/SKILL.md
davidkna-sap, line 1: [q] Doesn't the pick the latest internal version of the document instead of the latest public version?

.claude/skills/update-models/SKILL.md
```diff
@@ -0,0 +1,83 @@
...
+   - On success, **merge** the extracted rows into `scripts/sap-models.json`: read the existing file first, then update existing entries (matched by `model` field) and append any new ones.
+     Do NOT replace the whole file — use the Edit tool to apply targeted changes or add new entries.
+
+4. **Patch model-types.ts** by running the sync script:
+   ```bash
+   pnpm tsx scripts/sync-model-types.ts
+   ```
+   If the output ends with a `⚠ Skipped N model(s)` warning, show the user the listed model names and their `executableId` values and ask whether they want to add a mapping.
+   If yes, add the appropriate entry to `EXECUTABLE_ID_TO_TYPE` in `scripts/sync-model-types.ts` and re-run the script before proceeding.
+
```
davidkna-sap, line 32: [pp] It would be nice to add a step to run a script that checks if a new model can be used for chat/embedding inference via orchestration?

davidkna-sap (changes requested):

.claude/skills/update-models/SKILL.md
KavithaSiva: Didn't get you.

davidkna-sap: If we visit the SAP notes page we sometimes see an internal not-yet-public version instead of the published state for the public.

scripts/sync-model-types.ts
KavithaSiva: The `hasConcreteRetirementDate()` is a leftover, the `isRetiredSoon` function already covers cases where a concrete date is given and also when `not earlier than` phrase is used to define a date.
Right now, if the deprecation date is within the next two months, the model names eagerly removed.

.claude/commands/update-models.md
KavithaSiva: True, I tested it out, works as expected so switched to `playwright` plugin(+mcp) instead of `chrome-devtools`.

.claude/skills/update-models/SKILL.md
KavithaSiva: Okay, while I did not find an option to refer a released note, which is also available to customers, I think it is still okay, as we only expose model names, if it is available in orchestration and retirement dates, which I don't think is sensitive information.

.claude/skills/update-models/SKILL.md
KavithaSiva: It is a good idea in principle, but it would require AI Core credentials to be configured locally (e.g. a populated .env file) to run the inference call. And we would need additional scripts to load the environment variable in the shell, so Claude doesn't include it in their context.
For simplicity, for now, it's better suited as a separate manual verification step rather than part of the automated workflow.

.changeset/model-types-sync-2026-04-02.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/core': minor
+---
+
+[Improvement] Removed deprecated models `amazon--titan-embed-image`, `anthropic--claude-4.5-opus`.
```
marikaner:
```suggestion
[Improvement] Remove deprecated models `amazon--titan-embed-image` and `anthropic--claude-4.5-opus`.
```

marikaner (commented):
I like this, will try it out still.

.changeset/model-types-sync-2026-04-02.md
KavithaSiva: Hmm, will have to adjust the script to generate this instead.

marikaner (changes requested):
I found that commands have been replaced with skills: https://code.claude.com/docs/en/skills 

So, let's move it then?

scripts/sync-model-types.ts
```diff
@@ -0,0 +1,313 @@
...
+      if (isEmbedding && EMBEDDING_TYPE_MAP[typeName]) {
+        return EMBEDDING_TYPE_MAP[typeName];
+      }
+      return typeName;
+    }
+  }
+
+  return null;
+}
+function isRetiredSoon(retirementDate: string): boolean {
```
marikaner, line 86:
```suggestion
}

function isRetiredSoon(retirementDate: string): boolean {
```

scripts/sync-model-types.ts
```diff
@@ -0,0 +1,314 @@
...
+  );
+}
+
+function buildReleaseNote(
+  added: string[],
+  removed: Array<{ model: string; replacement: string; retirementDate: string }>
+): string {
+  const parts: string[] = [];
+
+  if (added.length > 0) {
```
marikaner:
```suggestion
  if (added.length) {
```

scripts/sync-model-types.ts
```diff
@@ -0,0 +1,314 @@
...
+    const names = added.map(m => `\`${m}\``).join(', ');
+    parts.push(`Added ${names} to the available model list.`);
+  }
+
+  if (removed.length > 0) {
```
marikaner:
```suggestion
  if (removed.length) {
```

scripts/sync-model-types.ts
```diff
@@ -0,0 +1,314 @@
...
+      return desc;
+    });
+    parts.push(`Removed deprecated model${removed.length > 1 ? 's' : ''} ${descriptions.join(', ')}.`);
+  }
+
+  return parts.join(' ');
+}
+
+async function syncModelTypes(): Promise<void> {
+  const { readFile } = await import('node:fs/promises');
```
marikaner: [q] Why not import it on top?

scripts/sync-model-types.ts
```diff
@@ -0,0 +1,314 @@
...
+  const skippedRows: Array<{ model: string; executableId: string }> = [];
+
+  for (const row of rows) {
+    if (isRetired(row)) {
+      retiredInfo[row.model] ??= {
+        replacement: row.suggestedReplacement,
+        retirementDate: row.retirementDate
+      };
+      continue;
+    }
```
marikaner, line 191: [q] can we use `.map()` instead?

scripts/sync-model-types.ts
```diff
@@ -0,0 +1,314 @@
...
+
+    typeToActiveModels[typeName] ??= new Set();
+    typeToActiveModels[typeName].add(row.model);
+  }
+
+  const currentContent = await import('node:fs/promises').then(fs =>
+    fs.readFile(MODEL_TYPES_PATH, 'utf8')
+  );
```
marikaner: [q] why not import "normally"?

scripts/sync-model-types.ts
```diff
@@ -0,0 +1,314 @@
...
+      }
+      return desc;
+    });
+    parts.push(`Removed deprecated model${removed.length > 1 ? 's' : ''} ${descriptions.join(', ')}.`);
+  }
+
+  return parts.join(' ');
+}
+
+async function syncModelTypes(): Promise<void> {
```
marikaner: [req] This function is very long, please consider if there is potential to break it up into multiple parts.

scripts/sync-model-types.ts
```diff
@@ -0,0 +1,314 @@
...
+
+  const releaseNote = buildReleaseNote(allAdded, allRemoved);
+  if (releaseNote) {
+    console.log('\n--- Release Note ---');
+    console.log(releaseNote);
+    console.log('--------------------');
+    await writeChangeset(releaseNote);
+  }
+
+  if (skippedRows.length > 0) {
```
marikaner:
```suggestion
  if (skippedRows.length) {
```

scripts/extract-model-table.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * Browser-side extraction script for the SAP Notes model table.
+ * Run via chrome-devtools-mcp evaluate_script on https://me.sap.com/notes/3437766.
+ * Returns a JSON-serializable array of model rows.
+ */
+(() => {
```
marikaner, line 6: [pp] Also consider splitting this up into multiple functions to improve legibility.

marikaner (dismissed):

.changeset/model-types-sync-2026-04-02.md
KavithaSiva: Fixed the script to generate the changesets.

.claude/skills/update-models/SKILL.md
```diff
@@ -0,0 +1,53 @@
...
+   - If the page shows a login screen, tell the user to log in to me.sap.com in the Playwright browser window and wait for their confirmation, then retry.
+
+2. **Extract the model table**:
+   - Read `scripts/extract-model-table.js` and pass its contents as the function body to `browser_evaluate`.
+   - The script returns either a JSON array of model rows (success) or `{ error: '...' }` (failure).
+   - If the result is an error or empty array, the session has expired — tell the user to log in to me.sap.com in the Playwright browser window and wait for their confirmation, then retry.
+   - On success, **merge** the extracted rows into `scripts/sap-models.json`: read the existing file first, then update existing entries (matched by `model` field) and append any new ones.
+     Do NOT replace the whole file — use the Edit tool to apply targeted changes or add new entries.
+
+3. **Patch model-types.ts** by running the sync script:
```
davidkna-sap, line 26: [nth] Maybe also ask the model to check if any deprecated: yes model is used in the code.

.claude/skills/update-models/SKILL.md
```diff
@@ -0,0 +1,53 @@
...
+argument-hint: "[no arguments needed]"
+---
+
+# Update Model Types from SAP Notes
+
+Syncs `packages/core/src/model-types.ts` with the current model table on SAP Notes (me.sap.com/notes/3437766).
+
+## Steps
+
+1. **Navigate to the SAP Notes page** using the Playwright MCP tool:
```
davidkna-sap, line 13: [pp] Please add `.playwright-mcp/` to `.gitignore`

scripts/sync-model-types.ts
KavithaSiva: I have added more functions.

.claude/skills/update-models/SKILL.md
KavithaSiva: Good idea, added it to the skill.

.claude/skills/update-models/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+3. **Patch model-types.ts** by running the sync script:
+   ```bash
+   pnpm tsx scripts/sync-model-types.ts
+   ```
+   If the output ends with a `⚠ Skipped N model(s)` warning, show the user the listed model names and their `executableId` values and ask whether they want to add a mapping.
+   If yes, add the appropriate entry to `EXECUTABLE_ID_TO_TYPE` in `scripts/sync-model-types.ts` and re-run the script before proceeding.
+
+4. **Check for deprecated model usage** in the codebase:
+   - From `scripts/sap-models.json`, collect the names of all models where `deprecated` is `"yes"` or the retirement date is set.
+   - Grep the repository for any of those model names (exclude `scripts/sap-models.json` itself and `packages/core/src/model-types.ts`).
```
davidkna-sap: [np]
```suggestion
   - Grep the repository for any of those model names (exclude `scripts/sap-models.json` itself, `packages/core/src/model-types.ts`, OpenAPI specs and generated files like `packages/*/src/client/**/*.ts`).
```

scripts/extract-model-table.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * Browser-side extraction script for the SAP Notes model table.
+ * Run via chrome-devtools-mcp evaluate_script on https://me.sap.com/notes/3437766.
```
davidkna-sap: [np]
```suggestion
 * Run via playwright-mcp evaluate_script on https://me.sap.com/notes/3437766.
```

scripts/extract-model-table.js
```diff
@@ -0,0 +1,67 @@
...
+      const cells = Array.from(row.querySelectorAll('td')).map(td => td.textContent);
+      const model = clean(cells[cols.modelCol]);
+      if (!model) return rows;
+      rows.push({
+        executableId: clean(cells[cols.executableIdCol]).split('\n')[0].trim(),
+        model,
+        availableInOrchestration: clean(cells[cols.orchestrationCol]),
+        deprecated: clean(cells[cols.deprecatedCol]),
+        retirementDate: clean(cells[cols.retirementCol]),
+        suggestedReplacement: clean(cells[cols.replacementCol])
```
davidkna-sap: [pp]
```suggestion
        suggestedReplacement: clean(cells[cols.replacementCol]).split('\n').map(clean).filter(Boolean).join(', ')
```

.claude/skills/update-models/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+
+3. **Patch model-types.ts** by running the sync script:
+   ```bash
+   pnpm tsx scripts/sync-model-types.ts
+   ```
+   If the output ends with a `⚠ Skipped N model(s)` warning, show the user the listed model names and their `executableId` values and ask whether they want to add a mapping.
+   If yes, add the appropriate entry to `EXECUTABLE_ID_TO_TYPE` in `scripts/sync-model-types.ts` and re-run the script before proceeding.
+
+4. **Check for deprecated model usage** in the codebase:
+   - From `scripts/sap-models.json`, collect the names of all models where `deprecated` is `"yes"` or the retirement date is set.
```
davidkna-sap, line 34: [pp] Isn't `deprecated`: `true` if retirement date is set? Sadly the check doesn't seem to be too reliable - it failed to find `text-embedding-ada-002` usages.

davidkna-sap (commented):
I have a few more nitpicks, but overall LGTM.

.claude/skills/update-models/SKILL.md
KavithaSiva: I have used a slightly different approach now, to get the list of removed models from the script run and grep against them instead.

scripts/sync-model-types.ts
```diff
@@ -0,0 +1,323 @@
...
+        console.error(`Updated ${typeName}`);
+        changed = true;
+        updated = next;
+      } else {
+        console.error(`No change for ${typeName}`);
+      }
+    }
+
+    return updated;
+  });
```
davidkna-sap, line 250: [pp] It would be better to directly format (requires `import * as prettier from 'prettier';` and ensure the models have consistent order in the file

```suggestion
  await transformFile(MODEL_TYPES_PATH, async content => {
    let updated = content;

    const sorted = Object.entries(typeToActiveModels).sort(([a], [b]) => a.localeCompare(b));
    for (const [typeName, models] of sorted) {
      if (models.size === 0) {
        continue;
      }

      const existingModels = [...extractCurrentModels(content, typeName)];
      const newBlock = buildLiteralUnionBlock(existingModels, models);
      const regex = new RegExp(
        `(export type ${typeName} =\\s*)LiteralUnion<[\\s\\S]*?>;`,
        'g'
      );
      const next = updated.replace(regex, `$1${newBlock};`);

      if (next !== updated) {
        console.error(`Updated ${typeName}`);
        changed = true;
        updated = next;
      } else {
        console.error(`No change for ${typeName}`);
      }
    }

    const options = await prettier.resolveConfig(MODEL_TYPES_PATH);
    return prettier.format(updated, options || undefined);
  });
```

davidkna-sap (approved):

PR #1718, opened 2026-04-07
chore: Separate api-docs PR and trigger another workflow for it

emincihangeri:
## Context

Closes SAP/ai-sdk-js-backlog#453.


.github/workflows/api-docs.yml
```diff
@@ -1,84 +1,89 @@
...
+    inputs:
+      version:
+        description: 'Version to generate docs for (e.g., 2.9.0).'
+        required: true
+        type: string
   workflow_dispatch:
     inputs:
       version:
-        description: 'Version to generate docs for (e.g., 2.9.0). Only the major version is used for the target folder (e.g., v2).'
+        description: 'Version to generate docs for (e.g., 2.9.0).'
```
marikaner: [q] Why do we even have to pass a version? Can't we take what is in the package.json?

.github/workflows/api-docs.yml
```diff
@@ -1,84 +1,89 @@
...
-          EOT
-
-      - name: Push generated API documentation to ai-sdk repo
+          # FULL version e.g. 2.9.0 (provided by input)
+          if [ -n "${{ github.event.inputs.version }}" ]; then
+            FULL_VERSION="${{ github.event.inputs.version }}"
+          else
+            # when called via workflow_call, inputs are available at ${{ inputs.version }}
+            FULL_VERSION="${{ inputs.version }}"
+          fi
```
marikaner: [q] How about that?
```suggestion
          FULL_VERSION="${{ github.event.inputs.version || inputs.version }}"
```

.github/workflows/api-docs.yml
```diff
@@ -1,84 +1,89 @@
...
-          # Extract major version from full version (e.g., "2.8.0" -> "v2")
-          FULL_VERSION="${{ inputs.version }}"
-          MAJOR_VERSION=$(echo "$FULL_VERSION" | cut -d. -f1)
-          VERSION="v${MAJOR_VERSION}"
+          # Determine FULL_VERSION (same logic as above)
+          if [ -n "${{ github.event.inputs.version }}" ]; then
+            FULL_VERSION="${{ github.event.inputs.version }}"
+          else
+            FULL_VERSION="${{ inputs.version }}"
+          fi
```
marikaner: [req] Instead of duplicating the code, please create a dedicated step, that retrieves the version and share the value(s) through the step output.

Please also compare how it is done [here](https://github.com/SAP/ai-sdk-js/pull/1690), maybe we can avoid bigger merge conflicts ;)

.github/workflows/bump.yml
```diff
@@ -88,25 +73,20 @@ jobs:
           BRANCH_NAME="update-release-notes-js-${{ needs.bump.outputs.version }}"
           git checkout -b "${BRANCH_NAME}"
           
-          # Stage both release notes and API docs
+          # Stage release notes only
```
marikaner:
```suggestion
          # Stage release notes
```

marikaner (changes requested):

.github/workflows/api-docs.yml
emincihangeri: This could be useful in very rare cases that there is something needs to be fixed in the API docs after a bump for a specific version but I am not sure if it is a must. :/

.github/workflows/api-docs.yml
```diff
@@ -1,84 +1,89 @@
 name: api-docs
 
-# Note: During normal releases, API docs are generated as part of the bump workflow
-# and included in the same PR as release notes.
+# Reusable workflow to generate API docs and open a PR in SAP/ai-sdk.
+# Can be triggered via workflow_call (from bump.yml) or manually via workflow_dispatch.
```
marikaner: [pp] I think we don't need this comment.

```suggestion
# Reusable workflow to generate API docs and open a PR in SAP/ai-sdk.
# Can be triggered via workflow_call (from bump.yml) or manually via workflow_dispatch.
```

.github/workflows/api-docs.yml
marikaner: Then this could be an optional field.

.github/workflows/api-docs.yml
```diff
@@ -1,21 +1,28 @@
 name: api-docs
 
-# Note: During normal releases, API docs are generated as part of the bump workflow
-# and included in the same PR as release notes.
+# Can be triggered via workflow_call (from bump.yml) or manually via workflow_dispatch.
```
marikaner: Still not needed ;) the config below already says it and it is not special.

```suggestion
```

.github/workflows/publish-release.yml
```diff
@@ -45,13 +45,13 @@ jobs:
...
-      - name: 'Check Whether Docs PR Can Be Merged'
+      - name: 'Check Whether Release Notes PR Can Be Merged'
         if: steps.check_pr_status.outputs.merged == 'false'
         uses: ./.github/actions/pr-is-mergeable
         with:
-          pr-ref: ${{ steps.determine-branch-name.outputs.docs_branch }}
-          repo: SAP/ai-sdk
-          token: ${{ steps.app-token.outputs.token }}
+          pr-ref: ${{ steps.determine-branch-name.outputs.release_notes_branch }}
+          repo: ${{ env.DOCS_REPO }}
```
marikaner: [pp] This adds unnecessary complexity + the variable does no longer exist.
```suggestion
          repo: SAP/ai-sdk
```

.github/workflows/publish-release.yml
```diff
@@ -45,13 +45,13 @@ jobs:
...
+          token: ${{ secrets.GH_CLOUD_SDK_JS_ADMIN_WRITE_TOKEN }}
```
marikaner: [req] This token no longer exists.
```suggestion
          token: ${{ steps.app-token.outputs.token }}
```

.github/workflows/publish-release.yml
```diff
@@ -87,9 +87,10 @@ jobs:
...
           permission-pull-requests: write
-      - name: 'Merge Docs PR'
+      - name: 'Merge Release Notes PR'
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
         run: |
-          echo "Attempting to merge PR for branch: ${{ needs.check-docs-pr.outputs.docs_branch }}"
-          gh pr merge --squash "${{ needs.check-docs-pr.outputs.docs_branch }}" --delete-branch --repo "${{ env.DOCS_REPO }}"
+          echo "Attempting to merge PR for branch: ${{ needs.check-release-notes-pr.outputs.release_notes_branch }}"
+          gh pr merge --squash "${{ needs.check-release-notes-pr.outputs.release_notes_branch }}" --delete-branch --repo "${{ env.DOCS_REPO }}"
```
marikaner:
```suggestion
          gh pr merge --squash "${{ needs.check-release-notes-pr.outputs.release_notes_branch }}" --delete-branch --repo "SAP/ai-sdk"
```

.github/workflows/publish-release.yml
```diff
@@ -4,11 +4,11 @@ on:
...
 
 jobs:
-  check-docs-pr:
-    name: Check Docs PR Status (Release Notes + API Docs)
+  check-release-notes-pr:
+    name: Check Release Notes PR Status
     runs-on: ubuntu-latest
     outputs:
-      docs_branch: ${{ steps.determine-branch-name.outputs.docs_branch }}
+      release_notes_branch: ${{ steps.determine-branch-name.outputs.release_notes_branch }}
```
marikaner: [pp] I didn't notice this before, but all other outputs use kebap case instead of snake case. Can you please adjust all variables here?

`release_notes_branch` => `release-notes-branch`

For env variables, e.g. `BRANCH_NAME` it is correct as is.

marikaner (changes requested):
We should remove the non-existent token + env variable, everything else is details, but would be nice to have before merging.

.github/workflows/api-docs.yml
marikaner: [req] You need to handle the case if nothing was passed still. I think there is a script to get the version from package.json. If not, I think this can be very easily be done on the command line.

marikaner: Otherwise, this is the result: https://github.com/SAP/ai-sdk/pull/459

marikaner (changes requested):

.github/workflows/api-docs.yml
```diff
@@ -28,6 +27,9 @@ jobs:
         env:
           FULL_VERSION: ${{ inputs.version }}
```
marikaner:
```suggestion
```

.github/workflows/api-docs.yml
```diff
@@ -1,15 +1,14 @@
...
 
 on:
-  workflow_dispatch:
+  workflow_call:
     inputs:
       version:
-        description: 'Version to generate docs for (e.g., 2.9.0). Only the major version is used for the target folder (e.g., v2).'
+        description: 'Version to generate docs for (e.g., 2.9.0).'
         required: true
         type: string
```
marikaner: remove this

.github/workflows/bump.yml
```diff
@@ -85,18 +80,8 @@ jobs:
...
-        run: |
-          # Extract major version from full version (e.g., "2.8.0" -> "v2")
-          FULL_VERSION="${{ needs.bump.outputs.version }}"
-          MAJOR_VERSION=$(echo "$FULL_VERSION" | cut -d. -f1)
-          VERSION="v${MAJOR_VERSION}"
-
-          # Copy generated docs to versioned folder
-          rsync -avz --delete knowledge-base/api-reference/ "${{ env.DOCS_CHECKOUT_PATH }}/static/api/${VERSION}/"
-
       - name: Open Release notes and API docs PR
```
marikaner:
```suggestion
      - name: Open release notes PR
```

.github/workflows/bump.yml
```diff
@@ -45,12 +45,12 @@ jobs:
           version: ${{ steps.bump.outputs.version }}
           user-name: ${{ vars.SAP_AI_SDK_BOT_NAME }}
 
-  prepare-release-notes-and-docs:
-    name: Prepare Release Notes and API Documentation
+  prepare-release-notes:
```
marikaner: [pp] This is not only preparation but also creating a PR.
```suggestion
 release-notes:
```

.github/workflows/api-docs.yml
marikaner: [req] After reviewing this once more, I noticed that you don't use the version for anything different than the branch name and title of the PR and commit. As discussed, you fixed this for the `workflow_dispatch`, but not the `workflow_call`.
When you call the workflow from bump as is right now, yes, you are not running on the bumped version, but on the state of `main` pre bump (https://docs.github.com/en/enterprise-server@3.19/actions/how-tos/reuse-automations/reuse-workflows?utm_source=chatgpt.com#calling-a-reusable-workflow). But changing the title of the api-docs commit, won't change that.

See my comments in bump for suggestions.

.github/workflows/bump.yml
```diff
@@ -107,30 +92,33 @@ jobs:
...
             fi
           fi
+
+  generate-api-docs:
+    name: Generate and Push API Documentation
+    needs: [bump]
+    uses: ./.github/workflows/api-docs.yml
+    secrets: inherit
+    with:
+      version: ${{ needs.bump.outputs.version }}
```
marikaner: [req] AFAIK you have 2 options:
1. Call the workflow using the gh cli.
2. trigger the workflow on push to the tag: https://docs.github.com/en/enterprise-server@3.19/actions/reference/workflows-and-actions/workflow-syntax?utm_source=chatgpt.com&versionId=enterprise-server%403.19&productId=actions&restPage=how-tos%2Creuse-automations%2Creuse-workflows#example-including-branches-and-tags

marikaner (changes requested):

marikaner (commented):
one more thought: I think it would be even better, if you made a reusable create-api-docs-pr action. Then you could use that directly in the api-docs.yml and bump.yml

emincihangeri:
> one more thought: I think it would be even better, if you made a reusable create-api-docs-pr action. Then you could use that directly in the api-docs.yml and bump.yml

So the bump workflow calling the api-docs.yml will no longer be the logic? but we still want to have the api-docs.yml for manual triggering then, right?

.github/workflows/bump.yml
emincihangeri: I tried to implement with the 2nd option now. Looks more clean.

.github/actions/create-api-docs-pr/action.yml
```diff
@@ -0,0 +1,98 @@
...
+    description: 'Name for the bot git config.'
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Print Action Start
+      shell: bash
+      run: echo ">>>>> Starting Create API Docs PR Action; Not printing inputs, as they might contain sensitive information"
+
```
marikaner: [pp] I think this is not necessary. Let's remove this.
```suggestion
```

.github/actions/create-api-docs-pr/action.yml
```diff
@@ -0,0 +1,98 @@
...
+        git config user.name "$BOT_NAME"
+        git checkout -b "$BRANCH_NAME"
+        git add -A
+        if git commit -m 'Update SAP AI SDK for JavaScript API documentation v${{ steps.version.outputs.major-version }}'; then
+          git push -u origin "$BRANCH_NAME"
+          echo "pushed=true" >> "$GITHUB_OUTPUT"
+        else
+          echo "No changes to commit. Skipping push."
+          echo "pushed=false" >> "$GITHUB_OUTPUT"
+        fi
```
marikaner: [req] I think the if else conditions here are redundant. We already have the check whether there are changes. If there are no changes, we never enter this step. So it would suffice to do the if block and not create a GITHUB_OUTPUT.

.github/actions/create-api-docs-pr/action.yml
```diff
@@ -0,0 +1,98 @@
...
+
+    - name: Create pull request
+      if: steps.check-changes.outputs.docs == 'true' && steps.commit-push.outputs.pushed == 'true'
```
marikaner:
```suggestion
      if: steps.check-changes.outputs.docs == 'true'
```

.github/actions/create-api-docs-pr/action.yml
```diff
@@ -0,0 +1,98 @@
...
+          --head "$BRANCH_NAME" \
+          --title "Update v${{ steps.version.outputs.major-version }} API documentation (Javascript)" \
+          --body "Update the v${{ steps.version.outputs.major-version }} API documentation for SAP Cloud SDK for AI, JavaScript.
+
+        Automatically generated by the [api-docs workflow run](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})."
+
+    - name: Print Action End
+      if: always()
+      shell: bash
+      run: echo "<<<<< Finished Create API Docs PR Action"
```
marikaner: [req] This brings no additional value.
```suggestion
```

.github/workflows/bump.yml
```diff
@@ -45,12 +45,12 @@ jobs:
           version: ${{ steps.bump.outputs.version }}
           user-name: ${{ vars.SAP_AI_SDK_BOT_NAME }}
 
-  prepare-release-notes-and-docs:
-    name: Prepare Release Notes and API Documentation
+  release-notes:
+    name: Prepare Release Notes
```
marikaner: [pp] It is not just preparation
```suggestion
    name: Create Release Notes PR
```

marikaner: [pp] Or just remove it?

```suggestion
```

.github/workflows/bump.yml
```diff
@@ -107,30 +92,25 @@ jobs:
...
+            git commit -m "Update JS release notes for ${{ needs.bump.outputs.version }}"
             git push -u origin "${BRANCH_NAME}"
-            PR_BODY="Auto-created by bump workflow.
-
-            This PR includes:
-            - Updated release notes for version ${{ needs.bump.outputs.version }}
-            - Updated API documentation"
-            PR_TITLE="Update JS Release Notes and API Docs for ${{ needs.bump.outputs.version }}"
+            PR_BODY="Auto-created by bump workflow."
+            PR_TITLE="Update JS Release Notes for ${{ needs.bump.outputs.version }}"
```
marikaner: [q] Shouldn't this be:
```suggestion
            PR_TITLE="chore: Update JS Release Notes for ${{ needs.bump.outputs.version }}"
```
or
```suggestion
            PR_TITLE="chore(docs): Update JS Release Notes for ${{ needs.bump.outputs.version }}"
```

marikaner (changes requested):
I still have some requests. I think it was not necessary to move the api-docs implementation to an action (my bad) but it is a good idea, if we want to align this between Cloud SDK and AI SDK, so let's keep it as is.

.github/actions/create-api-docs-pr/action.yml
```diff
@@ -0,0 +1,84 @@
...
+      run: |
+        FULL_VERSION=$(node -p "require('./package.json').version")
+        MAJOR_VERSION=$(echo "$FULL_VERSION" | cut -d. -f1)
+        echo "major-version=$MAJOR_VERSION" >> "$GITHUB_OUTPUT"
+
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        repository: SAP/ai-sdk
+        path: ai-sdk
+        token: ${{ inputs.token }}
```
davidkna-sap, line 30: [pp]
```suggestion
        token: ${{ inputs.token }}
        persist-token: true
```

.github/actions/create-api-docs-pr/action.yml
```diff
@@ -0,0 +1,84 @@
...
+
+    - name: Copy API docs
+      shell: bash
+      env:
+        MAJOR_VERSION: ${{ steps.version.outputs.major-version }}
+      run: |
+        rsync -avz --delete knowledge-base/api-reference/ ai-sdk/static/api/v${MAJOR_VERSION}/
```
davidkna-sap: [pp]
```suggestion
        rsync -avz --delete knowledge-base/api-reference/ "ai-sdk/static/api/v${MAJOR_VERSION}/"
```

.github/workflows/publish-release.yml
```diff
@@ -92,9 +92,10 @@ jobs:
...
           permission-pull-requests: write
-      - name: 'Merge Docs PR'
+      - name: 'Merge Release Notes PR'
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
         run: |
-          echo "Attempting to merge PR for branch: ${{ needs.check-docs-pr.outputs.docs_branch }}"
-          gh pr merge --squash "${{ needs.check-docs-pr.outputs.docs_branch }}" --delete-branch --repo "${{ env.DOCS_REPO }}"
+          echo "Attempting to merge PR for branch: ${{ needs.check-release-notes-pr.outputs.release-notes-branch }}"
+          gh pr merge --squash "${{ needs.check-release-notes-pr.outputs.release-notes-branch }}" --delete-branch --repo "SAP/ai-sdk"
```
davidkna-sap: [pp]
```suggestion
          gh pr merge --squash "${{ needs.check-release-notes-pr.outputs.release-notes-branch }}" --delete-branch --auto --repo "SAP/ai-sdk"
```

.github/workflows/publish-release.yml
```diff
@@ -25,20 +25,20 @@ jobs:
...
+          echo "release-notes-branch=$BRANCH_NAME" >> "$GITHUB_OUTPUT"
 
       - name: Check PR Status (Exists, Merged)
-        id: check_pr_status
+        id: check-pr-status
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
         run: |
-          BRANCH_NAME="${{ steps.determine-branch-name.outputs.docs_branch }}"
+          BRANCH_NAME="${{ steps.determine-branch-name.outputs.release-notes-branch }}"
```
davidkna-sap: [pp] move `BRANCH_NAME` assignment into `env:`

.github/workflows/bump.yml
```diff
@@ -107,30 +91,25 @@ jobs:
...
+          
           if git diff --staged --quiet; then
             echo "No changes to commit."
-            # Output an empty pr_url if no changes
-            echo "pr_url=" >> "$GITHUB_OUTPUT"
+            # Output an empty pr-url if no changes
+            echo "pr-url=" >> "$GITHUB_OUTPUT"
           else
-            git commit -m "Update JS release notes and API documentation for ${{ needs.bump.outputs.version }}"
+            git commit -m "Update JS release notes for ${{ needs.bump.outputs.version }}"
```
davidkna-sap: [pp] Avoid using action inputs directly and assign to an environment variable via `env:` to use it.

davidkna-sap (commented):
I've made some suggestions (also to pre-emptively reduce conflicts with #1822), please have a look - otherwise lgtm.

marikaner (approved):
LGTM. Thanks for putting that much effort into this 😊

PR #1740, opened 2026-04-14
feat(orchestration): Exclude EmbeddingMultiFormat from public EmbeddingData type

KavithaSiva:
## Summary

Temporary fix until we try out how multi format responses with Embedding looks like.

- `EmbeddingMultiFormat` was added to the generated `Embedding` union but is not yet ready for public API exposure
- Temporarily excludes it from `EmbeddingData.embedding` via `Exclude<Embedding, EmbeddingMultiFormat>`
- Will be widened back to `Embedding` once multi-format support is fully tested

## Test plan

- [x] No runtime changes — type-only modification verified by TypeScript compiler

### DIFF
--- a/.changeset/nice-emus-train.md
+++ b/.changeset/nice-emus-train.md
@@ -2,4 +2,5 @@
 '@sap-ai-sdk/orchestration': minor
 ---
 
-[feat] Update orchestration specification to 0.115.19. Multiple embedding output formats are not yet supported in the orchestration embedding client.
+[feat] Update orchestration specification to 0.115.19. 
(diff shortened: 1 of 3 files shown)


davidkna-sap (approved):
LGTM, but I would also consider restricting `EmbeddingsModelParams`'s `encoding_format`

KavithaSiva:
> LGTM, but I would also consider restricting `EmbeddingsModelParams`'s `encoding_format`

Good point, I excluded multiple arrays in input which probably triggers the `EmbeddingMultiFormat`. WDY think?

davidkna-sap (approved):

PR #1746, opened 2026-04-15
chore: Use `client-id` instead of `app-id`

marikaner:
## Context

`app-id` is deprecated.

Relates to SAP/ai-sdk-js-backlog#522.


KavithaSiva (approved):
LGTM, thanks for taking care.

PR #1759, opened 2026-04-20
feat: Add LLM-Batch Package

InjunPark-sap:
## Summary

Draft PR for the Batch API e2e testing task.

### Generated Client
- Created `@sap-ai-sdk/batch-api` package following the same pattern as `prompt-registry`
- Added `batch-service.yaml` spec from the llm-batch-service repo
- Generated TypeScript client using SAP Cloud SDK OpenAPI generator (`pnpm generate`)
- `BatchesApi` includes all 6 endpoints: list, create, get, delete, status, cancel


InjunPark-sap:
### Update: Shortened Batch API Method Names @KavithaSiva 
- Based on feedback from the team review meeting, the generated client method names 
were shorten (e.g. `batchServiceControllerBatchControllerListBatches`).
- Fixed by adding `x-sap-cloud-sdk-operation-name` to all endpoints in `batch-service.yaml` 
and regenerating the client

KavithaSiva:
> ### Update: Shortened Batch API Method Names @KavithaSiva
> * Based on feedback from the team review meeting, the generated client method names
>   were shorten (e.g. `batchServiceControllerBatchControllerListBatches`).
> * Fixed by adding `x-sap-cloud-sdk-operation-name` to all endpoints in `batch-service.yaml`
>   and regenerating the client

Thanks for the update.
We shouldn't modify the spec ourselves, but rather give this as a feedback to service owners.
Please revert this change and add this to a list we collect for feedback.
You can add this list as a comment here or in the BLI.

InjunPark-sap:
[a8efc4d](https://github.com/SAP/ai-sdk-js/pull/1759/commits/a8efc4d2126adf4faa56a18679a97e481948a06b)

While integrating the Batch API with SAP Cloud SDK's OpenAPI generator, we identified the following improvement that would benefit SDK consumers:

- Add x-sap-cloud-sdk-operation-name to all endpoints

The current spec generates very long method names from the operation IDs, e.g.:

```
batchServiceControllerBatchControllerListBatches()
batchServiceControllerBatchControllerCreateBatch()
batchServiceControllerBatchControllerGetBatch()
```

Adding the SAP Cloud SDK extension would allow generating shorter, more ergonomic names:

```
x-sap-cloud-sdk-operation-name: listBatches
x-sap-cloud-sdk-operation-name: createBatch
x-sap-cloud-sdk-operation-name: getBatch
```

This is already used in other SAP services (e.g., prompt-registry) and significantly improves developer experience.

InjunPark-sap:
fix: update batch-api spec with correct paths and required headers [29dd33d](https://github.com/SAP/ai-sdk-js/pull/1759/commits/29dd33da83de431f897a6d3998c2db227cf6fd8a)

Note: These are not spec enhancements but corrections required for the client to actually work against the live service.

While validating the integration against Bruno collection, I found two issues with the original spec:

1. Incorrect base path — The spec defined paths as `/batches`, but the actual AI Core gateway route is `/llm-batch-service/v1/batches`. Without this fix, all requests returned 404.

2. Missing required headers — Bruno sends `AI-Resource-Group` and `AI-Main-Tenant headers` on every request. These were not defined in the original spec, so the generated client had no way to pass them. I added them as global parameters referenced by all 6 endpoints.

InjunPark-sap:
feat: add batch-api sample code for service consumption ([b8f49d0](https://github.com/SAP/ai-sdk-js/commit/b8f49d0))

Added `sample-code/src/batch-api.ts` that wraps the generated `BatchesApi` client, replicating all 6 operations from the Bruno collection in code (`listBatches()`, `createBatch(inputUri, outputUri)`, `getBatchById(batchId)`, `getBatchStatus(batchId)`, `cancelBatch(batchId)`, `deleteBatch(batchId)`)

Note: The `createBatch` parameters were validated against the Bruno collection to match the live service configuration. (with object store URI format `ai://s3secret/...`, provider `azure-openai`)

InjunPark-sap:
Following @KavithaSiva 's feedback, the spec modifications were reverted and the approach
was aligned with existing patterns in the repo (e.g., `document-grounding`).

- Reverted spec paths from `/llm-batch-service/v1/batches` back to `/batches` (original)
- Created `options-per-service.json` to inject `basePath: /llm-batch-service/v1` into the generator without touching the spec directly

- Regenerated client and updated `sample-code` and unit test to pass `AI-Resource-Group` via `addCustomHeaders()` since it is not declared in the spec

---

> **Note for Junjie: Feedback for service owners**
>
> During integration, two header-related issues were identified in the OpenAPI spec:
>
> 1. **`AI-Resource-Group` is missing from the spec** — this is a required header for
>    all endpoints but is not declared. 
>
> 2. **`AI-Main-Tenant` is undocumented** — present in the Bruno collection but absent
>    from `help_doc.md` and the spec. Confirmed optional via testing (200 without it).
>    Please clarify if officially supported and document accordingly.

InjunPark-sap:
## OpenAI Batch SDK Comparison + s3_magic.ts Analysis

As discussed in our sync, I investigated how OpenAI handles batch output reading
in their SDK to inform potential convenience features for our SDK.

---

### Batch functionality of OpenAI SDK

**4 methods.**(**Source:** [`openai-node/src/resources/batches.ts`](https://github.com/openai/openai-node/blob/master/src/resources/batches.ts))

```typescript
batches.create(...)
batches.retrieve(batchId)
batches.cancel(batchId)
batches.list(...)
```

Input files are managed via `files.upload()` through the Files API — **S3 is
fully abstracted away from the user.**

---

### Convenience for reading output files in OpenAI SDK

**Via `files.content()`.**(**Source:** [OpenAI Batch Guide — Step 5: Retrieving the Results](https://platform.openai.com/docs/guides/batch#5-retrieving-the-results))

```typescript
// OpenAI: just 2 lines to get the full output
const file = await openai.files.content(batch.output_file_id);
const text = await file.text(); // full JSONL string
```

- Returns the full JSONL content at once (not streaming)

- Searched and confirmed in community discussion:
  [openai-node #898](https://github.com/openai/openai-node/issues/898),
  [openai-node #958](https://github.com/openai/openai-node/issues/958)

---

### OpenAI SDK **does not** auto-parse output into typed objects?

**Raw string only.**(**Source:** [OpenAI Batch Output Format](https://platform.openai.com/docs/guides/batch#the-response-body))

The official output example:
```jsonl
{"id":"batch_req_1","custom_id":"request-1","response":{"status_code":200,"body":{"id":"chatcmpl-abc","object":"chat.completion","choices":[{"message":{"role":"assistant","content":"Hello!"}}]}},"error":null}
```

Users receive this as a raw string and must manually call `JSON.parse(line)`.
**OpenAI does not cast `response.body` to a `ChatCompletion` type.**

---

### Structural Difference: OpenAI vs. SAP llm-batch-service

| | OpenAI | SAP llm-batch-service |
|---|---|---|
| Storage layer | Files API (fully managed) | BTP Object Store (S3-compatible) |
| S3 exposure | Hidden from user | User manages S3 credentials directly |
| Reading output | **2 lines** (see above) | Requires full S3 client setup |
| Output location | Internal file ID | `ai://secret-name/{batchId}/output.jsonl` |

KavithaSiva:
To summarise, the following are the feedback points for the Batch service owners:

- The endpoints have really long names, and must use the `x-sap-cloud-sdk-operation-name` flag to set readable names for generation. More details [here](https://github.com/SAP/ai-sdk-js/pull/1759#issuecomment-4282707569)
- `/llm-batch-service/v1` prefix sub-path is not available in any part of the specification, but is needed to access the batch service. 
- `AI-Resource-Group` header should be added to as a required parameter of the endpoint.
- `AI-Main-Tenant` header is not documented anywhere but is used in the Bruno collection. Is this header valid?

cc @jjtang1985

KavithaSiva:
Each line is a JSON(.jsonl) in the input and the output processed by the batch service.
The shapes of these JSON structure don't map exactly to existing types.
We can consider if we would like to introduce new wrapper types to make consumption easier for customers.

Input: The `body.messages` property matches `UserChatMessage`
```
{
  "custom_id": "request-1",
  "method": "POST",
  "url": "/v1/chat/completions",
  "body": {
    "model": "gpt-4.1",
    "messages": [{ "role": "user", "content": "What is machine learning?" }],
    "max_tokens": 150
  }
}
```

Output: The `response.body` property matches `LlmModuleResult`.

```
{
  "custom_id": "request-2",
  "response": {
    "body": {
      "id": "chatcmpl-DX3NsnqdSkPA5ZKav1Y4Y5JrKmfcL",
      "object": "chat.completion",
      "created": 1776770204,
      "model": "gpt-4.1-2025-04-14",
      "choices": [
        {
          "index": 0,
          "message": {
            "role": "assistant",
            "content": "Sure! Here’s a simple explanation of neural networks:\n\n**What is a neural network?**\n\nA neural network is a computer system designed to think a little bit like the human brain. It learns to recognize patterns and make decisions.\n\n**How does it work?**\n\n1. **Neurons:** Just like the brain has billions of nerve cells (neurons), a neural network has lots of little units called “artificial neurons.”  \n2. **Layers:** These neurons are arranged in layers. There’s an input layer (where information enters), one or more hidden layers (where information is processed), and an output layer (where results come out).  \n3. **Connections:** Each neuron is connected to many others, a bit like a web.",
            "refusal": null,
            "annotations": []
          },
          "finish_reason": "length",
          "logprobs": null,
          "content_filter_results": {
            "hate": { "filtered": false, "severity": "safe" },
            "self_harm": { "filtered": false, "severity": "safe" },
            "sexual": { "filtered": false, "severity": "safe" },
            "violence": { "filtered": false, "severity": "safe" }
          }
        }
      ],
      "usage": {
        "completion_tokens": 150,
        "prompt_tokens": 13,
        "total_tokens": 163,
        "completion_tokens_details": {
          "accepted_prediction_tokens": 0,
          "audio_tokens": 0,
          "reasoning_tokens": 0,
          "rejected_prediction_tokens": 0
        },
        "prompt_tokens_details": { "audio_tokens": 0, "cached_tokens": 0 }
      },
      "system_fingerprint": "fp_63d8723934",
      "prompt_filter_results": [
        {
          "prompt_index": 0,
          "content_filter_results": {
            "hate": { "filtered": false, "severity": "safe" },
            "self_harm": { "filtered": false, "severity": "safe" },
            "sexual": { "filtered": false, "severity": "safe" },
            "violence": { "filtered": false, "severity": "safe" }
          }
        }
      ]
    },
    "request_id": "70534487-ceb4-4934-b737-faaf09a2dacd",
    "status_code": 200
  },
  "error": null
}
```

tests/e2e-tests/src/batch-api.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    const result = await listBatches();
+
+    expect(result.count).toBeGreaterThanOrEqual(0);
+    expect(Array.isArray(result.resources)).toBe(true);
+  });
+
+  it('should create a batch job, cancel it, then delete it', async () => {
+    const response = await createBatch(
+      'ai://s3secret/input-batch.jsonl',
+      'ai://s3secret/'
```
ZhongpinWang: Check my comment for Java https://github.com/SAP/ai-sdk-java/pull/850#discussion_r3201326674
```suggestion
      'ai://s3secret/output/'
```

I suggest to set a "folder" name after the `/` to avoid the current bug behaviour that batch service prepend a `/` in the s3 file name.

ZhongpinWang: Also be aware that we are sharing the same s3 bucket, same input file. Also think if we want to have the same output folder name as the prefix?

tests/e2e-tests/src/batch-api.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      { retries: 10, minTimeout: 3000 }
+    );
+
+    await deleteBatch(id);
+  });
+
+  it('should create a batch job, wait for completion, download output, then delete it', async () => {
+    const response = await createBatch(
+      'ai://s3secret/input-batch.jsonl',
+      'ai://s3secret/'
```
ZhongpinWang: same here

ZhongpinWang (commented):
Try file api if you want to get the output. We better delete old output files in our e2e.

Also same as I suggested in the Java ticket, it is a good practice to not let your test depend on any system configuration outside. Meaning that `input-batch.jsonl` should be created on the fly inside the test instead of being a dependency to the test. Test may fail one day if the `input-batch.jsonl` is somehow gone or invalid. Think about next time the test fails because `gpt-4.1` is deprecated, we would need to change the file in S3 again to fix a test.

InjunPark-sap:
> Try file api if you want to get the output. We better delete old output files in our e2e.
> 
> Also same as I suggested in the Java ticket, it is a good practice to not let your test depend on any system configuration outside. Meaning that `input-batch.jsonl` should be created on the fly inside the test instead of being a dependency to the test. Test may fail one day if the `input-batch.jsonl` is somehow gone or invalid. Think about next time the test fails because `gpt-4.1` is deprecated, we would need to change the file in S3 again to fix a test.

Okay, good idea! Then it doesn't need to make input or output files in external, but just directly working inside of package code.  

So, if I get it correctly, my previous commit only added the functionality to "download the output using FileApi". However, your preference is that, this time around, it would be better to make the *entire* test lifecycle `upload → execute → download → delete` self-contained. Is that right? 

@ZhongpinWang

InjunPark-sap:
## Resolve CI conflict

- These fixes resolve CI failures caused by two issues found after merging origin/main:
URL double-prefixing in the generated client, and `@experimenta`l tag being lost on regeneration.

**options-per-service.json**
Remove basePath — batch-api spec already includes the full path `/llm-batch-service/v1/batches`, unlike document-grounding which uses short paths `/pipelines` and relies on basePath. Having both caused 404.


**package.json + `postgenerate-batch-api.ts`**
Inject `@experimental` via postgenerate so it survives re-generation.
Same pattern as `packages/rpt` make-rpt-internal + `scripts/postgenerate-rpt.ts`.


**batches-api.ts**
Regenerated. `_defaultBasePath: undefined (correct)`, `@experimental` auto-injected.

ZhongpinWang:
> it would be better to make the _entire_ test lifecycle `upload → execute → download → delete` self-contained. Is that right?

Exactly. 👍 Avoid pre-existing static test data to make the test self-contained. And also separate test data from Java.

InjunPark-sap:
Address @ZhongpinWang's feedback: remove S3 file dependency, clean up artifacts after each run.

**batch-api.ts**
- Add `uploadBatchInput` / `deleteFile` using `FileApi.fileUpload` / `fileDelete`
- Add `outputFolder` param to `downloadBatchOutput` — avoids leading-slash bug when output URI has no folder prefix (ref: ai-sdk-java#850)

**index.ts** export new/updated functions

**batch-api.test.ts** — each test now manages its full lifecycle:
- upload → create → poll → download → delete output → delete job → delete input (`finally`)

sample-code/src/llm-batch.ts
davidkna-sap, line 1: The file-api related functions should perhaps be added as a convenience layers to the ai-api package.
Likewise for the Batch API package itself (e.g. `createBatch()` albeit without sample-code-specific input handling.

tests/e2e-tests/src/batch-api.test.ts
davidkna-sap: For avoiding conflicts with Java we can also reconfigure the object store secret to include a sub-path of the bucket.

sample-code/src/llm-batch.ts
InjunPark-sap: I'd like to confirm the direction before making changes.
Based on your suggestion, I'm considering the following structure:
```
packages/ai-api/src/
  └── file-convenience.ts  ← uploadFile(), deleteFile(), downloadFile() 

packages/batch-api/src/
  └── batch-convenience.ts ← createBatch(), listBatches() 등 래퍼

sample-code/src/batch-api.ts
```

1. Move reusable FileApi wrappers (uploadFile, downloadFile, deleteFile) into the `@sap-ai-sdk/ai-api` package as a convenience layer (for any user).

2. Add convenience wrappers for BatchesApi operations (e.g. createBatch, listBatches) directly into the `@sap-ai-sdk/batch-api` package, without the sample-code-specific input handling.

3. Keep sample-code focused on test-specific logic, importing from the packages above.

Does this match what you had in mind? @davidkna-sap

davidkna-sap: Sorry I misremembered and thought the packages had some convenience layers already - let's keep things as-is for now.

packages/llm-batch/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@sap-ai-sdk/batch-api",
+  "version": "2.9.0",
```
KavithaSiva, line 3: [q] Can you investigate if with our release scripts it is possible to release this package alone with `2.9.0-beta.0` ?
Looks like you would need this additional entry:
```
 "publishConfig": {
    "tag": "beta",
    "access": "public"
  }
```
For e.g vue seems to be doing something like this https://www.npmjs.com/package/@vue/reactivity?activeTab=versions

tests/e2e-tests/src/llm-batch.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  deleteFile
+} from '@sap-ai-sdk/sample-code';
+import { loadEnv } from './utils/load-env.js';
+
+loadEnv();
+
+const secretName = 's3secret';
+const outputFolder = 'output/';
+
+describe('batch api', () => {
```
KavithaSiva, line 34: [q] I got a 415, when I tried running the tests locally, are they working for you locally?

KavithaSiva (commented):
Thanks for all the work, mostly LGTM,  I added two questions.

tests/e2e-tests/src/llm-batch.test.ts
InjunPark-sap: No, same issue here. It's the bug Charles reported like `http-client.ts` wraps the blob in `JSON.stringify`, sending a broken body with the wrong Content-Type into 415.

InjunPark-sap: I will fix it and ask you to review in this PR

InjunPark-sap:
### Blob upload sending corrupted body (HTTP 415)

- Issues causing `uploadBatchInput` to fail with HTTP 415:

1. http-client.ts applied `JSON.stringify()` to all non-`FormData` bodies. Passing a `Blob` resulted in `"{}"` being sent. The file content was completely lost.
2. `FileApi.fileUpload hardcodes `'content-type': '*/*'`, which overrides any MIME type set on the `Blob`. The server rejects this content type for file uploads (requires `text/csv`).


- Changes:

1. http-client.ts: Added `Blob` to the bypass condition alongside `FormData`, so raw binary content is passed through to axios unchanged.
2. http-client.test.ts: Added a test verifying that `Blob` body is transmitted as-is without stringification.
3. batch-api.ts: Changed Blob MIME type to `text/csv` and added explicit `content-type: text/csv` header override in `.execute()` to bypass the `*/*` default in `FileApi`.

> Issues reported by: @davidkna-sap , @CharlesDuboisSAP , @KavithaSiva

packages/batch-api/src/spec/options-per-service.json
```diff
@@ -0,0 +1,6 @@
+{
+  "src/spec/batch-service.yaml": {
+    "packageName": "batch-service",
+    "directoryName": "batch-service"
```
davidkna-sap: [req] But use something else in case we decide on different package name:
```suggestion
    "packageName": "llm-batch",
    "directoryName": "llm-batch"
```

packages/llm-batch/.gitignore
davidkna-sap, line 1: [q] I this file still needed?

packages/llm-batch/config.example.json
davidkna-sap, line 1: [q] I this file still needed?

packages/batch-api/jest.config.mjs
```diff
@@ -0,0 +1,5 @@
+import config from '../../jest.config.mjs';
+export default {
+  ...config,
+  displayName: 'batch-api',
```
davidkna-sap: [req] Don't forget to update name everywhere.

packages/batch-api/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@sap-ai-sdk/batch-api",
```
davidkna-sap: [req] Don't forget to update name everywhere.

sample-code/src/llm-batch.ts
```diff
@@ -0,0 +1,144 @@
...
+ * @returns Batch list response.
+ */
+export async function listBatches(): Promise<BatchListResponse> {
+  return BatchesApi.listBatches(defaultHeaders).execute();
+}
+
+/**
+ * Create a new batch job.
+ * @param inputUri - Object store URI of the input .jsonl file (e.g. ai://s3secret/input-batch.jsonl).
+ * @param outputUri - Object store URI of the output directory (e.g. ai://s3secret/).
```
davidkna-sap: [req] Also handle this elsewhere (keeping any runtime sample-code as-is is fine, but please use main placeholder syntax in comments/docs):
```suggestion
 * @param inputUri - Object store URI of the input .jsonl file (e.g. ai://OBJECT_STORE_SECRET/input-batch.jsonl).
 * @param outputUri - Object store URI of the output directory (e.g. ai://OBJECT_STORE_SECRET/).
```

sample-code/src/llm-batch.ts
```diff
@@ -0,0 +1,144 @@
...
+ * @param batchId - The ID of the completed batch job.
+ * @returns The output file as a Blob.
+ */
+export async function downloadBatchOutput(
+  secretName: string,
+  outputFolder: string,
+  batchId: string
+): Promise<Blob> {
+  return FileApi.fileDownload(
+    `${secretName}//${outputFolder}${batchId}/output.jsonl`,
```
davidkna-sap: [req]
Please confirm if `//` is still needed or consider moving to a sub-directory to keep it forward-compatible.

sample-code/src/batch-api.ts
```diff
@@ -0,0 +1,144 @@
...
+ * @returns The object store URI to use as the batch input URI.
+ */
+export async function uploadBatchInput(
+  secretName: string,
+  fileName: string
+): Promise<string> {
+  const content = [
+    '{"custom_id":"request-1","method":"POST","url":"/v1/chat/completions","body":{"model":"gpt-4.1","messages":[{"role":"user","content":"What is machine learning?"}],"max_tokens":150}}',
+    '{"custom_id":"request-2","method":"POST","url":"/v1/chat/completions","body":{"model":"gpt-4.1","messages":[{"role":"user","content":"Explain neural networks in simple terms"}],"max_tokens":150}}'
+  ].join('\n');
```
davidkna-sap: [pp] Readability
```suggestion
  const content = [
    {"custom_id":"request-1","method":"POST","url":"/v1/chat/completions","body":{"model":"gpt-4.1","messages":[{"role":"user","content":"What is machine learning?"}],"max_tokens":150}},
    {"custom_id":"request-2","method":"POST","url":"/v1/chat/completions","body":{"model":"gpt-4.1","messages":[{"role":"user","content":"Explain neural networks in simple terms"}],"max_tokens":150}}
  ].map(item => JSON.stringify(item)).join('\n');
```

sample-code/src/batch-api.ts
```diff
@@ -0,0 +1,144 @@
...
+  const blob = new Blob([content], { type: 'text/csv' });
```
davidkna-sap: [pp]
```suggestion
  // text/csv is needed to keep the API from type-checking the input
  const blob = new Blob([content], { type: 'text/csv' });
```

sample-code/src/batch-api.ts
```diff
@@ -0,0 +1,144 @@
...
+ */
+export async function downloadBatchOutput(
+  secretName: string,
+  outputFolder: string,
+  batchId: string
+): Promise<Blob> {
+  return FileApi.fileDownload(
+    `${secretName}//${outputFolder}${batchId}/output.jsonl`,
+    { 'AI-Resource-Group': defaultHeaders['AI-Resource-Group'] }
+  ).execute();
```
davidkna-sap: [pp] Parse the output
```suggestion
  const blob = FileApi.fileDownload(
    `${secretName}//${outputFolder}${batchId}/output.jsonl`,
    { 'AI-Resource-Group': defaultHeaders['AI-Resource-Group'] }
  ).execute();
  const text = blob.text();
  return text.trim().split('\n').map(line => JSON.parse(line));
```

sample-code/src/llm-batch.ts
```diff
@@ -0,0 +1,144 @@
...
+export async function deleteBatch(
+  batchId: string
+): Promise<BatchDeleteResponse> {
+  return BatchesApi.deleteBatch(batchId, defaultHeaders).execute();
+}
+
+/**
+ * Upload a JSONL input file for batch processing.
+ * @param secretName - The object store secret name (e.g. 's3secret').
+ * @param fileName - The file name / path within the secret (e.g. 'test-input-123.jsonl').
```
davidkna-sap: [req] Please also update this elsewhere 
```suggestion
 * @param fileName - The file name / path within the object store (e.g. 'test-input-123.jsonl').
```

tests/e2e-tests/src/batch-api.test.ts
```diff
@@ -0,0 +1,103 @@
...
+        { retries: 10, minTimeout: 3000 }
+      );
+
+      await deleteBatch(id);
+    } finally {
+      await deleteFile(secretName, inputFileName);
+    }
+  });
+
+  it('should create a batch job, wait for completion, download output, then delete it', async () => {
```
davidkna-sap: [req] The whole flow should not be handled in a single E2E test. The service reportedly needs up to 24h to process a request.

One way to handle this would be:
- One test that lists, downloads and deletes requests
- One test that creates request

Perhaps also check how Java handles E2E testing of this service.

davidkna-sap (changes requested):

packages/llm-batch/config.example.json
InjunPark-sap: I will be about to discuss with team in tomorrow daily

packages/llm-batch/README.md
davidkna-sap, line 1: Please add a `README.md`, you can use something similar to the RPT package.

packages/llm-batch/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "postgenerate": "pnpm update-imports && pnpm make-llm-batch-experimental && pnpm lint:fix",
+    "make-llm-batch-experimental": "node --no-warnings --loader ts-node/esm ../../scripts/postgenerate-llm-batch.ts ./src/client/batch-service/batches-api.ts",
+    "update-imports": "node --no-warnings --loader ts-node/esm ../../scripts/update-imports.ts ./src/client/batch-service"
+  },
+  "dependencies": {
+    "@sap-ai-sdk/core": "workspace:^"
+  },
+  "devDependencies": {
+    "ts-node": "^10.9.2"
+  }
```
davidkna-sap: The rest of the project is now using `tsx`, please prefer using that.

scripts/postgenerate-llm-batch.ts
```diff
@@ -0,0 +1,19 @@
...
+
+if (!filePath) {
+  console.error('Please provide the API file path as an argument.');
+  process.exit(1);
+}
+
+transformFile(resolve(filePath), file =>
+  file.replace(
+    "* This API is part of the 'batch-service' service.",
+    "* This API is part of the 'batch-service' service.\n * @experimental This API is experimental and may change at any time without prior notice."
```
davidkna-sap: [pp] I would prefer comments use the real service name
```suggestion
    "* This API is part of the 'llm-batch' service.\n * @experimental This API is experimental and may change at any time without prior notice."
```

```suggestion
    "* This API is part of the 'LLM Batch' service.\n * @experimental This API is experimental and may change at any time without prior notice."
```

packages/foundation-models/src/azure-openai/azure-openai-batch-input.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  /**
+   * Serializes the batch input to a JSONL Blob for upload via FileApi.
+   * @returns A Blob containing one JSON object per line.
+   */
+  toBlob(): Blob {
+    return new Blob([this.lines.map(line => JSON.stringify(line)).join('\n')]);
+  }
+}
```
davidkna-sap: [pp]

```suggestion
interface BatchInputLine {
  custom_id: string;
  method: 'POST';
  url: '/v1/chat/completions';
  body: AzureOpenAiCreateChatCompletionRequest;
}

function createAzureOpenAiBatchInput(
  requests: AzureOpenAiCreateChatCompletionRequest[]
): Blob {
  const lines = requests.map((body, i) =>
    JSON.stringify({
      custom_id: `request-${i + 1}`,
      method: 'POST',
      url: '/v1/chat/completions',
      body
    })
  );
  return new Blob([lines.join('\n')]);
}
```

packages/foundation-models/src/azure-openai/azure-openai-batch-output.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * Returns only failed output lines (error is not null).
+   * @returns Array of output lines where error is not null.
+   */
+  getFailed(): AzureOpenAiBatchOutputLine[] {
+    return this.lines.filter(line => line.error !== null);
+  }
+}
```
davidkna-sap, line 52:
```ts
function parseBatchOutput(): BatchOutputLine[] {
    const text = await blob.text();
    return  text
      .split('\n')
      .filter(line => line.trim().length > 0)
      .map(line => JSON.parse(line) as BatchOutputLine);
}
````

File: packages/llm-batch/config.example.json
Comment (davidkna-sap) [inline packages/llm-batch/config.example.json:1]: I would prefer removing this as it is not referenced anywhere afaict.

File: tests/e2e-tests/src/llm-batch.test.ts
Comment (davidkna-sap) [inline tests/e2e-tests/src/llm-batch.test.ts:1]: [req] Please ensure the e2e tests run successfully, in local testing these fail:

```
    ✕ should create a batch job, cancel it, then delete it (319 ms)
    ✕ should create a batch job and submit it for processing (126 ms)
    ✕ should download and delete a completed batch job if one exists (1043 ms)
```

File: tests/e2e-tests/src/batch-api.test.ts [hunk 19]
```diff
@@ -0,0 +1,107 @@
...
+    let output;
+    try {
+      output = await downloadBatchOutput(secretName, outputFolder, id);
+    } catch {
+      // Output file may have been deleted by a previous test run or created
+      // with a different output path — skip the download assertions.
+      return;
+    }
+    expect(output.length).toBeGreaterThan(0);
+    expect(output.filter(line => line.error === null).length).toBeGreaterThan(0);
```
davidkna-sap: This will still fail if `downloadBatchOutput()` fails due to the test condition.

packages/llm-batch/.gitignore
davidkna-sap, line 1: [pp] Remove file

tests/e2e-tests/src/batch-api.test.ts
davidkna-sap, line 1: [req] Remove the old file.

tests/e2e-tests/src/llm-batch.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    const inputFileName = `test-input-${Date.now()}.jsonl`;
+    const inputUri = await uploadBatchInput(secretName, inputFileName);
+
+    try {
+      const response = await createBatch(
+        inputUri,
+        `ai://${secretName}/${outputFolder}`
+      );
+      expect(response.id).toBeDefined();
+      const id = response.id!;
```
davidkna-sap, line 52: [req] All recent requests to the service have failed. Please investigate:

```json
  "status": {
    "current_status": "FAILED",
    "target_status": "COMPLETED",
    "updated_at": "2026-05-26T08:01:35.407000Z",
    "message": "Failed to prepare input file after multiple attempts. Error: "
  }
```

sample-code/src/llm-batch.ts
davidkna-sap: [req] I have confirmed that `//` is not needed anymore, and `//` does not work anymore. Why have you marked this as resolved?

Please update and handle everywhere.

davidkna-sap (changes requested):

InjunPark-sap:
### E2E Test Fixes

**Remove `//` from FileApi paths**
- `FileApi.fileUpload`, `fileDelete`, `fileDownload` all used `${secretName}//${path}` — the double slash is no longer supported by AI Core. Changed to `${secretName}/${path}` in all three call sites.

**Fix HTTP 415 on file upload**
- File upload was rejected with `Unsupported Media Type: application/octet-stream`. Setting `Blob.type` alone is not sufficient — Axios ignores it when serializing in Node.js. Fixed by passing `{ headers: { 'content-type': 'text/csv' } }` explicitly via `execute()`.

sample-code/src/llm-batch.ts
```diff
@@ -101,8 +101,8 @@ export async function uploadBatchInput(
   secretName: string,
   fileName: string
 ): Promise<string> {
-  // text/csv type is required: the AI Core file API rejects application/octet-stream.
-  // Axios uses the Blob's type property as Content-Type, so we set it explicitly.
+  // text/csv is required: the AI Core file API rejects application/octet-stream.
+  // Axios overrides Content-Type based on Blob.type, so the type must be set on the Blob.
```
davidkna-sap: [q] Please confirm this, iirc this is set via this project.

davidkna-sap:
> E2E results after fix: 6/8 passing. The 2 remaining download failures are due to stale COMPLETED batches (created in April with output.uri: ai://s3secret/) whose output files no longer exist at the output/ prefix path. These will resolve once new batches complete with the correct output URI.

Please delete the batch jobs in question.

tests/e2e-tests/src/llm-batch.test.ts
```diff
@@ -0,0 +1,103 @@
...
+describe('batch api', () => {
+  it('should list existing batch jobs', async () => {
+    const result = await listBatches();
+
+    expect(result.count).toBeGreaterThanOrEqual(0);
+    expect(Array.isArray(result.resources)).toBe(true);
+  });
+
+  it('should create a batch job, cancel it, then delete it', async () => {
+    const inputFileName = `test-input-${Date.now()}.jsonl`;
```
davidkna-sap:
```suggestion
    const inputFileName = `test-input-to-cancel-${Date.now()}.jsonl`;
```

Or remove uploading a batch input file as part of this test, as it is not strictly needed.

tests/e2e-tests/src/llm-batch.test.ts
```diff
@@ -0,0 +1,103 @@
...
+      );
+
+      await deleteBatch(id);
+    } finally {
+      await deleteFile(secretName, inputFileName);
+    }
+  });
+
+  it('should create a batch job and submit it for processing', async () => {
+    const inputFileName = `test-input-${Date.now()}.jsonl`;
```
davidkna-sap:
```suggestion
    const inputFileName = `test-input-complete-${Date.now()}.jsonl`;
```

tests/e2e-tests/src/llm-batch.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    try {
+      const response = await createBatch(
+        inputUri,
+        `ai://${secretName}/${outputFolder}`
+      );
+      expect(response.id).toBeDefined();
+    } finally {
+      await deleteFile(secretName, inputFileName);
+    }
```
davidkna-sap: [req] Only delete in completed handler.

sample-code/src/llm-batch.ts
```diff
@@ -0,0 +1,166 @@
...
+          model: 'gpt-4.1',
+          messages: [
+            { role: 'user', content: 'Explain neural networks in simple terms' }
+          ],
+          max_tokens: 150
+        }
+      ])
+    ],
+    { type: 'text/csv' }
+  );
```
davidkna-sap: [pp] Let's set this as part of `createBatchInput()` instead, it does appear to be the proper place to do that.
```suggestion
  const blob = createBatchInput([
        {
          model: 'gpt-4.1',
          messages: [{ role: 'user', content: 'What is machine learning?' }],
          max_tokens: 150
        },
        {
          model: 'gpt-4.1',
          messages: [
            { role: 'user', content: 'Explain neural networks in simple terms' }
          ],
          max_tokens: 150
        }
      ])
```

sample-code/src/llm-batch.ts
```diff
@@ -0,0 +1,166 @@
...
+  await FileApi.fileUpload(
+    `${secretName}/${fileName}`,
+    blob,
+    { overwrite: true },
+    { 'AI-Resource-Group': defaultHeaders['AI-Resource-Group'] }
+  ).execute();
+  return `ai://${secretName}/${fileName}`;
```
davidkna-sap: [pp]
```suggestion
  const result = await FileApi.fileUpload(
    `${secretName}/${fileName}`,
    blob,
    { overwrite: true },
    { 'AI-Resource-Group': defaultHeaders['AI-Resource-Group'] }
  ).execute();
  return result.url;
```

packages/foundation-models/src/azure-openai/azure-openai-batch-input.ts
```diff
@@ -0,0 +1,22 @@
...
+): Blob {
+  const lines = requests.map((body, i) =>
+    JSON.stringify({
+      custom_id: `request-${i + 1}`,
+      method: 'POST',
+      url: '/v1/chat/completions',
+      body
+    })
+  );
+  return new Blob([lines.join('\n')]);
```
davidkna-sap: [pp]
```suggestion
  return new Blob([lines.join('\n')], { 'type': 'text/csv' });
```

davidkna-sap (changes requested):
Please also (to keep up with other changes, in the llm batch package)L

- Delete `internal.d.ts`
- Delete `internal.js`
- Add `internal.ts` (like what the other packages have)
- Add this to package.json:

```
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "default": "./dist/index.js"
    },
    "./internal.js": {
      "types": "./dist/internal.d.ts",
      "default": "./dist/internal.js"
    }
  },
```

davidkna-sap:
Before I forget: Please also add two changesets covering the addition of the llm batch package, and a second one for the openai package that mentions the new helpers for llm-batch. 

If you haven't already, you can create a new changeset via `pnpm changeset`. Feel free to reach out to me or read the claude skill under my user on github.tools.sap if you have issues.

InjunPark-sap:
> Before I forget: Please also add two changesets covering the addition of the llm batch package, and a second one for the openai package that mentions the new helpers for llm-batch.
> 
> If you haven't already, you can create a new changeset via `pnpm changeset`. Feel free to reach out to me or read the claude skill under my user on github.tools.sap if you have issues.

---

Hey @davidkna-sap , here's what I'm planning for the two changesets. Let me know if this looks good or if you'd tweak anything!:

1. `@sap-ai-sdk/llm-batch`: minor — new LLM Batch API package
2. `@sap-ai-sdk/foundation-models`: minor — `createBatchInput()` and `parseBatchOutput()` helpers

davidkna-sap:
@InjunPark-sap Please add the proposed changesets, it will be easier to review them when they land here.

InjunPark-sap:
Added both changesets as planned:

`clever-foxes-jump.md` → `@sap-ai-sdk/llm-batch: minor`
`rapid-moons-sing.md`→ `@sap-ai-sdk/foundation-models: minor`

Also fixed a flaky e2e test where test 4 would 404 on output download if a previous run had already cleaned up the file.

tests/e2e-tests/src/llm-batch.test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    const id = completedBatch.id;
+    const details = await getBatchById(id);
+    const inputFilePath = details.input?.uri?.replace(
+      `ai://${secretName}/`,
+      ''
+    );
+
+    try {
+      const output = await downloadBatchOutput(secretName, outputFolder, id);
```
davidkna-sap: [nth] To replace `try {} catch {} finally {}`

```suggestion
      const output = await downloadBatchOutput(secretName, outputFolder, id).catch(retrowUnlessNotFound);
```

Requires this:
```ts
function retrowUnlessNotFound(e: any) {
  if (e.cause?.response?.status === 404) {
    return;
  }
  throw e;
}
```

tests/e2e-tests/src/llm-batch.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    if (!completedBatch?.id) {
+      return;
+    }
+
+    const id = completedBatch.id;
+    const details = await getBatchById(id);
+    const inputFilePath = details.input?.uri?.replace(
+      `ai://${secretName}/`,
+      ''
+    );
```
davidkna-sap: [req] The secret name still needs to be part of the path.
```suggestion
    const inputFilePath = details.input?.uri?.replace(/^ai:\/\//, '');
```

tests/e2e-tests/src/llm-batch.test.ts
```diff
@@ -0,0 +1,116 @@
...
+        0
+      );
+      await deleteFile(secretName, `${outputFolder}${id}/output.jsonl`);
+      if (inputFilePath) {
+        await deleteFile(secretName, inputFilePath);
+      }
+    } catch (e: any) {
+      // A COMPLETED batch may have its output file already deleted from a previous test run.
+      // In that case, skip assertions and proceed to cleanup.
+      if (!e.message?.includes('404')) {
```
davidkna-sap: [pp]
```suggestion
      if (e.cause?.response?.status !== 404) {
```

tests/e2e-tests/src/llm-batch.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      );
+
+      await deleteBatch(id);
+    } finally {
+      await deleteFile(secretName, inputFileName);
+    }
+  });
+
+  it('should create a batch job and submit it for processing', async () => {
+    const inputFileName = `test-input-complete-${Date.now()}.jsonl`;
```
davidkna-sap: [nth] For nicer file names (If applying this change please also replace the other `new Date()` call:
```suggestion
    const inputFileName = `test-input-complete-${runId()}.jsonl`;
```

Where `runId()` is:
```ts
function runId() {
  const timestamp = new Date().toISOString().replace(/[:T]/g, '-');
  const prRun = process.env.GITHUB_RUN_ID || 'local';

  return `${timestamp}-${prRun}`;
}
```

.changeset/clever-foxes-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/llm-batch': minor
+---
+
+[feat] Add new `@sap-ai-sdk/llm-batch` package for asynchronous batch processing of LLM requests via SAP AI Core Batch API.
```
davidkna-sap: [pp]
```suggestion
[feat] Add new `@sap-ai-sdk/llm-batch` package for asynchronous batch processing of LLM requests via SAP AI Core LLM Batch Service.
```

davidkna-sap (dismissed):

packages/foundation-models/src/azure-openai/azure-openai-batch-input.ts
```diff
@@ -0,0 +1,22 @@
...
+): Blob {
+  const lines = requests.map((body, i) =>
+    JSON.stringify({
+      custom_id: `request-${i + 1}`,
+      method: 'POST',
+      url: '/v1/chat/completions',
+      body
+    })
+  );
+  return new Blob([lines.join('\n')], { type: 'text/csv' });
```
KavithaSiva, line 22: [pp] Please also add the comment: ` // The AI Core file API requires content-type 'text/csv' for batch input files.` here.
This is strange though, have you reached out to the AI Core colleagues on why this is and why `application/jsonl` cannot be used?

packages/foundation-models/src/azure-openai/azure-openai-batch-output.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Parses a batch output Blob (JSONL format) into typed output lines.
+ * @param blob - The Blob returned from FileApi.fileDownload().
+ * @returns A Promise resolving to an array of parsed output lines.
+ * @experimental This API is experimental and may change at any time without prior notice.
+ */
+export async function parseBatchOutput(blob: Blob): Promise<BatchOutputLine[]> {
+  const text = await blob.text();
+  return text
+    .split('\n')
+    .filter(line => line.trim().length > 0)
```
KavithaSiva: [pp]
```suggestion
    .filter(line => line.trim().length)
```

packages/llm-batch/internal.d.ts
```diff
@@ -0,0 +1,2 @@
+export * from './dist/internal.js';
```
KavithaSiva: [req] For other packages we also deleted these files and updated `package.json`. Refer to David's comment [here](https://github.com/SAP/ai-sdk-js/pull/1759#pullrequestreview-4381826130)

KavithaSiva (changes requested):
Most are minor comments. Please also remove `internal.d.ts` and `internal.js` and add `internal.ts` like other packages.

packages/foundation-models/src/azure-openai/azure-openai-batch-input.ts
davidkna-sap: For the record the endpoint supports either `application/json` or `text/csv` (no `application/x-ndjson`). `application/json` does input validation which leads to upload errors for jsonl files with more than one line, so `text/csv` is the only choice.

InjunPark-sap:
- Replaced `internal.d.ts / internal.js` shims with explicit exports in `package.json`.
- Added text/csv comment to createBatchInput().

KavithaSiva (approved):
LGTM, thanks for all the effort @InjunPark-sap @davidkna-sap !

PR #1767, opened 2026-04-21
fix: Disallow providing both orchestration config ref and config object in orchestration client

ZhongpinWang:
## Context

Previously, it is possible to provide orchestration config ref together with other orchestration config modules, which is wrong. I used Xor and adapted the config property check to make the properties mutual exclusive.\ 

```ts
new OrchestrationClient({
    id: 'test-config-id',
    promptTemplating: {
      model: { name: 'gpt-5-mini' },
      prompt: { template: [{ role: 'user', content: 'Hello!' }] }
    }
  }).chatCompletion({
    placeholderValues: { name: 'Alice' },
    messagesHistory: [
      { role: 'system', content: 'You are a helpful assistant.' }
    ]
  })
```

For reviewer, please double check the behaviour and feel free to take the PR over.

### DIFF
--- a/packages/orchestration/src/orchestration-client.ts
+++ b/packages/orchestration/src/orchestration-client.ts
@@ -16,6 +16,7 @@ import {
   isOrchestrationModuleConfigList,
   assertIsOrchestrationModuleConfigList
 } from './orchestration-types.js';
+import type { Xor } from '@sap-cloud-sdk/util';
 import type { TemplatingChatMessage } from './client/api/schema/index.js';
 import type {
   HttpResponse,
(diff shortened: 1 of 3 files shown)


packages/orchestration/src/orchestration-client.ts
```diff
@@ -59,17 +60,21 @@ export class OrchestrationClient {
    */
   constructor(
     private config:
-      | OrchestrationModuleConfig
-      | OrchestrationModuleConfigList
       | string
-      | OrchestrationConfigRef,
+      | Xor<
+          OrchestrationConfigRef,
+          OrchestrationModuleConfig | OrchestrationModuleConfigList
```
davidkna-sap, line 66: [nth] `OrchestrationModuleConfigList`'s items should perhaps also `Omit` `OrchestrationConfigRef`'s keys

packages/orchestration/src/orchestration-client.ts
```diff
@@ -59,17 +60,21 @@ export class OrchestrationClient {
...
     private destination?: HttpDestinationOrFetchOptions
   ) {
     if (typeof config === 'string') {
       this.validateJsonConfig(config);
-    } else if (Array.isArray(config)) {
+    } else if (isOrchestrationModuleConfigList(config)) {
       this.config = this.parseModuleConfigList(config);
+    } else if (Array.isArray(config)) {
+      // Array that failed isOrchestrationModuleConfigList validation - let assertion throw with proper error
+      assertIsOrchestrationModuleConfigList(config);
```
davidkna-sap: [pp]
```suggestion
    } else if (Array.isArray(config)) {
      assertIsOrchestrationModuleConfigList(config);
      this.config = this.parseModuleConfigList(config);
```

davidkna-sap (commented):
Mostly LGTM, but I have made two suggestions

packages/orchestration/src/orchestration-client.ts
KavithaSiva: `OrchestrationModuleConfig` is a closed interface , so TypeScript's excess property checking already rejects any OrchestrationConfigRef key (id, scenario, name, version) inside a list item.

KavithaSiva (approved):
LGTM

PR #1769, opened 2026-04-22
chore: Re-enable pnpm trust-policy

davidkna-sap:
## Context

> Allows ignoring the trust policy check for packages published more than the specified number of minutes ago. This is useful when enabling strict trust policies, as it allows older versions of packages (which may lack a process for publishing with signatures or provenance) to be installed without manual exclusion, assuming they are safe due to their age.

Alternatively we could also use `trustPolicyExclude` but 3 months should give sufficient headroom and would not require maintaining a list of dependencies.

https://pnpm.io/settings#trustpolicyignoreafter


marikaner (approved):
lgtm

PR #1774, opened 2026-04-27
chore: Configure `save-exact` for smoke test package

davidkna-sap:
## Context

Add `save-exact` `npmrc` to smoke test package to ensure future updates stay locked to a single version.

### DIFF
--- a/tests/smoke-tests/.npmrc
+++ b/tests/smoke-tests/.npmrc
@@ -1,0 +1,2 @@
+save-exact=true
+save-prefix=''


tests/smoke-tests/.npmrc
```diff
@@ -0,0 +1,2 @@
+save-exact=true
```
davidkna-sap, line 1: pnpm 11 (released yesterday) pulls this into `pnpm-workspace.yaml`, it might be worth to switch to this, as I have noticed that this setting does not catch for some pnpm commands right now.

marikaner (approved):
let's adjust to pnpm 11 in a separate pr and move it there.

PR #1791, opened 2026-04-28
chore: Add suppression for OWASP dependency scan check

KavithaSiva:
## Context

For some reason, the security scan fails with:
```
One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '7.0': 
hono:=4.12.10 (pkg:npm/hono@4.12.10, cpe:2.3:a:hono:hono:4.12.10:*:*:*:*:*:*:*): CVE-2026-39408(5.9)
```
even though the hono was updated to later versions. Hence, adding a suppression.


davidkna-sap:
I've also started a scan on main, the cron check run was 7h ago, but Hono was updated 5h ago: https://github.com/SAP/ai-sdk-js/actions/runs/25039711568

PR #1793, opened 2026-04-28
chore(ci): Fix dependabot lockfile update action

davidkna-sap:
## Context

Uses new Cloud SDK setup feature.

Context: https://github.com/SAP/cloud-sdk-js/pull/6529#discussion_r3149520359

Should fix https://github.com/SAP/ai-sdk-js/actions/runs/24996752082/job/73195624912?pr=1777

### DIFF
--- a/.github/workflows/dependabot-pnpm-lock.yml
+++ b/.github/workflows/dependabot-pnpm-lock.yml
@@ -27,10 +27,15 @@ jobs:
         uses: sap/cloud-sdk-js/.github/actions/setup@main
         with:
           ref: ${{ github.head_ref }}
-          pnpm-install-args: --lockfile-only
+          skip-install: 'true'
           registry-token: ${{ secrets.NPM_TOKEN_ARTIFACTORY }}
           persist-credentials: 'true'
           token: ${{ steps.app-token.outputs.token }}
+
+      - name: Regenerate lockfile
+        run: pnpm install --lockfile-only
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN_ARTIFACTORY }}
 
       - name: Commit and push if changed
         env:
--- a/.github/workflows/dependabot-pnpm-lock.yml
+++ b/.github/workflows/dependabot-pnpm-lock.yml
@@ -27,15 +27,11 @@ jobs:
         uses: sap/cloud-sdk-js/.github/actions/setup@main
         with:
           ref: ${{ github.head_ref }}
-          skip-install: 'true'
+          pnpm-install-args: --lockfile-only
+          skip-build: 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-pnpm-lock.yml
```diff
@@ -27,11 +27,16 @@ jobs:
...
+          skip-install: 'true'
           registry-token: ${{ secrets.NPM_TOKEN_ARTIFACTORY }}
           persist-credentials: 'true'
           token: ${{ steps.app-token.outputs.token }}
 
+      - name: Regenerate lockfile
+        run: pnpm install --lockfile-only
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN_ARTIFACTORY }}
+
```
ZhongpinWang: I am probably a bit late to the party. But I guess this whole part is added only because there is no way to turn off the `--ignore-scripts` in cloud sdk setup action right? Can't we set this to default but provide a way to turn it off? Or just allow providing this as part of the `pnpm-install-args`?

ZhongpinWang: We are bypassing the setup action and do the install by ourselves. This feels similarly bad to the violation of Liskov Substitution Principle in OO programming.

ZhongpinWang (approved):
I can approve this. But please consider the comments.

.github/workflows/dependabot-pnpm-lock.yml
davidkna-sap: @ZhongpinWang The issue is that if `--lockfile-only` is used (as forwarded via the `pnpm-install-args` option) dependencies are not fetched at all and thus the `pnpm rebuild` after `pnpm install` in the action fails.

I have split `install` from `rebuild` to avoid having the artifactory key in the environment during builds. 

In my initial PR, I set the action to automatically avoid `rebuild` if `--lockfile-only` is used. Would you feel better about a `skip-rebuild` option instead of skipping all install steps?

ZhongpinWang: I find it currently a bit confusing. `skip-install` means
- skip the install which uses `--ignore-scripts`
- skip the rebuild

Can we avoid having one toggle with many hidden side effects?

I get the point that using `--lockfile-only` will cause issues for the upcoming `pnpm rebuild`, then we skip rebuild right? Also here in your workaround install, there is no `--ignore-scripts`.

davidkna-sap: @ZhongpinWang Without `--ignore-scripts` `pnpm install` is equivalent to `pnpm install --ignore-scripts && pnpm rebuild`, but I will open a PR to add a `skip-build` toggle.

PR #1806, opened 2026-04-30
chore(ci): Improve sonar coverage configuration

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#563.


sonar-project.properties
```diff
@@ -2,10 +2,37 @@ sonar.projectKey=SAP_ai-sdk-js
...
-sonar.typescript.lcov.reportPaths=packages/ai-api/coverage/lcov.info,packages/core/coverage/lcov.info,packages/document-grounding/coverage/lcov.info,packages/foundation-models/coverage/lcov.info,packages/langchain/coverage/lcov.info,packages/orchestration/coverage/lcov.info,packages/prompt-registry/coverage/lcov.info,packages/rpt/coverage/lcov.info
+sonar.sources=.
+sonar.tests=\
+  tests,\
+  test-util
+sonar.exclusions=\
+  **/node_modules/**,\
+  **/dist/**,\
+  **/dist-cjs/**\,
+  **/sample-code/resources/**
```
davidkna-sap:
```suggestion
  sample-code/resources/**
```

.github/workflows/build.yml
```diff
@@ -21,6 +21,9 @@ jobs:
           node-version: ${{ matrix.node-version }}
           registry-token: ${{ secrets.NPM_TOKEN_ARTIFACTORY }}
       - run: pnpm test:unit
+      - name: Fetch main for SonarQube blame
+        if: ${{ env.SONAR_TOKEN != '' && strategy.job-index == 0 }}
```
marikaner: [req] `env.SONAR_TOKEN` is never set, because you don't set the env for this step.

.github/workflows/build.yml
```diff
@@ -21,6 +21,9 @@ jobs:
...
+        run: git fetch --unshallow origin main
       - name: SonarQube Scan
```
marikaner, line 30: [req/q] If I am not mistaken sonarcube is a static code check. I would rather put it into the checks job below, then you don't need to check for `strategy.job-index == 0`.

marikaner: [pp] Consider also adding a comment to ensure it is clear that this check does not run on PRs from forks. I guess that is why we check whether the secret is there.

sonar-project.properties
```diff
@@ -2,10 +2,37 @@ sonar.projectKey=SAP_ai-sdk-js
...
+
+sonar.typescript.lcov.reportPaths=\
+  packages/ai-api/coverage/lcov.info,\
+  packages/core/coverage/lcov.info,\
+  packages/document-grounding/coverage/lcov.info,\
+  packages/foundation-models/coverage/lcov.info,\
+  packages/langchain/coverage/lcov.info,\
+  packages/orchestration/coverage/lcov.info,\
+  packages/prompt-registry/coverage/lcov.info,\
+  packages/rpt/coverage/lcov.info
```
marikaner: [q] Does this not work?
```suggestion
sonar.javascript.lcov.reportPaths=packages/**/coverage/lcov.info
```

marikaner (changes requested):

sonar-project.properties
davidkna-sap: Per the documentation, it should not work, but I will try to apply it anyway.

.github/workflows/build.yml
davidkna-sap: @marikaner It needs to run there to get access to the coverage reports.

sonar-project.properties
marikaner: Ok, just because the config above is using wildcards. ¯\\_(ツ)_/¯

.github/workflows/build.yml
marikaner: @davidkna-sap That's a pity. Consider adding a comment for this too. 

I am wondering if it is worth creating a separate job that needs the test job and then accesses the coverage reports as artifact. I don't like the `strategy.job-index == 0` check.
But I believe with one or two comments as suggested before, we can spare ourselves from over-engineering 😊

marikaner (approved):

marikaner:
@davidkna-sap I did not notice the auto-merge was on. Did you confirm the config worked?

davidkna-sap:
@marikaner The logs show that it picks up the lcov files correctly.

PR #1728, opened 2026-04-09
chore: Use new pnpm-based license checker

davidkna-sap:
# Update License Check Action to Use `main` Branch

### Chore

ð§ Updated the GitHub Actions workflow to use the `main` branch reference for the license check action instead of the `license-checker` branch, aligning it with the latest pnpm-based license checker.

### Changes

* `.github/workflows/build.yml`: Updated the `check-license` action reference from `@license-checker` to `@main` to use the new pnpm-based license checker implementation.

- [ ] 🔄 Regenerate and Update Summary

<details>
<summary>PR Bot Information</summary>

**Version:** `1.20.23`

- Event Trigger: `issue_comment.edited`
- Correlation ID: `c98a3908-7da2-4a47-b1ad-b217e0a0ee4b`
</details>

PR #1822, opened 2026-05-05
chore: Harden Github Actions configuration

davidkna-sap:
## Context

Part of SAP/ai-sdk-js-backlog#562.
Closes #1728


.github/workflows/build.yml
```diff
@@ -59,13 +59,13 @@ jobs:
       - name: Check dependencies
         run: pnpm check:deps
       - name: License Check
-        uses: sap/cloud-sdk-js/.github/actions/check-license@license-checker
+        uses: sap/cloud-sdk-js/.github/actions/check-license@main
```
davidkna-sap, line 62: I decided to also pull in #1728 to restrict action references to the main branch.

.github/workflows/build.yml
```diff
@@ -59,13 +59,13 @@ jobs:
...
       - name: API Doc Check
         run: pnpm doc
 
   dependabot:
     runs-on: ubuntu-latest
-    if: ${{ (github.actor == 'dependabot[bot]' || startsWith(github.event.pull_request.title, 'chore(deps')) && github.event_name == 'pull_request' }}
+    if: ${{ github.event_name == 'pull_request' && github.event.pull_request.user.login == 'dependabot[bot]' && github.repository == github.event.pull_request.head.repo.full_name }}
```
davidkna-sap, line 68: ` github.repository == github.event.pull_request.head.repo.full_name` is apparently a bit more robust than `is_fork`.

marikaner: [q] Just out of curiosity: What does "more robust" mean? Can `is_fork` be wrong?

davidkna-sap: I assumed static analyser default suggestions should have some reasoning behind their default recommendations and Claude claims the parent property of the `fork` property is `null` is some cases.

.github/workflows/auto-fix-lint.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           BOT_EMAIL: ${{ vars.SAP_AI_SDK_BOT_EMAIL }}
           BOT_NAME: ${{ vars.SAP_AI_SDK_BOT_NAME }}
         run: |
-          diff=`git diff`
+          diff=$(git diff)
```
davidkna-sap: I don't feel like this shellcheck warning is very useful. I am considering reverting this part of the changes.

.github/workflows/bump.yml
```diff
@@ -88,24 +89,27 @@ jobs:
...
           # Extract major version from full version (e.g., "2.8.0" -> "v2")
-          FULL_VERSION="${{ needs.bump.outputs.version }}"
+          FULL_VERSION="${NEEDS_BUMP_OUTPUTS_VERSION}"
           MAJOR_VERSION=$(echo "$FULL_VERSION" | cut -d. -f1)
           VERSION="v${MAJOR_VERSION}"
 
           # Copy generated docs to versioned folder
-          rsync -avz --delete knowledge-base/api-reference/ "${{ env.DOCS_CHECKOUT_PATH }}/static/api/${VERSION}/"
+          rsync -avz --delete knowledge-base/api-reference/ "${DOCS_CHECKOUT_PATH}/static/api/${VERSION}/"
+        env:
```
marikaner: [req] Please move this above run.

.github/zizmor.yml
```diff
@@ -0,0 +1,23 @@
+rules:
+  # already following SAP policies
+  dependabot-cooldown:
+    disable: true
```
marikaner, line 4: [q] Why disable it then? The check should pass, shouldn't it?

.github/workflows/gha-security-scan.yaml
```diff
@@ -0,0 +1,26 @@
...
+permissions: {}
+
+jobs:
+  zizmor:
+    name: Run zizmor 🌈
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write # Required for upload-sarif (used by zizmor-action) to upload SARIF files.
+      contents: read # Only needed for private repos. Needed to clone the repo.
+      actions: read # Only needed for private repos. Needed for upload-sarif to read workflow run info.
```
marikaner: [pp] I think these (assumingely AI-generated) comments are not necessary. 
```suggestion
      contents: read
      actions: read
```

marikaner (approved):
I left a few comments, but no major blocker, even though there is a [req].

.github/zizmor.yml
davidkna-sap: It wants 7 days, we use 4 days.

.github/workflows/gha-security-scan.yaml
davidkna-sap: This is a copy-paste from upstream, they have a disabled-by-default rule to require comments for any extra permissions.

.github/workflows/bump.yml
davidkna-sap: Already handled by merging `main`.

.github/zizmor.yml
marikaner: Then let's configure it for 4. That way if we add another config we won't forget to configure the cooldown too.

https://docs.zizmor.sh/audits/#configuration

.github/workflows/gha-security-scan.yaml
marikaner: But the [docs](https://docs.zizmor.sh/audits/#undocumented-permissions) state that for contents: read it is not checked as it is trivial.

Also, this is a "pedantic" rule by their definition. I agree, that it can make sense, however I could imagine that the comments will be outdated at some point, while the permission is still needed for something new. That is why I question the value a bit.

TLDR; Fine for me, but should not be needed for contents: read.

.github/workflows/auto-fix-lint.yml
marikaner: I thought so too 😊

PR #1823, opened 2026-05-05
feat(orchestration): Allow array for encoding_format in EmbeddingModelParams

InjunPark-sap:
## What changed

- `EmbeddingModelParams.encoding_format`: `EncodingFormat` → `EncodingFormat | EncodingFormat[]`
- Added type tests for single value, array, and invalid format rejection

## Why

The generated spec (`embeddings-model-params.ts`) already allows `EncodingFormat | EncodingFormat[]`, but the SDK public type was more restrictive.

Cohere Embed v4 accepts an array of `embedding_types` and returns `EmbeddingMultiFormat` — this change unblocks that use case.

### DIFF
--- a/packages/orchestration/src/orchestration-types.ts
+++ b/packages/orchestration/src/orchestration-types.ts
@@ -816,7 +816,7 @@ export type EmbeddingModelParams = Omit<
   OriginalEmbeddingsModelParams,
   'encoding_format'
 > & {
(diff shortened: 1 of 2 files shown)


davidkna-sap (commented):
I think we were discussing if this needs additional convenience.
In general, you probably want to revert all changes from #1740 as part of this PR.

PR #1836, opened 2026-05-07
chore: Add landscape model availability check to update models skill

emincihangeri:
## Context

Closes SAP/ai-sdk-js-backlog#566.


scripts/sync-model-types.ts
```diff
@@ -305,6 +306,50 @@ async function syncModelTypes(): Promise<void> {
...
+    return;
+  }
+
+  const syncedModels = new Set(
+    Object.values(typeToActiveModels).flatMap(s => [...s])
+  );
+
+  let landscapeModels: Set<string>;
+  try {
+    const result = await ScenarioApi.scenarioQueryModels('foundation-models', {
```
marikaner: [req] Can you find a more speaking name than result?

scripts/sync-model-types.ts
```diff
@@ -305,6 +306,50 @@ async function syncModelTypes(): Promise<void> {
...
+      'AI-Resource-Group': 'default'
+    }).execute();
+    landscapeModels = new Set(result.resources.map(r => r.model));
+  } catch (err) {
```
marikaner: [q] When will this throw? I assume only if `result`, `resources` or `model` are undefined, right? If this is the case, please check the variables before using them, instead of throwing an error to differentiate the cases.

scripts/sync-model-types.ts
```diff
@@ -305,6 +306,50 @@ async function syncModelTypes(): Promise<void> {
...
+    console.error('\n⚠ Landscape check failed:', err);
```
marikaner: [q] I believe nothing was "checked", only data was retrieved, right? If you specify more clearly at which step something unexpected happened, it will be easier to debug.

scripts/sync-model-types.ts
```diff
@@ -305,6 +306,50 @@ async function syncModelTypes(): Promise<void> {
...
+    return;
+  }
+
+  const missing = [...syncedModels].filter(m => !landscapeModels.has(m));
+  if (missing.length) {
+    console.error(
+      `\n⚠ ${missing.length} model(s) not available in your landscape:`
+    );
+    for (const m of missing) {
+      console.error(`  ${m}`);
```
marikaner: [nth] I believe it would be nicer to put this into one message, rather than multiple separate ones.

scripts/sync-model-types.ts
```diff
@@ -305,6 +306,50 @@ async function syncModelTypes(): Promise<void> {
...
+    // .env not found, AICORE_SERVICE_KEY must be set externally
+  }
+
+  if (!process.env['AICORE_SERVICE_KEY']) {
+    console.error('\n⚠ Skipping landscape check — AICORE_SERVICE_KEY not set.');
+    return;
+  }
+
+  const syncedModels = new Set(
+    Object.values(typeToActiveModels).flatMap(s => [...s])
```
marikaner, line 358: [q] I don't fully get this part `.flatMap(s => [...s])`, more specifically `s => [...s]`. What does that do? Destructure a list and then make a list again, only to flatten it afterwards with `flatMap()`.
I might totally be missing something, but consider to just use `.flat()` instead.

marikaner (changes requested):

scripts/sync-model-types.ts
emincihangeri: I think we have an array of sets instead of nested arrays so flat might not work. This was like an ugly workaround for that. We can use other structures though.

scripts/sync-model-types.ts
emincihangeri: I tried to treat the three cases differently and now the catch block is only for modelList. (previously called as result)

scripts/sync-model-types.ts
```diff
@@ -305,6 +306,55 @@ async function syncModelTypes(): Promise<void> {
...
+      'AI-Resource-Group': 'default'
+    }).execute();
+
+    if (!modelList.resources) {
+      console.error('\n⚠ Landscape check skipped — unexpected response: missing resources.');
+      return;
+    }
+
+    landscapeModels = new Set(
+      modelList.resources.flatMap(r => (r.model ? [r.model] : []))
```
davidkna-sap: [pp] Perhaps also filter whether `allowedScenarios` includes the `orchestration` `executableId`/`scenarioId` or even better the equivalent of `if (!isOrchestration && !isAzureOpenAi && !isSapRpt)` to catch models we don't have enabled for orchestration.

[nth] It would also be interesting to consider `"deprecated"` and `"retirementDate"`,  but let's skip that for this PR.

scripts/sync-model-types.ts
davidkna-sap: [req] For clarity, please restrict the `try-catch` block to the `ScenarioApi.scenarioQueryModels()` call. This would also allow making `landscapeModels` `const`.

scripts/sync-model-types.ts
```diff
@@ -331,6 +332,64 @@ async function syncModelTypes(): Promise<void> {
...
+    typeToActiveModels['SapRptModel']?.has(model);
+
+  const landscapeModels = new Set<string>();
+  for (const r of modelList.resources) {
+    if (!r.model) continue;
+    const isOrchestrationEnabled = r.allowedScenarios?.some(s => s.scenarioId === 'orchestration');
+    if (isAzureOrRpt(r.model) || isOrchestrationEnabled) {
+      landscapeModels.add(r.model);
+    }
+  }
```
davidkna-sap: [pp]

```suggestion
  const landscapeModels = new Set(
      modelList.resources
        .filter(r => r.model)
        .filter(
          r =>
            r.executableId === 'azure-openai' ||
            r.model.startsWith('sap-rpt-') ||
            r.allowedScenarios?.some(s => s.scenarioId === 'orchestration')
        )
        .map(r => r.model)
    );
```

scripts/sync-model-types.ts
```diff
@@ -331,6 +332,64 @@ async function syncModelTypes(): Promise<void> {
...
+  const modelList = await ScenarioApi.scenarioQueryModels('foundation-models', {
+    'AI-Resource-Group': 'default'
+  }).execute().catch((err: unknown) => {
+    console.error('\n⚠ Landscape check skipped — API request failed:', err);
+    return null;
+  });
+
+  if (!modelList) return;
+
+  if (!modelList.resources) {
```
davidkna-sap:
```suggestion
  if (!modelList.resources?.length) {
```

davidkna-sap (approved):
LGTM

scripts/sync-model-types.ts
marikaner: makes sense 👍

scripts/sync-model-types.ts
```diff
@@ -331,6 +332,62 @@ async function syncModelTypes(): Promise<void> {
...
+}
+
+async function checkLandscapeAvailability(
+  typeToActiveModels: Record<string, Set<string>>
+): Promise<void> {
+  try {
+    process.loadEnvFile(resolve(import.meta.dirname, '../sample-code/.env'));
+  } catch {
+    // .env not found, AICORE_SERVICE_KEY must be set externally
+  }
```
marikaner: [req] Please check whether the file is there and only load it if it exists. That way you don't have to use try catch. (You maybe should still do that and log a message that .env exists and there was an error when trying to load the env file though).

scripts/sync-model-types.ts
```diff
@@ -331,6 +332,62 @@ async function syncModelTypes(): Promise<void> {
...
+  );
+
+  const modelList = await ScenarioApi.scenarioQueryModels('foundation-models', {
+    'AI-Resource-Group': 'default'
+  }).execute().catch((err: unknown) => {
+    console.error('\n⚠ Landscape check skipped — API request failed:', err);
+    return null;
+  });
+
+  if (!modelList) return;
```
marikaner: [req] This is unnecessary, see below/

scripts/sync-model-types.ts
```diff
@@ -331,6 +332,62 @@ async function syncModelTypes(): Promise<void> {
...
+
+  if (!modelList.resources?.length) {
```
marikaner:
```suggestion
  if (!modelList?.resources?.length) {
```

scripts/sync-model-types.ts
```diff
@@ -331,6 +332,62 @@ async function syncModelTypes(): Promise<void> {
...
+  }
+
+  const landscapeModels = new Set(
+      modelList.resources
+        .filter(r => r.model)
+        .filter(
+          r =>
+            r.executableId === 'azure-openai' ||
+            r.model.startsWith('sap-rpt-') ||
+            r.allowedScenarios?.some(s => s.scenarioId === 'orchestration')
```
marikaner: [pp] You could merge those. 

```suggestion
        .filter(r => r.model)
        .filter(
          r =>
            r.model && (
            r.executableId === 'azure-openai' ||
            r.model.startsWith('sap-rpt-') ||
            r.allowedScenarios?.some(s => s.scenarioId === 'orchestration'))
```

marikaner: also [pp] it would be helpful if there was a comment or a variable that shortly describes what we are checking here, e.g. `isRelevantResource` (or hopefully something better)

emincihangeri: I added both a comment and and variable with a more descriptive name

marikaner (approved):

PR #1837, opened 2026-05-07
feat: Azure OpenAI client using official provider SDK 

KavithaSiva:
## Context

Closes SAP/ai-sdk-js-backlog#519.

<!-- Please provide a description summarizing your changes and their rationale -->
  
Introduces `@sap-ai-sdk/foundation-models-openai`, a thin integration layer that configures the official openai npm package to work with SAP AI Core.
                                                                                                                                                                                                                                                                                                                                                         
  The package exposes two entry points:                                                                                                                                                                                                                                                                                                                  
  - `createOpenAiConfig()` — returns an `AzureClientOptions` object for users who want to construct `AzureOpenAI` themselves                                                                                                                                                                                                                                   
  - `SapOpenAi.createClient` — returns a ready-to-use `SapOpenAi` instance with deployment resolution, authentication, and SAP-specific headers pre-configured                                                                                                                                                                                       
                                                                                                                                                                  
  `SapOpenAi` uses composition over inheritance: it wraps an internal `AzureOpenAI` instance and exposes only the three endpoints supported by SAP AI Core (chat, embeddings, responses). Unsupported endpoints simply do not exist on the class. The model parameter is removed from all `create()` and `parse()` signatures and injected automatically from `modelDeployment` at request time. 

Sample Usage:
```TS
//Chat completions
const client = await SapOpenAi.createClient({ modelDeployment: 'gpt-4.1' });
const response = await client.chat.completions.create({
    messages: [{ role: 'user', content: 'What is the capital of France?' }]
  });

//Responses
 const client = await SapOpenAi.createClient({ modelDeployment: 'gpt-5' });
 const response = await client.responses.create({
    instructions: 'You are a helpful assistant.',
    input: 'What is the capital of France?'
  });
```
  


<!-- Before raising this PR, please review the Definition of Done below and confirm that all applicable items are completed. 
## Definition of Done

- [ ] Documentation updated
- [ ] (Optional) Release notes updated -->


packages/openai/src/types.ts
```diff
@@ -0,0 +1,34 @@
...
+   */
+  modelDeployment: ModelDeployment<SapModelName>;
+  /**
+   * Optional custom destination. Defaults to the `aicore` service binding or `AICORE_SERVICE_KEY` env var.
+   */
+  destination?: HttpDestinationOrFetchOptions;
+  /**
+   * Azure OpenAI API version. Defaults to `'2024-10-21'`.
+   */
+  apiVersion?: string;
```
ZhongpinWang, line 33: [q/pp] I hope llm access can move chat completion to v1 api as well.. Funny thing, `{{apiurl}}/inference/deployments/{{deploymentid_gpt-5}}/v1/chat/completions` works without `api-version`... Can we bother them again and perhaps they will all internally switch to v1 api even using the without `v1`path? Thx.

packages/foundation-models-openai/src/completions.ts
```diff
@@ -0,0 +1,73 @@
...
+  parse<
+    Params extends WithoutModel<ChatCompletionCreateParamsNonStreaming>,
+    ParsedT = ExtractParsedContentFromParams<Params & { model: string }>
+  >(
+    body: Params,
+    options?: RequestOptions
+  ): APIPromise<ParsedChatCompletion<ParsedT>> {
+    return this.openAICompletions.parse(
+      {
+        model: this.defaultModel ?? '',
```
ZhongpinWang: [q/pp] I tried and for all ai core endpoints, model can be any string right? Then why do we still care about setting a `defaultModel` here? Maybe just set it empty? Or dummy and write a comment in this line? WDYT?

packages/openai/src/completions.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  parse<
+    Params extends WithoutModel<ChatCompletionCreateParamsNonStreaming>,
+    ParsedT = ExtractParsedContentFromParams<Params & { model: string }>
+  >(
+    body: Params,
+    options?: RequestOptions
+  ): APIPromise<ParsedChatCompletion<ParsedT>> {
+    return this.openAICompletions.parse(
```
ZhongpinWang: [q] Why don't we also support `parse()` in Response API?

packages/foundation-models-openai/src/types.ts
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Optional custom destination. Defaults to the `aicore` service binding or `AICORE_SERVICE_KEY` env var.
+   */
+  destination?: HttpDestinationOrFetchOptions;
+  /**
+   * Azure OpenAI API version. Defaults to `'2024-10-21'`.
+   */
+  apiVersion?: string;
+  /**
+   * Value appended to the `ai-client-type` header. Defaults to `'AI SDK JavaScript'`.
```
ZhongpinWang:
```suggestion
   * Additional client types appended to the `ai-client-type` header with the preconfigured `AI SDK JavaScript` value.
```
[pp] I feel `defaults to` is a bit miss-leading that user may think they can overwrite and remove the default client type?

packages/foundation-models-openai/src/client.ts
```diff
@@ -0,0 +1,72 @@
...
+      : 'modelName' in options.modelDeployment
+        ? options.modelDeployment.modelName
+        : undefined;
+
+  const azureOpenAI = new AzureOpenAI(config);
+
+  return new SapAzureOpenAI(
+    new SapChat(azureOpenAI, defaultModel),
+    new SapEmbeddings(azureOpenAI, defaultModel),
+    new SapResponses(azureOpenAI, defaultModel)
```
ZhongpinWang: [q/req] Maybe its on purpose? But why don't we just pass down `azureOpenAI` to `SapAzureOpenAI` and initialize the client within the constructor of `SapAzureOpenAI`?

packages/foundation-models-openai/src/token-provider.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export function createTokenProvider(
+  destination?: HttpDestinationOrFetchOptions
+): () => Promise<string> {
+  return async () => {
+    const dest = await getAiCoreDestination(destination);
+    const token = dest.authTokens?.[0]?.value;
+    if (!token) {
+      throw new Error(
+        'Could not retrieve authentication token from AI Core destination. Ensure the service binding or AICORE_SERVICE_KEY is configured correctly.'
```
ZhongpinWang: [req] Probably this is currently also miss-leading :) `Ensure the service binding or AICORE_SERVICE_KEY is configured correctly.` part does not mention the provided `destination` could be wrong?

sample-code/src/openai.ts
```diff
@@ -0,0 +1,81 @@
...
+  AsyncIterable<{ type: string; delta?: string }>
+> {
+  const client = await createAzureOpenAIClient({ modelDeployment: 'gpt-5' });
+
+  return client.responses.create({
+    instructions: 'You are a helpful assistant.',
+    input: 'Give me a short introduction of SAP Cloud SDK.',
+    stream: true
+  });
+}
```
ZhongpinWang, line 85: [pp] Maybe some examples for `parse()`

ZhongpinWang (commented):
Very nice solution using composition. I left some comments for potential improvements, but good enough as a PoC. Thanks for working on this.

packages/foundation-models-openai/src/client.ts
```diff
@@ -0,0 +1,72 @@
...
+ * `modelDeployment` option passed at construction.
+ *
+ * Only the endpoints supported by SAP AI Core are exposed (`chat`, `embeddings`, `responses`).
+ * Use {@link createAzureOpenAIClient} to create an instance.
+ */
+export class SapAzureOpenAI {
+  readonly chat: SapChat;
+  readonly embeddings: SapEmbeddings;
+  readonly responses: SapResponses;
+
```
davidkna-sap, line 46: [q] Why do you choose not to also have a user-facing constructor that takes `SapAzureOpenAIOptions` here?

packages/foundation-models-openai/src/token-provider.ts
KavithaSiva: True, updated the JSDoc to also reflect the provided destination could be wrong.

packages/openai/src/completions.ts
KavithaSiva: Added support now.

packages/foundation-models-openai/src/completions.ts
KavithaSiva: Yes, you are right, we can set the model as an empty string here, the right model is already targeted after the configuration creation time.

packages/openai/src/types.ts
KavithaSiva: I think they already switched to the v1 api without using the `v1` path. `{{apiurl}}/inference/deployments/{{deploymentid_gpt-5}}/chat/completions?api-version=latest` works but only with `api-version`

sample-code/src/openai.ts
KavithaSiva: Added an example with parse for chat completions api, but with responses api, parse method doesn't work, have to investigate to check why.

packages/foundation-models-openai/src/client.ts
KavithaSiva: constructors can't be async, so `new SapAzureOpenAI(options)` can't resolve the deployment ID internally. The factory function is the only way to handle that async initialization.

packages/openai/src/types.ts
ZhongpinWang: AFAIK, `api-version` is [not relevant any more when using v1 api](https://learn.microsoft.com/en-us/azure/foundry/openai/api-version-lifecycle).

> This article shows you how to use the v1 Azure OpenAI API. The v1 API simplifies authentication, removes the need for dated api-version parameters, and supports cross-provider model calls.

I didn't know that `latest` is also possible. Can you send me a link to where `latest` is documented?

packages/foundation-models-openai/src/client.ts
davidkna-sap: [pp] In that case I would prefer a `static` `create()` or `createClient()` method for this purpose or lazy just-in-time resolution.

sample-code/src/openai.ts
ZhongpinWang: Pretty weird, it should work. Maybe a gap in llm-access.

packages/openai/src/responses.ts
davidkna-sap, line 1: [pp] Cover stateful responses API, otherwise add to follow-up BLI.

packages/foundation-models-openai/src/token-provider.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Creates a token provider function compatible with `AzureOpenAI`'s `azureADTokenProvider` option.
+ * On each invocation, resolves the AI Core destination and returns the bearer token.
+ * @param destination - Optional HTTP destination or fetch options for the AI Core service.
+ * If omitted, falls back to the `aicore` service binding or the `AICORE_SERVICE_KEY` environment variable (local testing only).
+ * @returns A function that returns a promise resolving to the bearer token string.
+ */
+export function createTokenProvider(
+  destination?: HttpDestinationOrFetchOptions
+): () => Promise<string> {
+  return async () => {
```
davidkna-sap: [pp]
```suggestion
  return async function sapAiCoreTokenProvider() {
```

packages/foundation-models-openai/src/types.ts
```diff
@@ -0,0 +1,34 @@
...
+import type {
+  AzureOpenAiChatModel,
+  AzureOpenAiEmbeddingModel
+} from '@sap-ai-sdk/core';
+import type { HttpDestinationOrFetchOptions } from '@sap-cloud-sdk/connectivity';
+
+/**
+ * Union of all supported Azure OpenAI model names.
+ */
+export type SapModelName = AzureOpenAiChatModel | AzureOpenAiEmbeddingModel;
```
davidkna-sap: [pp] Add a `AzureOpenAiResponsesModel` and add `gpt-5.3-codex`

packages/openai/tsconfig.cjs.json
davidkna-sap, line 1: [q] Can we get rid of this now that `require(esm)` is part of oldest supported node (>= 22)? That said jest needs node 24 and users need a modern `tsconfig` module resolution.

sample-code/src/openai.ts
davidkna-sap, line 1: [pp] Use `gpt-5.4` and `gpt-5.4-nano` as elsewhere.

packages/foundation-models-openai/src/config.ts
```diff
@@ -0,0 +1,55 @@
...
+  const baseURL = `${dest.url.replace(/\/+$/, '')}/v2/inference/deployments/${deploymentId}`;
+
+  return {
+    baseURL,
+    apiVersion: apiVersion ?? defaultApiVersion,
+    azureADTokenProvider: createTokenProvider(destination),
+    defaultHeaders: {
+      'ai-resource-group': resourceGroup,
+      'ai-client-type': ['AI SDK JavaScript', clientType]
+        .filter(c => c)
```
davidkna-sap:
```suggestion
        .filter(Boolean)
```

packages/foundation-models-openai/src/embeddings.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.openAIEmbeddings = new Embeddings(client);
+  }
+
+  create(
+    body: WithoutModel<EmbeddingCreateParams>,
+    options?: RequestOptions
+  ): APIPromise<CreateEmbeddingResponse> {
+    return this.openAIEmbeddings.create(
+      // SAP AI Core routes via deployment URL; model is required by the SDK type but ignored by the API
+      { model: '', ...body } as EmbeddingCreateParams,
```
davidkna-sap: [pp] Assuming sonarcloud is right:
```suggestion
      { model: '', ...body } satisfies EmbeddingCreateParams,
```

packages/foundation-models-openai/src/completions.ts
```diff
@@ -0,0 +1,73 @@
...
+  >(
+    body: Params,
+    options?: RequestOptions
+  ): APIPromise<ParsedChatCompletion<ParsedT>> {
+    return this.openAICompletions.parse(
+      {
+        // SAP AI Core routes via deployment URL; model is required by the SDK type but ignored by the API
+        model: '',
+        ...body
+      } as ChatCompletionCreateParamsNonStreaming,
```
davidkna-sap: [pp] Assuming sonarcloud is right:

```suggestion
      } satisfies ChatCompletionCreateParamsNonStreaming,
```

sample-code/src/openai.ts
```diff
@@ -0,0 +1,102 @@
...
+export async function chatCompletionParse(): Promise<string | null> {
+  const client = await createAzureOpenAIClient({ modelDeployment: 'gpt-4.1' });
+
+  const response = await client.chat.completions.parse({
+    messages: [{ role: 'user', content: 'What is the capital of France?' }],
+    response_format: zodResponseFormat(CapitalResponse, 'capital_response')
+  });
+
+  return response.choices[0].message.parsed?.capital ?? null;
+}
```
davidkna-sap, line 154: [pp] Add file upload test:
```suggestion
}


/**
 * Pass a PDF file inline as base64 and use it as context in a Responses API call.
 * @returns The answer grounded in the file content.
 */
export async function responsesApiWithFile(): Promise<string | undefined> {
  const client = await createAzureOpenAIClient({
    modelDeployment: 'gpt-5.4-nano'
  });

  const filePath = join(
    import.meta.dirname,
    '..',
    '..',
    'resources',
    'test.pdf'
  );
  const base64String = await readFile(filePath, 'base64');

  const response = await client.responses.create({
    instructions: 'Transcribe the text content of the PDF document.',
    input: [
      {
        type: 'message',
        role: 'user',
        content: [
          {
            type: 'input_file',
            filename: 'test.pdf',
            file_data: `data:application/pdf;base64,${base64String}`
          },
          { type: 'input_text', text: 'Summarize the content of the file.' }
        ]
      }
    ]
  });

  return response.output_text;
}
```

packages/foundation-models-openai/src/completions.ts
```diff
@@ -0,0 +1,73 @@
...
+  ParsedChatCompletion
+} from 'openai/resources/chat/completions/completions';
+import type { ExtractParsedContentFromParams } from 'openai/lib/parser';
+import type { Stream } from 'openai/streaming';
+import type { APIPromise } from 'openai/api-promise';
+
+type RequestOptions = Parameters<Completions['create']>[1];
+
+/**
+ * Removes the `model` field from request param types so callers don't need to supply it.
```
davidkna-sap:
```suggestion
 * Removes the `model` field from request param types so callers won't be able to supply it.
```

packages/openai/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "files": [
+    "dist/**/*.js",
+    "dist/**/*.js.map",
+    "dist/**/*.d.ts",
+    "dist/**/*.d.ts.map"
+  ],
+  "scripts": {
+    "compile": "tsc",
+    "compile:cjs": "tsc -p tsconfig.cjs.json",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest --passWithNoTests",
```
davidkna-sap: [pp] Exclude the package from sonar coverage tracking.

packages/foundation-models-openai/src/config.ts
```diff
@@ -0,0 +1,55 @@
...
+ * @returns A promise that resolves to an AzureClientOptions object.
+ * @example
+ * ```ts
+ * import { AzureOpenAI } from 'openai';
+ * import { createOpenAIConfig } from '@sap-ai-sdk/foundation-models-openai';
+ *
+ * const config = await createOpenAIConfig({ modelDeployment: 'gpt-4.1' });
+ * const client = new AzureOpenAI(config);
+ * await client.chat.completions.create({
+ *   model: 'gpt-4.1',
```
davidkna-sap: [pp] Avoid `model` in sample code.
```suggestion
```

KavithaSiva: Missed this earlier.

packages/foundation-models-openai/src/config.ts
```diff
@@ -38,7 +38,7 @@ export async function createOpenAIConfig(
   );
   const resourceGroup = getResourceGroup(modelDeployment) ?? 'default';
   const dest = await getAiCoreDestination(destination);
-  const baseURL = `${dest.url.replace(/\/+$/, '')}/v2/inference/deployments/${deploymentId}`;
+  const baseURL = `${dest.url.replace(/\/$/u, '')}/v2/inference/deployments/${deploymentId}`;
```
davidkna-sap: [nth] Consider using `deploymentUrl` key returned by the deployment API (will be required for realtime API).

packages/openai/src/responses.ts
KavithaSiva: Added to the follow-up ticket instead: https://github.com/SAP/ai-sdk-js-backlog/issues/545

sample-code/src/openai.ts
KavithaSiva: Done.

packages/openai/tsconfig.cjs.json
KavithaSiva: Good point — Node 20 is EOL as of April 30, 2025. We could remove the CJS builds across all packages as a follow-up once we drop Node 20 from the build CI matrix. For now I'll keep `tsconfig.cjs.json` consistent with the other packages.

packages/foundation-models-openai/src/config.ts
KavithaSiva: Good idea, used `deploymentUrl` now.

sample-code/src/openai.ts
KavithaSiva: I am not adding this example for now, as with the given example the model was not able to recognise the uploaded file. We may need to implement the [Files API](https://developers.openai.com/api/reference/resources/files). 
Reference: https://developers.openai.com/api/docs/guides/file-inputs#uploading-files
It's better to do this in a follow-up after beta release.

packages/foundation-models-openai/src/client.ts
KavithaSiva: True, helps also with better discoverability.

packages/openai/src/types.ts
KavithaSiva: `latest` is not documented, I tried it out with Bruno, and it worked successfully, with `gpt-5`, using `latest` ends up using `gpt-5-2025-08-07`.

I am mistaken, regarding the v1 api, will reach out to them, I don't think they have switched internally to v1.
`{{apiurl}}/inference/deployments/{{deploymentid_gpt-5}}/chat/completions` works, only with `api-version`.

sample-code/src/openai.ts
KavithaSiva: ~I will create a follow-up item to support parse for responses, I don't think it should be a blocker for this beta client.~
parse with responses work, must have made a mistake with payload last time. Added the example now.

packages/openai/package.json
davidkna-sap, line 1: [pp] Set as private to avoid breaking canary releases.
[pp] Add package readme or mention in follow-up BLI.

sample-code/src/openai.ts
davidkna-sap, line 1: Note: This may break smoke tests until the package is released.


[pp] Add to e2e-test suite or mention in follow-up BLI.

sample-code/src/openai.ts
davidkna-sap: FWIW this worked for me locally. Files API does not appear to work right now.

davidkna-sap (commented):
LGTM, but we should perhaps hold merging given that this for now as this could break smoke-tests and canary releases.

packages/openai/package.json
KavithaSiva: Will do these in a follow-up for releasing the package as beta.

sample-code/src/openai.ts
KavithaSiva: Adding this in the follow-up too.

sample-code/src/openai.ts
KavithaSiva: When I try running it locally I get a response similar to:
```
I don’t have the PDF file or its extracted text in this chat, so I can’t summarize its contents yet.
```

davidkna-sap:
Please also:

- Delete `internal.d.ts`
- Delete `internal.js`
- Add `internal.ts` (like what the other packages have)
- Add this to package.json:

```
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "default": "./dist/index.js"
    },
    "./internal.js": {
      "types": "./dist/internal.d.ts",
      "default": "./dist/internal.js"
    }
  },
```

packages/openai/src/config.ts
```diff
@@ -0,0 +1,65 @@
...
+        ).execute(destination)
+      ).deploymentUrl
+    : await resolveDeploymentUrl({
+        scenarioId: 'foundation-models',
+        executableId: 'azure-openai',
+        model: translateToFoundationModel(modelDeployment),
+        resourceGroup,
+        destination
+      });
+
```
davidkna-sap, line 69: [pp]

Explicitly handle undefined `baseURL`?

sample-code/src/server.ts
```diff
@@ -280,6 +290,110 @@ app.get('/azure-openai/invoke-tool-chain', async (req, res) => {
   }
 });
 
+/* Foundation Models (Azure OpenAI via openai SDK) */
+app.get('/azure-openai-sdk/chat-completion', async (req, res) => {
```
davidkna-sap: [pp] `/openai-sdk` or `/openai`? Could also be handled in follow-up.

sonar-project.properties
```diff
@@ -22,6 +22,7 @@ sonar.javascript.lcov.reportPaths=packages/**/coverage/lcov.info
 sonar.coverage.exclusions=\
   **/sample-cap/**,\
   **/sample-code/**,\
+  packages/foundation-models-openai/**,\
```
davidkna-sap:
```suggestion
  packages/openai/**,\
```

sample-code/src/openai.ts
davidkna-sap, line 1: [pp] Rename to `openai.ts`/`openai-sdk.ts` and consider moving outside `foundation-models`.

davidkna-sap (dismissed):
LGTM, assuming follow-ups to handle any missing pieces.

sample-code/src/server.ts
KavithaSiva: Handled it here.

sonar-project.properties
KavithaSiva: Adding tests in this PR, so will remove the exclusion.

sample-code/src/openai.ts
KavithaSiva: Done.

KavithaSiva:
Should we reconsider the Sap prefix on the wrapper classes? The current names are:
  - `SapAzureOpenAI`
  - `SapChat`
  - `SapCompletions`
  - `SapEmbeddings`
  - `SapResponses`

Please discuss before releasing. @davidkna-sap

davidkna-sap (dismissed):
LGTM.

Please review @ZhongpinWang @marikaner and add any concerns you have.

davidkna-sap (commented):
TODO: Add to smoke-test package after release, but before merge

packages/openai/src/client.ts
```diff
@@ -0,0 +1,53 @@
...
+  readonly chat: SapChat;
+  readonly embeddings: SapEmbeddings;
+  readonly responses: SapResponses;
+
+  /** @internal — use {@link SapOpenAi.createClient} instead */
+  constructor(client: AzureOpenAI) {
+    this.chat = new SapChat(client);
+    this.embeddings = new SapEmbeddings(client);
+    this.responses = new SapResponses(client);
+  }
```
marikaner, line 52: [q/pp] The original AzureOpenAI client does not use the factory pattern as we do here. Why did you decide to use that?
I think for consistency it would be nice to just use `constructor(options: SapOpenAiInput)`.

[q] On a different note: In case we decide to keep the current structure: could the constructor be private?

packages/openai/src/completions.ts
```diff
@@ -0,0 +1,76 @@
...
+ * Removes the `model` field from request param types so callers won't be able to supply it.
+ * @internal
+ */
+export type WithoutModel<T> = Omit<T, 'model'>;
+
+/**
+ * Wraps `Completions` exposing only `create` and `parse`, with `model` omitted from the public API as SAP AI Core routes requests via the deployment URL.
+ * @experimental This class is experimental and may change at any time without prior notice.
+ */
+export class SapCompletions {
```
marikaner, line 28: [q] Should we add API docs to the constructor and functions?

packages/openai/src/config.ts
```diff
@@ -0,0 +1,77 @@
...
+  getResourceGroup,
+  translateToFoundationModel,
+  isDeploymentIdConfig
+} from '@sap-ai-sdk/ai-api/internal.js';
+import { DeploymentApi } from '@sap-ai-sdk/ai-api';
+import { createTokenProvider } from './token-provider.js';
+import type { AzureClientOptions } from 'openai/azure';
+import type { SapOpenAiInput, SapOpenAiOptions } from './types.js';
+
+const defaultApiVersion = '2024-10-21';
```
marikaner, line 13: [q] Is there a newer one?

packages/openai/src/config.ts
```diff
@@ -0,0 +1,77 @@
...
+ * import { createOpenAIConfig } from '@sap-ai-sdk/openai';
+ *
+ * const config = await createOpenAIConfig('gpt-5.4');
+ * const client = new AzureOpenAI(config);
+ * await client.chat.completions.create({
+ *   messages: [{ role: 'user', content: 'Hello!' }]
+ * });
+ * ```
+ */
+export async function createOpenAIConfig(
```
marikaner: [req] stylguide

```suggestion
export async function createOpenAiConfig(
```

packages/openai/src/config.ts
```diff
@@ -0,0 +1,77 @@
...
+  options: SapOpenAiInput
+): Promise<AzureClientOptions> {
+  const opts: SapOpenAiOptions =
+    typeof options === 'string' ? { deployment: options } : options;
+  const { deployment, destination, apiVersion, clientType } = opts;
+
+  const resourceGroup = getResourceGroup(deployment) ?? 'default';
+
+  const baseURL = isDeploymentIdConfig(deployment)
```
marikaner:
```suggestion
  const baseUrl = isDeploymentIdConfig(deployment)
```

packages/openai/src/config.ts
```diff
@@ -0,0 +1,77 @@
...
+    ? (
+        await DeploymentApi.deploymentGet(
+          deployment.deploymentId,
+          {},
+          { 'AI-Resource-Group': resourceGroup }
+        ).execute(destination)
```
marikaner: [q] Can this fail? Do we need error handling here?

packages/openai/src/config.ts
```diff
@@ -0,0 +1,77 @@
...
+ *   messages: [{ role: 'user', content: 'Hello!' }]
+ * });
+ * ```
+ */
+export async function createOpenAIConfig(
+  options: SapOpenAiInput
+): Promise<AzureClientOptions> {
+  const opts: SapOpenAiOptions =
+    typeof options === 'string' ? { deployment: options } : options;
+  const { deployment, destination, apiVersion, clientType } = opts;
```
marikaner: [nth] I think it would be nice to assign the default value here already:
```suggestion
  const { deployment, destination, apiVersion = defaultApiVersion, clientType } = opts;
```

That way should it be used somewhere in between it would already be the correct value.

packages/openai/src/responses.ts
```diff
@@ -0,0 +1,71 @@
...
+import type { WithoutModel } from './completions.js';
+
+type RequestOptions = Parameters<Responses['create']>[1];
+
+/**
+ * Wraps `Responses` exposing only `create` and `parse`, with `model` omitted from the public API as SAP AI Core routes requests via the deployment URL.
+ * @experimental This class is experimental and may change at any time without prior notice.
+ */
+export class SapResponses {
+  private readonly openAIResponses: Responses;
```
marikaner: [req] Generally, please check compliance with the style guide.
```suggestion
  private readonly openAiResponses: Responses;
```

marikaner (changes requested):
I think a few type tests could be nice

packages/openai/src/client.ts
davidkna-sap: Also discussed here: https://github.com/SAP/ai-sdk-js/pull/1837#discussion_r3234881144

packages/openai/src/config.ts
davidkna-sap: IIRC more recent versions are previews.

packages/openai/src/client.ts
KavithaSiva: I tried creating the client with a constructor instead which passes along a promise to other classes like `SapChat`, `SapEmbeddings`, `SapResponses`. But then this forces me to add aync to `create` and `parse` functions, which in changes the return type of these functions `Promise<T>` instead of `APIPromise<T>`.
This would strip away some convenience functions for the returned response like `withResponse()` and `asResponse()`.

We could argue that these functions may not be used by end users that much, but if OpenAI extends the `APIPromise` class with more functions, this would be lost for us.
I think using the static function to create the client once and then for all calls adhering to the same style as OpenAI is better.
Also, since our client explicitly has to wait for resolving deployment details, I think this is more evident if this is resolved while creating the client, rather than postponing it to later during `create`, where it is not so evident.

davidkna-sap: One idea I hade after looking at how azure handles it is to have users call a `resolveDeployment('my-gpt')` helper exposed by this package and use that as a static `deployment` in the constructor.

But I would prefer keeping this as-is and handling that in a follow-up.

packages/openai/src/completions.ts
KavithaSiva: Made the constructors internal and added JSDocs to other functions.

packages/openai/src/config.ts
KavithaSiva: When the deployment details are empty, we already identify throw an error below these lines.
When the HTTP call itself fails, the error thrown from HTTP client should itself suffice, WDY think?

packages/openai/src/client.ts
KavithaSiva: Hmm, WDY think about this idea @marikaner.
Personally, I would prefer to keep it as one function which resolves deployment and creates the client.

marikaner: I agree with leaving it as is. Thanks for taking a look at the details though.

packages/openai/src/config.ts
marikaner: I think it would be safer to catch it here and log a message + the error. Something along the lines "fetching deployment failed with error xy". Otherwise what does it look like at the moment?
Also, we might want to add a test for it.

.changeset/chatty-stamps-report.md
```diff
@@ -0,0 +1,7 @@
+---
+"@sap-ai-sdk/openai": minor
+---
+
+[feat] Add new `@sap-ai-sdk/openai` package with a SAP AI Core-aware OpenAI client.
```
marikaner: [q] Is it a beta/experimental version? Maybe mention it here.

packages/openai/src/config.ts
KavithaSiva: Done.

marikaner (dismissed):
👍

packages/openai/src/responses.ts
```diff
@@ -0,0 +1,85 @@
...
+    body: WithoutModel<ResponseCreateParamsBase>,
+    options?: RequestOptions
+  ): APIPromise<Response | Stream<ResponseStreamEvent>>;
+  create(
+    body: WithoutModel<ResponseCreateParamsBase>,
+    options?: RequestOptions
+  ): APIPromise<Response | Stream<ResponseStreamEvent>> {
+    return this.openAiResponses.create(
+      // SAP AI Core routes via deployment URL; model is required by the SDK type but ignored by the API
+      { model: '', ...body } satisfies ResponseCreateParamsBase,
```
davidkna-sap:
```suggestion
      { ...body, model: '' } satisfies ResponseCreateParamsBase,
```

packages/openai/src/embeddings.ts
```diff
@@ -0,0 +1,40 @@
...
+   * @param options - Optional request options.
+   * @returns A promise resolving to a {@link CreateEmbeddingResponse}.
+   */
+  create(
+    body: WithoutModel<EmbeddingCreateParams>,
+    options?: RequestOptions
+  ): APIPromise<CreateEmbeddingResponse> {
+    return this.openAiEmbeddings.create(
+      // SAP AI Core routes via deployment URL; model is required by the SDK type but ignored by the API
+      { model: '', ...body } satisfies EmbeddingCreateParams,
```
davidkna-sap:
```suggestion
      { ...body, model: '' } satisfies EmbeddingCreateParams,
```

davidkna-sap (dismissed):
LGTM, I just made two minor suggestions.

davidkna-sap (approved):

PR #1839, opened 2026-05-08
chore: Handle SAP Cloud SDK v4.7.0 and eslint 10 update

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#560.


pnpm-workspace.yaml
```diff
@@ -13,6 +13,11 @@ packages:
   - tests/smoke-tests
   - tests/type-tests
 
+allowBuilds:
+  better-sqlite3: false
+  esbuild: false
+  unrs-resolver: false
+
```
KavithaSiva, line 20: [q] Nice improvement? But why did you choose only these 3 packages?

KavithaSiva: Only, these 3 have install scripts.

KavithaSiva (approved):
LGTM

PR #1855, opened 2026-05-12
chore: Update model list

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#567.


scripts/sync-model-types.ts
davidkna-sap, line 1: I changed the logic here:

- not to consider not earlier than dates as retired
- add new models at the end of their respective model-family group rather than the end of their respective union.

KavithaSiva: Thank you for the changes, the model prefix change especially makes reviewing much easier.

KavithaSiva (approved):
LGTM, Thanks for taking care.

PR #1859, opened 2026-05-13
chore: Improve smoke-test deployments

davidkna-sap:
## Context

~~Part of https://github.com/SAP/ai-sdk-js-backlog/issues/553~~

Closes https://github.com/SAP/ai-sdk-js-backlog/issues/571

Use a new approach to handle smoke-test deployment that avoids persisting the node auth token by moving the deployed app into nested subdirectory that is hidden from cf.

https://github.com/SAP/ai-sdk-js/actions/runs/26097081688/job/76737603328


tests/smoke-tests/scripts/create-deployment.ts
```diff
@@ -0,0 +1,93 @@
...
+import { resolve } from 'node:path';
+import { promisify } from 'node:util';
+
+const pExecFile = promisify(execFile);
+
+const workspaceRoot = resolve(import.meta.dirname, '..', '..', '..');
+const appDir = resolve(import.meta.dirname, '..');
+const appName = '@sap-ai-sdk/smoke-tests';
+
+async function installDependencies(): Promise<string> {
```
marikaner: [req] This does a lot of things but installing dependencies. Maybe call it something like `prepareDeployment()` or maybe you have better ideas.

tests/smoke-tests/scripts/create-deployment.ts
```diff
@@ -0,0 +1,93 @@
...
+  const tmpDir = await mkdtemp(resolve(tmpdir(), 'cf-deploy-smoke-tests-'));
```
marikaner: [q] Is the trailing `-` on purpose?

tests/smoke-tests/package.json
```diff
@@ -20,16 +20,17 @@
     "@langchain/langgraph": "=1.3.0",
     "@langchain/mcp-adapters": "=1.1.3",
     "@langchain/textsplitters": "=1.0.1",
+    "@modelcontextprotocol/sdk": "=1.29.0",
```
marikaner: [q] Why do we have these dependencies here? Wouldn't it be easier to use the dependencies from the samples repo and overwrite the SDK dependencies with `canary`? I see that we currently don't have any dependencies in the deployment directory. Why do we do it this way? If there is a good reason we should document this in the readme.

marikaner (changes requested):

tests/smoke-tests/scripts/create-deployment.ts
davidkna-sap: It's on purpose, the string is used as a prefix of the temporary directory.

tests/smoke-tests/package.json
davidkna-sap: I am not sure why this was set up like this, but I will investigate ways to avoid the duplication.

davidkna-sap: Overrides would only work in the workspace config, I moved it to `peerDependencies` and used this opportunity to (mostly) make local smoke tests work.

tests/smoke-tests/scripts/create-deployment.ts
```diff
@@ -0,0 +1,85 @@
...
+import { promisify } from 'node:util';
+
+const pExecFile = promisify(execFile);
+
+const workspaceRoot = resolve(import.meta.dirname, '..', '..', '..');
+const appDir = resolve(import.meta.dirname, '..');
+const appName = '@sap-ai-sdk/smoke-tests';
+
+await using tmpDir = await mkdtempDisposable(
+  resolve(tmpdir(), 'cf-deploy-smoke-tests-')
```
marikaner, line 20: [q] Why the trailing slash?

tests/smoke-tests/jest.config.mjs
```diff
@@ -1,7 +1,8 @@
 import config from '../../jest.config.mjs';
+
 export default {
   ...config,
-  globalSetup: undefined,
-  globalTeardown: undefined,
+  globalSetup: import.meta.resolve('./scripts/local-server-setup.ts'),
```
marikaner: [q] is this only used locally? Looks like this would be loaded on cf too, right?

marikaner (commented):
I honestly don't fully understand why adding another directory is able to trick CF, can you elabore a little bit? I couldn't find where the artifactory token is used (because it needs to be used somewhere). I am sure I am just not seeing the bigger picture here.

tests/smoke-tests/scripts/create-deployment.ts
davidkna-sap: This leads to `cf-deploy-smoke-tests-RANDOM_STRING` otherwise there is no separation.

tests/smoke-tests/jest.config.mjs
davidkna-sap: It would, but I added a guard to avoid loading it in CI or cf:
```ts
if (process.env.VCAP_SERVICES || process.env.CI) {
  return;
}
```

davidkna-sap:
@marikaner The root wrapper directory has a package.json which has no dependencies at all, it just forwards scripts to the nested package. Cloud Foundry finds that root `package.json` and, stops looking deeper rather than recursing into the actual app directory.

The Artifactory token is used during the initial `pnpm install` (setup action). `pnpm deploy` then re-uses the already-downloaded dependencies from the store, so no token is needed at the CF staging step itself (or there would be an error due to missing tokens as the setup step persistently sets up the registry mirror).

tests/smoke-tests/scripts/create-deployment.ts
marikaner: sorry for asking the same question again. 🫠

marikaner:
> @marikaner The root wrapper directory has a package.json which has no dependencies at all, it just forwards scripts to the nested package. Cloud Foundry finds that root `package.json` and, stops looking deeper rather than recursing into the actual app directory.
> 
> The Artifactory token is used during the initial `pnpm install` (setup action). `pnpm deploy` then re-uses the already-downloaded dependencies from the store, so no token is needed at the CF staging step itself (or there would be an error due to missing tokens as the setup step persistently sets up the registry mirror).

Wouldn't this generally be the case (CF staging does not need the token), if we run `pnpm install` before the push? Or can we not configure the `node_modules` to already be there using [`.cfignore`](https://docs.cloudfoundry.org/devguide/deploy-apps/prepare-to-deploy.html#exclude) without having a subdirectory?
I am pretty sure we have done this before, but I currently have a hard time finding transparent documentation from CF on this.
I am only asking to avoid that we have a more complex setup than necessary. If the subdirectory is needed for something else, than it's also fine.

davidkna-sap:
@marikaner This would still leads to issues:

- CloudFoundry would try to rebuild all binary dependencies (running package scripts again, which we want to avoid executing at all)
- CloudFoundry unconditionally needs the token even if full `node_modules` is uploaded as it will query the registry
- CloudFoundry will try to re-resolve dependencies that have already been resolved by pnpm with minimumReleaseAge and excludes for Cloud SDK (can somewhat be avoided with `npm shrinkwrap`, see current smoke test setup)
- CloudFoundry-npm can get confused by the pnpm node_modules layout (can somewhat be avoided, see current smoke test setup)

marikaner:
> @marikaner This would still leads to issues:
> 
> * CloudFoundry would try to rebuild all binary dependencies (running package scripts again, which we want to avoid executing at all)
> * CloudFoundry unconditionally needs the token even if full `node_modules` is uploaded as it will query the registry
> * CloudFoundry will try to re-resolve dependencies that have already been resolved by pnpm with minimumReleaseAge and excludes for Cloud SDK (can somewhat be avoided with `npm shrinkwrap`, see current smoke test setup)
> * CloudFoundry-npm can get confused by the pnpm node_modules layout (can somewhat be avoided, see current smoke test setup)

Convinced. I think it wouldn't hurt adding these to some README or ADR (nothing big though). Just to avoid future confusion.

marikaner (dismissed):

tests/smoke-tests/jest.config.mjs
marikaner: Then maybe the name of the script is incorrect. It is just server setup, right?

marikaner (approved):

PR #1874, opened 2026-05-20
chore(orchestration): Export types from schema that are used in public API - Method 1

davidkna-sap:
## Context

This ensure the types show up/expand in tsdoc.
https://github.com/SAP/ai-sdk-js-backlog/issues/507

Found the missing types via CC + TS lsp.

Closes #1878

### DIFF
--- a/packages/orchestration/src/index.ts
+++ b/packages/orchestration/src/index.ts
@@ -89,5 +89,16 @@ export type {
   ChatCompletionTool,
   FunctionObject,
   Error as OrchestrationError,
-  Citation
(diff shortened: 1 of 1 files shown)


InjunPark-sap (commented):
LGTM!

packages/orchestration/src/index.ts
davidkna-sap, line 1: Reading some of the related documentation, not exporting some of these was a conscious choice - an alternative approach would be to properly expose `/internal` in the `package.json`.

PR #1876, opened 2026-05-20
chore: Replace python parquet generation script with ts

ZhongpinWang:
Update:

Since it is at the moment not possible to test if the generated parquet works with the context registry service, I would pause this PR for a moment.

---

As in the title. 

Also added a flag to the generation script to leave out items with `[PREDICT]` placeholder. This is a preparation step for the upcoming context registry feature support.


package.json
```diff
@@ -65,6 +65,7 @@
     "tsx": "^4.22.0",
     "typedoc": "^0.28.19",
     "typescript": "^6.0.3",
-    "zod": "^4.4.3"
+    "zod": "^4.4.3",
+    "@dsnp/parquetjs": "^1.8.7"
```
davidkna-sap, line 69: [pp] I would prefer something lighter (dependency-tree wise), e.g. `hyparquet-writer`, `parquet-wasm` or `@duckdb/duckdb-wasm`.

sample-code/resources/product_data_with_placeholders.parquet
davidkna-sap, line 1: [nth] Consider enabling compression. This file is quite a bit larger despite similar contents.

package.json
ZhongpinWang: I picked this one only because it is quite popular. Also different parquet library has different implementation of the protocol. For now I can't really test if the generated parquet works with our services (as they might expect parquet file exported from HANA). We can hold this PR for a moment and see if the package in the end works.

davidkna-sap:
Closing in favour of #1946

PR #1878, opened 2026-05-20
chore: Expose unexported types in tsdoc - Method 2

davidkna-sap:
## Context

Different way of handling #1874

Closes #1874


tsconfig.typedoc.json
```diff
@@ -52,6 +52,7 @@
       "packages/foundation-models/src/index.ts",
       "packages/langchain/src/index.ts",
       "packages/orchestration/src/index.ts",
+      "packages/orchestration/src/internal.ts",
```
davidkna-sap, line 55: This will lead to a separate entry with a few duplicate entries from `index.ts` in rendered docs.

PR #1885, opened 2026-05-26
chore: Set explicit workflow permissions

davidkna-sap:
## Context

Part of SAP/ai-sdk-js-backlog#475.


davidkna-sap (commented):
Regarding current failure to run checks: https://www.githubstatus.com/

.github/workflows/publish.yml
```diff
@@ -16,14 +16,22 @@ on:
...
+permissions: {}
+
 jobs:
   delegate_to_release_job:
     if: ${{ github.event_name == 'release' }}
+    permissions:
+      contents: read
+      id-token: write
     uses: ./.github/workflows/publish-release.yml
     secrets: inherit
```
marikaner: From what I understand this should be part of our hardening - we should avoid using `secrets: inherit`

marikaner (dismissed):

.github/workflows/publish.yml
davidkna-sap: I will handle it, but wanted to get this out to get ahead of the token permission change.

davidkna-sap (commented):
Some additional changes slipped into this PR. If necessary I can split them but it shouldn't hinder reviewing too much as all the changes related to this PR reside in `.github/`

packages/document-grounding/package.json
```diff
@@ -18,13 +18,21 @@
   "type": "module",
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
```
KavithaSiva, line 20: [pp] These can also be removed after we support only node 22+ onwards during https://github.com/SAP/ai-sdk-js-backlog/issues/542

KavithaSiva (approved):
LGTM

PR #1917, opened 2026-06-09
chore(sample-code): Handle server startup failures

davidkna-sap:
## Context

Handle startup failures.

Common errors include `EADDRINUSE`.

### DIFF
--- a/sample-code/src/server.ts
+++ b/sample-code/src/server.ts
@@ -103,8 +103,12 @@ import type {
 const app = express();
 const port = 8080;
 
-app.listen(port, () => {
+const server = app.listen(port, () => {
   console.log(`Server running at http://localhost:${port}`);
+});
+
+server.on('error', error => {
+  throw error;
 });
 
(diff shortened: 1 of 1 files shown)


sample-code/src/server.ts
```diff
@@ -103,10 +103,14 @@ import type {
...
 
-app.listen(port, () => {
+const server = app.listen(port, () => {
   console.log(`Server running at http://localhost:${port}`);
 });
 
+server.on('error', error => {
+  throw error;
+});
+
```
InjunPark-sap:
```suggestion
server.on('error', (error: Error) => {
  console.error(`Failed to start server on port ${port}`, error);
  process.exit(1);
});

```

InjunPark-sap (changes requested):
The current `throw error` reproduces node's default behavior when no error listener is
registered, so this adds no new handling. A more useful pattern for local development

InjunPark-sap (approved):

PR #1923, opened 2026-06-10
chore: Update `parseBatchOutput()` to handle `Buffer`

davidkna-sap:
## Context

Actually the generated client currently returns a `Buffer`

### DIFF
--- a/packages/foundation-models/src/azure-openai/azure-openai-batch-output.ts
+++ b/packages/foundation-models/src/azure-openai/azure-openai-batch-output.ts
@@ -51,9 +51,13 @@ export async function parseBatchOutput(
     text = data;
   } else if (Buffer.isBuffer(data)) {
     text = data.toString('utf-8');
-  } else {
-    // Blob case
+  } else if (data instanceof Blob) {
     text = await data.text();
+  } else {
+    // typeguard - all cases handled
+  	data satisfies never;
+  	// let if fail in a natural way
+  	text = data as any;
   }
   return text
     .split('\n')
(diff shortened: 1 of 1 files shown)


InjunPark-sap (dismissed):
LGTM!

InjunPark-sap (commented):
But I see there are no tests which are added for the Buffer and string branches.

Do you think we need to add cases to azure-openai-batch-output.test.ts?

e.g. parseBatchOutput(Buffer.from(jsonlString)), parseBatchOutput(jsonlString) (if string is kept)

davidkna-sap:
@InjunPark-sap The E2E-tests already cover this.

InjunPark-sap:
quick questions:
1. Is the `string` branch also covered by E2E, or is there an actual caller that passes a raw string? 
3. Once [#6663 in cloud-sdk-js](https://github.com/SAP/cloud-sdk-js/pull/6663) is released and `execute()` always returns `Blob`, the `Buffer` branch here becomes dead code. Is this intended as a temporary shim until then, or a permanent fallback?

davidkna-sap:
> Is the `string` branch also covered by E2E, or is there an actual caller that passes a raw string?

The `string` branch is handled reasonably well by TS type-safety.

> Once [#6663 in cloud-sdk-js](https://github.com/SAP/cloud-sdk-js/pull/6663) is released and `execute()` always returns `Blob`, the `Buffer` branch here becomes dead code. Is this intended as a temporary shim until then, or a permanent fallback?

Keeping it requires low maintenance effort, and removing would be breaking change that is hard to justify in this case (even for experimental code).

InjunPark-sap:
Fair enough on both points. It makes sense to keep it as a permanent fallback  given the breaking change concern. Thanks for the context!

marikaner (approved):
lgtm

Issue #1898, opened 2026-05-31
How do we compress the context?


SanthoshCheekoti:
### Checklist

- [x] My issue is not an [AI Core service issue](https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/support-process), but an SDK issue.
- [x] My issue is a JS/TS related issue, not a [Java related issue](https://github.com/SAP/ai-sdk-java/issues).
- [x] My issue is not [already reported or documented](https://github.com/SAP/ai-sdk-js/issues).

### Describe the Question

Hello - I'm using CloudSDK + Langgraph. I have a node in my agent that calls a tool hosted by a remote MCP server. The tool is called repeatedly as part of the agent loop and is generating a lot of context, and hence my token count is very large.  
1. Does SDK support the ability to compress the context based on a trigger ?
2. Secondly, we are having a system prompt that is a bit heavy. Does SDK prompt caching as shown below? If yes, please provide details on how to use it.
https://docs.langchain.com/oss/javascript/integrations/middleware/anthropic

davidkna-sap:
> Does SDK support the ability to compress the context based on a trigger ?

The SAP Cloud SDK for AI is currently not an agent framework by itself, so it does not automatically compress conversation history or summarize context based on a trigger within an agent loop.

You will need to add appropriate handling to your langgraph that automatically forward the message history to a summarisation agent.
It is likely that some implementations of this pattern for langgraph are publicly available, and e.g. when using langchain.js you can use the [built-in summarisation middleware](https://docs.langchain.com/oss/javascript/langchain/middleware/built-in#summarization).

> Does SDK prompt caching as shown below? If yes, please provide details on how to use it.

The SAP Cloud SDK for AI langchain package does not yet fully expose this feature.

Nevertheless, you can already cache your system prompt with something like the following:

```ts
import { OrchestrationClient } from '@sap-ai-sdk/langchain';

const client = new OrchestrationClient({
    promptTemplating: {
      model: {
        name: 'anthropic--claude-4.5-haiku'
      },
      prompt: {
        template: [
          {
            role: 'system',
            content: [
              {
                type: 'text',
                // Repeated filler text to exceed Anthropic's minimum cacheable token threshold (4096 tokens for Haiku 4.5).
                text:
                  'You are a knowledgeable assistant. ' +
                  'You have deep expertise in science, technology, history, literature, mathematics, and many other fields. '.repeat(
                    220
                  ),
[...]

PR #1926, opened 2026-06-11
fix: Prevent accumulation of duplicate tools on repeated invocations

davidkna-sap:
## Context

Closes https://github.com/SAP/ai-sdk-js-backlog/issues/588

Related https://github.com/SAP/ai-sdk-js/issues/1898

### DIFF
--- a/packages/langchain/src/orchestration/client.ts
+++ b/packages/langchain/src/orchestration/client.ts
@@ -462,31 +462,28 @@ export class OrchestrationClient extends BaseChatModel<
     };
 
     if (tools.length) {
-      config.promptTemplating.prompt ??= Object.create(null);
+      config.promptTemplating.prompt ??= {};
       if (
         typeof config.promptTemplating.prompt === 'object' &&
         !isTemplateRef(config.promptTemplating.prompt)
       ) {
         // Copy before mutating
-        config.promptTemplating.prompt = Object.assign(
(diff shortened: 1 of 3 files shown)


InjunPark-sap (dismissed):
LGTM!

packages/langchain/src/orchestration/client.ts
```diff
@@ -462,28 +462,31 @@ export class OrchestrationClient extends BaseChatModel<
...
       }
     }
 
     // Handle responseFormat for structured output
     if (responseFormat) {
       // Ensure prompt object exists
-      if (!config.promptTemplating.prompt) {
-        config.promptTemplating.prompt = {};
-      }
+      config.promptTemplating.prompt ??= Object.create(null);
```
marikaner: [q] does `Object.create(null)` bring a meaningful advantage? We never used it on our codebase and I am all for consistency. Generally `{}` is easier on the eyes and the advantages of `Object.create(null)` are not significant afaik. Maybe we should even add one of the options (`{}` or `Object.create(null)`) to the style guide.

davidkna-sap: It doesn't have a `prototype` which can avoid prototype pollution issues, but I think I will revert that for now, as it could have unintended consequences.

marikaner (approved):

Issue #1500, opened 2026-02-03
Feat: Add tool_choice support to Orchestration API


jerome-benoit:
### Checklist

- [x] My issue is not an [AI Core service issue](https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/support-process), but an SDK issue.
- [x] My issue is a JS/TS related issue, not a [Java related issue](https://github.com/SAP/ai-sdk-java/issues).
- [x] My issue is not [already reported or documented](https://github.com/SAP/ai-sdk-js/issues).

### Describe the Problem

The Foundation Models API explicitly types `tool_choice`:

```typescript
// packages/foundation-models/.../create-chat-completion-request.ts
tool_choice?: AzureOpenAiChatCompletionToolChoiceOption;
```

The Orchestration API does not:

```yaml
# packages/orchestration/src/spec/api.yaml
LLMModelDetails:
  params:
    type: object
    additionalProperties: true  # no tool_choice
```

This prevents applications using orchestration features (filtering, grounding, masking) from controlling tool invocation behavior.

### Propose a Solution

Add `tool_choice` to the Orchestration API schema if the backend supports it:

```typescript
interface LlmModelParams {
  tool_choice?: 'auto' | 'none' | 'required' | { type: 'function'; function: { name: string } };
}
```

### Describe Alternatives

Switch to Foundation Models API when `tool_choice` control is needed, losing orchestration features.

### Affected Development Phase

Development

### Impact

Impaired

### Additional Context

Related: #490 (tool_choice in Langchain Foundation Models client).

KavithaSiva:
@jerome-benoit Closing as stated by you, feel free to reopen if otherwise.

InjunPark-sap:
Hi @jerome-benoit,

Following up on @davidkna-sap's comments in this thread — the documentation is now addressed in PR #1939 (merged).

**What is now documented:**

`ChatCompletionRequest` (the argument to `.chatCompletion()`) only accepts `messages`, `messagesHistory`, and `placeholderValues`. TypeScript users passing additional fields will see a compile-time error; at runtime those fields are ignored since `constructCompletionPostRequest()` reconstructs the request body explicitly.

**The supported way to use `tool_choice`:**

Pass it via `LlmModelParams` at construction time. `LlmModelParams` accepts model-specific parameters beyond its named fields:

```typescript
const client = new OrchestrationClient({
  promptTemplating: {
    model: {
      name: 'gpt-4o',
[...]

PR #1939, opened 2026-06-16
chore: orchestration document unsupported parameters

InjunPark-sap:
## Context

Part of SAP/ai-sdk-js-backlog#511.

- `ChatCompletionRequest` (passed to `.chatCompletion()`) silently drops any field beyond `messages`, `messagesHistory`, and `placeholderValues` since`constructCompletionPostRequest()` reconstructs the request body explicitly.
- This was undocumented, causing repeated user confusion (#1500).


packages/orchestration/src/orchestration-types.ts
```diff
@@ -154,7 +157,16 @@ export interface TranslationModule {
 export type OrchestrationErrorResponse = ErrorResponse;
 
 /**
- * Model Parameters for LLM module configuration.
+ * Model parameters for the LLM module configuration.
+ * These are set once at client construction and apply to every request made by this client.
+ * To use different parameters for a request, create a new {@link OrchestrationClient} instance.
+ * Model-specific parameters not in the standard set (e.g. `tool_choice`, `reasoning_effort`)
+ * can be passed via the index signature, but behavior depends on the underlying model provider.
```
marikaner: [req] The `LlmModelParams` type does not have an index signature. I also think "index signature" is unnecessarily technical. Also, "standard set" is not really clear. Please rephrase to something along those lines:
```suggestion
 * You can pass model-specific parameters, that are not specified in `LlmModelParams` (e.g. `tool_choice`, `reasoning_effort`), but behavior depends on the underlying model provider.
```

packages/orchestration/src/orchestration-types.ts
```diff
@@ -154,7 +157,16 @@ export interface TranslationModule {
...
+ * @example
+ * params: { temperature: 0.7, max_tokens: 512 }
+ * @example
+ * // Model-specific parameters not in the standard set can be passed via the index signature.
```
marikaner: Same here ;)

packages/orchestration/src/orchestration-types.ts
```diff
@@ -167,6 +179,12 @@ export type LlmModelParams = {
 
 /**
  * Representation of the 'Template' schema.
+ * Includes `tools` and `response_format` inherited from the base template type.
+ * @remarks
+ * `tools` and `response_format` are fixed at construction time; create a new {@link OrchestrationClient}
+ * to use different tools or response format for a request.
+ * `tool_choice` is not part of the Orchestration Service template schema.
+ * It may be passed via {@link LlmModelParams} using the index signature, but behavior depends on the underlying model provider.
```
marikaner: And here

marikaner (dismissed):

.changeset/document-unsupported-params.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': patch
+---
+
+[impr] Document that model parameters, `tools`, and `response_format` must be set in the constructor config, not in `chatCompletion()`, and that `tool_choice` is not part of the Orchestration Service template schema but may be passed via `LlmModelParams`.
```
marikaner: I think changes to the API docs shouldn't go into the changelog. I would like to remove this.

marikaner (dismissed):

.changeset/document-unsupported-params.md
InjunPark-sap: will apply it!

packages/orchestration/src/orchestration-types.ts
```diff
@@ -154,7 +157,14 @@ export interface TranslationModule {
 export type OrchestrationErrorResponse = ErrorResponse;
 
 /**
- * Model Parameters for LLM module configuration.
+ * Model parameters for the LLM module configuration.
+ * These are set once at client construction and apply to every request made by this client.
+ * To use different parameters for a request, create a new {@link OrchestrationClient} instance.
+ * You can pass model-specific parameters, that are not specified in `LlmModelParams` (e.g. `tool_choice`, `reasoning_effort`), but behavior depends on the underlying model provider.
```
davidkna-sap, line 163: I am not sure we should word it like that, as there is some harmonisation on certain model parameters.

davidkna-sap: Some of these we may want to document in the future e.g. `reasoning_effort`

davidkna-sap (dismissed):
Despite my concerns, this feels like an improvement so I wouldn't be opposed to merging and discussing later (e.g. during review).

packages/orchestration/src/orchestration-types.ts
InjunPark-sap: how about deleting the entire sentence in line 163, since the [`Record<string, any>` type](https://github.com/SAP/ai-sdk-js/pull/1939/changes/BASE..196f0a546855b3cdf6ad73d663c1948d4ba76107#diff-185f8389ea59e9b83ffecbfef3cb6e55a0ea19b15c46b8d88bc59c3c8b10e3c9L166) itself implies that additional parameters are allowed, so no explanation is necessary.

Or do we keep it like that way?

davidkna-sap: Model parameters mostly are model-specific, but the service does have additional harmonisation for certain parameters.

davidkna-sap (approved):

PR #1943, opened 2026-06-17
fix: Improve error handling for non-JSON responses in SseStream

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#486.

Rewords the message a bit and includes a (window of) the string payload to better surface any server error messages.


InjunPark-sap (dismissed):
LGTM! It seems clear change and test cases cover the bug case

InjunPark-sap (approved):

marikaner (approved):

PR #1944, opened 2026-06-17
chore: Add review-spec-update claude skill

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#591.


marikaner (approved):
Thanks for adding that.

PR #1946, opened 2026-06-17
chore: Replace python parquet generation script with ts [differnt lib…

davidkna-sap:
…rary]

## Context

#1876 but using a lighter library that also produces smaller parquet files.


sample-code/resources/generate-parquet.ts
```diff
@@ -0,0 +1,88 @@
...
+ * This script generates a parquet file containing product data.
+ * By default, the generated file includes cells with '[PREDICT]'-placeholders for prediction with SAP RPT.
+ * Use the --no-predict flag to exclude rows with '[PREDICT]' placeholders.
+ * @example
+ * npx tsx generate-parquet.ts                    // Include all data (with [PREDICT] placeholders)
+ * npx tsx generate-parquet.ts --no-predict       // Exclude rows with [PREDICT] placeholders
+ */
+/* eslint-disable no-console */
+
+import { join } from 'path';
```
davidkna-sap:
```suggestion
import { join } from 'node:path';
```

sample-code/resources/generate-parquet.ts
```diff
@@ -0,0 +1,88 @@
...
+const rows = includePredictRows
+  ? [...predictRows, ...regularRows]
+  : regularRows;
+const filename = includePredictRows
+  ? 'product_data.parquet'
+  : 'product_data_no_predict.parquet';
+
+const outputPath = join(import.meta.dirname, filename);
+
+parquetWriteFile({
```
marikaner: [nth] Just an idea: maybe it would be nice to encapsulate this further so that we export a data-independent function. That would a) make it reusable, b) make it easier to change the data we have here.

davidkna-sap: done

KavithaSiva (approved):
LGTM

PR #1950, opened 2026-06-22
feat: Add cache control support in langchain orchestration client

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#575.


.github/workflows/build.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       - name: Set NODE_PATH
         run: echo "NODE_PATH=$(pwd)/node_modules" >> "$GITHUB_ENV"
       - name: Check public API
-        uses: sap/cloud-sdk-js/.github/actions/check-public-api@main
+        uses: sap/cloud-sdk-js/.github/actions/check-public-api@extend-generator-ts-options
```
davidkna-sap, line 65: To be reverted after merge of https://github.com/SAP/cloud-sdk-js/pull/6694

packages/langchain/package.json
```diff
@@ -20,6 +20,10 @@
       "types": "./dist/index.d.ts",
       "default": "./dist/index.js"
     },
+    "./orchestration/prompt-caching-middleware": {
```
davidkna-sap, line 23: I added this as explicit export to avoid having to dynamically import `langchain` as this would have also required some manual type definitions.

packages/langchain/src/orchestration/prompt-caching-middleware.ts
davidkna-sap, line 1: We could also reasonably re-implement what langchain currently does, but if it becomes smarter in upstream this way we don't have to maintain this more agentic capability ourselves.

PR #1960, opened 2026-06-24
chore: remove peer dependencies from sample-code

marikaner:
This fixes the bump. Not sure why or whether this is needed.


davidkna-sap (approved):
I think this is fine as a temporary fix, but this will likely break the smoke test set-up.

marikaner:
> I think this is fine as a temporary fix, but this will likely break the smoke test set-up.

Seems to work, but not sure if it really picks the canary version: https://github.com/SAP/ai-sdk-js/actions/runs/28080184186

PR #1964, opened 2026-06-25
fix(orchestration): handle Anthropic content-block arrays in getDeltaContent()

### LINKED ISSUE #1963: bug(orchestration): getDeltaContent() returns empty string for Anthropic Claude streaming chunks
## Describe the Bug

When using Anthropic Claude models via the Orchestration API in streaming mode, `OrchestrationStreamChunkResponse.getDeltaContent()` returns `undefined` (treated as `''`) for every chunk, causing `handleLLMNewToken` to fire with empty strings. No text is ever surfaced to the LangChain streaming caller.

## Root Cause

`ChatDelta.content` is typed as `string` in the generated schema (`packages/orchestration/src/client/api/schema/chat-delta.ts`):

```ts
export type ChatDelta = {
  content: string;
  // ...
};
```

`getDeltaContent()` returns it directly:

```ts
getDeltaContent(choiceIndex = 0): string | undefined {
  return this.findChoiceByIndex(choiceIndex)?.delta.content;
}
```

However, Anthropic Claude models return `delta.content` as `Array<{type: string, text: string}>` (content blocks) in streaming chunks — not a plain string. The TypeScript type is wrong, and the method has no handling for the array case. The result is that every streaming chunk from a Claude model is silently treated as an empty token.

## Steps to Reproduce

1. Configure `OrchestrationClient` with a Claude model (e.g. `claude-3-5-sonnet`)
2. Enable streaming (`{ streaming: true }` in `langchainOptions`)
3. Call `streamEvents({ version: "v3" })` or iterate `_streamResponseChunks`
4. Observe that all `handleLLMNewToken` callbacks fire with `content = ""`

## Expected Behavior

`getDeltaContent()` should extract and concatenate text from content-block arrays when `delta.content` is an array:

```ts
getDeltaContent(choiceIndex = 0): string | undefined {
  const content = this.findChoiceByIndex(choiceIndex)?.delta.content;
  if (Array.isArray(content)) {
    return content
      .filter(b => b.type === 'text')
      .map(b => b.text ?? '')
      .join('');
  }
  return content;
}
```

The `ChatDelta` type should also be widened:

```ts
export type ChatDelta = {
  content: string | Array<{ type: string; text?: string; [key: string]: any }>;
  // ...
};
```

## Actual Behavior

Every streaming chunk from Claude fires `handleLLMNewToken("")`. The full response is only available after stream completion via the non-streaming path.

## Confirmed Affected Version

`@sap-ai-sdk/langchain` 2.11.0 (latest at time of filing)

## Workaround

Downstream consumers can subclass `OrchestrationClient` and override `_streamResponseChunks` to extract text manually from `chunk._data`:

```ts
const rawContent =
  chunk._data?.intermediate_results?.llm?.choices?.[0]?.delta?.content ??
  chunk._data?.final_result?.choices?.[0]?.delta?.content;

const text = Array.isArray(rawContent)
  ? rawContent.filter(b => b.type === 'text').map(b => b.text ?? '').join('')
  : '';
```

This is the workaround we currently carry in our application. We'd like to see it fixed natively in the SDK.

madankumarpichamuthu:
## Summary

Fixes #1963.

Anthropic Claude models return `delta.content` as `Array<{type,text}>` (content blocks) in streaming chunks. `ChatDelta.content` was typed as `string`, so `getDeltaContent()` returned `undefined` for every Claude chunk — causing `handleLLMNewToken("")` to fire silently for each token. No text was surfaced to the LangChain streaming caller.

## Changes

**`packages/orchestration/src/spec/api.yaml`**
- Widen `ChatDelta.content` from `type: string` to `oneOf: [string, array of TextContent]` — consistent with how `ChatMessageContent` is already modelled

**`packages/orchestration/src/client/api/schema/chat-delta.ts`** (generated)
- Update to match spec: `content: string | TextContent[]`

**`packages/orchestration/src/orchestration-stream-chunk-response.ts`**
- `getDeltaContent()` now joins text blocks when `content` is an array; returns `undefined` for empty arrays (unchanged behaviour for non-array content)

**`packages/orchestration/src/util/stream.ts`**
- Extract `getDeltaText()` helper to coerce `string | TextContent[]` to `string`
- Use it in `mergeMessage()` and `transformStreamingChoice()` where `ResponseChatMessage.content` (`string | undefined`) is assigned

**`packages/orchestration/src/orchestration-stream-chunk-response.test.ts`**
- Two new unit tests: content-block array extraction (with mixed text/tool_use blocks) and empty-array → `undefined`

## Testing

```
Test Suites: 16 passed, 16 total
Tests:       1 skipped, 187 passed, 188 total
```

All existing tests pass. Two new tests added for the content-block array path.


madankumarpichamuthu:
Closing — raw trace confirms delta.content is always a plain string through SAP AI Core Orchestration. The Orchestration proxy normalizes all model responses to OpenAI format as expected. The fix is not needed. See issue #1963 for the trace evidence.

PR #1967, opened 2026-06-25
fix: Auto-route tool call results to messages_history

emincihangeri:
## Context

Closes SAP/ai-sdk-js-backlog#471.

Taken over from @emincihangeri by @InjunPark-sap


packages/orchestration/src/util/module-config.ts
```diff
@@ -358,16 +358,29 @@ export function constructCompletionPostRequest(
...
+  // 2. Tool results — their content comes from external systems and may contain {{?...}}
+  //    patterns that are not user-defined placeholders. We route through the last tool
+  //    message so the assistant+tool pair stays together for tool_call_id validation.
   const configs = Array.isArray(config) ? config : [config];
-  const routeMessagesToHistory = configs.some(c =>
+  const usesTemplateRef = configs.some(c =>
     isTemplateRef(c?.promptTemplating?.prompt || {})
   );
+  const messages = request?.messages || [];
+  const lastToolIndex = messages.reduce(
```
davidkna-sap: [pp] Use `findLastIndex()`

packages/orchestration/src/util/module-config.ts
```diff
@@ -389,18 +402,17 @@ export function constructCompletionPostRequest(
...
+    ...messages.slice(0, splitIndex)
+  ];
 
   return {
     config: configWithStream,
     ...(request?.placeholderValues && {
       placeholder_values: request.placeholderValues
     }),
-    ...(messagesHistory && {
+    ...(messagesHistory.length && {
```
davidkna-sap: [pp]
```suggestion
    ...((request?.messagesHistory || messagesHistory.length) && {
```

davidkna-sap (commented):
Just had a quick look, at a glance it mostly looks good.

packages/orchestration/src/util/module-config.ts
```diff
@@ -358,16 +358,29 @@ export function constructCompletionPostRequest(
...
-  // (the template lives remotely). Route them to messages_history instead.
+  // Determine where to split request.messages: everything before splitIndex is routed
+  // to messages_history (bypassing prompt templating); everything from splitIndex onward
+  // stays in messages and is merged into prompt.template.
+  //
+  // Two cases route messages out of templating:
+  // 1. TemplateRef configs — the template lives remotely, so messages cannot be merged in.
+  // 2. Tool results — their content comes from external systems and may contain {{?...}}
+  //    patterns that are not user-defined placeholders. We route through the last tool
+  //    message so the assistant+tool pair stays together for tool_call_id validation.
```
jerome-benoit: You are changing the initial intent and semantic of a content behind the scene. The user is not even aware of it and has no way to control it.  
What is the rationale behind such a debatable decision? 
How can you guarantee that a LLM will behave the same way if a content is moved from a designed container for an intent to another one with another intent? 
The whole point of an AI SDK is to interface with an API behind an structured abstraction and not to play with the specified structure of the abstraction.

davidkna-sap: Message history and prompt messages are joined and both sent to the LLM. The primary impact of moving the messages is that it changes how orchestration processes the messages.

jerome-benoit: > Message history and prompt messages are joined and both sent to the LLM. The primary impact of moving the messages is that it changes how orchestration processes the messages.

Where in the context window? The chronological ordering is altered or not? What is the diff of the context window sent to the LLM with or without it? The placement of content in the context window matters, really matters.  

Again, why the API is not fixed properly instead of forcing its consumer to workaround it? It's a simple boolean flag triggering two different code paths.

davidkna-sap: Assuming the split is done properly, the LLM context window will not change (for this PR I haven't done in-depth checks if this is the case here yet). If this is not the case we may have to re-evaluate doing this, and explore alternatives such as documentation changes, as the service rejected your proposed changes. I would also be hesistant to add escaping to avoid influencing the model with invisible unicode characters.

Anyway in some cases is already message history after consultation with the service.

jerome-benoit: > and explore alternatives such as documentation changes, as the service rejected your proposed changes. 

But I think it should be put again on the table by people working directly on AI tooling at SAP in a better way that could lead to a proper fix. 

> I would also be hesistant to add escaping to avoid influencing the model with invisible unicode characters.

I agree that inserting invisible unicode characters is an uglier workaround than moving the culprit content to  a place without template syntax interpretation if the moving it not altering the chronological content in the context window, or at least at a minimal level.

packages/orchestration/src/util/module-config.ts
emincihangeri: I tried to use findLastIndex, but it requires bumping our target to ES2023. I am not sure if it is a local issue.

davidkna-sap: Node 22 does support ES2023 (and more), but bumping it will require https://github.com/SAP/cloud-sdk-js/pull/6694 for check-public-api to pass.
Please try adding `"es2023.array"` or `'es2023'` wholly to the `tsconfig` `lib` key.

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     "target": "ES2022",
+    "lib": ["es2023", "dom"],
```
davidkna-sap:
```suggestion
    "lib": ["es2023"],
```

emincihangeri: this causes an error on weather-mcp-server.ts. should I work on it in this PR?

.changeset/yummy-ducks-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': patch
+---
+
+[Fix] Route `role: "tool"` messages from `messages` to `messages_history` automatically.
```
davidkna-sap:
```suggestion
[Fix] Route messages preceding tool messages from the `promptTemplating` section `messages` property to `messages_history` automatically if  no templating placeholder values are provided.
```

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -332,4 +334,33 @@ describe('orchestration', () => {
...
+      const response = await orchestrationToolResultInMessages();
+      expect(response.getContent()).toEqual(expect.any(String));
+
+      const templating = response.getIntermediateResults().templating;
+      expect(templating).toBeDefined();
+      const roles = templating!.map(m => m.role);
+      const toolIdx = roles.lastIndexOf('tool');
+      const userIdx = roles.lastIndexOf('user');
+      expect(toolIdx).toBeGreaterThan(-1);
+      expect(userIdx).toBeGreaterThan(toolIdx);
```
davidkna-sap: I would prefer something like this (might have gotten names wrong):

```ts
const roles = templating!.map(m => m.role);
expect(roles).toEqual(['assistant', 'tool', 'user']);
```

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -332,4 +334,29 @@ describe('orchestration', () => {
...
   });
+
+  describe('tool result prompt templating interference', () => {
+    it('should succeed when tool result containing {{?...}} syntax is sent via messages', async () => {
+      const response = await orchestrationToolResultInMessages();
+      expect(response.getContent()).toEqual(expect.any(String));
+
+      const templating = response.getIntermediateResults().templating;
+      const roles = templating!.map(m => m.role);
+      expect(roles).toEqual(['assistant', 'tool', 'user']);
```
davidkna-sap: [req] This test fails, please resove. The `prompt.template` with the system prompt is not being properly handled in the routing, and also consider disabling autorouting if something like that is provided.

https://github.com/SAP/ai-sdk-js/actions/runs/28664711694

davidkna-sap (changes requested):

tests/e2e-tests/src/orchestration.test.ts
InjunPark-sap: All failures are external 503 errors (AI-External-Failure: true) — not related to our changes. Our new tests passed. https://github.com/SAP/ai-sdk-js/actions/runs/28792702001/job/85375187656

packages/orchestration/src/util/module-config.ts
```diff
@@ -358,17 +358,39 @@ export function constructCompletionPostRequest(
...
+  const splitIndex = usesTemplateRef ? messages.length : lastToolIndex + 1;
+
+  const remainingMessages = messages.slice(splitIndex);
+  let moduleRequest = request;
+  if (splitIndex > 0 && request) {
+    moduleRequest = {
+      ...request,
+      messages: remainingMessages.length ? remainingMessages : undefined
+    };
+  }
```
davidkna-sap: [pp] Please extract the handling and detection of different conditions related to this into a separate function.

packages/orchestration/src/util/module-config.ts
```diff
@@ -358,17 +358,39 @@ export function constructCompletionPostRequest(
...
-  const routeMessagesToHistory = configs.some(c =>
+  const usesTemplateRef = configs.some(c =>
     isTemplateRef(c?.promptTemplating?.prompt || {})
   );
-
-  const moduleRequest =
-    routeMessagesToHistory && request
-      ? { ...request, messages: undefined }
-      : request;
+  const hasStaticPrompt = configs.some(c => {
```
davidkna-sap: [req] Please avoid splitting if any placeholder values are set.

.changeset/yummy-ducks-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': patch
+---
+
+[Fix] Route messages preceding tool messages from the `promptTemplating` section `messages` property to `messages_history` automatically if  no templating placeholder values are provided.
```
davidkna-sap: [req] Update to accurately reflect when this will trigger.

packages/orchestration/src/util/module-config.ts
```diff
@@ -358,17 +358,39 @@ export function constructCompletionPostRequest(
...
+    ? -1
+    : messages.findLastIndex(msg => msg.role === 'tool');
+  const splitIndex = usesTemplateRef ? messages.length : lastToolIndex + 1;
+
+  const remainingMessages = messages.slice(splitIndex);
+  let moduleRequest = request;
+  if (splitIndex > 0 && request) {
+    moduleRequest = {
+      ...request,
+      messages: remainingMessages.length ? remainingMessages : undefined
```
davidkna-sap: [req] If prompt is defined, `prompt.template` may not be empty.

davidkna-sap (changes requested):

packages/orchestration/src/util/module-config.ts
InjunPark-sap: I will keep the `remaining.length ? remaining : undefined` check since the `usesTemplateRef` case always produces an empty `remaining`. Is your intent to remove this check entirely, or just to confirm the current handling is correct?

packages/orchestration/src/util/module-config.ts
InjunPark-sap: Extracted into `getMessageSplitIndex()`.

packages/orchestration/src/util/module-config.ts
InjunPark-sap: Done. Routing is skipped when `placeholderValues` is non-empty.

.changeset/yummy-ducks-run.md
InjunPark-sap: Updated to reflect the placeholder values condition.

tests/e2e-tests/src/orchestration.test.ts
InjunPark-sap: e2e test completed! https://github.com/SAP/ai-sdk-js/actions/runs/28802437550/job/85409114988

packages/orchestration/src/util/module-config.ts
```diff
@@ -358,17 +359,18 @@ export function constructCompletionPostRequest(
...
-      : request;
+  const messages = request?.messages || [];
+  const splitIndex = getMessageSplitIndex(configs, messages, request);
+
+  let moduleRequest = request;
+  if (splitIndex > 0 && request) {
+    const remaining = messages.slice(splitIndex);
+    moduleRequest = {
+      ...request,
+      messages: remaining.length ? remaining : undefined
```
davidkna-sap: Only add messages key if `remaining.length`

packages/orchestration/src/util/module-config.ts
```diff
@@ -443,6 +445,53 @@ function buildCompletionModulesConfig(
...
+ * @returns The index at which to split messages between messages_history and prompt.template.
+ */
+function getMessageSplitIndex(
+  configs: OrchestrationModuleConfig[],
+  messages: ChatMessage[],
+  request?: ChatCompletionRequest
+): number {
+  const usesTemplateRef = configs.some(c =>
+    isTemplateRef(c?.promptTemplating?.prompt || {})
+  );
```
davidkna-sap: Prompt is only valid iff at least one message is provided: 

```ts
  const usesTemplate = configs.some(c => c?.promptTemplating?.hasOwnProperty('prompt'));
```

packages/orchestration/src/util/module-config.ts
```diff
@@ -443,6 +445,53 @@ function buildCompletionModulesConfig(
...
+
+  const hasStaticPrompt = configs.some(c => {
+    const prompt = c?.promptTemplating?.prompt;
+    return (
+      isTemplate(prompt) && (prompt.template?.length || prompt.tools?.length)
+    );
+  });
+  if (hasStaticPrompt) {
+    return 0;
+  }
```
davidkna-sap: Handled above:
```suggestion
```

packages/orchestration/src/util/module-config.ts
```diff
@@ -358,17 +359,19 @@ export function constructCompletionPostRequest(
...
+
+  let moduleRequest = request;
+  if (splitIndex > 0 && request) {
+    const remaining = messages.slice(splitIndex);
+    const { messages: _messages, ...rest } = request;
+    moduleRequest = {
+      ...rest,
+      ...(remaining.length && { messages: remaining })
+    };
+  }
```
davidkna-sap, line 366: [req] You also need to get rid of the handling `buildCompletionModulesConfig()` that defaults `prompt` to `{}`.

packages/orchestration/src/util/module-config.ts
```diff
@@ -358,17 +359,19 @@ export function constructCompletionPostRequest(
...
-  const routeMessagesToHistory = configs.some(c =>
-    isTemplateRef(c?.promptTemplating?.prompt || {})
-  );
-
-  const moduleRequest =
-    routeMessagesToHistory && request
-      ? { ...request, messages: undefined }
-      : request;
+  const messages = request?.messages || [];
+  const splitIndex = getMessageSplitIndex(configs, messages, request);
```
davidkna-sap: [q] Let's maybe simplify and pass everything to history if `prompt` and `placeholder_values` are not set. WDYT?

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -332,4 +334,29 @@ describe('orchestration', () => {
...
+    });
+
```
davidkna-sap, line 350: [req] Add e2e where the last message is of type tool.

.changeset/yummy-ducks-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': patch
+---
+
+[Fix] Automatically route `role: 'tool'` messages (and all preceding messages) from `messages` to `messages_history` to bypass prompt templating, unless placeholder values are provided.
```
davidkna-sap: Also update changeset as implementation changes.

davidkna-sap (changes requested):

packages/orchestration/src/util/module-config.ts
```diff
@@ -414,15 +416,16 @@ function buildCompletionModulesConfig(
...
   // prompt is not a string here as it is already parsed in `parseAndMergeTemplating` method
-  const prompt = {
-    ...(promptTemplating.prompt as Template | TemplateRef)
-  };
-
-  // If promptTemplating.prompt is not defined, we initialize it with an empty Template object
-  promptTemplating.prompt = promptTemplating.prompt || { template: [] };
+  const prompt: Template | TemplateRef = promptTemplating.prompt
+    ? { ...(promptTemplating.prompt as Template | TemplateRef) }
+    : { template: [] };
```
davidkna-sap: Not providing a template is allowed, so we shouldn't set a default if a user provides nothing related to templates.
We still have the issue that if the full message block is split and moved to message_history, it errors here (whether at runtime or in the backend server validation).

packages/orchestration/src/util/module-config.ts
davidkna-sap: Please avoid resolving questions without comment.

packages/orchestration/src/util/module-config.ts
```diff
@@ -443,19 +446,56 @@ function buildCompletionModulesConfig(
...
+    const p = c?.promptTemplating?.prompt;
+    return !!p && typeof p === 'object' && 'template_ref' in p;
+  });
+  if (usesTemplateRef) {
+    return messages.length;
+  }
+
+  const hasPlaceholderValues =
+    !!request?.placeholderValues &&
+    Object.keys(request.placeholderValues).length > 0;
```
davidkna-sap: [pp] styleguide
```suggestion
    Object.keys(request.placeholderValues).length;
```

davidkna-sap (changes requested):

packages/orchestration/src/util/module-config.ts
InjunPark-sap: Let's maybe simplify and pass everything to history if prompt and placeholder_values are not set. WDYT?

packages/orchestration/src/util/module-config.ts
InjunPark-sap: @davidkna-sap 
When `prompt` is not set and all messages move to `messages_history`, what should be sent to the service — omit the `prompt` key entirely from `prompt_templating`, or is there another expected structure?

packages/orchestration/src/util/module-config.ts
InjunPark-sap: @davidkna-sap 
I implemented the simplification you suggested: when no `prompt` property is set on the config, all messages are routed to `messages_history`. Is this the intended direction, or did you have a different approach in mind?

packages/orchestration/src/util/module-config.ts
davidkna-sap: Omit the prompt key, it's optional.

packages/orchestration/src/util/module-config.ts
davidkna-sap: I formulated it as a question, so this was not a requirement as-such. Of course I appreciate follow-up questions if something is unclear. (BTW template_ref would still need to go into messages_history as they have nowhere else to go in that case).

As implemented currently, you still need to ensure that the tool message search does not lead to an empty list being appended to the prompt section.

My main suggestion was to replace `getMessageSplitIndex()` with a function that returns a boolean that returns whether to route messages to messages_history e.g. `shouldRouteMessagsToHistory()`.
That would also mean not peeking at tool messages.

InjunPark-sap: I will consider it

InjunPark-sap:
@davidkna-sap 

- Replaced `getMessageSplitIndex` with `shouldRouteMessagesToHistory` (boolean); routes to `messages_history` when no `prompt` or `TemplateRef` is used, skips when `Template` or `placeholderValues` are set. No longer peeking at tool messages.
- Omits `prompt` key entirely when not configured.
- Fixed conflict with `cache_control` (from main): added `prompt ??= {}` in `mergeOrchestrationConfig` to keep messages in `prompt.template`.

packages/langchain/src/orchestration/client.ts
```diff
@@ -509,6 +509,12 @@ export class OrchestrationClient extends BaseChatModel<
       }
     }
 
+    // Ensure prompt exists when cache_control is used so messages are merged
+    // into prompt.template rather than routed to messages_history.
+    if (options.cache_control) {
```
davidkna-sap: As far as I can tell `messages_history` also supports `cache_control`

packages/orchestration/src/util/module-config.ts
```diff
@@ -436,26 +448,50 @@ function buildCompletionModulesConfig(
...
+ * - placeholder_values are provided (user has opted into templating).
+ * @param configs - The orchestration module configurations.
+ * @param request - The optional chat completion request.
+ * @returns True if all messages should be routed to messages_history.
+ */
+function shouldRouteMessagesToHistory(
+  configs: OrchestrationModuleConfig[],
+  request?: ChatCompletionRequest
+): boolean {
+  const hasPlaceholderValues =
```
davidkna-sap: Even if it has placeholder values, it needs to be send to messages_history in the case of `prompt_ref`

davidkna-sap (changes requested):

packages/orchestration/src/util/module-config.ts
```diff
@@ -436,26 +448,56 @@ function buildCompletionModulesConfig(
...
+  // placeholderValues means the user opted into templating — do not route.
+  if (
+    !!request?.placeholderValues &&
+    Object.keys(request.placeholderValues).length > 0
+  ) {
+    return false;
+  }
+
+  // No prompt configured — messages have nowhere to be merged into.
+  return configs.some(c => !c?.promptTemplating?.prompt);
```
davidkna-sap: [req] If any module has templating, disable.
```suggestion
  return  configs.every(c => !c?.promptTemplating?.prompt);
```

packages/orchestration/src/util/module-config.ts
```diff
@@ -389,11 +386,10 @@ export function constructCompletionPostRequest(
...
     : { modules: moduleConfigurations };
 
-  // When routing messages to history, append request.messages after messagesHistory
   const messagesHistory =
-    routeMessagesToHistory && request?.messages?.length
-      ? [...(request.messagesHistory || []), ...request.messages]
-      : request?.messagesHistory;
+    routeToHistory || request?.messagesHistory?.length
+      ? [...(request?.messagesHistory || []), ...(request?.messages || [])]
+      : undefined;
```
davidkna-sap: [req] If both `messagesHistory` and `messages` are provided, `messages` may be duplicated into `messagesHistory`.

davidkna-sap (changes requested):
Please also fix [E2E tests](https://github.com/SAP/ai-sdk-js/actions/runs/29344511026/job/87124568282) and review whether comments and change sets are still up-to-date.

InjunPark-sap:
This CodeQL failure (`v4.37.0 config vs v4.36.3 runner`) is a pre-existing issue in main unrelated to this PR — it was introduced before this branch was updated.

InjunPark-sap:
@davidkna-sap 
Your feedbacks are applied and also e2e test is passed: "https://github.com/SAP/ai-sdk-js/actions/runs/29415289767"
Please let me know if there is anything to do.

davidkna-sap (changes requested):
Please avoid large-scale changes in the final stages of the review changes without justification, as these may (re)introduce new issues and increases review burden.

I believe the issues raised on 780ad79f74ac1d8fa00e10f3841fea36c58fac18 could be solved with some minor adjustments. If you disagree I would be open to discuss this.

InjunPark-sap:
Replaced the `splitIndex` approach with a minimal extension of `780ad79`: extracted routing into `shouldRouteMessagesToHistory()` and added `role === 'tool'` detection to route all messages to `messages_history` when tool results are present.

Also fixed two issues found during review:

- `prompt.template: []` caused service rejection with `tools`-only configs — now removed when empty
- Message duplication when both `messages` and `messagesHistory` are provided — now only appended when routing is active

.changeset/yummy-ducks-run.md
davidkna-sap, line 1: [req] Make sure this is up to date.

packages/orchestration/src/util/module-config.ts
```diff
@@ -413,36 +408,96 @@ function buildCompletionModulesConfig(
...
+  if (configs.some(c => isTemplateRef(c?.promptTemplating?.prompt || {}))) {
+    return true;
+  }
+
+  // Skip routing when any config has prompt.tools — the service requires template
+  // to be present alongside tools, so routing would strip messages and break the request.
+  if (
+    configs.some(c => {
+      const prompt = c?.promptTemplating?.prompt;
+      return isTemplate(prompt) && !!(prompt as { tools?: unknown }).tools;
```
davidkna-sap:
```suggestion
      const prompt = c?.promptTemplating?.prompt;
      return isTemplate(prompt) || !!(prompt as { tools?: unknown }).tools;
```

packages/orchestration/src/util/module-config.ts
```diff
@@ -413,36 +408,96 @@ function buildCompletionModulesConfig(
...
+    })
+  ) {
+    return false;
+  }
+
+  return !!request?.messages?.some(m => m.role === 'tool');
```
davidkna-sap: [pp]
```suggestion
  return true;
```

packages/orchestration/src/util/module-config.ts
```diff
@@ -413,36 +408,96 @@ function buildCompletionModulesConfig(
...
+  if (
+    !!request?.placeholderValues &&
+    Object.keys(request.placeholderValues).length > 0
+  ) {
+    return false;
+  }
+
+  if (configs.some(c => isTemplateRef(c?.promptTemplating?.prompt || {}))) {
+    return true;
+  }
```
davidkna-sap: [req] Swap these.

packages/orchestration/src/util/module-config.ts
davidkna-sap: Regarding the requirement when tools are passed, we should consider making a request to lift the requirement.

tsconfig.base.json
davidkna-sap: [req] Remove duplicate `lib`, and it is now possible to replace this with `target: "ES2024"` fully.

davidkna-sap (changes requested):
Putting this on hold.

The auto-routing is skipped whenever prompt.tools is set, which is exactly the scenario that produces tool call results, so the fix is inert in the real tool-calling flow.

.changeset/yummy-ducks-run.md
InjunPark-sap: Updated

packages/orchestration/src/util/module-config.ts
InjunPark-sap: Applied. Also added `!!prompt &&` guard to prevent TypeError when `prompt` is `undefined`.

tsconfig.base.json
InjunPark-sap: @davidkna-sap 
Tried `target: "ES2024"` but the `check-public-api` CI tool doesn't recognize ES2024 and falls back to ES2021, causing compilation errors in `packages/core`. Used `target: "ES2022"` with `lib: ["es2024"]` as a compromise — removes the duplicate `lib` and enables ES2024 type definitions. Let me know if there's a preferred approach.

InjunPark-sap:
The current fix covers routing for the no-prompt and TemplateRef cases, but you're right that it doesn't address the core tool-calling scenario where `prompt.tools` is set — routing is skipped there to avoid the service's `template` required constraint.

Looking forward to the service-side requirement being lifted so we can extend routing to cover that case as well.

PR #1970, opened 2026-06-26
chore: add release environment to workflows

BrigittaK307:
Added environment tag to the release workflows.


marikaner (approved):
nice 👍 can you also remove the auth.yml and your test workflow?

BrigittaK307:
> nice 👍 can you also remove the auth.yml and your test workflow?

I have removed the test workflow in this commit: e657fc0847b08a72b103af1322fdd60b154e5fa5.
The `auth.yml` is also used in the spec-update-cleanup.yaml, that's why I only removed its usage from the `bump.yml`.

davidkna-sap:
> > nice 👍 can you also remove the auth.yml and your test workflow?
> 
> I have removed the test workflow in this commit: [e657fc0](https://github.com/SAP/ai-sdk-js/commit/e657fc0847b08a72b103af1322fdd60b154e5fa5). The `auth.yml` is also used in the spec-update-cleanup.yaml, that's why I only removed its usage from the `bump.yml`.

I think that can actually use [this instead](https://github.blog/changelog/2026-06-18-control-who-and-what-triggers-github-actions-workflows/).

marikaner:
> > > nice 👍 can you also remove the auth.yml and your test workflow?
> > 
> > 
> > I have removed the test workflow in this commit: [e657fc0](https://github.com/SAP/ai-sdk-js/commit/e657fc0847b08a72b103af1322fdd60b154e5fa5). The `auth.yml` is also used in the spec-update-cleanup.yaml, that's why I only removed its usage from the `bump.yml`.
> 
> I think that can actually use [this instead](https://github.blog/changelog/2026-06-18-control-who-and-what-triggers-github-actions-workflows/).

Good idea!

PR #1980, opened 2026-06-29
feat: allow setting model(-deployment) in requests

marikaner:
## Context

Closes SAP/ai-sdk-js-backlog#596.
This PR is currently based on #1978.

- change `WithoutModel` to `WithOptionalModel` to allow setting a model
- extend `AzureOpenAI` with an additional class
- override `buildRequestOptions`

## Edits

I noticed a few issues with how resource groups were handled and fixed them accordingly.
Also, now the model on the request will not trigger deployment resolution, if it is the same as on the client.


packages/openai/src/client.ts
```diff
@@ -36,16 +36,16 @@ export class SapOpenAi {
    * ```
    */
   static async createClient(options: SapOpenAiInput): Promise<SapOpenAi> {
```
davidkna-sap, line 38: [pp] Consider making the model optional, not sure how Azure handles this on the type level with optional models.

packages/openai/src/azure-openai.ts
```diff
@@ -0,0 +1,76 @@
...
+  return false;
+}
+
+// Implementation copied from https://github.com/openai/openai-node/blob/71d2347c59007124e7b317db0205b98a69d747ee/src/internal/utils/values.ts#L36-L38
+function isObj(obj: unknown): obj is Record<string, unknown> {
+  return obj != null && typeof obj === 'object' && !Array.isArray(obj);
+}
+
+// Implementation copied and modified based on https://github.com/openai/openai-node/blob/71d2347c59007124e7b317db0205b98a69d747ee/src/azure.ts#L156
+const _deployments_endpoints = new Set([
```
davidkna-sap: [q] I think for us this should include `/responses`? IIRC Azure doesn't require deploymentIds for Responses API.

packages/openai/src/azure-openai.test.ts
davidkna-sap, line 1: [req] E2E tests.

packages/openai/src/responses.ts
```diff
@@ -67,17 +80,17 @@ export class SapResponses {
...
     body: Params,
     options?: RequestOptions
   ): APIPromise<ParsedResponse<ParsedT>> {
     return this.openAiResponses.parse(
       {
-        // SAP AI Core routes via deployment URL; model is required by the SDK type but ignored by the API
         ...body,
-        model: ''
+        // SAP AI Core routes via deployment URL; model is required by the OpenAI SDK type but ignored by the API
+        model: body.model || ''
```
davidkna-sap, line 85: [q] Please check if this is needed per-client, it should already be optional: `model?: Shared.ResponsesModel`.

packages/openai/src/azure-openai.ts
marikaner: Agreed, I have a todo for it already, just wasn't sure if there is more than just `/responses`

davidkna-sap: I think for AI Core everything is a deployment endpoint anyway.

packages/openai/src/client.ts
marikaner: I am not sure, what you are referring to. The new type is `WithOptionalModel`, but that's not what you mean, right?

davidkna-sap: I mean that making `SapOpenAiInput` and the deployment/model name in `createClient()` optional. For `AzureOpenai` `opts.deployment` can be `undefined`.

packages/openai/src/azure-openai.ts
marikaner: I now wonder, does this even make sense for us? Everything is a deployment for us, right? So maybe I just remove that alltogether?

packages/openai/src/azure-openai.test.ts
marikaner: done

packages/openai/src/azure-openai.ts
marikaner: Haha, I should have reloaded before answering. ;)

packages/openai/src/azure-openai.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+
+      const { model } = options.body;
+      if (isModelDeployment(model)) {
+        const resourceGroup = getResourceGroup(model) ?? this.resourceGroup;
+        const deploymentUrl = await resolveDeploymentUrlForModel(model, {
+          scenarioId: 'foundation-models',
+          executableId: 'azure-openai',
+          resourceGroup,
+          destination: this.destination
```
marikaner, line 53: @KavithaSiva @davidkna-sap We could consider allowing to set the destination per request as well, if I didn't miss anything.

packages/openai/src/azure-openai.ts
marikaner: Removed.

packages/openai/src/client.ts
marikaner: Ah ok, do we want that? I guess if you want to use the per request approach, you might not be interested in the per client approach... On the other hand, when you set don't set it on the client, you HAVE yo set it on the request. Not sure this can be reflected (nicely) in the types...

packages/openai/src/azure-openai.ts
davidkna-sap: I am not sure if that would be worth it and I am not sure how easy it would be to replace the authentication token at that stage, but if implemented it shouldn't cause too many issues.

 Additionally it looks like the deployment cache is currently not multi-destination-aware.

packages/openai/src/client.ts
davidkna-sap: I agree that it doesn't fit well into type-restrictions but it would move the whole API surface closer to what upstream (documentation) expects.

packages/openai/src/azure-openai.ts
marikaner: Then I think it is not worth it. I didn't know about the cache. Thanks for clearing that up. 👍

packages/openai/src/client.ts
marikaner: I will try to add this on a separate PR 👍

packages/openai/src/azure-openai.test.ts
```diff
@@ -0,0 +1,209 @@
...
+          messages: []
+        }
+      });
+
+      expect(result.req.headers.get('ai-resource-group')).toBe(
+        customResourceGroup
+      );
+    });
+
+    it('does not send ai-resource-group header when a per-request model is used', async () => {
```
davidkna-sap, line 178: [req] Requests without `ai-resource-group` will fail.

sample-code/src/openai.ts
```diff
@@ -37,6 +37,39 @@ export async function chatCompletionStream(): Promise<
...
+export async function chatCompletionPerRequestModel(): Promise<string | null> {
+  const client = await SapOpenAi.createClient({
+    deployment: 'gpt-5.4'
+  });
+  const response = await client.chat.completions.create({
+    model: 'gpt-5.4-nano',
+    messages: [{ role: 'user', content: 'What is the capital of France?' }]
+  });
+
+  return response.choices[0].message.content;
```
davidkna-sap, line 70: [pp] also return `response.model`

packages/openai/src/azure-openai.test.ts
marikaner: `buildRequest` only builds the diff for the given request. When there is no new resource group per request, it should not be in the built request. Maybe I could improve the test titles?

davidkna-sap: It seemed unusual to test it this way at the network level, I think it would be fine to add a comment here.

KavithaSiva (dismissed):
LGTM, would be nice if we can ship this by next week. Then can I also showcase this version in the AI Core meeting next week.

davidkna-sap (dismissed):

KavithaSiva (approved):
LGTM

PR #1991, opened 2026-07-01
chore: rename tsconfig files for VSCode type discovery

marikaner:
Currently when working with jest and vscode none of the jest globals are found, which makes it hard to work with the tests.
I didn't find a way to set the tsconfig directly, but given the new typecheck I have the impression that this way around it is more correct anyways.

Changes: 
* rename `tsconfig.json` => `tsconfig.base.json`
* rename `tsconfig.test.json` => `tsconfig.json`


tsconfig.json
```diff
@@ -1,15 +1,10 @@
...
-    "esModuleInterop": true,
-    "forceConsistentCasingInFileNames": true,
-    "strict": true,
-    "skipLibCheck": true,
-    "isolatedModules": true,
-    "types": ["node"]
-  }
+    "rootDir": ".",
+    "noEmit": true,
+    "types": ["node", "jest"]
```
davidkna-sap, line 6: The reason I didn't enable it in the first place is that e.g. the type tests were accidentally using jest globals for some time. We should consider using `@jest/globals` everywhere.

marikaner: What are the implications of using jest globals in the type tests? And do you remember why this happend? This configuration should not bleed into the type tests unless something else is wrong, right?

davidkna-sap: It didn't actually break the type tests - the type tests used `expect<Type>(Something)` which still makes failures obvious in the VSCode highlighting, but I haven't verified if `tsd` would recognise potential errors in that case.
It's probably an acceptable risk either way, as this only leaked into those tests and smoke tests should catch any production code issues.

davidkna-sap (dismissed):

davidkna-sap (approved):

PR #2009, opened 2026-07-07
chore: Added logger warning and modified tests

BrigittaK307:
## Context

Closes [SAP/ai-sdk-js-backlog#438](https://github.com/SAP/ai-sdk-js-backlog/issues/438).

  Improve warnings for template and template_ref usage in orchestration clients


davidkna-sap (changes requested):
As per the issue, also warn issues if they use normal prompt templates (via the constructor) as these are unexpectedly pre-pended to every message.

In case users already do follow the recommendation of two separate clients, consider adding a way to hide the message, as `messages_history` is not easily accessible for langchain users.

BrigittaK307:
> As per the issue, also warn issues if they use normal prompt templates (via the constructor) as these are unexpectedly pre-pended to every message.
> 
> In case users already do follow the recommendation of two separate clients, consider adding a way to hide the message, as `messages_history` is not easily accessible for langchain users.

Right now I have implemented a `callCount` mechanism to detect the cases where warnings should appear. If you can think of and better suggestions to approach this.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -259,9 +264,35 @@ export class OrchestrationClient {
...
+    );
+    if (hasInlineTemplate) {
+      logger.warn(
+        'A prompt template is defined and messages are provided. The template will always be prepended to the messages on every request. ' +
+          'When reusing the same client across multiple turns, this causes the template to appear in every call. ' +
+          'To avoid duplication, use two separate clients: one with the template for the first turn, and one without for subsequent turns.'
+      );
+    }
+  }
+
```
InjunPark-sap, line 403: `warnInlineTemplateOnReuse` never fires via the LangChain client — `_generate` and `_streamResponseChunks` each create a new `OrchestrationClientBase` instance, so `this.callCount` is always 1 here.

Would it be probably Intentional? If so, a short comment noting LangChain users are covered by `warnTemplateUsage` in `langchain/client.ts` would help clarify the split.

packages/orchestration/src/util/module-config.ts
```diff
@@ -365,6 +365,13 @@ export function constructCompletionPostRequest(
     isTemplateRef(c?.promptTemplating?.prompt || {})
   );
 
+  if (routeMessagesToHistory && request?.messages?.length) {
+    logger.warn(
+      'Messages passed alongside a template_ref are sent as messages_history, not as part of the prompt template. ' +
+        'The prompt template is defined remotely and cannot be extended inline.'
+    );
+  }
```
InjunPark-sap, line 389: This fires a second time for LangChain users — `warnTemplateUsage` in `langchain/client.ts:70` already warns for the same case before reaching here. 

Consider removing this block; direct `OrchestrationClient` users are covered by the `isConfigReference` warn at `orchestration-client.ts:101`.

InjunPark-sap (commented):
some pps and questions

packages/orchestration/src/orchestration-client.ts
BrigittaK307: Yes, it is intentional. The LangChain adapter creates a fresh `OrchestrationClientBase` instance on every `_generate / _streamResponseChunks` call because LangChain's BaseChatModel contract treats each invocation as stateless — the base client is a local variable inside `caller.callWithOptions` and is discarded after each request. That means its `callCount` is always 1 and  `warnInlineTemplateOnReuse` can only ever emit the first-call info log there, never the reuse warning.

Reuse warnings are handled by the LangChain client's own `callCount` via `warnTemplateUsage` in `langchain/client.ts`.

  This split has now been made explicit with comments in both places.

InjunPark-sap (dismissed):

packages/orchestration/src/orchestration-client.ts
```diff
@@ -96,11 +97,15 @@ export class OrchestrationClient {
     }
   ): Promise<OrchestrationResponse> {
     requestConfig?.signal?.throwIfAborted();
+    this.callCount++;
```
davidkna-sap: It might be easier to handle this with a global or class-level boolean, also to avoid printing log messages at every point.
Also if user re-uses the same client multiple times to start the same agentic workflow they will get issues.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -96,11 +97,15 @@ export class OrchestrationClient {
...
     if (isConfigReference(this.config) && request?.messages?.length) {
-      logger.debug(
-        'Messages provided with an orchestration config reference will be sent as messages_history.'
+      logger.warn(
+        'Messages passed alongside an orchestration config reference are sent as messages_history, not as part of the prompt template. ' +
+          'The prompt template is defined remotely and cannot be extended inline. ' +
+          'In LangGraph workflows, consider using two separate clients: one with the config reference for the first node, and one without for subsequent conversational nodes.'
```
davidkna-sap: [pp] I would avoid mentioning langgraph in the base client.

```suggestion
          'In agentic workflows, consider using two separate clients: one with the config reference for the first node, and one without for subsequent conversational nodes.'
```

packages/orchestration/src/util/module-config.ts
davidkna-sap: If you don't remove this please handle `hasWarnedTemplateRef`/`hasWarnedInlineTemplate `

packages/orchestration/src/orchestration-client.test.ts
davidkna-sap, line 1: [req] Please fix the indentation/closing blocks in this file.

packages/langchain/src/orchestration/client.ts
```diff
@@ -439,6 +466,58 @@ export class OrchestrationClient extends BaseChatModel<
     }
   }
 
+  private warnTemplateUsage(
```
davidkna-sap, line 483: [pp] To reduce complexity consider, forwarding this to base client `warnInlineTemplateOnReuse` if feasible.

packages/orchestration/src/util/module-config.ts
InjunPark-sap, line 1: The warn here has no once-guard — it fires on every call. `constructCompletionPostRequest` is stateless so there's no instance flag to deduplicate against.

Also, `warnInlineTemplateOnReuse` in `orchestration-client.ts` explicitly excludes `template_ref` configs (`!('template_ref' in prompt)`), so removing this block without a fix would silently drop the warning for `template_ref` users entirely.

Suggested fix:
1. Remove this block.
2. In `warnInlineTemplateOnReuse`, drop the `!('template_ref' in prompt)` exclusion so it covers both inline template and `template_ref` — the existing `hasSeenInlineTemplateCall`/`hasWarnedInlineTemplateReuse` guards then handle deduplication correctly.

Note: LangChain adapter creates a fresh `OrchestrationClientBase` on every call, so the same leak affects LangChain users via this path too.

packages/orchestration/src/util/module-config.ts
```diff
@@ -366,6 +366,13 @@ export function constructCompletionPostRequest(
     isTemplateRef(c?.promptTemplating?.prompt || {})
   );
 
+  if (routeMessagesToHistory && request?.messages?.length) {
+    logger.warn(
```
InjunPark-sap, line 385: The `orchestration-client.ts` ConfigRef warning ends with actionable guidance:
> "In agentic workflows, consider using two separate clients..."

This message stops at the symptom with no next step. If you keep this warn (with a once-guard), please align the message to include the same guidance.

Resolved automatically if fix for the above is applied (warn moves into `warnInlineTemplateOnReuse` which already uses the full message text).

packages/langchain/src/orchestration/client.ts
```diff
@@ -455,6 +480,58 @@ export class OrchestrationClient extends BaseChatModel<
...
+    configs: LangChainOrchestrationModuleConfig[],
+    hasMessages: boolean
+  ): void {
```
davidkna-sap, line 486: [pp] Let's add some guards to avoid checking on every request.

```suggestion
  ): void {
  if (this.isFirstCall || (this.hasWarnedTemplateRef && this.hasWarnedInlineTemplate))
```

packages/langchain/src/orchestration/client.ts
```diff
@@ -455,6 +480,58 @@ export class OrchestrationClient extends BaseChatModel<
...
+    const hasTemplateRef = configs.some(
+      c =>
+        typeof c.promptTemplating.prompt === 'object' &&
+        isTemplateRef(c.promptTemplating.prompt)
+    );
```
davidkna-sap, line 499:
```suggestion
    );
    // To avoid rechecking if there is a TemplateRef next time.
    if (!hasTemplateRef) {
      this.hasWarnedTemplateRef = true;
    }
```

packages/langchain/src/orchestration/client.ts
```diff
@@ -455,6 +480,58 @@ export class OrchestrationClient extends BaseChatModel<
...
+      );
+    }
+
+    const hasInlineTemplate = configs.some(
+      c =>
+        typeof c.promptTemplating.prompt === 'object' &&
+        !isTemplateRef(c.promptTemplating.prompt) &&
+        Array.isArray((c.promptTemplating.prompt as any).template) &&
+        (c.promptTemplating.prompt as any).template.length > 0
+    );
```
davidkna-sap, line 522:
```suggestion
    );
    // To avoid rechecking if there is a inline template next time.
    if (!hasInlineTemplate) {
      this.hasWarnedInlineTemplate = true;
    }
```

packages/orchestration/src/orchestration-client.ts
```diff
@@ -338,9 +353,52 @@ export class OrchestrationClient {
...
+   *
+   * Note: when this client is used via the LangChain adapter, each _generate /
+   * _streamResponseChunks call constructs a fresh OrchestrationClient instance,
+   * so hasSeenInlineTemplateCall/hasWarnedInlineTemplateReuse are always false and the reuse warning never fires.
+   * The LangChain client tracks its own state and delegates to warnTemplateUsage
+   * in langchain/src/orchestration/client.ts — that is the intentional split.
+   * @param request - The chat completion request to check.
    */
+  private warnInlineTemplateOnReuse(request?: ChatCompletionRequest): void {
+    if (!request?.messages?.length) {
```
davidkna-sap: Perf:
```suggestion
    if (!request?.messages?.length || this.hasWarnedInlineTemplateReuse)
```

packages/orchestration/src/orchestration-client.ts
```diff
@@ -338,9 +353,52 @@ export class OrchestrationClient {
...
+    const hasInlineTemplate = configs.some(
+      c =>
+        c.promptTemplating.prompt &&
+        typeof c.promptTemplating.prompt === 'object' &&
+        !('template_ref' in c.promptTemplating.prompt) &&
+        Array.isArray((c.promptTemplating.prompt as any).template) &&
+        (c.promptTemplating.prompt as any).template.length > 0
+    );
+    if (!hasInlineTemplate) {
+      return;
```
davidkna-sap, line 387:
```suggestion
    // To enable short circuit above.
    if (!hasInlineTemplate) {
      this.hasWarnedInlineTemplateReuse = true;
      return;
```

davidkna-sap (commented):
LGTM, I have made some suggestions to enable early-exits, to avoid this having any impact on overall performance.

davidkna-sap (approved):

PR #2012, opened 2026-07-07
feat(langchain): add cache_control support and richer token usage met…

davidkna-sap:
## Context

…adata

Split from https://github.com/SAP/ai-sdk-js/pull/1950
Part of https://github.com/SAP/ai-sdk-js-backlog/issues/575

- Add `cache_control` call option to `OrchestrationCallOptions` that automatically places a cache breakpoint on the last cacheable content block of the last message in both invoke and stream paths.
- Expose `cached_tokens` and `cache_creation_tokens` via LangChain's `usage_metadata.input_token_details` for orchestration responses.
- Extract `buildUsageMetadata` helper for non-streaming and streaming token usage mapping; add `response_metadata.tokenUsage` to match LangChain conventions.

### DIFF
--- a/packages/langchain/src/orchestration/util.ts
+++ b/packages/langchain/src/orchestration/util.ts
@@ -360,7 +360,7 @@ function buildUsageMetadata(usage: TokenUsage): {
     input_tokens: usage.prompt_tokens,
     output_tokens: usage.completion_tokens,
     total_tokens: usage.total_tokens,
-    ...(usage.prompt_tokens_details && {
+    ...(usage.prompt_tokens_details?.cached_tokens !== undefined && {
       input_token_details: {
         cache_read: usage.prompt_tokens_details.cached_tokens ?? 0,
         cache_creation: usage.prompt_tokens_details.cache_creation_tokens ?? 0
--- a/packages/langchain/src/orchestration/util.ts
+++ b/packages/langchain/src/orchestration/util.ts
@@ -12,6 +12,7 @@ import type {
 } from '@sap-ai-sdk/orchestration';
 import type {
   AssistantChatMessage,
+  CacheControl,
   ChatCompletionTool,
   ChatMessage,
   ChatMessageContent,
@@ -19,6 +20,7 @@ import type {
   FunctionObject,
   MessageToolCalls,
   SystemChatMessage,
+  TokenUsage,
   ToolChatMessage,
   UserChatMessage,
   TemplateRef,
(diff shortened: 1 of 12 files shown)


packages/langchain/src/orchestration/util.ts
```diff
@@ -263,6 +322,36 @@ function mapOrchestrationToLangChainToolCallChunk(
...
+    ...(usage.prompt_tokens_details && {
+      input_token_details: {
+        cache_read: usage.prompt_tokens_details.cached_tokens ?? 0,
+        cache_creation: usage.prompt_tokens_details.cache_creation_tokens ?? 0
+      }
+    }),
+    ...(usage.completion_tokens_details && {
+      output_token_details: {
+        reasoning: usage.completion_tokens_details.reasoning_tokens ?? 0
+      }
```
InjunPark-sap: When `completion_tokens_details `exists but `reasoning_tokens` is absent, this emits `{ reasoning: 0 }`, which may signal reasoning capability to downstream consumers. Consider only including `output_token_details` when `reasoning_tokens` is actually present.

packages/langchain/src/orchestration/util.ts
```diff
@@ -228,6 +230,63 @@ export function mapLangChainMessagesToOrchestrationMessages(
...
+    if (lastMessage.role === 'system' || lastMessage.role === 'tool') {
+      (lastMessage as SystemChatMessage | ToolChatMessage).content = [
+        {
+          type: 'text',
+          text: lastMessage.content,
+          cache_control: cacheControl
+        }
+      ];
+      return;
+    }
```
InjunPark-sap:
```suggestion
    if (lastMessage.role === 'system' || lastMessage.role === 'tool' || lastMessage.role === 'developer') {
      (lastMessage as SystemChatMessage | ToolChatMessage | DeveloperChatMessage).content = [
        {
          type: 'text',
          text: lastMessage.content,
          cache_control: cacheControl
        }
      ];
      return;
    }
```

InjunPark-sap: `ChatMessage` union includes `DeveloperChatMessage` (`role: 'developer'`), but the string-content branch only checks `'system' | 'tool' | 'user'`. When a developer message with string content is the last message, it falls through to the final `return` silently — no `cache_control` is applied and the request is sent without a cache breakpoint.

packages/langchain/src/orchestration/util.ts
```diff
@@ -228,6 +230,63 @@ export function mapLangChainMessagesToOrchestrationMessages(
...
+  const block = lastMessage.content.findLast(
+    b =>
+      b &&
+      typeof b === 'object' &&
+      (b.type === 'text' ||
+        (isUserMessage && (b.type === 'image_url' || b.type === 'file')))
+  );
+  if (block) {
+    (block as { cache_control?: CacheControl }).cache_control = cacheControl;
+  }
```
InjunPark-sap, line 307:
```suggestion
  const isUserMessage = lastMessage.role === 'user';
  lastMessage.content = [...lastMessage.content];
  const block = lastMessage.content.findLast(
    b =>
      b &&
      typeof b === 'object' &&
      (b.type === 'text' ||
        (isUserMessage && (b.type === 'image_url' || b.type === 'file')))
  );
  if (block) {
    (block as { cache_control?: CacheControl }).cache_control = cacheControl;
  }
```

InjunPark-sap: Bug: `mapSystemMessageToOrchestrationSystemMessage` returns `content: message.content` without copying, so `block` points directly into the caller's original `BaseMessage` object. The `cache_control` assignment leaks back into the caller's LangChain message. If the same `messages` array is reused across calls — common in chat history scenarios — subsequent `invoke()` calls that do not pass `cache_control` will still carry the breakpoint from the previous call.

InjunPark-sap (changes requested):

InjunPark-sap (approved):

PR #2013, opened 2026-07-07
feat(langchain): add orchestrationPromptCachingMiddleware

davidkna-sap:
## Context

Exports `orchestrationPromptCachingMiddleware()` from `@sap-ai-sdk/langchain/orchestration/prompt-caching-middleware`.

The middleware wraps the upstream `anthropicPromptCachingMiddleware` from `langchain` with an OrchestrationClient identity proxy so the upstream middleware's model-name guard passes, then delegates actual caching to the `cache_control` call option added in the previous PR.

The `langchain` package is added as an optional peer dependency.

Split from https://github.com/SAP/ai-sdk-js/pull/1950
Part of https://github.com/SAP/ai-sdk-js-backlog/issues/575


packages/langchain/src/orchestration/prompt-caching-middleware/prompt-caching-middleware.ts
davidkna-sap, line 1: We could also reasonably re-implement what langchain currently does, but if it becomes smarter in upstream this way we don't have to maintain this more agentic capability ourselves.

packages/langchain/src/orchestration/prompt-caching-middleware/prompt-caching-middleware.ts
```diff
@@ -0,0 +1,142 @@
+import { createLogger } from '@sap-cloud-sdk/util';
+import { z } from 'zod';
+import { createMiddleware, anthropicPromptCachingMiddleware } from 'langchain';
```
InjunPark-sap, line 2: `langchain` is an optional peer, but the static import is resolved at module load time. Any consumer who imports this entry point without `langchain` installed gets `ERR_MODULE_NOT_FOUND` with no indication that installing the optional peer fixes it.

InjunPark-sap: The entry point name makes the dependency implicit enough that this may be acceptable — I'll leave it to you to decide. At minimum, a `@throws` in the JSDoc noting that the import itself will fail without `langchain` would help.

packages/langchain/src/orchestration/prompt-caching-middleware.ts
```diff
@@ -0,0 +1,142 @@
+import { createLogger } from '@sap-cloud-sdk/util';
+import { z } from 'zod';
```
InjunPark-sap:
```suggestion
```

InjunPark-sap: `z` is only used for the local `contextSchema` block below. Once that block is removed (see next suggestion), this import becomes unused.

packages/langchain/src/orchestration/prompt-caching-middleware.ts
```diff
@@ -0,0 +1,142 @@
...
+  package: 'langchain',
+  messageContext: 'prompt-caching-middleware'
+});
+
+const contextSchema = z.object({
+  enableCaching: z.boolean().optional(),
+  ttl: z.enum(['5m', '1h']).optional(),
+  minMessagesToCache: z.number().optional(),
+  unsupportedModelBehavior: z.enum(['ignore', 'warn', 'raise']).optional()
+});
```
InjunPark-sap:
```suggestion

```

InjunPark-sap: This is a manual copy of langchain's internal `contextSchema`, but written in `zod v4` while `langchain@1.5.2` internally uses `zod/v3`. If langchain adds a new context field in a future version, this outer schema will silently strip it from `runtime.context` before the inner middleware sees it. Using `inner.contextSchema` directly (see next suggestion) keeps things in sync automatically.

packages/langchain/src/orchestration/prompt-caching-middleware.ts
```diff
@@ -0,0 +1,142 @@
...
+    name: 'OrchestrationPromptCachingMiddleware',
+    contextSchema,
+    wrapModelCall: (request, handler) => {
+      const inner = anthropicPromptCachingMiddleware(middlewareOptions);
+
+      if (!inner.contextSchema) {
+        throw new OrchestrationPromptCachingMiddlewareError(
+          'Failed to extract contextSchema from langchain anthropicPromptCachingMiddleware.'
+        );
+      }
```
InjunPark-sap:
```suggestion
export function orchestrationPromptCachingMiddleware(
  middlewareOptions?: OrchestrationPromptCachingMiddlewareConfig
): AgentMiddleware {
  const inner = anthropicPromptCachingMiddleware(middlewareOptions);

  if (!inner.wrapModelCall) {
    throw new Error(
      'anthropicPromptCachingMiddleware did not return a wrapModelCall hook. ' +
      'Please check your langchain version.'
    );
  }

  return {
    name: 'OrchestrationPromptCachingMiddleware',
    contextSchema: inner.contextSchema,
    wrapModelCall: (request, handler) => {
```

InjunPark-sap: 1. `inner` moves to factory level. So `middlewareOptions` never changes, so recreating it on every `wrapModelCall` is unnecessary.
2. `contextSchema: inner.contextSchema` replaces the local copy, eliminating the zod version mismatch and the schema drift risk described above. The `contextSchema` guard is replaced with a `wrapModelCall` guard, which is the field that's actually accessed with `!` assertion later.

InjunPark-sap (dismissed):

.github/workflows/build.yml
```diff
@@ -78,7 +78,7 @@ jobs:
             uses: sap/cloud-sdk-js/.github/actions/check-public-api@main
             with:
               force_internal_exports: 'false'
-              ignored_path_pattern: '.*?/client/.*?/schema|.*?/zod/.*?'
+              ignored_path_pattern: '.*?/client/.*?/schema|.*?/zod/.*?|.*?/prompt-caching-middleware.*?'
```
davidkna-sap:
```suggestion
              ignored_path_pattern: '.*?/client/.*?/schema|.*?/zod/.*?'
```

.changeset/short-waves-worry.md
```diff
@@ -0,0 +1,8 @@
+---
+"@sap-ai-sdk/langchain": minor
+---
+
+[feat] Export `orchestrationPromptCachingMiddleware()` middleware for the LangChain Orchestration client from `@sap-ai-sdk/langchain/orchestration/prompt-caching-middleware`.
```
marikaner: [pp] "Export" sounds like it was there before.

packages/langchain/src/orchestration/prompt-caching-middleware/prompt-caching-middleware.ts
```diff
@@ -0,0 +1,142 @@
...
+      const middleware = createMiddleware({
+        name: 'OrchestrationPromptCachingMiddleware',
+        contextSchema,
+        wrapModelCall: proxiedRequestHandler =>
+          parentMiddleware.wrapModelCall!(proxiedRequestHandler, req =>
+            handler({ ...req, model: realModel })
+          )
+      });
+
+      return middleware.wrapModelCall!(
```
marikaner: [pp] I prefer not to use ! if somehow possible. If not, I would like to have a short note why this is needed in the code.

marikaner (commented):
left a few comments, but generally looks good to me

.changeset/short-waves-worry.md
davidkna-sap: Rewrote the message.

packages/langchain/src/orchestration/prompt-caching-middleware/prompt-caching-middleware.ts
davidkna-sap: I've added additional comments/runtime checks.

marikaner (dismissed):

PR #2036, opened 2026-07-15
chore: add e2e tests for Qwen models via orchestration service

InjunPark-sap:
## Context

Closes SAP/ai-sdk-js-backlog#615.

- Verifies that models not listed in ChatModel can be used with the orchestration service by passing their name as a plain string. Three models available in the Canary landscape are tested for basic chat completion (all three), structured output, and function calling.
- Also adds a JSDoc note to LlmModelDetails.name clarifying that any string is accepted — not only the values listed in ChatModel.
- Out of scope: Adding the models to ChatModel for autocomplete (follow-up ticket) and image input (pending vision support confirmation in Canary).

## Definition of Done

- [x] Documentation updated
- [ ] (Optional) Release notes updated -->


tests/e2e-tests/src/orchestration.test.ts
davidkna-sap, line 1: Not sure we want this in the e2e tests, other models don't have it, otherwise maybe restrict to the flash model.

.changeset/qwen-models-orchestration.md
davidkna-sap, line 1: I think we can skip a changeset for this.

davidkna-sap (approved):

tests/e2e-tests/src/orchestration.test.ts
KavithaSiva: [q] Is there a reason why you picked flash model ?

KavithaSiva (approved):
LGTM.

tests/e2e-tests/src/orchestration.test.ts
InjunPark-sap: based on david's feedback, I assume Flash is the lightest variant in the Qwen family, which makes it appropriate for an e2e smoke test. (need @davidkna-sap consent)

Using a heavier model (plus or max) would add latency to CI without testing anything additional.

PR #2038, opened 2026-07-16
feat: Add OpenAI Realtime API support

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#625.

This is using the sub-path export method, which should be easy to refactor into a separate package if necessary.
I think either option could work, but I wanted to present how that would look like.

~~The demo sample application isn't working right now, will investigate.~~ E2E are green.


packages/openai/package.json
```diff
@@ -50,9 +58,17 @@
     "@sap-cloud-sdk/util": "^4.7.0"
   },
   "peerDependencies": {
-    "openai": "^6.45.0"
+    "openai": "^6.45.0",
+    "ws": "^8.20.0"
```
davidkna-sap, line 62: The main OpenAI package may optionally require in the future `ws` too if WebSocket connections to OpenAI models get support.

.github/workflows/build.yml
```diff
@@ -75,7 +75,7 @@ jobs:
       # Independent read-only checks — run them concurrently.
       - parallel:
           - name: Check public API
-            uses: sap/cloud-sdk-js/.github/actions/check-public-api@main
+            uses: sap/cloud-sdk-js/.github/actions/check-public-api@davidkna-sap/fix-subpath-public-api-check
```
davidkna-sap: Temporary requirement to support subpaths in the check

davidkna-sap:
```suggestion
            uses: sap/cloud-sdk-js/.github/actions/check-public-api@main
```

packages/openai/src/realtime/realtime-client.ts
```diff
@@ -0,0 +1,122 @@
...
+   * });
+   * ```
+   */
+  static async createClient(
+    options: SapOpenAiRealtimeInput
+  ): Promise<SapOpenAiRealtime> {
+    const { url, tokenProvider, resourceGroup, clientType } =
+      await createRealtimeContext(options);
+    const token = await tokenProvider();
+    return new SapOpenAiRealtime(url, token, resourceGroup, clientType);
```
InjunPark-sap: The token is fetched once at createClient time. For long-lived sessions (e.g., multi-turn voice conversations), what happens when the token expires?

Is there a reconnect / token-refresh strategy planned, or is this a known limitation for the experimental release?

packages/openai/src/realtime/realtime-client.ts
```diff
@@ -0,0 +1,122 @@
...
+        event = JSON.parse(data.toString());
+      } catch (err) {
+        this._onError(null, 'could not parse websocket event', err);
+        return;
+      }
+      this._emit('event', event);
+      if (event.type === 'error') {
+        this._onError(event);
+      } else {
+        // @ts-expect-error TS isn't smart enough to get the relationship right here
```
InjunPark-sap: What's the exact type mismatch behind the `@ts-expect-error` here?

Is this a limitation of the openai SDK's base class types, or something that could be resolved with a more precise cast?

packages/openai/src/realtime/realtime-client.ts
```diff
@@ -0,0 +1,122 @@
+import { OpenAIRealtimeEmitter } from 'openai/realtime/internal-base.js';
+import WebSocket from 'ws';
```
InjunPark-sap: `ws` is declared as an optional peer dependency, but it's imported unconditionally. 

This means the `/realtime` sub-path won't load in a browser at all. 

Is this intentionally Node.js-only? If so, should the README and `engines` field document the Node.js requirement explicitly?

- ref: [package.json](https://github.com/SAP/ai-sdk-js/pull/2038/changes#diff-840c47490796dab375e84692342a73c960ef8e8db8fa0adb16339b7099580387R66)

InjunPark-sap (commented):
some quick questions needed to be clear

packages/openai/src/realtime/realtime-client.ts
davidkna-sap: The readme documents this.

packages/openai/src/realtime/realtime-client.ts
davidkna-sap: This has been discussed in the PoC.
IIRC this was `globalThis.Websocket` vs `ws`, but `globalThis.WebSocket` will not work here.

packages/openai/src/realtime/realtime-client.ts
davidkna-sap: The realtime client should open a single long-lived connection.

sample-code/src/index.ts
```diff
@@ -108,3 +108,17 @@ export {
   responsesApiStateful,
   responsesApiMultiTurn
 } from './openai.ts';
+
+export {
+  realtimeTextToAudio,
+  realtimeAudioToAudio,
+  realtimeSpeechToSpeech,
```
KavithaSiva: [q] Is `realtimeSpeechToSpeech` export needed?

sample-code/package.json
KavithaSiva, line 27: [q] Shouldn't we also list `ws` as an explicit dependency?

KavithaSiva (dismissed):
Mostly looks good to me, added a few questions.

sample-code/package.json
davidkna-sap: I wanted to avoid pulling it in for users of `openai`-only and who wouldn't need it, I believe it's also optional for `openai`, that said it's also pulled in via e.g. langchain.
So we either have to:
- Move into extra package
- Make it explicit

I will make create a vote on Slack.

sample-code/src/index.ts
davidkna-sap: I've removed it.

packages/openai/src/realtime/realtime-client.ts
```diff
@@ -0,0 +1,122 @@
...
+    const { url, tokenProvider, resourceGroup, clientType } =
+      await createRealtimeContext(options);
+    const token = await tokenProvider();
+    return new SapOpenAiRealtime(url, token, resourceGroup, clientType);
+  }
+
+  /** The resolved WebSocket URL, including the `/v1/realtime` suffix. */
+  readonly url: URL;
+
+  private readonly socket: WebSocket;
```
KavithaSiva: [q] `socket` is private readonly here, whereas the [upstream variant](https://github.com/openai/openai-node/blob/v7.3.0/src/realtime/ws.ts#L15)(`OpenAIRealtimeWS`) keeps it public. The base `OpenAIRealtimeEmitter` only re-emits message- and error-derived events, so with the socket private a consumer can hear every server message and transport error, but has no way to observe a clean server-initiated close.

packages/openai/src/realtime/realtime-client.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    this.socket.on('message', (data: WebSocket.RawData) => {
+      let event: RealtimeServerEvent;
+      try {
+        event = JSON.parse(data.toString());
+      } catch (err) {
+        this._onError(null, 'could not parse websocket event', err);
+        return;
+      }
+      this._emit('event', event);
```
KavithaSiva: [suggestion] The try/catch only fires when the input is unparseable; a frame that's valid JSON
but not an event object (the literal null, a number, a bare string) parses fine, skips the catch, and reaches `event.type` below. 
For a null frame that throws a TypeError, and because it's thrown after the try/catch has exited, it escapes the listener as an unhandled exception.

KavithaSiva (commented):
Few more questions after review.

packages/openai/src/realtime/realtime-client.ts
davidkna-sap: I've made the socket public.

packages/openai/src/realtime/realtime-client.ts
davidkna-sap: I changed this to match upstream more.

davidkna-sap:
I pushed some minor refactoring for the sample-code to try to make SonarQube pass by reducing code duplication.

packages/openai/src/realtime/realtime-client.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  /** The underlying WebSocket connection. */
+  socket: WebSocket;
+
+  /** @internal — use {@link SapOpenAiRealtime.createClient} instead */
+  private constructor(
+    deploymentUrl: string,
+    token: string,
+    resourceGroup: string,
+    clientType?: string
```
marikaner: [pp] Not a fan of more than 3 parameters, even it if is an internal function.

packages/openai/src/realtime/realtime-client.ts
```diff
@@ -0,0 +1,126 @@
...
+   *   });
+   * });
+   * ```
+   */
+  static async createClient(
+    options: SapOpenAiRealtimeInput
+  ): Promise<SapOpenAiRealtime> {
+    const { url, tokenProvider, resourceGroup, clientType } =
+      await createRealtimeContext(options);
+    const token = await tokenProvider();
```
marikaner: [q] What happens if the token expires? Should the client have a reference to the tokenProvider? I am not sure how we handle this in the OpenAI client, but I believe the original AzureOpenAI client takes the tokenprovider, is something in that direction also applicable here?

packages/openai/src/realtime/realtime-client.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  /** The resolved WebSocket URL, including the `/v1/realtime` suffix. */
+  readonly url: URL;
+
+  /** The underlying WebSocket connection. */
+  socket: WebSocket;
+
+  /** @internal — use {@link SapOpenAiRealtime.createClient} instead */
+  private constructor(
```
marikaner: [q] I haven't checked this in detail, but is it not possible to extend the original [`OpenAIRealtimeWS`](https://github.com/openai/openai-node/blob/main/src/realtime/ws.ts) class?
[req] Also, naming-wise, why are we not adhering to the example of OpenAI? Aka. calling it `SapOpenAiRealtimeWs` + locating it under `/realtime/ws`?

marikaner: [q] For a feature like this one, I would have expected some decision document or ADR before implementation. Did I miss that? If not, it might be unnecessary now, though.

sample-code/package.json
marikaner: I think I missed the vote: I think it would be appropriate to follow OpenAI's lead: https://github.com/openai/openai-node/blob/928fa2badc0fc509a67dfe284c5c2076c09c7e7e/package.json#L68 => aka. make it a real dependency

marikaner (changes requested):
For me the biggest question is, why are we not extending the original OpenAIRealtimeWS client and should we use the Azure version of that?

packages/openai/src/realtime/realtime-client.ts
KavithaSiva: We didn't add an ADR, but after a PoC, the findings were listed [here](https://github.tools.sap/I771208/realtime-api-poc/blob/main/findings.md). But good point that we should have added the summary to our repo.

KavithaSiva: And for extending the `OpenAIRealtimeWS` class, the `buildRealtimeURL()` was not overridable and was appending query params, more details [here](https://github.tools.sap/I771208/realtime-api-poc/blob/main/findings.md#openai-sdk-openairealtimews)

packages/openai/src/realtime/realtime-client.ts
KavithaSiva: From my understanding, a WebSocket connection is different from an HTTP one. We only
  need the token once, during the handshake, to establish the connection. After that the
  socket stays open long-term and the token is never re-checked, so token expiry doesn't
  affect a live connection. A fresh token is only needed if we reconnect after the
  socket closes, which `createClient` handles by resolving a new one each time it's called.

davidkna-sap:
@marikaner I extend that class because we cannot override `buildRealtimeURL()` which adds unsupported query parameters which are not accepted by AI Core.

sample-code/package.json
davidkna-sap: It is optional there: https://github.com/openai/openai-node/blob/928fa2badc0fc509a67dfe284c5c2076c09c7e7e/package.json#L108-L110

davidkna-sap:
@marikaner @KavithaSiva I've found a way to simplify the client anyway and also created an upstream feature request to improve the situation:
- Simpler way to extend the client: #2149
- Feature request: [openai/openai-node#2305](https://www.github.com/openai/openai-node/issues/2305)

sample-code/package.json
marikaner: oh and what I linked is a devDependency, not a real dependency. Than this is the way to go in my opinion 😊

marikaner (approved):

KavithaSiva:
> For me the biggest question is, why are we not extending the original OpenAIRealtimeWS client and should we use the Azure version of that?

Solved by picking https://github.com/SAP/ai-sdk-js/pull/2149 instead.

packages/openai/README.md
```diff
@@ -110,6 +117,56 @@ const client = new AzureOpenAI(config);
 `createTokenProvider` returns an `azureADTokenProvider`-compatible function that resolves the bearer token from the AI Core destination on each call.
 It is used internally by `createOpenAiConfig`.
 
+### Realtime API (experimental)
+
+> [!WARNING]
```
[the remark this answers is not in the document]
KavithaSiva: Looks like a false positive.

packages/openai/README.md
```diff
@@ -110,6 +117,56 @@ const client = new AzureOpenAI(config);
...
+> Only WebSocket connections are supported and browser usage is not supported.
+
+> [!NOTE]
```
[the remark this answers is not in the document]
KavithaSiva: Other occurances in the repo also do not have space.

PR #2048, opened 2026-07-17
fix: Per model type fixes

marikaner:
## Context

Closes SAP/ai-sdk-js-backlog#630.


packages/openai/src/completions.ts
```diff
@@ -60,7 +60,7 @@ export class SapCompletions {
       {
         ...body,
         // SAP AI Core routes via deployment URL; model is required by the OpenAI SDK type but ignored by the API
-        model: body.model || ''
+        model: body.model as any
```
davidkna-sap, line 63: [q] What do you think about smuggling model as stringified JSON?

```suggestion
        model: JSON.stringify(body.model),
```

tests/type-tests/test/openai.test-d.ts
```diff
@@ -52,41 +52,185 @@ expectType<SapResponses>(client.responses);
...
+  client.embeddings.create({ input: 'hello' })
+);
+
+// --- SapResponses ---
+
+/**
+ * SapResponses.create accepts a model name string.
+ */
+expectAssignable<Promise<unknown>>(
+  client.responses.create({ model: 'gpt-4o', input: 'Hello' })
```
davidkna-sap: [pp] Avoid deprecated models, use e.g. `gpt-5.4-nano`, `gpt-5.5`

davidkna-sap (dismissed):

tests/type-tests/test/openai.test-d.ts
marikaner: done

packages/openai/src/completions.ts
marikaner: I think this would mask the fact that this is not a string and potentially make it harder to investigate failures. "as any" honestly acknowledges that there is a type mismatch.

davidkna-sap: Okay, we reconsider if OpenAI ever asserts that this should be a string.

davidkna-sap (approved):

PR #2049, opened 2026-07-17
chore: add .d.ts files to sonar exclusions

marikaner:
.d.ts files that are not under dist are our test files and do not require coverage. they ARE coverage 😊


davidkna-sap:
See also #2039 😄

davidkna-sap (approved):

PR #2055, opened 2026-07-20
chore: tiny refactoring

marikaner:
I noticed that this could be simplified. However, if `messagesHistory` was `undefined` it would now be `[]`.


davidkna-sap (approved):
LGTM, but this could have some perf impact.

PR #2057, opened 2026-07-20
feat(orchestration): support reasoning content in request/responses

InjunPark-sap:
## Context

Closes SAP/ai-sdk-js-backlog#616.

- Add typed reasoning content support to the Orchestration client. Adds `ReasoningBlock` schema, `reasoning_content` field to request/response/delta types, `getReasoningContent()` / `getDeltaReasoningContent()` convenience functions, and streaming merge logic.
- `getReasoningContent()` returns `string[]` (text only, signature hidden per Orchestration team decision). Full `ReasoningBlock[]` with signature flows through `getAllMessages()` for multi-turn use automatically.


InjunPark-sap:
Pending before merge: Awaiting Zhongpin's confirmation that getReasoningContent() returning string[] is aligned with the Java SDK design.

InjunPark-sap:
Added JSDoc documentation and sample code (`orchestrationReasoningContent`, `orchestrationReasoningContentStream`) as requested.

**Design decisions (from cross-team discussion)**:

- `getReasoningContent()` returns `string[]` (reasoning text only) — per Orchestration team's decision not to expose `signature` to users
- `signature` is preserved internally in `ReasoningBlock[]` and flows automatically through `getAllMessages()` for multi-turn use
- One `ReasoningBlock` always has exactly one `content` + one `signature` per the spec — multiple blocks per response is the array scenario (already confirmed by Nourhan)

**Still pending**:

- Nourhan (Java team) to share the model/prompt that triggered multiple blocks — will add a concrete test once received
- @ZhongpinWang to confirm Java SDK return type alignment

InjunPark-sap:
Added JSDoc documentation and sample code (`orchestrationReasoningContent`, `orchestrationReasoningContentStream`).

Also aligned `getReasoningContent()` return type with the Java SDK — now returns `string | undefined` (all reasoning blocks joined with a space) instead of `string[]`, per @ZhongpinWang  's confirmation.

packages/orchestration/src/util/stream.ts
```diff
@@ -207,6 +212,32 @@ function mergeToolCalls(
...
+        content: (merged[i].content ?? '') + (incoming[i].content ?? ''),
+        signature: incoming[i].signature
+          ? incoming[i].signature
+          : merged[i].signature
+      };
+    } else {
+      merged.push(incoming[i]);
+    }
+  }
+  return merged;
```
davidkna-sap: [pp] Simplify:
```suggestion
  if (!incoming?.length) return existing;
  if (!existing?.length) return incoming;

  const shared = existing.map((block, i) => ({
    content: (block.content ?? '') + (incoming[i].content ?? ''),
    signature: incoming[i].signature || block.signature
  }));
  return incoming.length > existing.length
    ? [...shared, ...incoming.slice(existing.length)]
    : shared;
```

packages/orchestration/src/orchestration-response.ts
```diff
@@ -79,10 +79,25 @@ export class OrchestrationResponse {
...
+   * Parses the orchestration response and returns the reasoning content generated by the model.
+   * Only available for models that support extended thinking (e.g. Anthropic Claude).
+   * To preserve reasoning context in multi-turn conversations, pass the result of {@link getAllMessages} as messagesHistory in the next request instead of using this method directly.
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns All reasoning blocks joined with a space, or undefined if not present.
+   */
+  getReasoningContent(choiceIndex = 0): string | undefined {
+    const blocks =
+      this.findChoiceByIndex(choiceIndex)?.message?.reasoning_content;
+    return blocks?.map(b => b.content ?? '').join(' ') || undefined;
```
davidkna-sap: [pp] Return empty string as-is.
```suggestion
    return blocks?.map(b => b.content ?? '').join(' ');
```

packages/orchestration/src/spec/api.yaml
davidkna-sap, line 1: [pp] Please match how the upstream spec specifies this (https://github.com/SAP/ai-sdk-js/pull/2057), or exclude it form this PR and target https://github.com/SAP/ai-sdk-js/pull/2057.

sample-code/src/orchestration.ts
```diff
@@ -1319,3 +1319,64 @@ export async function orchestrationCacheControl(): Promise<
...
+    }
+  });
+  return orchestrationClient.stream(
+    {
+      messages: [
+        { role: 'user', content: 'Explain step by step: what is 15 * 17?' }
+      ]
+    },
+    controller.signal
+  );
```
davidkna-sap, line 1391: [req] Call your helpers for reading reasoning information.

sample-code/src/orchestration.ts
```diff
@@ -1319,3 +1319,64 @@ export async function orchestrationCacheControl(): Promise<
...
+ * Demonstrates reasoning content support with streaming and Anthropic extended thinking.
+ * @param controller - The AbortController to cancel the stream.
+ * @returns The orchestration stream response.
+ */
+export async function orchestrationReasoningContentStream(
+  controller: AbortController
+): Promise<OrchestrationStreamResponse<OrchestrationStreamChunkResponse>> {
+  const orchestrationClient = new OrchestrationClient({
+    promptTemplating: {
+      model: { name: 'anthropic--claude-3-7-sonnet' }
```
davidkna-sap: [req] Use more recent models, this model is not in models types anymore.
Same issue in the other sample.

packages/orchestration/src/index.ts
```diff
@@ -89,5 +89,6 @@ export type {
   ChatCompletionTool,
   FunctionObject,
   Error as OrchestrationError,
-  Citation
+  Citation,
+  ReasoningBlock
```
davidkna-sap, line 93: [q] Verify, I think we wanted to avoid exposing types for this sort of thing.

```suggestion
```

.changeset/orchestration-reasoning-content.md
```diff
@@ -0,0 +1,8 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[feat] Add reasoning content support to the Orchestration client.
+`ReasoningBlock` type and `reasoning_content` field are added to request, response, and streaming delta types.
+`getReasoningContent()` and `getDeltaReasoningContent()` convenience functions return the reasoning text from model responses.
+Reasoning blocks can be supplied in `AssistantChatMessage.reasoning_content` for multi-turn conversations.
```
davidkna-sap: [req] Many models don't support user-supplied reasoning.
```suggestion
```

.changeset/orchestration-reasoning-content.md
```diff
@@ -0,0 +1,8 @@
+---
+'@sap-ai-sdk/orchestration': minor
+---
+
+[feat] Add reasoning content support to the Orchestration client.
+`ReasoningBlock` type and `reasoning_content` field are added to request, response, and streaming delta types.
+`getReasoningContent()` and `getDeltaReasoningContent()` convenience functions return the reasoning text from model responses.
```
davidkna-sap:
```suggestion
`reasoning_content` fields were added on response, streaming delta, and assistant message types.
The `getReasoningContent()` and `getDeltaReasoningContent()` convenience functions return the reasoning text from model responses.
```

davidkna-sap (changes requested):

packages/orchestration/src/spec/api.yaml
InjunPark-sap: @davidkna-sap the link in this comment seems to point to the current PR itself. Can you clarify which upstream spec you're referring to?

davidkna-sap: Sorry, it's #2002

sample-code/src/orchestration.ts
InjunPark-sap: Updated model to anthropic--claude-4.5-sonnet

sample-code/src/orchestration.ts
InjunPark-sap: Fixed, added getDeltaReasoningContent() call in the streaming sample.

packages/orchestration/src/index.ts
InjunPark-sap: @davidkna-sap  That makes sense — we're on board with not exposing it if that's the intent.

One concern though: `AssistantChatMessage` is publicly exported and has `reasoning_content?: ReasoningBlock[].` Without `ReasoningBlock` exported, users can't reference the type directly and would need workarounds like `NonNullable<AssistantChatMessage['reasoning_content']>[number]`.

Should we keep `ReasoningBlock` unexported and accept that, or would it make more sense to either export it or make `reasoning_content` internal? 
CC: @ZhongpinWang

packages/orchestration/src/orchestration-response.ts
davidkna-sap: @InjunPark-sap 
You reverted this, what is the rationale for this? Treating `''` and `undefined` would not match existing API contract - even if doesn't matter much overall.

In particular also consider something the following to avoid superfluous spaces:
```ts
blocks?.map(b => b.content).filter(Boolean).join(' ')
```

packages/orchestration/src/util/stream.ts
davidkna-sap: @InjunPark-sap You edited this, as to handle `existing.length > incoming.length`, but as a consequence the code no longer handles `incoming.length > existing.length`, which I would always expect to be the case.
Please verify the actual behaviour and implement a fix that handles differing lengths appropriately.

InjunPark-sap:
Meeting notes with Zhongpin:

- **Return type decision**: Reverting `getReasoningContent()` from `string | undefined` back to `string[] | undefined`. Returning the array directly is cleaner for users — they can see the block structure and join themselves if needed. No need to align with Java on this one.
- **Open investigation**: streaming multiple blocks: When a model calls a tool mid-reasoning and starts a second reasoning block, the behavior of the first block in the streaming array is unclear (placeholder `""`, `undefined`, or completed string?). Need to test with a tool-call request against `anthropic--claude-4.5-sonnet` to verify `mergeReasoningBlocks` handles this correctly. Will follow up with findings.

packages/orchestration/src/client/api/schema/reasoning-block.ts
```diff
@@ -0,0 +1,21 @@
...
+  /**
+   * The reasoning text generated by the model.
+   */
+  content?: string;
+  /**
+   * Encrypted signature for redacted reasoning blocks. Treat as opaque; pass back as-is in subsequent turns.
+   *
+   */
+  signature?: string;
+};
```
KavithaSiva, line 13: [req] Looks like you added JSDoc comments on top of the generated file, please revert them as they will be overridden anyway during the next generation.

[req] Could you also target your PR to #2002 ?

packages/orchestration/src/orchestration-response.ts
```diff
@@ -79,10 +79,25 @@ export class OrchestrationResponse {
...
   }
 
+  /**
+   * Parses the orchestration response and returns the reasoning content generated by the model.
+   * Only available for models that support extended thinking (e.g. Anthropic Claude).
+   * To preserve reasoning context in multi-turn conversations, pass the result of {@link getAllMessages} as messagesHistory in the next request instead of using this method directly.
+   * @param choiceIndex - The index of the choice to parse.
+   * @returns All reasoning blocks joined with a space, or undefined if not present.
+   */
+  getReasoningContent(choiceIndex = 0): string | undefined {
```
KavithaSiva: [req] I would be okay with `undefined` for also when reasoning contents are omitted. But, if you choose this option could you also document this behaviour in JSDoc.

sample-code/src/orchestration.ts
KavithaSiva: You still did not use helper methods you introduced in the code. Traverse the chunk and consume them using `getDeltaReasoningContent()`

sample-code/src/orchestration.ts
```diff
@@ -1319,3 +1319,65 @@ export async function orchestrationCacheControl(): Promise<
...
 }
+
+/**
+ * Demonstrates reasoning content support with Anthropic extended thinking.
+ * Logs the reasoning content and final answer, then performs a follow-up turn
+ * passing the full message history to preserve reasoning context.
+ * @param signal - An optional AbortSignal to cancel the request.
+ * @returns The orchestration service response from the follow-up turn.
+ */
+export async function orchestrationReasoningContent(
```
KavithaSiva, line 1330: [req] Wire up these functions in `server.ts` so that we have endpoints targeting these functions.

packages/orchestration/src/client/api/schema/reasoning-block.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright (c) 2026 SAP SE or an SAP affiliate company. All rights reserved.
+ *
+ * This is a generated file powered by the SAP Cloud SDK for JavaScript.
+ */
+
+/**
+ * A reasoning block containing the model's internal reasoning content. Applicable to models that support reasoning or extended thinking (e.g. Anthropic Claude). When present in a response, pass back verbatim in the next turn's AssistantChatMessage.reasoning_content to maintain reasoning context.
+ *
+ */
+export type ReasoningBlock = {
```
KavithaSiva, line 10: [q] Do you know or did you test if Orchestration service makes this works across all providers? Like anthropic, openai?

KavithaSiva (changes requested):

packages/orchestration/src/client/api/schema/reasoning-block.ts
davidkna-sap: I know that at least gemini should use this too.

packages/orchestration/src/util/stream.ts
InjunPark-sap: Fixed. Rewrote using `minLength` to merge the overlapping portion, then append the remaining blocks from whichever array is longer — handles both `existing > incoming` and `incoming > existing`.

packages/orchestration/src/client/api/schema/reasoning-block.ts
InjunPark-sap: Removed. Missed that this is a generated file — JSDoc would be overwritten on the next `pnpm generate`.

- Regarding targeting #2002: could you clarify — do you mean changing the base branch of this PR to #2002's branch, or just linking it as a related PR?

sample-code/src/orchestration.ts
InjunPark-sap: Added `orchestrationReasoningContent` endpoint (`reasoningContent`) to `server.ts`.

packages/orchestration/src/index.ts
InjunPark-sap: After reconsidering with zhongpin's meeting, keeping `ReasoningBlock` exported.

In our sync with Zhongpin, the core principle was "let users decide" — which is why we returned `string[]` instead of a joined string. Hiding the type while keeping `reasoning_content?: ReasoningBlock[]` public breaks that same principle: users can't even name the type they're working with.

Signature is already hidden at the convenience function level. The type itself doesn't expose anything unsafe.

packages/orchestration/src/client/api/schema/reasoning-block.ts
davidkna-sap: @InjunPark-sap Please target the PR's branch.

InjunPark-sap: Done — reverted the JSDoc to the generated default, and rebased onto `spec-update/orchestration/release-0.138.12` (#2002). The generated files (reasoning-block.ts, api.yaml schema ordering) now match #2002 exactly.

packages/orchestration/src/spec/api.yaml
InjunPark-sap: Fixed. `api.yaml` now follows the same field order as #2002 (items before description) for all three reasoning_content array blocks.

InjunPark-sap:
Confirmed with the Orchestration team (@Marc-Fabio): `reasoning_content` can contain multiple `ReasoningBlocks` in a single response. The array-based return type (`string[] | undefined`) and the `mergeReasoningBlocks` accumulation logic already handle this correctly.

packages/orchestration/src/orchestration-response.ts
InjunPark-sap: Done. The JSDoc now reads: "or undefined if no reasoning content is present."

sample-code/src/orchestration.ts
InjunPark-sap: Done — the streaming function now iterates the stream and calls `getDeltaReasoningContent()` on each chunk.

packages/orchestration/src/client/api/schema/reasoning-block.ts
InjunPark-sap: Confirmed with the Orchestration team (Marc-Fabio): the ReasoningBlock schema is harmonized across all providers. The signature field may be an empty string but is always present.

packages/orchestration/src/orchestration-response.ts
InjunPark-sap: After a sync with Zhongpin, the return type changed from `string | undefined` to `string[] | undefined` — returning an array preserves the 1:1 block structure needed for multi-turn (each string maps back to a `ReasoningBlock` entry in `AssistantChatMessage.reasoning_content`). 

The join suggestion was valid for the string approach, but is now superseded. Sorry for the back-and-forth while the design was being finalized.

InjunPark-sap:
Follow-up on the open investigation from the Zhongpin sync:

mergeReasoningBlocks handles all three cases regardless of what the first block's content looks like when a second block starts:
- If it arrives as "" (closed marker): concatenating "" has no effect on the accumulated content
- If it arrives as undefined: handled by the `?? ''` fallback
- If it arrives as a completed string: normal accumulation

The `incoming.length > existing.length` branch appends the new block via incoming.slice(minLength). The E2E streaming test with tool calls confirmed the accumulation works correctly end-to-end.

If the Java team can share the specific model/prompt that reliably triggers multiple blocks, I can add a concrete unit test with a mocked multi-block delta sequence.

ZhongpinWang:
Let me clarify few things.

1. Orchestration service / llm access service currently does not support server side tool calling. (But may be supported in the future).
2. Server side tool calling is mostly the reason for multiple reasoning blocks. Orchestration may support this in the future.
3. So we should consider the case that there can be multiple blocks to handle the future change.
4. Regarding your last comment, I was not saying that you should check "", undefined, or null etc in the block list. What I meant is that in the first delta you get [block_1, block_2], in the second delta you get [block_2], so no placeholder for block 1 at all. But it is hard to test at the moment with orchestration or any SAP proxied llm access. So we can assume at the moment that there will always be a placeholder.
5. For rare case, anthropic redacted reasoning can happen, but in that case, returning a string list is also fine. **Users should any way expect the reasoning text to be a list**.

As for this PR, the discussion regarding this point lasted a bit too long. I prefer we return a list of string for now and it is functionally correct. In case future changes, we can provide fix if server side tool is supported.

InjunPark-sap:
> Let me clarify few things.
> 
> 1. Orchestration service / llm access service currently does not support server side tool calling. (But may be supported in the future).
> 2. Server side tool calling is mostly the reason for multiple reasoning blocks. Orchestration may support this in the future.
> 3. So we should consider the case that there can be multiple blocks to handle the future change.
> 4. Regarding your last comment, I was not saying that you should check "", undefined, or null etc in the block list. What I meant is that in the first delta you get [block_1, block_2], in the second delta you get [block_2], so no placeholder for block 1 at all. But it is hard to test at the moment with orchestration or any SAP proxied llm access. So we can assume at the moment that there will always be a placeholder.
> 5. For rare case, anthropic redacted reasoning can happen, but in that case, returning a string list is also fine. **Users should any way expect the reasoning text to be a list**.
> 
> As for this PR, the discussion regarding this point lasted a bit too long. I prefer we return a list of string for now and it is functionally correct. In case future changes, we can provide fix if server side tool is supported.

Thanks for the detailed clarification.

Understood — the concern was about the delta array contracting (e.g. delta 1: [block_1, block_2], delta 2: [block_2] with no placeholder for block_1), not about the content values within each block. My previous comment addressed the wrong scenario.

For the current Orchestration service (no server-side tool calling), the placeholder assumption holds, so the minLength merge logic is safe. Noted that this will need revisiting when server-side tool calling is supported.

packages/orchestration/src/orchestration-stream-chunk-response.test.ts
```diff
@@ -153,4 +153,55 @@ describe('Orchestration chat completion stream chunk response', () => {
...
+    });
+
+    it('should return undefined when no reasoning blocks in delta', () => {
+      const chunkResponse = new OrchestrationStreamChunkResponse({
+        request_id: 'test-id',
+        final_result: {
+          id: 'test',
+          object: 'chat.completion.chunk',
+          created: 0,
+          model: 'gpt-4o',
```
davidkna-sap:
```suggestion
          model: 'gpt-5.4-nano',
```

packages/orchestration/src/orchestration-stream-chunk-response.test.ts
```diff
@@ -153,4 +153,55 @@ describe('Orchestration chat completion stream chunk response', () => {
...
+
+  describe('getDeltaReasoningContent', () => {
+    it('should return delta reasoning blocks when present', () => {
+      const chunkWithThinking = new OrchestrationStreamChunkResponse({
+        request_id: 'test-id',
+        final_result: {
+          id: 'test',
+          object: 'chat.completion.chunk',
+          created: 0,
+          model: 'claude-3-7-sonnet',
```
davidkna-sap:
```suggestion
          model: 'anthropic--claude-4.5-haiku',
```

sample-code/src/orchestration.ts
```diff
@@ -1319,3 +1319,79 @@ export async function orchestrationCacheControl(): Promise<
...
+    messagesHistory: response.getAllMessages()
+  });
+}
+
+/**
+ * Demonstrates reasoning content support with streaming and Anthropic extended thinking.
+ * Iterates over the stream and logs the delta reasoning content from each chunk.
+ * @param controller - The AbortController to cancel the stream.
+ */
+export async function orchestrationReasoningContentStream(
```
davidkna-sap, line 1371: [pp] Wire this up to the sample server.

sample-code/src/orchestration.ts
```diff
@@ -1319,3 +1319,79 @@ export async function orchestrationCacheControl(): Promise<
...
+      ]
+    },
+    { signal }
+  );
+
+  const reasoning = response.getReasoningContent();
+  if (reasoning) {
+    logger.info('Reasoning:', reasoning);
+  }
+  logger.info('Answer:', response.getContent());
```
davidkna-sap: This should be part of the response body for the sample-code server.

packages/orchestration/src/orchestration-stream-response.ts
```diff
@@ -136,10 +136,29 @@ export class OrchestrationStreamResponse<T> {
     return choice?.message?.refusal;
   }
 
+  /**
+   * Parses the orchestration stream response and returns the accumulated reasoning content generated by the model.
+   * Only available after the stream has been fully consumed.
```
davidkna-sap, line 141: [pp] Mention that this only returns the user-visible reasoning.
Also handle this like this for similar functions on the other classes.

sample-code/src/orchestration.ts
```diff
@@ -1319,3 +1319,79 @@ export async function orchestrationCacheControl(): Promise<
...
+ * Iterates over the stream and logs the delta reasoning content from each chunk.
+ * @param controller - The AbortController to cancel the stream.
+ */
+export async function orchestrationReasoningContentStream(
+  controller: AbortController
+): Promise<void> {
+  const orchestrationClient = new OrchestrationClient({
+    promptTemplating: {
+      model: {
+        name: 'anthropic--claude-4.5-sonnet',
```
davidkna-sap: [pp] Try using a diff model (provider) like `gemini-3.5-flash`

tests/e2e-tests/src/orchestration.test.ts
```diff
@@ -272,6 +272,78 @@ describe('orchestration', () => {
...
+    const reasoning = response.getReasoningContent();
+    // Reasoning + tool-call interleaving is non-deterministic; assert loosely.
+    if (reasoning) {
+      expect(reasoning.length).toBeGreaterThanOrEqual(1);
+      reasoning.forEach(block => expect(typeof block).toBe('string'));
+      // If delta reasoning was streamed, the accumulated result must reflect it.
+      if (deltaReasoningChunks.length > 0) {
+        expect(reasoning.join('')).not.toBe('');
+      }
+    }
```
davidkna-sap: Try removing the conditional tests by moving to e.g. a lower temperature, and change the prompt to think before calling the tool, as afaict the post-tool-call state is currently unhandled.

davidkna-sap (commented):
I have a few more nitpicks, but this mostly looks good.

tests/e2e-tests/src/orchestration.test.ts
davidkna-sap: Also consider using tool_choice if you really want a tool call.

sample-code/src/orchestration.ts
InjunPark-sap: After investigation, the `thinking: { type: 'enabled', budget_tokens }` param is Anthropic-specific — Orchestration passes model params through verbatim without normalizing the "enable reasoning" trigger across providers (Gemini uses thinkingConfig, OpenAI uses reasoning_effort). 

Passing the Anthropic shape to Gemini would fail, so the sample stays on an Anthropic model to remain runnable. 

if you would prefer it, I will add a provider-specific variant. :)

davidkna-sap: @InjunPark-sap, this is not quite true, it is harmonised: https://help.sap.com/docs/sap-ai-core/generative-ai/reasoning?locale=en-US#enabling-reasoning

InjunPark-sap: You're right, my mistake — I was looking at the provider-native shapes and missed that Orchestration exposes a harmonized reasoning_effort param that it maps to each provider's native format.

Switched both the sample and the e2e tests to `gemini-3.5-flash` with `params: { reasoning_effort: 'high' }`, and updated the JSDoc to reference `reasoning_effort` instead of Anthropic extended thinking. Verified the e2e tests pass against the live service with Gemini. 

Thanks for the pointer to the docs!

InjunPark-sap: Verified `reasoning_effort` across all three providers on the live service (non-streaming test, same `params: { reasoning_effort: 'high' }`, only model name changed):

| Provider  | Model                        | reasoning_content |
|-----------|------------------------------|-------------------|
| Gemini    | gemini-3.5-flash             | returned ✅       |
| Anthropic | anthropic--claude-4.5-sonnet | returned ✅       |
| OpenAI    | gpt-5.4                      | returned ✅       |

`reasoning_effort` is accepted by all three (harmonized, no 400s) and all three return reasoning_content. (An earlier test on `gpt-5` came back empty, but `gpt-5.4` exposes the reasoning blocks — it was a model-version quirk, not a provider limitation.)

Sample + e2e stay on gemini-3.5-flash.

tests/e2e-tests/src/orchestration.test.ts
InjunPark-sap: Good point — but forced `tool_choice (any / tool)` returns a 400 with extended thinking enabled `(type: 'enabled');` only auto is allowed. 

Since I can't guarantee a deterministic tool call that way, I dropped the tool from this test and kept it focused on what it actually verifies: reasoning content accumulation across the stream. 

The assertions are now unconditional. Interleaved (post-tool-call) reasoning would need the interleaved-thinking beta header, which is out of scope here.

packages/orchestration/src/orchestration-stream-response.ts
InjunPark-sap: Done

sample-code/src/orchestration.ts
InjunPark-sap: Done

sample-code/src/orchestration.ts
InjunPark-sap: Done

sample-code/src/orchestration.ts
davidkna-sap: `gpt-5.4` does use the reasoning blocks.

InjunPark-sap: Confirmed — retested with gpt-5.4 and it returns reasoning_content. Updated the table above. Thanks.

davidkna-sap (approved):

PR #2058, opened 2026-07-21
fix(langchain): skip cache_control for tool messages to prevent HTTP 400

### LINKED ISSUE #2047: Cache control fails for Orchestration service when last message is tool result
### Checklist

- [x] My issue is not an [AI Core service issue](https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/support-process), but an SDK issue.
- [x] My issue is a JS/TS related issue, not a [Java related issue](https://github.com/SAP/ai-sdk-java/issues).
- [x] My issue is not [already reported or documented](https://github.com/SAP/ai-sdk-js/issues).

### Describe the Bug

When adding cache_control to an langchain OrchestrationClient invocation, the call throws a 400 error if the last message in the message list is a tool result:

{"error":{"request_id":"2a273afa-b802-908b-85e7-816f29aea923","code":400,"message":"400 - LLM Module: Tool message content must be a string for Anthropic harmonization. Received: list."

However, the same call to a direct inference call of a claude 4.6 sonnet model deployment is successful. 

### Steps to Reproduce

0. Have AI Core credentials as environment variables and an instance of anthropic--claude-4.6-sonnet deployed in the AI Core account
1. Call the Orchestration service with a list of messages where the last one is a tool result (code below)
2. Call the deployed claude model directly with the same list of messages

### Expected Behavior

Both calls succeed and token usage shows some amount of tokens written to cache.

### Screenshots

_No response_

### Used Versions

- Node version: v25.6.1
- NPM version: 11.18.0
- SAP Cloud SDK for AI version: 2.13.0

### Code Examples

1. Call to Orchestration service fails
```ts
import { OrchestrationClient } from "@sap-ai-sdk/langchain";

// Claude 4.6 sonnet has a minimim of 4096 tokens for caching. Create a long system prompt
const system_prompt = "Roses are red. Violets are Blue.".repeat(512);

try {
	const client = new OrchestrationClient(
		{ promptTemplating: { model: { name: "anthropic--claude-4.6-sonnet" } } },
		{},
		{ resourceGroup: "default" },
	);

	const response = await client
		.bindTools([
			{
				type: "function",
				function: {
					name: "get_data",
					description: "Gets data",
					parameters: { type: "object", properties: {} },
				},
			},
		])
		.invoke(
			[
				{ role: "system", content: system_prompt },
				{ role: "user", content: "Call get_data and tell me what it returned" },
				{
					role: "assistant",
					content: "",
					tool_calls: [
						{ id: "tc1", name: "get_data", args: {}, type: "tool_call" },
					],
				},
				{ role: "tool", content: "the result", tool_call_id: "tc1" },
			],
			{ cache_control: { type: "ephemeral" } },
		);

	console.log("✓ PASS (unexpected)");
	console.log("Response:", response.content);
	console.log("Token usage:", response.usage_metadata);
} catch (e) {
	const err = e as Error & {
		cause?: Error & { response?: { data?: unknown } };
	};
	const sapError = err.cause?.response?.data;
	console.error("✗ FAIL:", sapError ? JSON.stringify(sapError) : err.message);
}
```

2. A direct inference call to a deployment of the same model succeeds:

```ts
export {}

// ─── auth ─────────────────────────────────────────────────────────────────────

const sk = JSON.parse(process.env.AICORE_SERVICE_KEY ?? "{}") as {
	clientid: string;
	clientsecret: string;
	url: string;
};
const basic = btoa(`${sk.clientid}:${sk.clientsecret}`);
const tokenRes = await fetch(
	`${sk.url}/oauth/token?grant_type=client_credentials`,
	{
		method: "GET",
		headers: { Authorization: `Basic ${basic}` },
	},
);
const { access_token: token } = (await tokenRes.json()) as {
	access_token: string;
};


// Claude 4.6 sonnet has a minimim of 4096 tokens for caching. Create a long system prompt
const system_prompt = "Roses are red. Violets are Blue.".repeat(512);

const INFERENCE_URL = `${process.env.AICORE_BASE_URL}/v2/inference/deployments/d859cd1b783479e3`;
const res = await fetch(`${INFERENCE_URL}/invoke`, {
	method: "POST",
	headers: {
		"Content-Type": "application/json",
		Authorization: `Bearer ${token}`,
		"AI-Resource-Group": "default",
	},
	body: JSON.stringify({
		anthropic_version: "bedrock-2023-05-31",
		max_tokens: 50,
		tools: [
			{
				name: "get_data",
				description: "Gets data",
				input_schema: { type: "object", properties: {} },
			},
		],
		system: system_prompt,
		messages: [
			{
				role: "user",
				content: [
					{ type: "text", text: "Call get_data and tell me what it returned" },
				],
			},
			{
				role: "assistant",
				content: [{ type: "tool_use", id: "tc1", name: "get_data", input: {} }],
			},
			{
				role: "user",
				content: [
					{
						type: "tool_result",
						tool_use_id: "tc1",
						content: [
							{
								type: "text",
								text: "the result",
							},
						],
						cache_control: { type: "ephemeral" },
					},
				],
			},
		],
	}),
});

if (res.ok) {
	console.log("✓ PASS: HTTP 200");

	// Check the token usage in the response
	const body = await res.json();
	console.log("Response:", body.content)
	console.log("Token usage:", body.usage);
} else {
	console.error("✗ FAIL:", await res.text());
}
```

### Log File

_No response_

### Region

EU12 Canary

Development

### Impact

Inconvenience

### Timeline

_No response_

### Additional Context

_No response_

InjunPark-sap:
## Context

Closes #2047 .

- `applyCacheControlToLastMessage` was converting tool message string content into a `TextContent[]` array when `cache_control` was passed as a call option.
- The Orchestration service's Anthropic harmonization layer requires tool message content to remain a plain string and rejects arrays with HTTP 400.

- Fix adds an early-return guard for `role === 'tool'` in `applyCacheControlToLastMessage`, skipping cache_control injection on tool messages entirely. This also aligns with Anthropic's caching spec — cache breakpoints on tool messages are not supported by Anthropic models.


packages/langchain/src/orchestration/util.ts
```diff
@@ -258,16 +258,16 @@ export function applyCacheControlToLastMessage(
     return;
   }
 
+  // Skip tool messages: cache_control is not supported by Anthropic models on tool messages.
+  // Additionally, the Orchestration service's Anthropic harmonization layer
+  // requires tool message content to be a plain string (rejects arrays with HTTP 400).
```
davidkna-sap: [req] Update the comment please: This is inaccurate. It should be supported for Claude according to the documentation - this is likely an orchestration bug.
The reason we still want to avoid doing this for tool messages is that amazon nova models will not support that.

packages/langchain/src/orchestration/util.ts
```diff
@@ -258,16 +258,16 @@ export function applyCacheControlToLastMessage(
...
+  if (lastMessage.role === 'tool') {
+    return;
+  }
```
davidkna-sap: [req] Avoid returning directly, instead try using `findLastIndex` or similar to find the last applicable message.

packages/langchain/src/orchestration/util.ts
davidkna-sap: [pp] Sine we are touching this, please replace this with `messageSupportsCacheControl`, checking `lastMessage.role === 'system' || lastMessage.role === 'developer')`

davidkna-sap (changes requested):

InjunPark-sap:
@davidkna-sap 
Thanks for the review! Two quick questions before implementing:

1. [util.ts line 264-266](https://github.com/SAP/ai-sdk-js/pull/2058/changes#diff-5b851e2acb0300d26449d596b034804c267edd1dcda3de53bb470749bc2d95acR264-R266) + [line 289](https://github.com/SAP/ai-sdk-js/pull/2058/changes#diff-5b851e2acb0300d26449d596b034804c267edd1dcda3de53bb470749bc2d95acL289)
: `findLastIndex` applicable roles — in a sequence like `[system, user, assistant (with tool_calls), tool]`, should `assistant` be considered applicable, or should we skip it too? 
(Asking because `assistant` string content is already a no-op per the existing comment on line 289.)
3. [util.ts line 297-303](https://github.com/SAP/ai-sdk-js/pull/2058/changes#diff-5b851e2acb0300d26449d596b034804c267edd1dcda3de53bb470749bc2d95acR297-R303): `messageSupportsCacheControl` scope — should it include `user` as well, or just `system | developer`? The current `isUserMessage` flag also gates `image_url`/`file` blocks as cacheable (line 303), so if `user` is excluded from the helper that logic would need to be handled separately.

davidkna-sap:
@InjunPark-sap Please restrict cache_control markers to positions supported by all models, [as documented by the service](https://help.sap.com/docs/sap-ai-core/generative-ai/prompt-caching?locale=en-US).
`messageSupportsCacheControl` should filter according to the same rules set by the service.

packages/langchain/src/orchestration/util.ts
InjunPark-sap: Updated — the real reason is Nova models don't support it, not Anthropic.

packages/langchain/src/orchestration/util.ts
InjunPark-sap: Done

packages/langchain/src/orchestration/util.ts
InjunPark-sap: Extracted as requested. Kept isUserMessage in the array branch since it still gates image_url/file blocks.

.changeset/fix-cache-control-tool-message.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ai-sdk/langchain': patch
+---
+
+[fix] Fix https? 400 when `cache_control` is used and the last message is a tool result.
+The SDK now skips tool and assistant messages when applying `cache_control`, finding the last applicable message (system, user, or developer) instead.
```
davidkna-sap: [pp]
```suggestion
[fix] Skip tool and assistant messages when applying `cache_control`, using the last applicable message (system, user, or developer) instead.
Some models do not support `cache_control` on `tool` messages.
```

davidkna-sap (approved):

PR #2065, opened 2026-07-22
chore: replace depcheck with knip

davidkna-sap:
## Context

- Faster than depcheck
- More modern dependency footprint
- Uncovers additional issues
- Does not require ignoring global dependencies like jest.
- Workspace-aware


package.json
```diff
@@ -53,19 +53,19 @@
     "@types/jsonwebtoken": "^9.0.10",
     "@types/mock-fs": "^4.13.4",
     "@types/node": "^22.20.1",
-    "depcheck": "^1.4.7",
+    "@typescript/native": "npm:typescript@^7.0.2",
     "eslint": "^10.7.0",
     "glob": "^13.0.6",
     "jest": "^30.4.2",
     "jsonwebtoken": "^9.0.3",
+    "knip": "https://pkg.pr.new/knip@1902",
```
davidkna-sap: TODO: move to stable version once released

package.json
```diff
@@ -34,7 +34,7 @@
     "smoke-tests": "pnpm -F=@sap-ai-sdk/smoke-tests run",
     "sample-code": "pnpm -F=@sap-ai-sdk/sample-code run",
     "sample-cap": "pnpm -F=@sap-ai-sdk/sample-cap run",
-    "check:deps": "pnpm -r -F !. -F !./tests/smoke-tests -F !./sample-cap exec depcheck --ignores=\"nock,@jest/globals\" --quiet"
+    "check:deps": "knip --config knip.config.mjs -W '!.' -W '!./sample-cap'"
```
InjunPark-sap, line 37: [q] The previous depcheck command explicitly excluded tests/smoke-tests. Please confirm whether this omission is intentional.

`"check:deps": "knip --config knip.config.mjs -W '!.' -W '!./sample-cap' -W '!./tests/smoke-tests'"`

package.json
```diff
@@ -53,19 +53,19 @@
     "@types/jsonwebtoken": "^9.0.10",
     "@types/mock-fs": "^4.13.4",
     "@types/node": "^22.20.1",
-    "depcheck": "^1.4.7",
+    "@typescript/native": "npm:typescript@^7.0.2",
     "eslint": "^10.7.0",
     "glob": "^13.0.6",
     "jest": "^30.4.2",
     "jsonwebtoken": "^9.0.3",
+    "knip": "6.29.0",
```
InjunPark-sap, line 55: [minor] knip is pinned to an exact version "6.29.0", inconsistent with the ^x.y.z range pattern used for all other devDependencies.

package.json
davidkna-sap: Smoke tests work now.

package.json
davidkna-sap: This would change if we do switch to renovate as planned.

marikaner: [pp] Then I would propose to make this change on the renovate branch instead. But no strong opinions.

knip.config.mjs
```diff
@@ -0,0 +1,14 @@
...
+    'sample-code': {
+      entry: [
+        'src/tutorials/mcp/weather-mcp-server.ts',
+        'resources/generate-parquet.ts'
+      ]
+    },
+    'tests/type-tests': {
+      entry: ['test/**/*.test-d.ts']
+    }
+  }
```
marikaner, line 22: [pp] This seems oddly selective. If this is intended, can you maybe add a comment why those entrypoints are here, but others aren't?

pnpm-workspace.yaml
```diff
@@ -27,6 +27,7 @@ minimumReleaseAgeExclude:
   - '@sap-cloud-sdk/*'
   # Add minimum age exceptions for audit issues here
   # - sample-package@0.1.0
+  - 'knip@6.29.0'
```
marikaner: [req] No. ;) (and no longer needed)

sample-code/src/server.ts
marikaner, line 1: [req] The changes on this file seem like they accidently were merged from another branch. This is not related to knip, is it?

marikaner (changes requested):

sample-code/src/server.ts
davidkna-sap: It found an orphan unused function, so I just added it to the sample server (exporting it from `index.js` would likely have also worked).

knip.config.mjs
davidkna-sap: For sample-code they need to be listed because they are not directly referenced in the `package.json` and the files are otherwise orphans, e.g. adding `generate-parquet` as a script should also fix it.
`src/tutorials/mcp/weather-mcp-server.ts` isn't really meant as a full mcp server but part of the larger tutorial, and `generate-parquet` is mostly an internal thing.

For the type tests it is need because it does not have handling for `tsd` "entry" files.

marikaner: Can you add this as a (maybe a little bit more general) comment to this file?

So knip understands the pnpm-workspace.yml and reads all package.jsons and adds this config on top?

davidkna-sap: I've added a few comments.

marikaner (dismissed):

PR #2068, opened 2026-07-22
chore: migrate to oxfmt for faster formatting

davidkna-sap:
## Context

| | prettier+oxc | prettier | oxfmt |
|---|---:|---:|---:|
| Mean (wall) | 1.503 s | 1.727 s | 0.569 s |
| ± σ | 0.028 s | 0.170 s | 0.019 s |
| Min | 1.467 s | 1.619 s | 0.546 s |
| Max | 1.537 s | 2.028 s | 0.586 s |
| User CPU | 1.813 s | 2.577 s | 3.885 s |
| System CPU | 0.275 s | 0.297 s | 0.507 s |
| Peak memory | 270 MB | 350 MB | 350 MB |


davidkna-sap (commented):
check-public-api requires prettier, but it is not available, likely due to issues with ncc.

davidkna-sap (commented):
might need https://github.com/SAP/cloud-sdk-js/pull/6806 for checks not to fail

.oxfmtrc.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$schema": "./node_modules/oxfmt/configuration_schema.json",
```
InjunPark-sap, line 2: $schema points to a local ./node_modules/... path, which breaks editor JSON IntelliSense on fresh clones and CI cache misses. Use an HTTPS URL instead.

scripts/sync-model-types.ts
```diff
@@ -302,8 +304,9 @@ async function patchModelTypes(
         console.error(`No change for ${typeName}`);
       }
     }
-    const options = await prettier.resolveConfig(MODEL_TYPES_PATH);
-    return prettier.format(updated, options || undefined);
+    const oxfmtrc = JSON.parse(await readFile(OXFMTRC_PATH, 'utf8'));
```
InjunPark-sap, line 312: Passing the full .oxfmtrc.json to format() includes CLI-only fields like $schema and ignorePatterns, which may cause unexpected behavior. Since only model-types.ts is being formatted, pass only the required formatting options directly.

InjunPark-sap (commented):
Clear motivation (3x speed gain) and consistent script updates

scripts/sync-model-types.ts
davidkna-sap: They don't cause unexpected behaviour.

.oxfmtrc.json
davidkna-sap: An HTTPS URL would also break, as the editor would have to trust the source first.

marikaner: I think it is legit to expect installing dependencies before formatting. Also HTTP URLS might point to a different version. Currently it is coupled to the version we use.

scripts/sync-model-types.ts
marikaner: @InjunPark-sap can you elaborate? Currently there is a single source of truth, which is better than additional configuration for one file. I don't fully understand why `@schema` or `ignorePatterns` would be CLI-only?

scripts/sync-model-types.ts
```diff
@@ -302,8 +304,9 @@ async function patchModelTypes(
...
+    const { code } = await oxfmt(MODEL_TYPES_PATH, updated, oxfmtrc);
```
marikaner, line 313: [q] Do we even need this step? I would expect auto-lint to do this.

marikaner (dismissed):
lgtm

.oxfmtrc.json
davidkna-sap, line 1: If oxlint is merged, it would make sense to enable import sorting here.

scripts/sync-model-types.ts
davidkna-sap: It avoids some unnecessary noise when Claude tries to look at the diff in the skill.

InjunPark-sap (dismissed):

scripts/sync-model-types.ts
marikaner: this could also be improved in the skill, i guess?

.github/workflows/build.yml
```diff
@@ -75,7 +72,7 @@ jobs:
       # Independent read-only checks — run them concurrently.
       - parallel:
           - name: Check public API
-            uses: sap/cloud-sdk-js/.github/actions/check-public-api@main
+            uses: sap/cloud-sdk-js/.github/actions/check-public-api@davidkna-sap_internal-ts-shim
```
marikaner: [req] no longer needed

marikaner (dismissed):

.github/workflows/build.yml
davidkna-sap:
```suggestion
            uses: sap/cloud-sdk-js/.github/actions/check-public-api@main
```

marikaner (dismissed):

marikaner (approved):

PR #2069, opened 2026-07-22
chore: [js] move to oxlint

davidkna-sap:
## Context

3.1x times faster (0.6s vs 1.9s)

Some rules not supported (e.g. stylistic -> prettier, some import/no-internal (use restricted instead), sorting imports (use oxfmt/prettier instead)

It did find some rule violations, some of them worse than others.


packages/core/src/stream/sse-stream.test.ts
```diff
@@ -325,7 +325,7 @@ describe('streaming decoding', () => {
     expect(event.value.event).toBeNull();
     expect(JSON.parse(event.value.data)).toEqual({ content: 'culpa ' });
 
-    expect(stream.next()).rejects.toThrow(
+    void expect(stream.next()).rejects.toThrow(
```
InjunPark-sap:
```suggestion
    return expect(stream.next()).rejects.toThrow(
```

InjunPark-sap: void silences the linter but discards the Promise entirely — rejects.toThrow() never executes, so the test passes regardless of whether the stream throws.

.oxlintrc.json
```diff
@@ -0,0 +1,320 @@
...
+        "no-unused-vars": [
+          "error",
+          {
+            "argsIgnorePattern": "^_",
+            "varsIgnorePattern": "^_",
+            "destructuredArrayIgnorePattern": "^_"
+          }
+        ],
+        "import/default": "error",
+        "import/namespace": "error",
```
InjunPark-sap, line 230:
```suggestion
        "import/namespace": "off",
```

.oxlintrc.json
```diff
@@ -0,0 +1,320 @@
...
+            "varsIgnorePattern": "^_",
+            "argsIgnorePattern": "^_"
+          }
+        ],
+        "import/namespace": "off"
+      },
+      "plugins": [
+        "import"
+      ]
+    },
```
InjunPark-sap:
```suggestion
```

InjunPark-sap: This block exists only to flip `import/namespace` to `"off"`, but re-declaring `no-unused-vars` here without `destructuredArrayIgnorePattern: "^_"` silently drops it — oxlint applies later overrides last, so this block wins over the main `**/*.ts` block above.

InjunPark-sap: Delete this entire block, and instead change line 236 from `"import/namespace": "error"` to `"import/namespace": "off"` directly in the main `**/*.ts` override. Two-line fix total.

InjunPark-sap (changes requested):

scripts/generate-docs.ts
```diff
@@ -195,4 +199,4 @@ process.on('unhandledRejection', reason => {
   process.exit(1);
 });
 
-generateDocs();
\ No newline at end of file
+void generateDocs();
```
davidkna-sap:
```suggestion
await generateDocs();
```

sample-code/src/tutorials/mcp/weather-mcp-server.ts
```diff
@@ -88,4 +88,4 @@ server.registerTool(
 );
 
 const transport = new StdioServerTransport();
-server.connect(transport);
+void server.connect(transport);
```
davidkna-sap:
```suggestion
await server.connect(transport);
```

.oxlintrc.json
davidkna-sap: Not needed.

eslint-plugins/no-uppercase-internal-tag.js
```diff
@@ -0,0 +1,22 @@
...
+        return {
+          Program() {
+            for (const comment of context.sourceCode.getAllComments()) {
+              if (comment.value.includes('@Internal')) {
+                context.report({
+                  loc: comment.loc,
+                  message: 'You are not allowed to use @Internal. Please use @internal.'
+                });
+              }
+            }
```
marikaner: [pp] Consider .foreach

packages/core/src/stream/sse-stream.test.ts
```diff
@@ -325,7 +325,7 @@ describe('streaming decoding', () => {
     expect(event.value.event).toBeNull();
     expect(JSON.parse(event.value.data)).toEqual({ content: 'culpa ' });
 
-    expect(stream.next()).rejects.toThrow(
+    return expect(stream.next()).rejects.toThrow(
```
marikaner: [q] Why return?

test-util/mock-http.ts
```diff
@@ -66,12 +66,12 @@ export function createDestinationTokens(
...
   destination = aiCoreDestination
 ): HttpDestination {
-  const mockDestination: HttpDestination = {
+  const mockedDestination: HttpDestination = {
     ...destination,
     authentication: 'OAuth2ClientCredentials',
     ...createDestinationTokens()
   };
-  return mockDestination;
+  return mockedDestination;
```
marikaner: [pp] Just return directly.

sample-cap/.oxlintrc.json
```diff
@@ -0,0 +1,155 @@
...
+    "Request": "readonly",
+    "Response": "readonly",
+    "ReadableStream": "readonly",
+    "WritableStream": "readonly",
+    "URL": "readonly",
+    "URLSearchParams": "readonly"
+  },
+  "ignorePatterns": ["**/@cds-models/"],
+  "rules": {
+    "constructor-super": "error",
```
marikaner, line 51: [q] Can we reuse the global config?

marikaner (dismissed):
lgtm

packages/core/src/stream/sse-stream.test.ts
davidkna-sap: This should ensure the test runner can await any returned promise.

sample-cap/.oxlintrc.json
davidkna-sap: I think it should be possible, but this was also using a special CAP eslint configuration before.

packages/core/src/stream/sse-stream.test.ts
marikaner: I thought `.rejects.toThrow()` already ensured that, but apparently not (https://vitest.dev/api/expect.html#rejects). The behavior is the same for jest.
I would still prefer to await it though. In my opinion it describes what the test is doing while `return` describes that this is the end of the test (which it might not be in the future).

WDYT?

In case you agree, I think this could be a nice addition to the styleguide.

sample-cap/.oxlintrc.json
marikaner: Didn't see that. Then I won't ask for extra changes to not introduce other potential sources of errors.

packages/core/src/stream/sse-stream.test.ts
davidkna-sap: I feel like `return await` can be a bit superfluous, but for tests preferring `await` sounds good to me.

```suggestion
    await expect(stream.next()).rejects.toThrow(
```

marikaner (approved):

PR #2072, opened 2026-07-22
chore: move to vitest, but leverage injected globals

davidkna-sap:
## Context

Alternative to #1662


packages/orchestration/src/__snapshots__/orchestration-client.test.ts.snap
InjunPark-sap, line 1: The snapshot was regenerated as part of the Jest → Vitest migration, but the new value only captures `error.message`. The old snapshot also asserted .cause content (`request_id, code: 400, message, location`). 

The new value only captures `error.message` — Vitest's `toThrowErrorMatchingSnapshot()` does not traverse `.cause` by default. If `.cause` propagation breaks, this test will still pass. Consider replacing `toThrowErrorMatchingSnapshot()` in the corresponding test with explicit `.cause` assertions.

davidkna-sap: Fixed, that said eslint should catch that sort of issue.

CONTRIBUTING.md
```diff
@@ -23,19 +23,6 @@ Run `pnpm install` in the root directory instead.
 
 All (new) functionality shall be covered by tests.
 
-### Jest Runner Set-up
-
-If you're using the Jest Runner extension in Visual Studio Code, you'll need to add the following settings to your `settings.json` file before running the tests from VS Code:
```
marikaner, line 28: [q] I assume there is a vitest runner and it doesn't need extra setup?

vitest.config.ts
marikaner, line 1: [q] Where is this config used?

marikaner (dismissed):
lgtm

CONTRIBUTING.md
davidkna-sap: It shouldn't, also it's the same extensions (no need for extra NODE_OPTIONS).

vitest.config.ts
davidkna-sap: It doesn't fully work right now for smoke tests/e2e etc, but it allows running all tests from the root without narrowing to the package level.

InjunPark-sap (dismissed):
LGTM

vitest.config.ts
davidkna-sap: The workspace configuration now works, it can be run with `vitest run` from the workspace root.

marikaner (dismissed):
lgtm

marikaner (approved):

PR #2074, opened 2026-07-23
chore: forward requestconfig to deployment api calls

davidkna-sap:
## Context

Part of SAP/ai-sdk-js-backlog#621.


packages/ai-api/src/utils/deployment-resolver.ts
```diff
@@ -279,13 +285,15 @@ export async function resolveDeploymentUrlForModel(
...
+ * @param requestConfig - The request configuration.
  * @returns The ID of the deployment, if found.
  * @internal
  */
 export async function getFoundationModelDeploymentId(
   modelDeployment: ModelDeployment,
   executableId: string,
-  destination?: HttpDestinationOrFetchOptions
+  destination?: HttpDestinationOrFetchOptions,
+  requestConfig?: CustomRequestConfig
```
marikaner: [pp] I don't like it when functions have more than 3 parameters, especially when the last two are optional. Given, that this is an internal function, I would prefer it, if we combined `destination` and `requestConfig` into one parameter. Maybe even adding `modelDeployment`.
What do you think about that?

PR #2075, opened 2026-07-23
chore: only allow erasable syntax

davidkna-sap:
## Context

This PR gets rid of `tsx`, non-erasable syntax or anything not supported by node (public x arg setting x on class, replacing js with ts extension). Some of the generated files still use js extensions, but if those are removed we could investigate publishing our typescript source files without a separate compilation step.


sample-code/src/tutorials/mcp/mcp-adapter.ts
```diff
@@ -12,8 +12,8 @@ export const mcpClient = new MultiServerMCPClient({
   useStandardContentBlocks: true,
   mcpServers: {
     weather: {
-      command: 'npx',
-      args: ['tsx', './src/tutorials/mcp/weather-mcp-server.ts']
+      command: 'node',
+      args: ['--env-file=.env', './src/tutorials/mcp/weather-mcp-server.ts']
```
InjunPark-sap:
```suggestion
      args: ['--env-file-if-exists=.env', './src/tutorials/mcp/weather-mcp-server.ts']
```

InjunPark-sap: Two unrelated behavior changes were introduced together. `--env-file=.env` crashes Node with `Error: .env: not found` if the file is absent (new clone, CI). The original `npx tsx` had no such requirement. 

If intentional, document the prerequisite; otherwise use `--env-file-if-exists`.

InjunPark-sap (changes requested):

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream.ts
```diff
@@ -188,7 +188,7 @@ export class AzureOpenAiChatCompletionStream<Item> extends SseStream<Item> {
   }
 
   constructor(
-    public iterator: () => AsyncIterator<Item>,
+    iterator: () => AsyncIterator<Item>,
```
marikaner, line 191: [req] This is no longer public, which would be a breaking change. I guess you just forgot to add a public field?

packages/orchestration/src/orchestration-stream.ts
```diff
@@ -101,7 +101,7 @@ export class OrchestrationStream<Item> extends SseStream<Item> {
   }
 
   constructor(
-    public iterator: () => AsyncIterator<Item>,
+    iterator: () => AsyncIterator<Item>,
```
marikaner, line 104: [req] same here. mistake or no mistake? ;)

marikaner (changes requested):
Only two comments, not sure if this is on purpose, but I believe this is a breaking change.

General: I would have wished that the .ts changes and the erasable syntax changes had been separated in two PRs. The .ts changes call for an audit rather than a review, while the erasable syntax changes needed a proper review. Now I looked at all 168 files 🤓

sample-code/src/tutorials/mcp/mcp-adapter.ts
marikaner: good to know!

packages/foundation-models/src/azure-openai/azure-openai-chat-completion-stream.ts
davidkna-sap: I forgot about that, I will update the code to fix the regression.

davidkna-sap: Fixed via change to base class.

davidkna-sap (commented):
May need sap/cloud-sdk-js#6832

marikaner (dismissed):
lgtm

InjunPark-sap (dismissed):
LGTM

marikaner (approved):

PR #2076, opened 2026-07-23
chore: replace changeset-cli with pnpm change

davidkna-sap:
## Context

`pnpm change` is also supposed to get a check feature, which would be handle some of our pr-checks.

Mutually exclusive with changeset/action.

Advantages of this:
- included in pnpm
- configuration for keeping everything on same version
- lane feature, to properly mark certain packages as -beta.


.github/workflows/publish-canary.yml
```diff
@@ -22,31 +22,19 @@ jobs:
...
         env:
           INPUTS_TAG: ${{ inputs.tag }}
         run: |
-          date=`date +%Y%m%d%H%M%S`
-          pnpm changeset pre enter "${date}"
-          pnpm changeset version
-          pnpm changeset pre exit
-          pnpm changeset publish --tag "${INPUTS_TAG}"
+          date=$(date +%Y%m%d%H%M%S)
+          pnpm -r exec pnpm version prepatch --preid "${date}" --no-git-tag-version
```
InjunPark-sap:
```suggestion
          pnpm -r --filter='./packages/**' exec pnpm version prepatch --preid "${date}" --no-git-tag-version
```

InjunPark-sap: pnpm -r exec runs across every workspace package, including private ones (sample-cap, sample-code). These should never be published or version-bumped.

InjunPark-sap (changes requested):

.github/workflows/publish-canary.yml
davidkna-sap: Adding, but this shouldn't matter as we should be setting `private: true`, and this only sets version afaict.

davidkna-sap: Reverted because this would change existing behaviour.

marikaner: i think this is fine, as we never adjust the versions of private packages.

.github/workflows/bump.yml
```diff
@@ -67,6 +65,7 @@ jobs:
       - name: Checkout current repository
         uses: sap/cloud-sdk-js/.github/actions/setup@main
         with:
+          pnpm-version: ''
```
marikaner, line 67: [req/q] this is a big ugly. why is this needed? my intuitive understanding: take the default version, but then I would expect I don't need to set it.

marikaner (changes requested):
Nice 👍

.github/workflows/bump.yml
davidkna-sap: The default is pnpm 10, so I unset it there in order to have it use pnpm 11 from devEngines.

marikaner: I thought we removed the dev engines setting parsing... (well apparently, not).
I think the order should be:
1. `pnpm-version` set explicitly
2. devEndinges setting
3. default value in action

'' is an invalid value IMO

WDYT?

davidkna-sap: If we provide `''` to the pnpm action it will parse `devEngines` on behalf of us, but `'11'` should also work instead of `''`.
`pnpm-version` defaults to `'10'`.

marikaner:
@davidkna-sap is this still a draft on purpose?

davidkna-sap:
@marikaner This is mostly draft because, because 'approach-to-pnpm11' is not quite defined yet.

.github/workflows/bump.yml
davidkna-sap: Moved to the in-repo action to fix this.

PR #2097, opened 2026-07-29
feat: Overload Orchestration client constructor

marikaner:
I noticed that (even with v4.8.0 of the SAP Cloud SDK) I don't get very helpful autocompletion, when instantiating an orchestration client. Overloads make this much clearer and equally don't allow setting both types.


packages/orchestration/src/orchestration-client.ts
```diff
@@ -52,23 +52,59 @@ const logger = createLogger({
...
     private config:
       | string
-      | Xor<
-          OrchestrationConfigRef,
-          OrchestrationModuleConfig | OrchestrationModuleConfigList
-        >,
+      | OrchestrationConfigRefById
+      | OrchestrationConfigRefByName
+      | OrchestrationModuleConfig
+      | OrchestrationModuleConfigList,
```
davidkna-sap, line 142: [pp] Xor might provide additional type safety here for the internal implementation, IIRC this implementation construct wouldn't be part of the external API.

.changeset/clear-spies-grin.md
davidkna-sap, line 1: [nth] Treat it as `patch`? But the check could complain in that case.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -52,23 +52,59 @@ const logger = createLogger({
...
+  /* eslint-enable @typescript-eslint/unified-signatures */
   constructor(
     private config:
       | string
-      | Xor<
-          OrchestrationConfigRef,
-          OrchestrationModuleConfig | OrchestrationModuleConfigList
-        >,
+      | OrchestrationConfigRefById
+      | OrchestrationConfigRefByName
```
InjunPark-sap:
```suggestion
      | Xor<OrchestrationConfigRefById, OrchestrationConfigRefByName>
```

packages/orchestration/src/orchestration-client.ts
InjunPark-sap: + The public overloads already guide users to the right type. The implementation signature is not part of the external API, so keeping `Xor` here restores the structural guard against `{ id, scenario, name, version }` being passed simultaneously via `any` — which the plain union no longer prevents.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -270,36 +270,42 @@ export type OrchestrationModuleConfigList = [
...
+  scenario: string;
+  /** Configuration name. */
+  name: string;
+  /** Configuration version. */
+  version: string;
+}
+
+/**
+ * Reference to an orchestration configuration created via the Prompt Registry API.
+ * @deprecated - Use {@link OrchestrationConfigRefById} or {@link OrchestrationConfigRefByName} instead.
```
InjunPark-sap:
```suggestion
 * @deprecated Use {@link OrchestrationConfigRefById} or {@link OrchestrationConfigRefByName} instead.
```

InjunPark-sap: TSDoc/JSDoc `@deprecated` does not take a leading dash. Some tooling parses ` - Use` as literal text in the deprecation notice.

packages/orchestration/src/orchestration-client.ts
```diff
@@ -52,23 +52,59 @@ const logger = createLogger({
...
+      | OrchestrationModuleConfig
+      | OrchestrationModuleConfigList,
     private deploymentConfig?: ResourceGroupConfig | DeploymentIdConfig,
     private destination?: HttpDestinationOrFetchOptions
   ) {
```
InjunPark-sap, line 145: The `string` overload is listed first, but `OrchestrationModuleConfig` is by far the most common usage. VS Code shows overloads in declaration order — consider moving the `ModuleConfig` overload to position 1 for better discoverability.

InjunPark-sap, line 145: Overloads 1–2 use `config`, overloads 3–4 use `configRef` for the same positional parameter. IDE hints show the parameter name per overload — I think using `config` consistently (or `configRef` consistently) could avoid confusion.

.changeset/clear-spies-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/orchestration': minor
```
InjunPark-sap, line 2: Two changesets for one PR, both `minor` — is this intentional? Changesets takes the highest bump so no double-increment, but it produces two separate `CHANGELOG` entries. Fine if the separate entries are desired; worth confirming.

InjunPark-sap (dismissed):
some questions and code suggestions with my investigation

packages/orchestration/src/orchestration-client.ts
marikaner: good idea 😊

marikaner: I actually did it on purpose because 1 and 2 are configurations and 3 and 4 are references for configurations. I prefer this more exact name. WDYT?

.changeset/clear-spies-grin.md
marikaner: Yes, they are separate entries in separate categories. One describes the feature, the other the deprecation. So later it will show up like this:

## Compatibility Notes

[orchestration] xy is now deprecated in favor of z

## New Functionality

[orchestration] z is cool

packages/orchestration/src/orchestration-client.ts
marikaner: True, theoretically we should also xor `OrchestrationModuleConfig` otherwise I find this a little inconsistent. `Xor` only takes 2 parameters though...

.changeset/clear-spies-grin.md
marikaner: Didn't we have a rule, that a deprecation is always a minor bump?

packages/orchestration/src/orchestration-client.ts
marikaner: how do you like the new order?

.changeset/clear-spies-grin.md
InjunPark-sap: Yes, deprecations are always `minor` per our convention. The `[compat]` changeset is correct as-is.

packages/orchestration/src/orchestration-client.ts
InjunPark-sap: Makes sense. The distinction is intentional and the JSDoc explains each overload clearly. 👍

InjunPark-sap: Looks good! 

One quick question: `OrchestrationModuleConfigList` is now a separate 5th overload — intentional, or should it stay combined with `OrchestrationModuleConfig` as `OrchestrationModuleConfig | OrchestrationModuleConfigList`?

packages/orchestration/src/orchestration-client.ts
InjunPark-sap: `Xor<RefById, Xor<RefByName, ModuleConfig>>` means "RefById XOR (RefByName XOR ModuleConfig)" — this actually allows RefById + ModuleConfig simultaneously, which isn't the intent. Also `string` and `OrchestrationModuleConfigList` are still outside.

@davidkna-sap's concern was specifically about mixing both ref types. `string` and `ModuleConfig` don't overlap with refs structurally, so they don't need Xor. Simpler and correct:

```
| string
| Xor<OrchestrationConfigRefById, OrchestrationConfigRefByName>
| OrchestrationModuleConfig
| OrchestrationModuleConfigList
```

.changeset/clear-spies-grin.md
davidkna-sap: We already have other minor bumps anyway.

packages/orchestration/src/orchestration-client.ts
davidkna-sap: That is not the case.

InjunPark-sap: You're right, I was wrong. `Xor<RefById, Xor<RefByName, ModuleConfig>>` is a valid 3-way mutual exclusion — verified with the TypeScript compiler. `{ id, promptTemplating } and { id, scenario, name, version } `both produce compile errors as intended.

The current implementation is correct. Thanks for the correction.

packages/orchestration/src/orchestration-client.ts
marikaner: Intentional, because I thought this would look nicer in the API. 5 overloads might too much though. WDYT?

InjunPark-sap: yes make sense

marikaner:
@InjunPark-sap, @davidkna-sap I am now getting errors and I am unsure how to proceed. The issue lies in how TypeScript resolves overloads: a union argument A | B won't match two separate overloads that each take A or B individually (maybe the lint rule was there for a reason...).

Anyways, I find the discoverability much better with the overloads, but it can cause real issues and apparently is a breaking change if someone passes on a parameter that combines any of the existing values.

I am currently leaning towards just closing this PR, although, I really think that the API would benefit from better discoverability. (There might be other options, like factory functions 😕 )

marikaner:
An other option could be to add another overload that combines everything in one (same signature as the implementation).

davidkna-sap:
I would say that adding a new overload that combines everything would be fine, though please confirm if we can `@deprecate` it, I am not sure how well our tooling handles that.

InjunPark-sap:
The actual breaking case:

```
function getRef(): ConfigRef { return { id: 'x' } }
new OrchestrationClient(getRef())  // ❌ no overload matches
```

- Fix: add @deprecated catch-all as the last specific overload (before implementation). TS resolves top-to-bottom — putting it last keeps inline literals hitting specific overloads, while ConfigRef-typed variables fall through to the deprecated one and get the IDE strikethrough hint.

@deprecated on individual overloads is supported since TS 3.8.

marikaner:
> I would say that adding a new overload that combines everything would be fine, though please confirm if we can `@deprecate` it, I am not sure how well our tooling handles that.

Do we want to deprecate this though? Should that be a workaround for users to not use it directly? Users who have a case as described before HAVE to use this overload.

packages/orchestration/src/orchestration-types.ts
```diff
@@ -299,19 +290,43 @@ export type OrchestrationConfigRefOverride = Omit<
...
+  /** Configuration name. */
+  name: string;
+  /** Configuration version. */
+  version: string;
+} & {
+  /**
+   * Optional partial configuration to override parts of the stored orchestration config at request time.
+   */
+  overrideConfig?: OrchestrationConfigRefOverride;
+};
```
davidkna-sap, line 322: [nth] I would prefer a `OrchestrationConfigBaseT>` that expands to `T & { overrideConfig?: }` or a new `CommonOverrideOptions` type for `overrideConfig?`.

davidkna-sap:
> > I would say that adding a new overload that combines everything would be fine, though please confirm if we can `@deprecate` it, I am not sure how well our tooling handles that.
> 
> Do we want to deprecate this though? Should that be a workaround for users to not use it directly? Users who have a case as described before HAVE to use this overload.

I don't feel too strongly about deprecating, leaving as-is should be fine.

packages/orchestration/src/orchestration-types.ts
marikaner: Just to make sure I understand correctly. The purpose would be to not repeat:

```
& {
  /**
   * Optional partial configuration to override parts of the stored orchestration config at request time.
   */
  overrideConfig?: OrchestrationConfigRefOverride;
};
```

I was considering that too, but I fear that we would clutter the API unnecessarily. I also find it hard to find a good name (that is not too generic).

Let me know if you have more arguments pro such a change.

davidkna-sap: Let's leave it as-is for now.

davidkna-sap (approved):

PR #2104, opened 2026-08-03
chore: speed up orchestration timeout tests

davidkna-sap:
## Context

This should also make the tests more reliable.


packages/langchain/src/orchestration/client.test.ts
```diff
@@ -134,19 +136,34 @@ describe('orchestration service client', () => {
...
+        controller.abort(
+          new DOMException('The operation timed out.', 'TimeoutError')
+        );
+        await vi.advanceTimersByTimeAsync(0);
+        await expect(response).rejects.toThrow(
+          expect.objectContaining({
+            stack: expect.stringMatching(/Timeout/)
+          })
+        );
+        await settled;
```
marikaner: [q] Is this needed?

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -120,12 +118,16 @@ describe('orchestration service client', () => {
...
     it('throws when delay exceeds timeout', async () => {
-      mockInferenceWithResilience(mockResponse, { delay: 2000 });
+      mockInferenceWithResilience(mockResponse, { delay: 5000 });
+      const controller = new AbortController();
       const client = new OrchestrationClient(config, { maxRetries: 0 });
-      const response = client.invoke(messages, { timeout: 1000 });
+      const response = client.invoke(messages, { signal: controller.signal });
+      controller.abort(
+        new DOMException('The operation timed out.', 'TimeoutError')
+      );
```
marikaner: [q] Does this really reflect the same situation as having a timeout?

marikaner (commented):
Left a few questions

packages/langchain/src/orchestration/client.test.ts
davidkna-sap: unhandled rejection can cause runtime issues, removed this by refactoring the test. `await` is not strictly necessary but avoids leaving behind stale promises.

packages/langchain/src/orchestration/client.test.ts
davidkna-sap: Updated this to be more realistic.

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -120,32 +122,42 @@ describe('orchestration service client', () => {
...
+      const response = client.invoke(messages, {
+        signal: AbortSignal.timeout(100)
+      });
+      const assertion = expect(response).rejects.toThrow(
         expect.objectContaining({
           stack: expect.stringMatching(/Timeout/)
         })
       );
+      await vi.advanceTimersByTimeAsync(100);
+      await assertion;
```
marikaner: [req] This is unusual. I would expect it to be:

```suggestion
      await expect(response).rejects.toThrow(
        expect.objectContaining({
          stack: expect.stringMatching(/Timeout/)
        })
      );;
```

If this is not possible for some reason, please add a comment.

packages/langchain/src/orchestration/client.test.ts
```diff
@@ -120,32 +122,42 @@ describe('orchestration service client', () => {
...
     it('throws when delay exceeds timeout', async () => {
-      mockInferenceWithResilience(mockResponse, { delay: 2000 });
+      vi.useFakeTimers();
+      mockInferenceWithResilience(mockResponse, { delay: 5000 });
       const client = new OrchestrationClient(config, { maxRetries: 0 });
-      const response = client.invoke(messages, { timeout: 1000 });
-      await expect(response).rejects.toThrow(
+      const response = client.invoke(messages, {
+        signal: AbortSignal.timeout(100)
+      });
```
marikaner: [q] Maybe this wasn't clear from my previous comment, but what I meant is:
This test now checks, that we handle a custom signal that sends a timeout correctly. Previously, we tested, that the langchain orchestration's client handles the timeout setting correctly. Those are 2 different settings. While it sounds legit, I am not sure whether they behave the same internally, Even if they do, we just test something similar, not the same thing from before.

That said, is not possible to keep the timeout? When advancing the timers, I would expect the response to throw as expected?

marikaner (changes requested):

packages/langchain/src/orchestration/client.test.ts
davidkna-sap: I've reverted it there, but it was necessary for `throws when delay exceeds timeout during streaming`.

packages/langchain/src/orchestration/client.test.ts
davidkna-sap: I've reverted it back to using the timeout property.

packages/langchain/src/orchestration/client.test.ts
marikaner: Makes sense, thanks for adding the comment. I think you forgot to push the revert here though.

marikaner (changes requested):

packages/langchain/src/orchestration/client.test.ts
davidkna-sap: Reverted.

marikaner (approved):

PR #2105, opened 2026-08-03
feat: PoC support sap-rpt-1.5 (don't merge)

ZhongpinWang:
Try out rpt 1.5 spec. 

Key findings:

The spec contains OAS 3.1 features. Meaning the new openapi-generator in the soon released Cloud SDK JS 4.9.0 is needed.

One noticeable breaking change (for SDK, not necessarily for our users) and "weird behaviour":

```json
        "prediction_config": {
            "type": "string",
            "title": "Prediction Config",
            "description": "JSON string containing the prediction configuration (see PredictionConfig schema).",
            "example": "{\"target_columns\":[{\"name\": \"PRICE\",\"prediction_placeholder\": null,\"task_type\": \"regression\"}]}",
            "contentMediaType": "application/json",
            "contentSchema": {
              "$ref": "#/components/schemas/PredictionConfig"
            }
          },
```

In the rpt 1.5 spec, `prediction_config` now has type `string`. I tried it out and the remote server accepts both JSON string and JSON object. It is questionable if they should change the type to oneOf `string` or `PredictionConfig` ref.

This is currently causing issues in our SDK as we need to then either JSON parse the object before sending it to remote, or just mark the object type as any as the remote reads object as well.

Other changes are mostly feature additions. We need to think if we

- have two separate packages (feels overkill to me)
- have two separate clients (okay)
- or have one client but do type inference based on the model deployment name (but how do we know at design time whether we should show rpt 1.5 specific types if user provides directly deployment id? Also this approach is quite complicated by mixing two specs with the assumption that they won't diverge significantly)

### DIFF
--- a/packages/document-grounding/src/client/api/retrieval-api.ts
+++ b/packages/document-grounding/src/client/api/retrieval-api.ts
@@ -38,14 +38,14 @@ export const RetrievalApi = {
   /**
    * List data repository by id
    * @param repositoryId - Repository ID
-   * @param headerParameters - Object containing the following keys: AI-Resource-Group.
    * @param queryParameters - Object containing the following keys: remoteName.
+   * @param headerParameters - Object containing the following keys: AI-Resource-Group.
    * @returns The request builder, use the `execute()` method to trigger the request.
    */
   getDataRepositoryById: (
     repositoryId: string,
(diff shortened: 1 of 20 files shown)


marikaner:
@ZhongpinWang can we close this?

PR #2132, opened 2026-08-07
chore: Realtime demo UI

davidkna-sap:
## Context

image

Served on http://localhost:8080/openai-realtime


marikaner (changes requested):
I got this error while trying it out.
There is a dropdown to change the color of the ui (I assume) but it isn't working. When I say something it only prints a few words (sometimes different words than I said) but apparently understands the complete sentence.

Image

marikaner:
> I got this error while trying it out. There is a dropdown to change the color of the ui (I assume) but it isn't working. When I say something it only prints a few words (sometimes different words than I said) but apparently understands the complete sentence.
> 
> Image

I believe the error came from my first try when I just clicked the mic without holding it. Still shouldn't error I think.

davidkna-sap:
> I got this error while trying it out. There is a dropdown to change the color of the ui (I assume) but it isn't working. When I say something it only prints a few words (sometimes different words than I said) but apparently understands the complete sentence.
> 
> Image

It's a bit difficult to work around that because IIRC it does not count silence, I will have a look again on how to improve it.

Currently the transcription of the user-voice is handled via browser-APIs and not the gpt-realtime model itself, so the quality is a bit lower - ai core currently doesn't host a real transcription model.

The dropdown is for changing the model's voice.

davidkna-sap:
I improved the error message and added a label to the voice select menu.

davidkna-sap:
Initial browser transcriptions are now replaced with gemini-based transcription after recording has stopped.

BrigittaK307:
I first tried out some written prompts and then some with using a mic. It transcripted gibberish and it acted like it didn't get anything from my message. Then I switched voices and tried again with a mic (both with push to talk and VAD) and it worked. (I don't know if this is a bug or it just worked in a weird way for me now).

BrigittaK307:
image
When holding the mic button for a small amount of time and releasing it, an error message appears, but in the chat a random message appears as if it would have been written by the user (as shown on the screenshot).

davidkna-sap:
@BrigittaK307 

> I first tried out some written prompts and then some with using a mic. It transcripted gibberish and it acted like it didn't get anything from my message. Then I switched voices and tried again with a mic (both with push to talk and VAD) and it worked. (I don't know if this is a bug or it just worked in a weird way for me now).

Should be fixed with improved connection handling now.

> When holding the mic button for a small amount of time and releasing it, an error message appears, but in the chat a random message appears as if it would have been written by the user (as shown on the screenshot).

Errored messages now display with distinct styling to show that they are not used. I also improved the gemini transcription to print [silence] if it detects silence.

BrigittaK307:
When having VAD active sometimes the transcription isn't showing what I really asked, but is calling the right tools. e.g. I have asked what is the weather in Paris? On the screenshot it can be seen that the tool call has been called with the right parameters, but the text shows something else.
Screenshot 2026-09-23 102119

PR #2148, opened 2026-08-11
chore: don't require approval on scheduled canary releases

davidkna-sap:
## Context

This ensure scheduled event do not require approval.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,5 +1,10 @@
 # For trusted publishing releases, it is required to start all releases from a single workflow file.
 # This workflow delegates to the appropriate release workflow based on the event that triggered it.
+#
+# Environment protection matrix:
+# - Regular release: needs approval, needs to run on main/protected branch.
+# - Scheduled canary release: no approval, needs to run on main/protected branch.
+# - Manual canary release: needs approval, no branch restrictions.
 name: publish
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -20,8 +20,24 @@ on:
 permissions: {}
 
 jobs:
+  approval:
+    name: Approve Release
+    runs-on: ubuntu-latest
+    # Scheduled releases may continue without approval, but releases triggered by a GitHub Release require approval.
+    environment: ${{ case(github.event_name == 'schedule', 'ensure-main', 'approve-release') }}
```
davidkna-sap, line 34: I am making this conditional here, because otherwise `delegate_to_canary_job` will need a more complicated `if` expression.

marikaner (changes requested):
Is there a benefit to adding these additional environments and enforcing them here?
Why don't we just split the release environment into two?

release:
- needs approval, runs on main, registered on npm

canary-release:
- registered on npm


Also:
- The current `ensure-main` environment only ensures that this runs on a protected branch, not necessarily `main`.
- I think it would be nice to allow publishing canary versions from a different branch than main, this is helpful if you need to try something out before merging it.

davidkna-sap:
@marikaner We can only register one environment on npm afaik.

The way this should work now:
- Scheduled release: no approval, needs to run on main/protected branch.
- Regular release: approval, needs to run on main/protected branch.
- Manual canary release: approval, no branch restrictions.

PR #2150, opened 2026-08-12
chore: also migrate ai-sdk to memfs

davidkna-sap:
## Context

Cloud SDK is already migrated, handle this repo too.


scripts/get-package-version.test.ts
```diff
@@ -1,25 +1,23 @@
...
+import { fs, vol } from 'memfs';
+
+vi.mock('fs/promises', () => fs.promises);
+
 import { getPackageVersion } from './get-package-version.ts';
 
 describe('get package version', () => {
-  afterEach(() => {
-    mock.restore();
+  beforeEach(() => {
```
marikaner: [req] Why before? Sounds like we did not clean up somewhere else correctly + we leave it "unclean" after the last test ran.

marikaner (changes requested):

scripts/get-package-version.test.ts
davidkna-sap: I did have an agent do the change, and for this test file it does not make a difference and left it as-is.
I have reverted that change now.

marikaner (approved):

PR #2153, opened 2026-08-12
chore(sample-code): replace deprecated GitHub MCP server with native-fetch tools

InjunPark-sap:
## Context

Closes SAP/ai-sdk-js-backlog#641.

Replaces the deprecated `@modelcontextprotocol/server-github` MCP with four native-`fetch` tools (`github-tools.ts`), reusing the same tool names so the agent loop is untouched.
Repo scope is enforced in the URL — the model only passes a term / issue number / path, never `owner`, `repo`, or a raw query, so scope-widening tricks can't work. No new dependency.

Also folds in the post-merge follow-ups: `throwOnLoadError: false` + context7 libraryId allowlist (`agent.ts`), head+tail code-block truncation (`reply.ts`), and knowledge entries for prompt-registry / document-grounding / llm-batch (`knowledge.ts`).

No Changeset (private package), no REUSE header (wildcard), no public API change.


InjunPark-sap:
Deprecated `@modelcontextprotocol/server-github` MCP → four native-`fetch` tools (`github-tools.ts`), same tool names so the agent loop is untouched. context7 MCP retained.

**Guardrail:** repo scope enforced at the API call — repo hardcoded in every URL, `scopedQuery()` strips injected `repo:`/`org:`/`user:` + booleans. Model passes only a term / issue number / path, so scope-widening is impossible.

sample-code/src/support-bot/github-tools.ts
```diff
@@ -0,0 +1,150 @@
...
+  return res.json();
+}
+
+// SEC-3: scope enforced at the API call. Strip any repo/org/user qualifiers and boolean
+// operators the model slipped into its term, then force-prepend repo:SAP/ai-sdk-js — this
+// makes "OR repo:other" bypasses impossible because the model never controls the full query.
+function scopedQuery(term: string): string {
+  const stripped = (term ?? '')
+    .replace(/(?:^|\s)(?:repo|org|user):[^\s]*/gi, '')
+    .replace(/\b(?:OR|AND|NOT)\b/g, '')
```
davidkna-sap: Verify this, but it might also be possible to add 5 AND/NOT/OR instead, any more and searches are rejected. (e.g. `AND  repo:sap/ai-sdk-js`).

sample-code/src/support-bot/github-tools.ts
```diff
@@ -0,0 +1,150 @@
...
+    .replace(/\s{2,}/g, ' ')
+    .trim();
+  return `repo:${ALLOWED_REPO} ${stripped}`.trim();
+}
+
+const searchIssues = tool(
+  async ({ query }) => {
+    const q = encodeURIComponent(`${scopedQuery(query)} is:issue`);
+    const data = (await gh(`/search/issues?q=${q}`)) as {
```
davidkna-sap: [pp]
```suggestion
    const data = (await gh(`/search/issues?search_type=hybrid&q=${q}`)) as {
```

sample-code/src/support-bot/github-tools.ts
InjunPark-sap: Verified. The 5-operator cap is real (422 on the 6th). But the only actual scope-bypass is an injected `repo:/org:/user: qualifier` — `repo:SAP/ai-sdk-js foo OR repo:golang/go bar` leaks the other repo (9754 hits in my test). 

Bare `OR/AND/NOT` never widen scope; the leading repo: binds the whole query. So I dropped the OR/AND/NOT strip and kept only the qualifier strip — boolean search works now, GitHub's 5-op limit handles the rest.

InjunPark-sap:
### context7 → hosted HTTP transport

Per David's review: switched context7 from local stdio spawn to the hosted HTTP endpoint (`https://mcp.context7.com/mcp`, the variant listed in mcp-registry).

- Dropped `@upstash/context7-mcp` devDep — no local binary to spawn.
- API key optional (free tier works); `CONTEXT7_API_KEY` header wired for higher rate limits when needed.
- Side effect: removes the stdio `ENOENT` failure mode — a broken/missing binary can no longer kill the run.
- Also resolved & allowlisted the SAP AI SDK llms.txt libraryId (`/llmstxt/sap_github_io_ai-sdk_llms_txt`, 1305 snippets).

Verified: `tsc` + lint pass; `--frozen-lockfile` clean; local reply.ts run loads `resolve-library-id, query-docs` from the hosted endpoint and cites real issues. Lockfile diff is context7-removal only (canary snapshot churn excluded).

InjunPark-sap:
Widened the issue-reply trigger to all opened issues (was `MEMBER/OWNER/COLLABORATOR` only).

It's an open-source repo — most support questions are from external users, exactly who the old filter skipped. Abuse surface stays bounded: Slack-only, no auto issue reply, so a spam filing costs one model call + one Slack line. Per-issue concurrency guard unchanged.

.github/workflows/issue-reply.yml
```diff
@@ -20,9 +20,10 @@ permissions: {}
 
 jobs:
   reply:
-    # SEC-1: only respond to trusted contributors — blocks quota abuse from anonymous filers
-    # workflow_dispatch is always triggered by an authenticated user (team member)
-    if: github.event_name == 'workflow_dispatch' || contains(fromJSON('["MEMBER","OWNER","COLLABORATOR"]'), github.event.issue.author_association)
+    # Trigger on every opened issue (and manual dispatch). This is an open-source repo, so most
+    # support questions come from external users — restricting to team members would skip the bot's
+    # actual audience. Blast radius of an abusive issue is bounded: the bot posts to Slack only (no
+    # auto issue reply), so a spam filing costs one model call + one Slack line, never a public reply.
```
marikaner: Agreed, but don't leave the comment here.

sample-code/src/support-bot/github-tools.ts
```diff
@@ -0,0 +1,165 @@
+import { tool } from '@langchain/core/tools';
+import { z } from 'zod';
+import type { StructuredToolInterface } from '@langchain/core/tools';
+
+// Native-fetch GitHub tools — replaces the deprecated @modelcontextprotocol/server-github.
```
marikaner, line 5: [q] This seems odd to me. The deprecation note states that there seems to be a replacement: https://www.npmjs.com/package/@modelcontextprotocol/server-github can't we use that somehow? (=>https://github.com/github/github-mcp-server)

.github/workflows/issue-reply.yml
InjunPark-sap: Done. dropped the comment.

sample-code/src/support-bot/github-tools.ts
InjunPark-sap: The successor is `github/github-mcp-server` which is a Go binary + hosted endpoint, not a Node drop-in. 

The hosted `/readonly` URL fits our transport but has no single-repo scoping (only toolset/token), whereas this module hardcodes `repo:SAP/ai-sdk-js` so the model can't reach other repos. 

Keeping that would need a fine-grained single-repo PAT — for a read-only Slack-only bot the 166-LOC module is the smaller bet.

.github/workflows/issue-reply.yml
KavithaSiva, line 63: [q] Which slack channel does this get posted to?

KavithaSiva (commented):
I would like to try out alternate implementation with opencode that could simplify the agent loop and uses github mcp server instead. I will open another PR, and then we can evaluate.

.github/workflows/issue-reply.yml
InjunPark-sap: #ai-sdk-js-support

PR #2173, opened 2026-08-18
chore: Update model list (2026-08-18)

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#ISSUENUMBER..


packages/core/src/model-types.ts
```diff
@@ -96,4 +95,6 @@ export type AiCoreOpenSourceEmbeddingModel =
 /**
  * SAP RPT models.
  */
-export type SapRptModel = LiteralUnion<'sap-rpt-1-small' | 'sap-rpt-1-large'>;
+export type SapRptModel = LiteralUnion<
+  'sap-rpt-1-small' | 'sap-rpt-1-large' | 'sap-rpt-1.5' | 'sap-rpt-1.5-large'
```
davidkna-sap, line 99: We might want to skip this until we have full support, on the other hand RPT 1.5 should be a superset of RPT 1 so it shouldn't break anything. @marikaner?

BrigittaK307 (dismissed):
Besides the question regarding the RPT models, it LGTM.

packages/core/src/model-types.ts
marikaner: we can merge it i think.

marikaner (dismissed):

marikaner (approved):

PR #2178, opened 2026-08-20
chore: use branch lock files

marikaner:
While researching solutions to the pnpm merge conflicts we have, I found this: https://pnpm.io/git_branch_lockfiles

I would like to propose trying it out for a while and keep this setting if it doesn't cause problems.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -21,6 +21,10 @@ allowBuilds:
   esbuild: false
   unrs-resolver: false
 
+gitBranchLockfile: true
+mergeGitBranchLockfilesBranchPattern:
+  - main
+
 blockExoticSubdeps: true
 minimumReleaseAge: 5760
 minimumReleaseAgeExclude:
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -24,6 +24,7 @@ allowBuilds:
 gitBranchLockfile: true
 mergeGitBranchLockfilesBranchPattern:
   - main
+  - renovate/*
 
 blockExoticSubdeps: true
 minimumReleaseAge: 5760
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 dist
 dist-cjs
 node_modules
+pnpm-lock.*.yaml
 coverage
 
 # Logs
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 dist
 dist-cjs
 node_modules
-pnpm-lock.*.yaml
 coverage
 
 # Logs


davidkna-sap:
@marikaner I also made a PR for this at some point I cannot find right now, but I think we would also need a workflow on main that actively merges the branch lockfiles.

Edit: https://github.com/SAP/cloud-sdk-js/commit/48f453f87

pnpm-workspace.yaml
```diff
@@ -21,6 +21,10 @@ allowBuilds:
   esbuild: false
   unrs-resolver: false
 
+gitBranchLockfile: true
+mergeGitBranchLockfilesBranchPattern:
+  - main
```
davidkna-sap, line 26: This should ensure that the weekly lockfile updates also merge all the branch lockfiles.

```suggestion
  - main
  - renovate/lock-file-maintenance
```

We could also include all renovate branches to ensure cleanup happens more often and dependency updates don't pollute the main branch as much:

```suggestion
  - main
  - renovate/*
```

marikaner: good idea 👍

marikaner:
> @marikaner I also made a PR for this at some point I cannot find right now, but I think we would also need a workflow on main that actively merges the branch lockfiles.
> 
> Edit: [SAP/cloud-sdk-js@48f453f87](https://github.com/SAP/cloud-sdk-js/commit/48f453f87)

I expected that you already tried it 😊
I thought `mergeGitBranchLockfilesBranchPattern` does that automatically, but forgot that of course that's not merged back to main...

davidkna-sap (approved):
LGTM.
My only concern would be that `main` in `mergeGitBranchLockfilesBranchPattern` could lead to issues with the frozen lockfile setting but we will see, and that would be easy to fix.

marikaner:
> LGTM. My only concern would be that `main` in `mergeGitBranchLockfilesBranchPattern` could lead to issues with the frozen lockfile setting but we will see, and that would be easy to fix.

We still definitely need a workflow that updates the lockfile on main, if there are branch lockfiles, right? without that it is possible that we have multiple branch lockfiles on main at the time of a release.
this could be handled through an auto-merge-lockfile workflow on main. wdyt?

davidkna-sap:
@marikaner This could also be handled along the changelog updates.

InjunPark-sap (approved):
LGTM. Two follow-ups before this settles:

1. Merge workflow on main — as discussed, branch lockfiles will accumulate without it. Suggest tracking this as a follow-up PR (reference:https://github.com/SAP/cloud-sdk-js/commit/48f453f87).
2. CI --frozen-lockfile compat — worth verifying main CI doesn't fail after first branch lockfile is merged in.

KavithaSiva (commented):
I looked into this now, and we have a workflow that also updates pnpm lock files `pnpm-update.yml`.
Since it also creates a branch, the lock file updates might be done in the branch specific lock file, and right now it is not staged.

Either that or we add `bot/monthly-transitive-dependency-update*` pattern to `mergeGitBranchLockfilesBranchPattern`.

davidkna-sap:
> I looked into this now, and we have a workflow that also updates pnpm lock files pnpm-update.yml.
Since it also creates a branch, the lock file updates might be done in the branch specific lock file, and right now it is not staged.

As configured right now every renovate PR including lockfile updates should merge the lockfiles.

KavithaSiva:
> > I looked into this now, and we have a workflow that also updates pnpm lock files pnpm-update.yml.
> > Since it also creates a branch, the lock file updates might be done in the branch specific lock file, and right now it is not staged.
> 
> As configured right now every renovate PR including lockfile updates should merge the lockfiles.

I don't understand, only branches created by renovate merges the lockfile correct? I am talking about the workflow which updates transitive dependencies in lock files, that creates branches of style `bot/monthly-transitive-dependency-update*`

davidkna-sap:
Ok I never removed `.github/workflows/pnpm-update.yml`, but renovate also does weekly refreshes as part of the best-practices preset we use.

KavithaSiva:
> Ok I never removed `.github/workflows/pnpm-update.yml`, but renovate also does weekly refreshes as part of the best-practices preset we use.

Just read upon this, very cool, so we can remove the workflow.

pnpm-workspace.yaml
```diff
@@ -21,6 +21,11 @@ allowBuilds:
...
```
davidkna-sap, line 26: [pp] I would prefer leaving out `main`, as any merged lockfile would not be committed back.
```suggestion
```
Frozen Lockfiles do not appear to cause any issues on investigation.

marikaner: then let's try it out.... 🤞

PR #2180, opened 2026-08-20
chore: add rpt-1.5 adr and poc

marikaner:
## Context

Closes SAP/ai-sdk-js-backlog#632.
Part of SAP/ai-sdk-js-backlog#662.
Closes #2105


adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,112 @@
...
+
+# Open Questions / Notes
+
+- **Do 1.5 deployments use new model names or replace existing ones?** The RPT team needs to confirm whether RPT-1.5 is deployed under new model identifiers (e.g. `sap-rpt-1.5-small`) or reuses the existing `sap-rpt-1-small`/`sap-rpt-1-large` names. If the model names are reused, a per-version package strategy (Option 3) breaks down — the same deployment name would need to be handled by two different packages simultaneously.
+
+- **If old deployments are retired or stop working**, there is no reason to keep old version sub-paths in the package. In that case Options 2 and 3 simplify significantly.
+
+- **Is `prediction_config` as a JSON string in `predict_parquet` intentional?** In 1.5 the spec changes `prediction_config` in the parquet multipart form from a structured object to a JSON-encoded string. This may be a mistake in the spec — needs confirmation from the RPT team before deciding how the SDK should handle it. If intentional, the SDK should serialize transparently in `predictParquet()` so consumers are not exposed to the wire format change.
```
marikaner: @cbiehlsap

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,112 @@
...
+Cons:
+
+- Subpath exports are less discoverable than top-level package names — consumers need to know to look for `@sap-ai-sdk/rpt/1.5`.
+- Internal package complexity grows with each version: multiple specs, multiple generated clients, multiple sets of hand-written types all in one package.
+- The RPT team expects further breaking changes — each new version still requires duplicating the full client stack inside the package.
+- Requires careful `exports` and `typesVersions` configuration to ensure TypeScript resolves subpath types correctly.
+
+# Open Questions / Notes
+
+- **Do 1.5 deployments use new model names or replace existing ones?** The RPT team needs to confirm whether RPT-1.5 is deployed under new model identifiers (e.g. `sap-rpt-1.5-small`) or reuses the existing `sap-rpt-1-small`/`sap-rpt-1-large` names. If the model names are reused, a per-version package strategy (Option 3) breaks down — the same deployment name would need to be handled by two different packages simultaneously.
```
marikaner: @cbiehlsap

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,112 @@
...
+## Context
+
+The current `@sap-ai-sdk/rpt` package targets RPT-1.0. RPT-1.5 introduces breaking changes at the spec level:
+
+1. `POST /predict_parquet`: `prediction_config` changes from a structured object to a JSON-serialized string on the wire.
+2. `POST /predict_parquet`: `parse_data_types` default flips from `true` to `false`.
+3. `ColumnType` enum expands from 3 to 16 values — any exhaustive switch/conditional in consumer code gains unhandled cases.
+4. `TargetColumnConfig` and `SchemaFieldConfig` gain `additionalProperties: false` — extra properties previously ignored now cause validation errors.
+
+Additionally, the RPT team has signaled that further breaking changes are expected in future versions (e.g. RPT-1.6 is already in progress). We need to decide how to structure our package(s) going forward.
```
marikaner:
```suggestion
Additionally, the RPT team has signaled that further breaking changes are expected in future versions. We need to decide how to structure our package(s) going forward.
```

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Status
+
+proposed
+
+## Context
+
+The current `@sap-ai-sdk/rpt` package targets RPT-1.0. RPT-1.5 introduces breaking changes at the spec level:
+
+1. `POST /predict_parquet`: `prediction_config` changes from a structured object to a JSON-serialized string on the wire.
```
davidkna-sap: I think this was only part of our specification edits, as it is part of `FormData` for the parquet endpoint, this should be equivalent to what was there before.

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,112 @@
...
+## New Features in 1.5
+
+<!-- prettier-ignore -->
+| Feature | SDK handling | Notes |
+| ------- | ------------ | ----- |
+| Explainability support (`PredictionConfig.explanations`, `PredictResponsePayload.explanations`) | Expose generated types as-is | Pass `explanations` in `prediction_config`; read from response |
+| `PredictionResult.confidence_interval` | Expose generated types as-is | `[number, number] \| null`; regression only |
+| `GET /health` endpoint | Expose via generated `RptApi` | Internal client only |
+| `TargetColumnConfig.prediction_placeholder` accepts `null` | No action | Additive |
+| `Content-Encoding: gzip` formally documented on `/predict` | No action — compression middleware already handles this | Additive |
```
davidkna-sap, line 52: This was requested by Java AFAIK.

adr/008-rpt-versioning-strategy.md
davidkna-sap: The SAP note 343776 lists it as with new model identifiers.

adr/008-rpt-versioning-strategy.md
davidkna-sap: I think this was only part of our specification edits, as it is a FormData field for the parquet endpoint which will always be a "string" on the wire, this should be equivalent to what was there before.

davidkna-sap (commented):
I would be somewhat in favour of Option 2 or Option 3. To decrease duplication we could have some sort of base class with some level of generics, and it would be easier to deprecate old clients, and it could avoid support issues related to people using new features with old models, which are still more common than with OpenAI models.
I feel like using per-model generics could lead to issues down the line.

If the current client still has few users I feel like Option 1 would be fine too (for this bump at least).

marikaner:
> I would be somewhat in favour of Option 2 or Option 3. To decrease duplication we could have some sort of base class with some level of generics, and it would be easier to deprecate old clients, and it could avoid support issues related to people using new features with old models, which are still more common than with OpenAI models. I feel like using per-model generics could lead to issues down the line.
> 
> If the current client still has few users I feel like Option 1 would be fine too (for this bump at least).

Very interesting, I now finalized the ADR and gave a different recommendation for the decision. Given my new understanding of the situation, I found that the other options are not really viable, but I am open to discuss.

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,91 @@
...
+
+The model name is always **required** — no default value is provided. The model name determines which generation of RPT behavior the SDK applies.
+
+The TypeScript type for the model name is a **strict union of known strings**:
+
+```ts
+'sap-rpt-1-small' | 'sap-rpt-1-large' | 'sap-rpt-1.5' | 'sap-rpt-1.5-large'
+```
+
+The 1.0 names (`sap-rpt-1-small`, `sap-rpt-1-large`) are marked `@deprecated` to nudge consumers toward 1.5 without breaking their build.
```
davidkna-sap: IIRC this does not work as-is and we would need something like this:

```ts
const RptModels = {
  /** @deprecated Since v1.5. Use 'sap-rpt-1.5' instead. */
  'rpt-1-small': 'sap-rpt-1-small',
  /** @deprecated Since v1.5. Use 'sap-rpt-1.5-large' instead. */
  'rpt-1-large': 'sap-rpt-1-large',
  'rpt-1.5': 'sap-rpt-1.5',
  'rpt-1.5-large': 'sap-rpt-1.5-large',
} as const
```

and users would use `RptModel['rpt-1']`. Another way might be to deprecate certain model names in the constructor etc via overloads.

marikaner: yeah, or let's just not do this at all and reflect it in a comment or docs 😊

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,91 @@
...
+
+The SDK infers the RPT generation from the model name at runtime. For RPT-1.0 model names the SDK silently injects `parse_data_types: true` as a request default (restoring the RPT-1.0 server default, which flipped to `false` in 1.5). An explicit user-supplied value always takes precedence over this injection.
+
+All TypeScript request/response types are the RPT-1.5 types regardless of model name. RPT-1.0 consumers who pass 1.5-only fields (e.g. `explanations`) will compile successfully but receive a server-side error.
+
+### Breaking Changes and SDK Handling
```
marikaner: @cbiehlsap

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,91 @@
...
+- Follows the precedent set by other packages in this repo (e.g. `@sap-ai-sdk/foundation-models` ships all Azure OpenAI versions under one package).
+- Consumers on RPT-1.0 deployments can upgrade the SDK without migrating their deployment immediately.
+
+Cons:
+
+- The package must maintain version-detection and request-patching logic until the RPT-1.0 retirement date (no earlier than 2026-12-31).
+
+### RPT-1.0 Compatibility Behavior
+
+The model name is always **required** — no default value is provided. The model name determines which generation of RPT behavior the SDK applies.
```
marikaner:
```suggestion
The model name is always **required** — no default value is provided.
The model name determines which generation of RPT behavior the SDK applies.
```

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,91 @@
...
+
+The TypeScript type for the model name is a **strict union of known strings**:
+
+```ts
+'sap-rpt-1-small' | 'sap-rpt-1-large' | 'sap-rpt-1.5' | 'sap-rpt-1.5-large'
+```
+
+The 1.0 names (`sap-rpt-1-small`, `sap-rpt-1-large`) are marked `@deprecated` to nudge consumers toward 1.5 without breaking their build.
+
+The SDK infers the RPT generation from the model name at runtime. For RPT-1.0 model names the SDK silently injects `parse_data_types: true` as a request default (restoring the RPT-1.0 server default, which flipped to `false` in 1.5). An explicit user-supplied value always takes precedence over this injection.
```
marikaner:
```suggestion
The SDK infers the RPT generation from the model name at runtime.
For RPT-1.0 model names the SDK silently injects `parse_data_types: true` as a request default (restoring the RPT-1.0 server default, which flipped to `false` in 1.5).
An explicit user-supplied value always takes precedence over this injection.
```

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,91 @@
...
+
+All TypeScript request/response types are the RPT-1.5 types regardless of model name. RPT-1.0 consumers who pass 1.5-only fields (e.g. `explanations`) will compile successfully but receive a server-side error.
```
marikaner:
```suggestion
All TypeScript request/response types are the RPT-1.5 types regardless of model name.
RPT-1.0 consumers who pass 1.5-only fields (e.g. `explanations`) will compile successfully but receive a server-side error.
```

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,95 @@
...
+The TypeScript type for the model name is a **strict union of known strings**:
+
+```ts
+'sap-rpt-1-small' | 'sap-rpt-1-large' | 'sap-rpt-1.5' | 'sap-rpt-1.5-large'
+```
+
+The 1.0 names (`sap-rpt-1-small`, `sap-rpt-1-large`) are marked `@deprecated` to nudge consumers toward 1.5 without breaking their build.
+
+The SDK infers the RPT generation from the model name at runtime.
+For RPT-1.0 model names the SDK silently injects `parse_data_types: true` as a request default (restoring the RPT-1.0 server default, which flipped to `false` in 1.5).
```
marikaner:
```suggestion
For RPT-1.0 model names the SDK injects `parse_data_types: true` as a request default (restoring the RPT-1.0 server default, which flipped to `false` in 1.5).
```

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,95 @@
...
+The model name is always **required** — no default value is provided.
+The model name determines which generation of RPT behavior the SDK applies.
+
+The TypeScript type for the model name is a **strict union of known strings**:
+
+```ts
+'sap-rpt-1-small' | 'sap-rpt-1-large' | 'sap-rpt-1.5' | 'sap-rpt-1.5-large'
+```
+
+The 1.0 names (`sap-rpt-1-small`, `sap-rpt-1-large`) are marked `@deprecated` to nudge consumers toward 1.5 without breaking their build.
```
marikaner:
```suggestion
```

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,94 @@
...
+The current `@sap-ai-sdk/rpt` package targets RPT-1.0.
+RPT-1.5 introduces breaking changes (detailed below).
+
+The existing RPT-1.0 model names (`sap-rpt-1-small`, `sap-rpt-1-large`) are not yet deprecated and have a retirement date of no earlier than 2026-12-31, meaning 1.0 and 1.5 deployments will coexist for at least the remainder of 2026.
+
+We need to decide, how to handle such updates in the future.
+
+## Decision
+
+Bump `@sap-ai-sdk/rpt` to support RPT-1.5, issue a major version changeset documenting breaking changes, and maintain backward compatibility for RPT-1.0 deployments within the same package.
```
KavithaSiva: [q] If we make a major bump for this package, aren't we also forced to major bump for all other ai sdk packages too (with the current changeset setting)?

marikaner: You are right, this comment was AI generated and I didn't review it thoroughly enough. There is no need for a major version bump as there are no problematic breaking changes. There is basically only one relevant change, that is easy to workaround through a tiny fix on our part (the default flip from true to false).

KavithaSiva: But we are removing the default ` RptClient()` , isn't that a breaking change ?
But, I think that should be okay to be removed under experimental API.

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,85 @@
...
+- Follows the precedent set by other packages in this repo (e.g. `@sap-ai-sdk/foundation-models` ships all Azure OpenAI versions under one package).
+- Consumers on RPT-1.0 deployments can upgrade the SDK without migrating their deployment immediately.
+
+Cons:
+
+- The package must maintain version-detection and request-patching logic until the RPT-1.0 retirement date (no earlier than 2026-12-31).
+
+### RPT-1.0 Compatibility Behavior
+
+- The model name is **required** and determines the RPT generation (`sap-rpt-1-*` → RPT-1.0 behavior; `sap-rpt-1.5*` → RPT-1.5 behavior).
```
ZhongpinWang: [q/req] I don't understand the word "generation"? Do you mean the actual behaviour in our client will be determined by the model name? Like the `parse_data_type` thing?

ZhongpinWang (commented):
I agree now it makes more sense to just have one spec and one client for all models after reading Clemens message

> Thanks, having a single service spec for models with an almost equivalent API makes a lot of sense to me (but maybe Vikash has another perspective here). Yes, this is the same web service code on server-side and the API is equivalent except for the two added optional parameters I referred to above ("explanations" added with rpt-1.5, "context_mode" added with rpt-1.6; one would need to omit bot of them for any request to RPT-1) - let me know in case you have more questions about these.

Especially after knowing there is just one api server which maps the request to different models, and there will (hopefully) be only one up-to-date spec for RPT service as a whole.

adr/008-rpt-versioning-strategy.md
marikaner: which generation, aka. version of rpt was used. I will rephrase that.

marikaner:
```suggestion
- The model name is **required** and determines the RPT version (`sap-rpt-1-*` → RPT-1.0 behavior; `sap-rpt-1.5*` → RPT-1.5 behavior).
```

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,85 @@
...
+
+Cons:
+
+- The package must maintain version-detection and request-patching logic until the RPT-1.0 retirement date (no earlier than 2026-12-31).
+
+### RPT-1.0 Compatibility Behavior
+
+- The model name is **required** and determines the RPT version (`sap-rpt-1-*` → RPT-1.0 behavior; `sap-rpt-1.5*` → RPT-1.5 behavior).
+
+- For RPT-1.0 model names, the SDK injects `parse_data_types: true` as a request default; explicit values always take precedence. If a `deploymentId` is passed without a `modelName`, generation detection is not possible and the workaround is skipped.
```
marikaner:
```suggestion
- For RPT-1.0 model names, the SDK injects `parse_data_types: true` as a request default; explicit values always take precedence.
If a `deploymentId` is passed without a `modelName`, generation detection is not possible and the workaround is skipped.
```

davidkna-sap:
@marikaner I think there was some issue when you merged main.

adr/008-rpt-versioning-strategy.md
marikaner: Yes, I will reflect this in the compatibility notes. It wasn't very considerate to have a default value in the first place...

marikaner:
> @marikaner I think there was some issue when you merged main.

fixed

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,86 @@
...
+If a `deploymentId` is passed without a `modelName`, generation detection is not possible and the workaround is skipped.
+
+- All TypeScript types are RPT-1.5 types regardless of model name — RPT-1.0 consumers passing 1.5-only fields (e.g. `explanations`) will compile but receive a server-side error.
+
+### Behavioral Changes and SDK Handling
+
+<!-- prettier-ignore -->
+| Change | SDK handling | Consumer impact |
+| ------ | ------------ | --------------- |
+| `predict_parquet`: `parse_data_types` default flips `true` → `false` | Follow the spec | None for RPT-1.0 consumers (SDK injects `parse_data_types: true`); RPT-1.5 consumers default to `false` |
```
davidkna-sap: [q] You mentioned we cannot always reliably detect the model version in runtime? In that case I would rather prefer not injecting any values at the SDK layer.

marikaner: Yes, that's correct. When I noticed this, I was also leaning towards this. @ZhongpinWang do you also agree?

marikaner: changed the implementation.

packages/rpt/src/spec/rpt.json
davidkna-sap, line 1: [pp] Use 1.6 spec? I think it should release rather soon, and IIRC has few changes.

.changeset/rpt-1.5-support.md
davidkna-sap, line 1: Split the changeset into change-per-file, unless the changeset parser supports this.

.changeset/rpt-1.5-support.md
```diff
@@ -0,0 +1,14 @@
...
+[feat] **Explainability.** Add an `explanations` field to `prediction_config`; read feature importance scores per query row from the response.
+
+[feat] **Confidence intervals.** Add a `confidence_interval` field to `PredictResponsePayload` predictions for regression tasks.
+
+[feat] **Health endpoint.** Expose `RptApi.health()` via the internal client.
+
+[feat] **Extended column types.** Expand `ColumnType` from 3 to 16 values (`integer`, `int16`, `int32`, `int64`, `uint8`, `decimal`, `double`, `boolean`, `largestring`, `uuid`, `time`, `datetime`, `timestamp`).
+Map the new numeric variants to `number`; all other new variants (including `datetime` and `timestamp`, which are full ISO strings) map to `string`.
+
+[feat] **New model names.** Add `'sap-rpt-1.5'` and `'sap-rpt-1.5-large'` to `SapRptModel`.
```
davidkna-sap: Remove if #2173 is merged first.

marikaner:
```suggestion
```

packages/core/src/model-types.ts
```diff
@@ -93,10 +93,10 @@ export type AiCoreOpenSourceEmbeddingModel =
   LiteralUnion<'nvidia--llama-3.2-nv-embedqa-1b'>;
 
 /**
- * SAP RPT models.
+ * All SAP RPT models.
```
marikaner:
```suggestion
 * SAP RPT models.
```

packages/rpt/src/spec/rpt.json
marikaner: i just noticed that we are not exposing the health endpoint. i guess we should still do that, right...

davidkna-sap: It doesn't seem to be exposed.

.changeset/rpt-1.5-require-model-name.md
```diff
@@ -0,0 +1,15 @@
...
+Remove the default model name.
+Pass one of the four known model names explicitly:
+
+```ts
+// before
+new RptClient();
+
+// after
+new RptClient('sap-rpt-1.5');
+```
```
davidkna-sap: Avoid complex formatting, with RPT-1.6 it will soon be 6 models:
```suggestion
Pass one of the known model names explicitly, e.g. `'sap-rpt-1.5'`.
```

packages/rpt/src/spec/rpt.json
ZhongpinWang: Normally, `health` endpoint is just for AI Core internal cluster only. I think the spec later needs further improvement and the service team should know it is a public facing spec.

marikaner: so, should we not expose it?

ZhongpinWang: no we should not expose it and RPT service should remove it from public spec. The endpoint is also not reachable from the deployment url.

packages/rpt/src/types.ts
```diff
@@ -26,14 +28,50 @@ export type DateString =
...
+
+/**
+ * Represents a timestamp as a numeric string with up to 7 digits (µs precision).
+ */
+export type TimestampString = `${number}`;
+
+/**
+ * Represents a UUID string in the RFC 4122 format xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx.
+ */
+export type UuidString = `${string}-${string}-${string}-${string}-${string}`;
```
marikaner, line 49: @cbiehlsap

packages/core/src/model-types.ts
```diff
@@ -94,9 +94,10 @@ export type AiCoreOpenSourceEmbeddingModel =
 
 /**
  * SAP RPT models.
+ * @remarks `sap-rpt-1-small` and `sap-rpt-1-large` are deprecated and will be removed no earlier than 2026-12-31. Use `sap-rpt-1.5` or `sap-rpt-1.5-large` instead.
  */
 export type SapRptModel = LiteralUnion<
-  'sap-rpt-1-small' | 'sap-rpt-1-large' | 'sap-rpt-1.5' | 'sap-rpt-1.5-large'
+  'sap-rpt-1.5' | 'sap-rpt-1.5-large' | 'sap-rpt-1-small' | 'sap-rpt-1-large'
```
davidkna-sap, line 100: Not a blocker, but [usually](https://github.com/SAP/ai-sdk-js/blob/main/scripts/sync-model-types.ts#L117-L138](https://github.com/SAP/ai-sdk-js/blob/main/scripts/sync-model-types.ts#L140-L155)) we remove deprecated models fully.

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,86 @@
...
+
+Cons:
+
+- The package must maintain version-detection and request-patching logic until the RPT-1.0 retirement date (no earlier than 2026-12-31).
+
+### RPT-1.0 Compatibility Behavior
+
+- The model name is **required** and determines the RPT version (`sap-rpt-1-*` → RPT-1.0 behavior; `sap-rpt-1.5*` → RPT-1.5 behavior).
+
+- For RPT-1.0 model names, the SDK injects `parse_data_types: true` as a request default; explicit values always take precedence.
```
KavithaSiva: [req] This is no longer relevant correct? Consumers have to explicitly pass in this field?

packages/rpt/src/types.ts
```diff
@@ -109,7 +131,7 @@ export type PredictionData<T extends DataSchema> = {
   index_column?: ColumnNames<T>;
   /**
    * Whether to parse the data types of the columns. If set to True, numeric columns will be parsed to float or integer and dates in ISO format YYYY-MM-DD will be parsed.
-   * Default: true.
+   * The RPT-1.5 server default is `false`. The SDK automatically uses `true` for RPT-1.0 model names unless explicitly overridden.
```
KavithaSiva: [req] Change this as we don't pass `true` for RPT-1.0 model.

packages/core/src/model-types.ts
marikaner: I think my comment is misleading: The models are not deprecated from the service owners. I wanted to mark them `@deprecated` in the sdk. Technically this is not possible, so I tried to reflect it in the comment. I could remove this comment (and changeset) altogether. WDYT?

packages/core/src/model-types.ts
```diff
@@ -94,9 +94,10 @@ export type AiCoreOpenSourceEmbeddingModel =
 
 /**
  * SAP RPT models.
+ * @remarks `sap-rpt-1-small` and `sap-rpt-1-large` are deprecated and will be removed no earlier than 2026-12-31. Use `sap-rpt-1.5` or `sap-rpt-1.5-large` instead.
```
marikaner:
```suggestion
 * @remarks `sap-rpt-1-small` and `sap-rpt-1-large` are deprecated and will be removed 2026-12-31. Use `sap-rpt-1.5` or `sap-rpt-1.5-large` instead.
```

.changeset/rpt-1.5-deprecate-old-models.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-ai-sdk/rpt': minor
+---
+
+[compat] Deprecate `sap-rpt-1-small` and `sap-rpt-1-large`.
+These model names remain functional until their retirement date (no earlier than 2026-12-31).
```
marikaner:
```suggestion
These model names remain functional until their retirement date (2026-12-31).
```

adr/008-rpt-versioning-strategy.md
marikaner:
```suggestion
```

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,84 @@
...
+- Consumers on RPT-1.0 deployments can upgrade the SDK without migrating their deployment immediately.
+
+Cons:
+
+- The package must maintain version-detection and request-patching logic until the RPT-1.0 retirement date (no earlier than 2026-12-31).
+
+### RPT-1.0 Compatibility Behavior
+
+- The model name is **required** and determines the RPT version (`sap-rpt-1-*` → RPT-1.0 behavior; `sap-rpt-1.5*` → RPT-1.5 behavior).
+  If a `deploymentId` is passed without a `modelName`, generation detection is not possible and the workaround is skipped.
```
marikaner:
```suggestion
- The model name is **required** and determines the RPT version.
```

adr/008-rpt-versioning-strategy.md
```diff
@@ -0,0 +1,83 @@
...
+
+Cons:
+
+- The package must maintain version-detection and request-patching logic until the RPT-1.0 retirement date (no earlier than 2026-12-31).
+
+### RPT-1.0 Compatibility Behavior
+
+- The model name is **required** and determines the RPT version.
+
+- All TypeScript types are RPT-1.5 types regardless of model name — RPT-1.0 consumers passing 1.5-only fields (e.g. `explanations`) will compile but receive a server-side error.
```
marikaner:
```suggestion
- Breaking change: The model name is **required** and determines the RPT version.
- All TypeScript types are RPT-1.5 types regardless of model name — RPT-1.0 consumers passing 1.5-only fields (e.g. `explanations`) will compile but receive a server-side error.
```

packages/rpt/src/spec/rpt.json
marikaner: removed 😊

packages/rpt/src/types.ts
marikaner: This is what happens if you don't review your own AI generated code after each change...😅

marikaner:
```suggestion
```

davidkna-sap:
Please also remove the beta/experimental information from package.json and Readme.md.

packages/core/src/model-types.ts
```diff
@@ -94,20 +94,8 @@ export type AiCoreOpenSourceEmbeddingModel =
...
- * The full list is managed by running the `update-models` skill.
- */
-export type LlmBatchModel = LiteralUnion<
-  | 'gpt-4.1'
-  | 'gpt-4.1-mini'
-  | 'gpt-4.1-nano'
-  | 'gpt-5'
-  | 'gpt-5.4'
-  | 'gpt-5.4-mini'
+  'sap-rpt-1.5' | 'sap-rpt-1.5-large' | 'sap-rpt-1-small' | 'sap-rpt-1-large'
```
marikaner, line 100:
```suggestion
  'sap-rpt-1.5' | 'sap-rpt-1.5-large' | 'sap-rpt-1-small' | 'sap-rpt-1-large'
>;

/**
 * Models supported for LLM batch processing.
 * The full list is managed by running the `update-models` skill.
 */
export type LlmBatchModel = LiteralUnion<
  | 'gpt-4.1'
  | 'gpt-4.1-mini'
  | 'gpt-4.1-nano'
  | 'gpt-5'
  | 'gpt-5.4'
  | 'gpt-5.4-mini'
```

packages/rpt/src/client.ts
```diff
@@ -155,6 +147,53 @@ export class RptClient {
...
+        columns: Object.fromEntries<(string | number | null)[]>(
+          Object.entries(columns).map(([k, vals]) => [
+            k,
+            vals.map(serializeBooleans)
+          ])
+        )
+      } satisfies PredictRequestPayload;
+    }
+
+    return base as unknown as PredictRequestPayload;
```
davidkna-sap: [nth]
```suggestion
    return {
      ...base,
      columns: Object.fromEntries<(string | number | null)[]>(
        Object.entries(columns || {}).map(([k, vals]) => [
          k,
          vals.map(serializeBooleans)
        ])
      )
    };
```

sample-code/src/rpt.ts
davidkna-sap, line 1: [nth] Sample-code with `top_k`

davidkna-sap (dismissed):

packages/rpt/src/client.ts
marikaner: Yes, `satisfies` makes no sense here. I fixed, but slightly differently.

sample-code/src/rpt.ts
marikaner: done.

packages/rpt/src/client.test.ts
```diff
@@ -113,4 +120,150 @@ describe('rpt', () => {
     expect(requestScope.isDone()).toBe(true);
     expect(result.predictions).toEqual([{ SALESGROUP: 'test' }]);
   });
+
+  it('should respect explicit parse_data_types: false for RPT-1.0 models', async () => {
```
davidkna-sap: [pp] Remove or simplify `parse_data_types`-related tests.

davidkna-sap (approved):

PR #2182, opened 2026-08-20
chore(support-bot): rewrite with opencode and remote GitHub MCP

KavithaSiva:
## Context

Rewrites the support bot workflow with `opencode `
Also moves the license from cloud sdk repo to in-house.


sample-code/src/support-bot/opencode.json
```diff
@@ -0,0 +1,24 @@
...
+      "name": "SAP AI Core"
+    }
+  },
+  "mcp": {
+    "github": {
+      "type": "remote",
+      "url": "https://api.githubcopilot.com/mcp/readonly",
+      "headers": {
+        "Authorization": "Bearer {env:GITHUB_TOKEN}",
+        "X-MCP-Toolsets": "repos,issues,code_search"
```
davidkna-sap, line 17: `X-MCP-Lockdown` is also interesting but would require a gh token with write access.

sample-code/src/support-bot/opencode.json
```diff
@@ -0,0 +1,24 @@
...
+  "provider": {
+    "sap-ai-core": {
+      "name": "SAP AI Core"
+    }
+  },
+  "mcp": {
+    "github": {
+      "type": "remote",
+      "url": "https://api.githubcopilot.com/mcp/readonly",
+      "headers": {
```
davidkna-sap, line 15: [q] I wonder if some MCP proxy/gateway would make sense here to further filter any tool calls, even though the risk is lower with slack-only.
Either way, I assume opencode has shell access so either way hard to secure without a real (container-based or likewise) sandbox.
But it might still be able smuggle the AI core credentials if it wants to, https://github.com/SAP/ai-sdk-js-backlog/issues/583 would reduce the risk a bit here, I assume AI core will get IAS soon?

sample-code/src/support-bot/reply.ts
```diff
@@ -1,15 +1,241 @@
...
+import { fileURLToPath } from 'node:url';
+import { SDK_KNOWLEDGE } from './knowledge.ts';
 
-// GitHub Action passes title, body, and issue number as separate args
-const title = process.argv[2];
-const rawBody = process.argv[3] ?? '';
-const issueNumber = process.argv[4]
-  ? Number.parseInt(process.argv[4], 10)
-  : undefined;
+const __dirname = dirname(fileURLToPath(import.meta.url));
```
davidkna-sap:
```suggestion
const __dirname = import.meta.dirname;
```

sample-code/src/support-bot/reply.ts
```diff
@@ -39,7 +265,6 @@ function extractErrorMessages(body: string): string[] {
 }
 
 function stripBoilerplate(body: string): string {
```
davidkna-sap, line 257: I assume opencode has native ways to handle large inputs, consider providing the issue and so on as files.

.github/workflows/issue-reply.yml
```diff
@@ -35,18 +35,22 @@ jobs:
         with:
           registry-token: ${{ secrets.NPM_TOKEN_ARTIFACTORY }}
 
+      - name: Install support-bot deps
+        working-directory: ./sample-code/src/support-bot
+        run: pnpm install
+
```
davidkna-sap: This shouldn't be necessary:

```suggestion
```

.github/workflows/issue-reply.yml
```diff
@@ -35,18 +35,22 @@ jobs:
...
+        working-directory: ./sample-code/src/support-bot
         env:
           AICORE_SERVICE_KEY: ${{ secrets.AI_CORE_PRODUCTION }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           ISSUE_TITLE: ${{ github.event.inputs.issue_title || github.event.issue.title }}
           ISSUE_BODY: ${{ github.event.inputs.issue_body || github.event.issue.body }}
           ISSUE_NUMBER: ${{ github.event.inputs.issue_number || github.event.issue.number }}
         run: |
-          pnpm exec node src/support-bot/reply.ts "$ISSUE_TITLE" "$ISSUE_BODY" "$ISSUE_NUMBER" > /tmp/answer.md
+          node reply.ts "$ISSUE_TITLE" "$ISSUE_BODY" "$ISSUE_NUMBER"
```
davidkna-sap: You handled this separately, but previously `pnpm` ensure any tools (e.g. opencode) are in path.

.github/workflows/issue-reply.yml
```diff
@@ -35,18 +35,22 @@ jobs:
         with:
           registry-token: ${{ secrets.NPM_TOKEN_ARTIFACTORY }}
```
davidkna-sap:
```suggestion
```

sample-code/src/support-bot/reply.ts
```diff
@@ -1,15 +1,241 @@
...
+  SDK_KNOWLEDGE,
+  '',
+  '## Tools available',
+  '- context7__query-docs  — search official SAP AI SDK documentation (libraryId: "' +
+    LIBRARY_ID +
+    '")',
+  '- github__search_issues — search GitHub issues (pass keywords only; repo scope is automatic)',
+  '- github__get_issue     — fetch full body of a specific issue by number',
+  '- github__search_code   — search code examples (pass keywords only; repo scope is automatic)',
+  '',
```
davidkna-sap, line 47: [pp] Also tell the model to look at the local checkout?

sample-code/src/support-bot/reply.ts
```diff
@@ -1,15 +1,241 @@
...
+  '  the same API/method, error, or feature; if you cannot confirm, omit it — never guess. A',
+  '  shared broad theme (e.g. "model parameters", "credentials", "error handling") is NOT enough —',
+  '  the specific API/method/error/feature must match. You',
+  '  MUST have run github__search_issues before this section; only write "No related issues found."',
+  '  if a search genuinely returned nothing on-topic.',
+  '  Do NOT include dependency bumps, unrelated chore PRs, or the issue being answered itself.'
+].join('\n');
+
+// Escape {{ to prevent potential template parsing issues in system prompts
+const esc = (s: string) => s.replaceAll('{{', '{ {');
```
davidkna-sap: [pp] Please verify if this is needed. OpenCode should already handle this using a similar approach.

sample-code/src/support-bot/reply.ts
```diff
@@ -81,7 +305,18 @@ function parseIssueBody(body: string) {
...
 }
 
-// ── Main ──────────────────────────────────────────────────────────────────────
+// ── Entry point ───────────────────────────────────────────────────────────────
+
+const issueTitle = process.argv[2];
+const rawBody = process.argv[3] ?? '';
+const issueNumber = process.argv[4]
+  ? Number.parseInt(process.argv[4], 10)
+  : undefined;
```
davidkna-sap: [nth] Fetch issue body/title from API using the number.

sample-code/src/support-bot/reply.ts
```diff
@@ -1,15 +1,241 @@
...
+  const binDir = resolve(__dirname, '../../../node_modules/opencode-ai/bin');
+  const exePath = resolve(binDir, 'opencode.exe');
+  const binPath = resolve(binDir, 'opencode');
+  if (existsSync(exePath) && !existsSync(binPath)) {
+    symlinkSync('opencode.exe', binPath);
+    console.log('[support-bot] created opencode symlink in bin/');
+  }
+  if (!process.env.PATH?.includes(binDir)) {
+    process.env.PATH = `${binDir}:${process.env.PATH ?? ''}`;
+  }
```
davidkna-sap: [pp] This wouldn't work on windows as it uses a different $path separator, and I think it is safe to assume `node_modules/.bin` has the opencode binary (if executed via `pnpm exec` it would be in `$PATH`), also consider `join(require.resolve('opencode-ai'), 'bin')`. I am not sure if that impacts node, but the variable is `Path` and not `PATH` on windows, and symlinks could lead to issues if a user has not enabled developer mode.

sample-code/src/support-bot/reply.ts
```diff
@@ -1,15 +1,241 @@
...
-const issueNumber = process.argv[4]
-  ? Number.parseInt(process.argv[4], 10)
-  : undefined;
+const __dirname = dirname(fileURLToPath(import.meta.url));
 
-if (!title) {
-  console.error('Usage: tsx src/support-bot/reply.ts "<title>" ["<body>"]');
-  process.exit(1);
+const PROVIDER_ID = 'sap-ai-core';
+const MODEL_ID = 'anthropic--claude-4.6-sonnet';
```
davidkna-sap: [nth] Might be a better fit, it is fairly cheap but capable.
```suggestion
const MODEL_ID = 'gpt-5.6-luna';
```

.github/workflows/issue-reply.yml
KavithaSiva: Was a leftover from when I had a separate `package.json` file, removed now.

sample-code/src/support-bot/opencode.json
KavithaSiva: From what I read and understood the lockdown mode filters content from external contributors who don't have push access. But we need to read issues created by everyone, so I don't think it's useful here. 
Please explain if otherwise.

davidkna-sap: Okay, that makes sens.

sample-code/src/support-bot/opencode.json
KavithaSiva: 1.  I have added `"permission": { "bash": "deny" }` to `opencode.json` to close the most direct exfiltration channel. The bot doesn't need shell access, it only uses context7 and GitHub MCP tools.

1. The GitHub MCP endpoint is /readonly, which limits write channels.

1.  Output is Slack-only, so blast radius is limited.

KavithaSiva: I am not sure about the timeline of AI Core with IAS.

sample-code/src/support-bot/reply.ts
KavithaSiva: I have replaced the hardcoded path and dropped the symlink.

sample-code/src/support-bot/reply.ts
KavithaSiva: Can be improved in a follow-up.

sample-code/src/support-bot/reply.ts
KavithaSiva: Opencode has a way to provide files as attachments, but I would prefer to do this in a follow-up.

sample-code/src/support-bot/reply.ts
KavithaSiva: Removed now, as not needed.

davidkna-sap (commented):
still issues spawning in CI: https://github.com/SAP/ai-sdk-js/actions/runs/33169168848/job/98841805850

package.json
```diff
@@ -36,6 +36,9 @@
     "sample-cap": "pnpm -F=@sap-ai-sdk/sample-cap run",
     "check:deps": "knip --config knip.config.mjs -W '!.' -W '!./sample-cap'"
   },
+  "pnpm": {
+    "onlyBuiltDependencies": ["opencode-ai"]
+  },
```
davidkna-sap: [req] Move to pnpm-workspace file, this position is deprecated.

KavithaSiva: There is `opencode-ai:true` in allowBuilds in pnpm-workspace, so removing this.

KavithaSiva:
> still issues spawning in CI: https://github.com/SAP/ai-sdk-js/actions/runs/33169168848/job/98841805850

Runs successfully now: https://github.com/SAP/ai-sdk-js/actions/runs/33186979692

sample-code/src/support-bot/install-opencode.mjs
davidkna-sap, line 1: [pp] Create BLI to investigate if we can avoid this.

.github/actions/check-license/check-license.ts
```diff
@@ -0,0 +1,90 @@
...
+interface BlueOakCategory {
+  name: string;
+  licenses: BlueOakLicense[];
+}
+interface PackageInfo {
+  name: string;
+  versions: string[];
+}
+
+const blueOakList: BlueOakCategory[] = require('@blueoak/list');
```
davidkna-sap, line 20: [nth/q] Any reason `import blueOakList from '@blueoak/list/index.json' with { type: 'json' };` doesn't work?

.github/workflows/build.yml
```diff
@@ -81,7 +81,7 @@ jobs:
           - name: Check dependencies
             run: pnpm check:deps --reporter=github-actions
           - name: License Check
-            uses: sap/cloud-sdk-js/.github/actions/check-license@main
+            uses: ./.github/actions/check-license
```
davidkna-sap, line 84: [pp] Please try this, though this may not work:
```suggestion
            uses: $/.github/actions/check-license
```
Otherwise please silence the zizmor warning.

knip.config.mjs
```diff
@@ -6,14 +6,10 @@ export default {
...
-        'src/support-bot/agent.ts'
+        'resources/generate-parquet.ts'
       ],
-      // spawned as CLI processes by the MCP client, so knip can't trace them
-      ignoreDependencies: [
-        '@modelcontextprotocol/server-github',
-        '@upstash/context7-mcp'
-      ]
+      // spawned as CLI processes, so knip can't trace the import
+      ignoreDependencies: ['opencode-ai']
```
davidkna-sap, line 12: [q] Knip says it can trace this in the logs, please confirm.

davidkna-sap (commented):
I would prefer to have the new zizmor warning/lint handled, the other stuff could also be handled in follow-ups.

PR #2184, opened 2026-08-21
chore: Add ADRs for tabular orchestration SDKs

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#643. See ticket for details.


adr/009-tabular-orchestration-release-strategy.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Context
+
+Providing SDK clients for the Tabular Orchestration APIs is a significant undertaking.
+The work consists of two main public surfaces:
+
+- the prediction client, generated from one OpenAPI specification; and
+- the context-registry client, generated from nine specifications covering synchronous and asynchronous operations.
+
+Neither client is strictly required to implement the other. The prediction client can accept existing scenario configuration names, while the context-registry client is useful independently for managing data destinations, tabular artifacts, and scenarios.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Neither client is strictly required to implement the other.
The prediction client can accept existing scenario configuration names, while the context-registry client is useful independently for managing data destinations, tabular artifacts, and scenarios.
```

adr/009-tabular-orchestration-release-strategy.md
```diff
@@ -0,0 +1,96 @@
...
+However, an end-to-end workflow commonly uses both: create or discover context, configure a scenario, and submit a prediction.
+
+Release order and implementation order are separate decisions. The prediction and context-registry clients can be tracked, implemented, reviewed, and released as independent efforts.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Release order and implementation order are separate decisions.
The prediction and context-registry clients can be tracked, implemented, reviewed, and released as independent efforts.
```

adr/009-tabular-orchestration-release-strategy.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Decision
+
+No decision has been accepted yet.
+
+The current recommendation is **Option D**: create independent tickets for the prediction and context-registry clients and release each client as soon as it satisfies its own release criteria. Neither client should wait for the other solely to preserve a predetermined release order.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
The current recommendation is **Option D**: create independent tickets for the prediction and context-registry clients and release each client as soon as it satisfies its own release criteria.
Neither client should wait for the other solely to preserve a predetermined release order.
```

adr/009-tabular-orchestration-release-strategy.md
```diff
@@ -0,0 +1,96 @@
...
+
+#### Option B: Context Registry First
+
+Release the context-registry client before the prediction client.
+
+**Pros:**
+
+- Establishes resource-management primitives before prediction workflows depend on them.
+- Gives users value for provisioning and managing data destinations, artifacts, and scenarios independently of prediction.
+- Exercises the more complex specification merge and async behavior early.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
- Exercises the more complex specification merge and asynchronous behavior early.
```

adr/009-tabular-orchestration-release-strategy.md
```diff
@@ -0,0 +1,96 @@
...
+
+**Cons:**
+
+- The less mature contract blocks the entire release.
+- Increases the initial review, test, and documentation scope.
+- Delays user feedback that could improve either client independently.
+
+#### Option D: Independent Efforts, Release When Ready
+
+Create separate delivery tickets for the prediction and context-registry clients. Each effort has its own owner, scope, review, tests, documentation, and release criteria. Work may proceed concurrently or at different times depending on available capacity. The first client to become release-ready is released first; the other follows when ready.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Create separate delivery tickets for the prediction and context-registry clients.
Each effort has its own owner, scope, review, tests, documentation, and release criteria.
Work may proceed concurrently or at different times depending on available capacity.
The first client to become release-ready is released first; the other follows when ready.
```

adr/010-context-registry.md
```diff
@@ -0,0 +1,243 @@
...
+- how the specifications are grouped for generation;
+- whether shared schemas are duplicated or unified;
+- which package boundary communicates the stability of the API;
+- how much behavior is added on top of the generated clients;
+- how the clients are released; and
+- how internal API elements are excluded or approved for public use.
+
+## Decision
+
+No decision has been accepted yet. Current recommendations and unresolved questions are recorded below.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
No decision has been accepted yet.
Current recommendations and unresolved questions are recorded below.
```

adr/010-context-registry.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Discussion
+
+### Question 1: Specification Layout
+
+**Recommendation: Option B.2.** Generate one synchronous and one asynchronous client from two merged specifications.
+
+If upstream later provides an authoritative combined specification, replacing a locally merged contract may still cause breaking changes. Keeping the merge deterministic and limiting it to two output surfaces should make that transition easier to assess.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
If upstream later provides an authoritative combined specification, replacing a locally merged contract may still cause breaking changes.
Keeping the merge deterministic and limiting it to two output surfaces should make that transition easier to assess.
```

adr/010-context-registry.md
```diff
@@ -0,0 +1,243 @@
...
+ctx-registry/
+  sync-specification.yaml
+  async-specification.yaml
+```
+
+Definitions used by both variants are included in each merged specification.
+
+**Pros:** Produces only two coherent client surfaces and keeps generation inputs self-contained.
+
+**Cons:** Duplicates shared definitions across the sync and async clients and requires a deterministic merge step.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
**Cons:** Duplicates shared definitions across the sync and asynchronous clients and requires a deterministic merge step.
```

adr/010-context-registry.md
```diff
@@ -0,0 +1,243 @@
...
+
+**Pros:** Avoids local merge logic and keeps the generated SDK aligned with its source of truth.
+**Cons:** Delays SDK availability and has not been provided despite having already been requested some time ago.
+Even if published on API Hub, multiple specifications may still be uploaded there.
+
+### Question 2: Shared Types Across Specifications
+
+**Recommendation: Option C.** The PoC already normalizes structurally identical schemas across synchronous and asynchronous resource pairs, then consolidates identical generated schema files into one shared schema directory.
+
+The merge step considers a schema shared only when its complete referenced-schema closure is also shared. The generation step fails if two variants emit different content for the same schema filename. The full generation pipeline succeeds with the current specifications.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
The merge step considers a schema shared only when its complete referenced-schema closure is also shared. The generation step fails if two variants emit different content for the same schema filename.
The full generation pipeline succeeds with the current specifications.
```

adr/010-context-registry.md
```diff
@@ -0,0 +1,243 @@
...
+
+#### Option B: Use OpenAPI `$ref` to share types across specifications
+
+**Pros:** Removes duplication and makes incompatibilities visible during generation.
+
+**Cons:** Requires upstream specification changes and generator support for cross-file references.
+
+#### Option C: Normalize and consolidate during local generation
+
+Before generation, detect structurally identical sync/async schemas and normalize their component names in both merged specifications. After generation, move identical emitted schema files into a shared directory and rewrite both clients to import from it.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Before generation, detect structurally identical sync/async schemas and normalize their component names in both merged specifications.
After generation, move identical emitted schema files into a shared directory and rewrite both clients to import from it.
```

adr/010-context-registry.md
```diff
@@ -0,0 +1,243 @@
...
+
+**Pros:** Avoids upstream specification changes, has been demonstrated by this PoC, and fails generation when same-named generated schemas differ between variants.
+
+**Cons:** Adds custom merge and post-generation logic. Structural equality cannot prove semantic equivalence, and newly introduced resource pairs must be added to the normalization configuration.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
**Cons:** Adds custom merge and post-generation logic.
Structural equality cannot prove semantic equivalence, and newly introduced resource pairs must be added to the normalization configuration.
```

adr/010-context-registry.md
```diff
@@ -0,0 +1,243 @@
...
+- B.1: `@sap-ai-sdk/rpt/ctx-registry`
+- B.2: `@sap-ai-sdk/tabular-orchestration/rpt/ctx-registry`
+
+**Pros:** Makes current coupling and compatibility expectations explicit.
+
+**Cons:** Requires a rename or sibling package if the service contract becomes more provider-neutral.
+
+### Question 4: Convenience Methods
+
+Generated SDKs are low-level by nature. The question is whether to layer convenience methods on top, particularly for polling asynchronous operations.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Generated SDKs are low-level by nature.
The question is whether to layer convenience methods on top, particularly for polling asynchronous operations.
```

adr/010-context-registry.md
```diff
@@ -0,0 +1,243 @@
...
+
+Ship only the generated client as-is.
+
+**Pros:** Minimizes maintenance and keeps behavior aligned with the specification.
+
+**Cons:** Repeats polling, retry, and terminal-state handling in every consumer.
+
+#### Option B: Add targeted convenience methods
+
+Add helpers for operations poorly supported by the raw generated client, e.g. polling async jobs to completion.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Add helpers for operations poorly supported by the raw generated client, e.g. polling asynchronous jobs to completion.
```

adr/010-context-registry.md
```diff
@@ -0,0 +1,243 @@
...
+
+- B.1 Postpone convenience methods until a future release, to reduce initial effort and maintenance burden.
+- B.2 Add convenience methods in the initial release, to improve developer experience.
+
+**Pros:** Improves common workflows while keeping the handwritten API small.
+
+**Cons:** Requires an explicit contract for timeouts, retries, cancellation, and error propagation.
+
+The exact helper API is still open. A targeted helper should make terminal and failure states explicit rather than hiding them in an operation-specific loop:
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
The exact helper API is still open.
A targeted helper should make terminal and failure states explicit rather than hiding them in an operation-specific loop:
```

adr/010-context-registry.md
```diff
@@ -0,0 +1,243 @@
...
+
+**Cons:** Makes the release dependent on upstream changes and their schedule.
+
+#### Option C: Expose the specifications unchanged
+
+Generate internal headers and endpoints as part of the public client.
+
+**Pros:** Requires no filtering or upstream changes.
+
+**Cons:** Creates an unsupported public surface and risks exposing internal implementation details. This option is not recommended.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
**Cons:** Creates an unsupported public surface and risks exposing internal implementation details.
This option is not recommended.
```

adr/011-tabular-orchestration-prediciton.md
```diff
@@ -0,0 +1,266 @@
...
+The prediction SDK is generated from one OpenAPI specification, but the checked-in contract and the current Tabular Orchestration v2 documentation are not demonstrably the same contract.
+The checked-in specification identifies itself as version `1.6.0` and already uses `/v1/predict`, but it still:
+
+- generates a request builder that calls `/predict` rather than `/v1/predict`;
+- exposes only `sap-rpt-*` values through the closed `TFMEnum`;
+- includes `contextSelectionConfig.filterConditions`, which is absent from the referenced v2 documentation;
+- places heuristic `indexColumn` under `strategyConfig`, while the v2 documentation also describes a top-level location; and
+- requires the internal `ai-main-tenant` header.
+
+The exact documentation and specification revisions used for this comparison have not been pinned. They must be captured before this ADR is accepted.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
The exact documentation and specification revisions used for this comparison have not been pinned.
They must be captured before this ADR is accepted.
```

adr/011-tabular-orchestration-prediciton.md
```diff
@@ -0,0 +1,266 @@
...
+
+The current client is therefore strongly coupled to an RPT-style contract. Significant changes may be required to support other models, putting both the public client design and package boundary at risk.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
The current client is therefore strongly coupled to an RPT-style contract.
Significant changes may be required to support other models, putting both the public client design and package boundary at risk.
```

adr/011-tabular-orchestration-prediciton.md
```diff
@@ -0,0 +1,252 @@
...
+Expose the generated operation and schema types as the public API.
+Consumers perform deployment discovery, destination adjustment, and header selection explicitly, as in the PoC.
+
+This has the smallest maintenance surface and stays closest to the service contract,
+but repeats AI Core integration code in every application and exposes generated names
+such as `predictV1PredictPost`.
+
+#### Option B: Prediction convenience client
+
+Add a thin client that roughly mirrors the convenience of the current RPT client.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Add a thin client that mirrors the convenience of the current RPT client.
```

adr/011-tabular-orchestration-prediciton.md
```diff
@@ -0,0 +1,252 @@
...
+  model: "other",
+  scenarioConfigName,
+  rows,
+  targets: [{ name: "salesgroup", task: "classification" }],
+});
+```
+
+`rpt` and `otherModel` can expose unrelated provider-specific inputs and outputs.
+The shared interface standardizes only the act of prediction; it does not imply a shared wire request or generated operation.
+Authentication, destination handling, request execution, and generic HTTP behavior still come from the SAP Cloud SDK used by each client.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Authentication, destination handling, request execution, and generic http behavior still come from the SAP Cloud SDK used by each client.
```

adr/011-tabular-orchestration-prediciton.md
```diff
@@ -0,0 +1,252 @@
...
+  rows,
+  targets: [{ name: "salesgroup", task: "classification" }],
+  explanations: {
+    topColumnScores: 3,
+    topRelevantContextRows: 2,
+  },
+});
+```
+
+Here the neutral generated client defines the shared operation and wire schema, while the SAP Cloud SDK provides request building, destination execution, and generic HTTP behavior.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Here the neutral generated client defines the shared operation and wire schema, while the SAP Cloud SDK provides request building, destination execution, and generic http behavior.
```

adr/011-tabular-orchestration-prediciton.md
```diff
@@ -0,0 +1,252 @@
...
+The current client is therefore strongly coupled to an RPT-style contract.
+Significant changes may be required to support other models, putting both the public client design and package boundary at risk.
+The model-name extensibility concern shared by generated clients is tracked in [ADR-009](./009-tabular-orchestration-release-strategy.md).
+
+## Decision
+
+No decision has been accepted yet.
+
+The working direction is conditional: use a neutral generated client with typed provider profiles if the authoritative service contract is provider-neutral; otherwise publish an explicitly RPT-specific client.
+Provider-specific clients may share a small public interface, but do not require a shared orchestration transport because the SAP Cloud SDK already supplies the common HTTP and destination infrastructure.
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Provider-specific clients may share a small public interface, but do not require a shared orchestration transport because the SAP Cloud SDK already supplies the common http and destination infrastructure.
```

adr/011-tabular-orchestration-prediciton.md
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Authentication, destination handling, request execution, and generic http behavior still come from the SAP Cloud SDK used by each client.
```

adr/011-tabular-orchestration-prediciton.md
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Provider-specific clients may share a small public interface, but do not require a shared orchestration transport because the SAP Cloud SDK already supplies the common http and destination infrastructure.
```

adr/011-tabular-orchestration-prediciton.md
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
Here the neutral generated client defines the shared operation and wire schema, while the SAP Cloud SDK provides request building, destination execution, and generic http behavior.
```

adr/011-tabular-orchestration-prediction.md
InjunPark-sap, line 1: file name misspell "prediciton" -> "prediction"

adr/010-context-registry.md
```diff
@@ -0,0 +1,261 @@
...
+
+Release clients incrementally, enabling E2E testing of each before moving to the next.
+
+**Pros:** Enables earlier feedback and isolates integration issues.
+
+**Cons:** Consumers may need multiple release-cycle upgrades, and temporarily incompatible client versions may coexist.
+
+### Question 6: Public Surface Filtering
+
+The owners of the upstream specifications must clean internal headers and endpoints before publication.
```
InjunPark-sap: Unlike Questions 1–5, this section has no recommendation. Is that intentional? 

If the decision is genuinely blocked on upstream spec changes, a short note would make that explicit.

adr/011-tabular-orchestration-prediciton.md
```diff
@@ -0,0 +1,251 @@
+# Tabular Orchestration Prediction SDK
```
marikaner:
```suggestion
# Tabular Orchestration Prediction Client
```

adr/010-context-registry.md
```diff
@@ -0,0 +1,261 @@
...
+- how internal API elements are excluded or approved for public use.
+
+## Decision
+
+No decision has been accepted yet.
+Current recommendations and unresolved questions are recorded below.
+
+## Discussion
+
+### Question 1: Specification Layout
```
davidkna-sap, line 39: I have noticed that overall combined the specifications aren't too big so for now at least manual/agentic merging of the files could be possible, if we assume they eventually provide combined spec.
I will also recheck if `redocly` can be configured to handle the conflicts.

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,251 @@
...
+
+await PredictApi.predictV1PredictPost(request, {
+  'ai-resource-group': resourceGroup
+}).execute(destination);
+```
+
+The neutral portion is the invocation and the basic data contract: a model identifier, scenario, optional context and query rows, and shared target-column configuration.
+The RPT coupling is visible in three places:
+
+- `modelName` is a closed `TFMEnum` containing only `sap-rpt-*` values.
```
marikaner: [q] If currently only SAP RPT is supported, this is no surprise. I wouldn't understand that as "coupling".

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,251 @@
...
+- `modelConfig` forwards arbitrary fields instead of describing a provider-independent capability.
+- The client does not represent explainability consistently: the OpenApi description directs model-specific fields to `modelConfig`.
+- Context selection and target-column semantics are only agnostic if every supported model gives them the same meaning.
+  The current RPT docs document RPT limits, jointly predicted columns, and RPT-specific explainability; Other model providers may not share an equivalent shared contract.
+
+That means changing the enum to accept another string would make the client appear agnostic without making its types or behavior agnostic.
```
marikaner: [q] ~~What enum?~~ Ah, TFMEnum. To make this more clear:
```suggestion
That means, once the TFMEnum other models, it would make the client appear agnostic without making its types or behavior agnostic.
```

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,251 @@
...
+const client = new TabularOrchestrationClient({ destination });
+await client.predict({
+  model: { name: 'sap-rpt-1.5' },
+  scenarioConfigName: 'product-prediction-scenario-lowercase',
+  rows,
+  targets: [{ name: 'salesgroup', task: 'classification' }]
+});
+```
+
+This is the smallest neutral API, but `providerConfig` remains an escape hatch.
```
marikaner: [q] What is the difference between `providerConfig` and `modelConfig`? To me it seems like it is basically the same as the current contract, only hoping that it will be model agnostic. 😊

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,251 @@
...
+It is appropriate for a transport SDK and for newly introduced providers, not a good developer experience for provider-specific features.
```
marikaner: [q] Aren't there cases in the Orchestration service where we have the same issue? I think it might be difficult to harmonize everything.

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,251 @@
...
+and optional features:
+
+```ts
+const client = new TabularOrchestrationClient({ destination });
+await client.predict({
+  model: { name: 'sap-rpt-1.5' },
+  scenarioConfigName,
+  rows,
+  targets: [{ name: 'salesgroup', task: 'classification' }]
+});
```
marikaner: [q] I think this is a copy paste error and should be removed, right?
```suggestion
});
```

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,251 @@
...
+  rows,
+  targets: [{ name: 'salesgroup', task: 'classification' }],
+  explanations: {
+    topColumnScores: 3,
+    topRelevantContextRows: 2
+  }
+});
+```
+
+Here the neutral generated client defines the shared operation and wire schema, while the SAP Cloud SDK provides request building, destination execution, and generic http behavior.
```
marikaner: My intuitive thoughts on this were: definitely no. But in case harmonization turns out to be difficult or not happening, I could imagine such an approach. Then, however, I would like to understand whether it is possible to integrate the existing RPT client.
Possibly this could also be turned around by directly just using the exising client with a specifc configuration.

Just thoughts.

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,251 @@
...
+This preserves provider-specific types and allows polymorphic consumers without introducing a neutral transport abstraction.
+The interface is useful only if such consumers exist; otherwise it adds generic types without reducing implementation work.
+
+#### Option D: Share nothing
+
+Each provider can ship a separate generated client, public API, and package without a shared orchestration interface, request type, or adapter.
+They still reuse the SAP Cloud SDK runtime for destination resolution, request execution, middleware, and error handling.
+
+This is the simplest boundary when provider contracts differ substantially or evolve independently.
+It avoids a lowest-common-denominator abstraction, but consumers that support multiple providers must define their own abstraction and provider switching is not uniform.
```
marikaner: But then, what is the purpose of Tabular AI?

marikaner (commented):
First glance at tabular orchestration prediction only.

My impression is, if we separate clients per model provider, then we shouldn't have extra ones, but reuse the existing one(s). If harmonization is intended by Tabular AI, but not "good enough", we should share that feedback.

I also have the impression, it might be too early to tell.

adr/010-context-registry.md
```diff
@@ -0,0 +1,261 @@
...
+Several of the specifications share type definitions, leading to duplication in the generated output.
+Some specifications also contain internal headers (`AI-Main-Tenant`) and internal endpoint references that we should avoid releasing.
+
+The SDK design must therefore determine:
+
+- how the specifications are grouped for generation;
+- whether shared schemas are duplicated or unified;
+- which package boundary communicates the stability of the API;
+- how much behavior is added on top of the generated clients;
+- how the clients are released; and
```
marikaner: [q] I assume this will be discussed in the release strategy adr.

adr/011-tabular-orchestration-prediction.md
davidkna-sap: I assume for RPT-style models at least a unified client would still be possible, so some of the orchestration promise would be valid and users get access to inputs via context registry.

adr/011-tabular-orchestration-prediction.md
davidkna-sap: I thought of `providerConfig` as potentially mapping to more fields than what currently encompasses `modelConfig` if necessary.

adr/010-context-registry.md
marikaner: [q] Do you know how the sync and async services relate to each other? E.g. the Tabular Artifact API is basically the same, except for the async/sync part. Is it intended to use only one of them or both in the same deployment? How do I have to envision that?

adr/010-context-registry.md
```diff
@@ -0,0 +1,261 @@
...
+## Discussion
+
+### Question 1: Specification Layout
+
+**Recommendation: Option B.2.** Generate one synchronous and one asynchronous client from two merged specifications.
+
+If upstream later provides an authoritative combined specification, replacing a locally merged contract may still cause breaking changes.
+Keeping the merge deterministic and limiting it to two output surfaces should make that transition easier to assess.
+
+#### Option A: Separate package per specification
```
marikaner: [req] I would like to also consider an option A.2. where we just consider separate modules, not necessarily packages.

adr/010-context-registry.md
```diff
@@ -0,0 +1,261 @@
...
+  async-specification.yaml
+```
+
+Definitions used by both variants are included in each merged specification.
+
+**Pros:** Produces only two coherent client surfaces and keeps generation inputs self-contained.
+
+**Cons:** Duplicates shared definitions across the sync and asynchronous clients and requires a deterministic merge step.
+
+#### Option C: Wait for an upstream combined specification
```
marikaner: [q] Do we know if this is planned?

adr/010-context-registry.md
marikaner: Actually, I found that the two specs are mutually exclusive alternatives — only one is loaded at runtime, based on the `ASYNC_CXT_ENABLED` env var, e.g: https://github.tools.sap/AI/context-registry/blob/main/src/controllers/tabular_artifact/application.py#L23-L25.

`false` (default) => sync, `true`=> async

marikaner (commented):
context registry: I think whether we merge the specs shouldn't influence the API design ultimately.

With my current understanding, that you can either use the async or sync API, it would even be wrong to merge those APIs together.

What I am missing in this ADR is, how "good" is the generated API, does it lack convenience? What are the use cases and how would they be implemented with a purely generated version (e.g. polling, but maybe there are other cases that we can include).
That would help envisioning the API and understand the need of additional convenience.

adr/010-context-registry.md
davidkna-sap: We have asked for this some time ago, but they haven't provided anything yet.

adr/010-context-registry.md
davidkna-sap: Okay, looking at the repository I think we can likely ignore the sync version, our deployment appears to support async polling so it would be that.

davidkna-sap:
> My impression is, if we separate clients per model provider, then we shouldn't have extra ones, but reuse the existing one(s). If harmonization is intended by Tabular AI, but not "good enough", we should share that feedback.

The service has apparently not been particularly designed for other models with test/learn/predict split and the way other models handle predication data without placeholder sentinels - I think it would be possible to extend the current API to that without too much in breaking changes, but it is not apparent if this will happen.

> What I am missing in this ADR is, how "good" is the generated API, does it lack convenience? What are the use cases and how would they be implemented with a purely generated version (e.g. polling, but maybe there are other cases that we can include).

The PoC repo in the backlog item has a few samples. IIRC the generated API-names are a bit verbose but overall the generated SDK is usable, I will expand the ADRs.

adr/010-context-registry.md
davidkna-sap: Added.

adr/010-context-registry.md
davidkna-sap: Added additional context.

adr/011-tabular-orchestration-prediction.md
davidkna-sap: Add some additional context - but that strategy mostly does work in case of the Orchestration service.

adr/011-tabular-orchestration-prediction.md
davidkna-sap: I think this is also changing in the RPT spec, but traditionally we have tried to keep such enums open, so I thought it is somewhat worth mentioning.

adr/010-context-registry.md
davidkna-sap: I've updated it.

adr/011-tabular-orchestration-prediction.md
davidkna-sap: I think this is also somewhat too close to providing basic destination helpers for the models.

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,252 @@
...
+The client-design concerns are the closed RPT-specific types (`TFMEnum`, `ContextSelectionConfig.filterConditions`, and untyped `ModelConfig`) and deployment handling.
+Prediction runs against a discovered deployment URL rather than the AI Core base destination, so consumers must resolve the deployment and override the destination URL or create a custom destination.
+Question 2 addresses this with a convenience layer.
+
+Supporting only SAP RPT at this stage is expected; the concern is that the contract shape, not just its model list, is RPT-specific.
+Another model might require a different contract, such as training-and-test payloads or relational tables instead of RPT's scenario, context-selection, and flat-row model.
+The service could reconcile these differences server-side, but no finalized specification commits to doing so in a vendor-neutral way.
+Until or unless it does, adding a structurally different model could significantly change the public client and package boundary.
+The model-name extensibility concern shared by generated clients is tracked in [ADR-009](./009-tabular-orchestration-release-strategy.md).
```
KavithaSiva: [q] Looks like a dead reference, model name extensibility is not talked about in ADR 009.

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,252 @@
...
+  scenarioConfigName: 'product-prediction-scenario-lowercase',
+  rows,
+  contextSelectionConfig: {
+    numRows: 500,
+    strategy: 'heuristic',
+    strategyConfig: { indexColumn: 'id' }
+  },
+  predictionConfig: {
+    targetColumns: [{ name: 'salesgroup', task_type: 'classification' }]
+  }
```
KavithaSiva: [req] Add a comment here that `modelConfig` is an additional property that can be configured in this request.

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,252 @@
...
+- `modelName` is a closed `TFMEnum` containing only `sap-rpt-*` values.
+- `modelConfig` forwards arbitrary fields instead of describing a provider-independent capability.
+- The client does not represent explainability consistently: the OpenApi description directs model-specific fields to `modelConfig`.
+- Context selection and target-column semantics are neutral only if every model gives them the same meaning; other providers may not share RPT's limits, jointly predicted columns, or explainability model.
+
+Adding models to `TFMEnum` would therefore make the client appear agnostic while moving incompatibilities to runtime validation and the untyped `modelConfig`.
+
+#### Option A: Generic generated client only
+
+If the service specification becomes genuinely provider-neutral, the generated surface
```
KavithaSiva: [q] Can we reach out to the service owners to ask them if they plan to make the service provider-neutral? And if the Context selection and target-column semantics would also be applied to other vendors?

adr/010-context-registry.md
```diff
@@ -0,0 +1,156 @@
...
+Add helpers for operations not well served by the raw generated client, e.g. polling asynchronous jobs to completion.
+
+- B.1 Postpone convenience methods until a future release, to reduce initial effort and maintenance burden.
+- B.2 Add convenience methods in the initial release, to improve developer experience.
+
+**Pros:** Improves common workflows while keeping the handwritten API small.
+
+**Cons:** Requires an explicit contract for timeouts, retries, cancellation, and error propagation.
+
+Asynchronous creation returns http `202 Accepted` and a `Location` URL whose subsequent `GET` requests return http `200`.
```
KavithaSiva, line 106: [q] Does the service also respond with the `Retry-After` header in the initial response? Would also be nice if this header is added to subsequent polling responses.

adr/010-context-registry.md
```diff
@@ -0,0 +1,156 @@
...
+The response lifecycle status is `PROCESSING`, `ACTIVE`, or `ERROR`; a helper should follow the location and expose terminal and failure states.
+The PoC's `wait-for-async-resource.mts` helper demonstrates this polling loop with the generated client.
+
+The exact helper API is still open:
+
+```ts
+// Note: Subject to change
+const artifact = await pollAsyncResource({
```
KavithaSiva, line 120: [suggestion] I like the idea of the async poll utility, would also be helpful for other async client methods we already have for e.g `VectorApi.createCollection()`. We should then ship it in a generic package like `core`.

adr/010-context-registry.md
```diff
@@ -0,0 +1,156 @@
...
+Nonparallel specifications require validation and reconciliation of conflicting or unique schemas.
+
+#### Option B: Wait for an upstream combined specification
+
+**Pros:** Avoids local merge logic and keeps the generated SDK aligned with its source of truth.
+**Cons:** Delays the SDK for a specification that has already been requested, and API Hub may still represent the service with multiple specifications.
+
+### Question 2: Package Name
+
+The context registry is mostly vendor-neutral but retains SAP RPT ties and some breaking-change risk.
```
KavithaSiva, line 50: [q] Could you elaborate or give some examples on which fields map mostly to RPT in this service?

davidkna-sap: I've added some additional text.

adr/010-context-registry.md
davidkna-sap: They do not send this currently.

adr/010-context-registry.md
davidkna-sap: I've added the suggestion to the document.

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,268 @@
...
+Adding models to `TFMEnum` would therefore make the client appear agnostic while moving incompatibilities to runtime validation and the untyped `modelConfig`.
+A typed overlay could narrow `modelConfig` per model instead of leaving it fully open: a mapped type keyed by `TFMEnum` would give each model name its own config shape (e.g. `sap-rpt-1.5` → RPT explainability options).
+Adding a model then becomes a typed extension rather than another entry in an untyped bag.
+
+> **Open question for the service owners.** Before committing to a neutral client shape, we should ask the Tabular AI service owners:
+>
+> - whether the service is intended to become provider-neutral, and if so on what timeline;
+> - whether context-selection and target-column semantics are expected to keep the same meaning for other vendors.
+>
+> The available evidence suggests the contracts already diverge structurally: TabPFN sends a stateless `x_train`/`y_train`/`x_test`/`task_config` payload with no scenario, context selection, or artifact references, whereas RPT uses `scenarioConfigName`, `contextSelectionConfig`, and `predictionConfig`.
```
davidkna-sap:
```suggestion
> The available evidence suggests the contracts already diverge somewhat: TabPFN needs a test/train x/y split whereas RPT follows a more freeform shape`.
```

adr/009-tabular-orchestration-release-strategy.md
```diff
@@ -0,0 +1,96 @@
...
+
+proposed
+
+## Context
+
+Providing SDK clients for the Tabular Orchestration APIs is a significant undertaking.
+The Tabular Orchestration services have two main entry points:
+
+- the prediction client, generated from one OpenApi specification; and
+- the context-registry client, generated from nine specifications covering synchronous and asynchronous operations.
```
ZhongpinWang: This part is a bit misleading.

There are two components: Context registry and Tabular AI Orchestration.

So context registry endpoint is not part of the Tabular AI Orchestration but a component in parallel.

- **Context registry** handles **data destination**, **tabular artifact**, **tabular scenario**, and **HANA DB**. (tabular data preparation, comparing to grounding service)
- **Tabular AI Orchestration** handles its **virtual deployment** in AI Core, routes request to the **actual prediction** model, and **context selector** to pick the relevant artifacts. (model inference with orchestration, comparing to llm orchestration service using grounding)

adr/009-tabular-orchestration-release-strategy.md
```diff
@@ -0,0 +1,96 @@
...
+- Allows package names, shared concepts, examples, and compatibility expectations to be reviewed together.
+- Avoids a temporary state in which documentation relies on unsupported setup paths.
+
+**Cons:**
+
+- The less mature contract blocks the entire release.
+- Increases the initial review, test, and documentation scope.
+- Delays user feedback that could improve either client independently.
+
+#### Option D: Independent Efforts, Release When Ready
```
ZhongpinWang, line 99: [pp] I completely agree with this option.

Reasons: 

1. Context registry and Tabular AI Orchestration are two independent components handled by two teams, one from Govind and Nama, the other from AI Services SIN 1 team (PO: Benjamin Tan, Manager: Benjamin Yap). Michael Still is our common high-level manager.

2. Just like grounding service and orchestration service, each of them focuses different aspects. Context registry handles data preparation and tabular AI orchestration focuses on the actual model routing and orchestration part.
    - But one slight difference from llm orchestration, tabular AI orchestration always requires a context scenario. Meaning you must have context registry configured properly first. Although you can provide additional context (aka. columns / rows) in the payload to tabAI orchestration as an addition. 

3. Both services will have separated specs and different release dates.

Once we have both SDKs ready, we can build an E2E case but it doesn't mean one should block the other.

ZhongpinWang: I would now prefer decoupling the BLI into two and handle them separately as the overall picture is getting clearer.

adr/010-context-registry.md
```diff
@@ -0,0 +1,175 @@
+# Context Registry Client
+
+## Status
+
+proposed
+
+## Context
+
+The tabular orchestration context registry service is shared via nine separate OpenApi specifications.
```
ZhongpinWang: [req] Please consider my previous comment to use the terms correctly, also everywhere else. Tabular AI Orchestration and Context Registry are two independent components.

adr/010-context-registry.md
```diff
@@ -0,0 +1,175 @@
...
+No combined specification is currently available at this time.
```
ZhongpinWang: Having a merged spec will be the prerequisite and it will happen. Currently under discussion where to put the merged spec (i.e. as part of AI API or as a separate service spec). Stay tuned and consider this task currently blocked.

adr/010-context-registry.md
```diff
@@ -0,0 +1,175 @@
...
+The specifications cover health, provisioning, a SQL API, and three resources that each ship a synchronous and an asynchronous variant: data destinations, scenario configuration, and tabular artifacts.
+
+The synchronous and asynchronous variants are mutually exclusive deployments, so merging both into one client is not coherent.
+The synchronous client is considered legacy, and we can likely avoid supporting it in the SDK.
```
ZhongpinWang: It is confirmed by Nama. So we only need to support async endpoints. There will be no sync endpoints in the final merged spec. Cheers.

adr/010-context-registry.md
```diff
@@ -0,0 +1,175 @@
...
+
+The asynchronous client still requires merging its resource specifications with the common health, provisioning, and SQL API specifications if we want to source from a single specification.
+They repeat some type definitions: every resource redefines `ApiError` and `DetailsErrorResponse`, while `ColumnInfo` and `Url` each appear more than once.
+
+Some specifications also contain internal headers (`AI-Main-Tenant`) and internal endpoint references that we should avoid releasing.
```
ZhongpinWang: Let's wait and see their final merged spec. The current specs contain too many internal stuff. It could need some iterations with them to improve the spec quality.

ZhongpinWang (commented):
I will continue my reviewing...

adr/009-tabular-orchestration-release-strategy.md
davidkna-sap: done.

adr/010-context-registry.md
```diff
@@ -0,0 +1,178 @@
...
+Consumers also provide the destination and resource-group header unless a helper is added (Question 4).
+
+The SDK design must therefore determine:
+
+- how the (asynchronous) specifications are grouped for generation;
+- whether types shared across specifications are duplicated or unified;
+- which package boundary communicates the stability of the API;
+- how much behavior is added on top of the generated clients;
+- how the clients are released relative to the prediction client (see [ADR-009](./009-tabular-orchestration-release-strategy.md)); and
+- how internal API elements are excluded or approved for public use.
```
ZhongpinWang: > - how the (asynchronous) specifications are grouped for generation;
> - whether types shared across specifications are duplicated or unified;
> - how internal API elements are excluded or approved for public use.

✅ Won't be a problem once we have a merged spec.

> - which package boundary communicates the stability of the API;

❓This I don't understand what you mean?

> - how much behavior is added on top of the generated clients;

✅ Aligned with Junjie. A purely generated client is acceptable. Approach similar to document grounding service.

Regarding async call, I don't think it is in general a good idea to do polling for user as the wait time, polling frequency can vary quite a lot from service to service. Although, it would be nice to have a destination constructed for user to call based on the 202 status instead of using `executeRaw()`. But well, nice to have.

> - how the clients are released relative to the prediction client (see [ADR-009](./009-tabular-orchestration-release-strategy.md)); and

✅ Already addressed in the other ADR.

adr/010-context-registry.md
ZhongpinWang: As stated before, it is confirmed by Nama that only async will be relevant. Sync endpoints will no longer be used.

Please wait for an answer from Govind. They are fully aware of the merge spec request and anyway need to provide one. Some points still need discussion with lead architects.

adr/010-context-registry.md
```diff
@@ -0,0 +1,178 @@
...
+The RPT ties are concentrated in the scenario-configuration and tabular-artifact resources, which mirror RPT's context-selection model rather than a generic registry:
+
+- `ContextSelectionStrategy` is a closed enum of `random` and `embedding`, both RPT context-selection strategies; a different provider may not select context the same way, or at all.
+- The scenario configuration is built around `tabularArtifacts` referenced by name, an RPT-specific grouping for in-context learning; TabPFN, for example, is stateless and carries its training rows inline per request instead.
+- Tabular artifacts carry `csnMetadata` with a `definitionType` of `DOCUMENT`, `REFERENCE`, or `AUTO`, reflecting RPT's CSN (CDS Standard Notation) metadata model; another provider's data intake need not follow CSN.
+- The data-destination and SQL-API resources are comparatively neutral (a named data store and a SQL surface), so the RPT coupling is not uniform across the registry.
+
+Because this risk is smaller than for prediction, a vendor-neutral package name is reasonable.
+
+**Recommendation: Option A.1.** Use `@sap-ai-sdk/ctx-registry`.
```
ZhongpinWang: Sounds good 👍. But avoid using `ctx` and write the full word `context`.

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,226 @@
...
+
+proposed
+
+The prediction SDK is generated from one OpenApi specification, but the checked-in contract and the current Tabular Orchestration v2 documentation are not demonstrably the same.
+The checked-in specification identifies itself as version `1.6.0` and already uses `/v1/predict`, but it still:
+
+- generates a request builder that calls `/predict` rather than `/v1/predict`;
+- exposes only `sap-rpt-*` values through the closed `TFMEnum`;
+- includes `contextSelectionConfig.filterConditions`, which is absent from the referenced v2 documentation;
+- places heuristic `indexColumn` under `strategyConfig`, while the v2 documentation also describes a top-level location; and
```
ZhongpinWang: Please ignore all v2 relevant information. They are on early stage and considered internal only.

adr/010-context-registry.md
```diff
@@ -0,0 +1,248 @@
...
+
+**Cons:** Requires a rename or sibling package if the service contract becomes more provider-neutral.
+
+### Question 4: Convenience Methods
+
+The question is whether to add convenience methods, especially for polling asynchronous operations, the default surface for this client.
+
+**Recommendation: Option B.2.** Include only targeted helpers for workflows that are cumbersome or unsafe with the generated client, starting with asynchronous polling.
+
+#### Option A: No convenience methods
```
jjtang1985, line 87: 👍 for the first version

adr/010-context-registry.md
davidkna-sap: I've rewritten this to avoid outdated information more.

adr/011-tabular-orchestration-prediction.md
```diff
@@ -0,0 +1,239 @@
...
+`Record<string, unknown>` always matches, so a mistyped RPT config passes without error:
+
+```ts
+// ✗ Not caught — second union member accepts anything
+await client.predict({ modelName: 'rpt-1', modelConfig: { temperature: 0.8 } });
+```
+
+The drawback is that the union grows with each vendor and cannot span package boundaries cleanly.
+
+#### Option C: Generic request type with a config registry
```
marikaner, line 135: I am pro options B and C.

adr/010-context-registry.md
```diff
@@ -0,0 +1,236 @@
...
+
+// An async generator covers progress-reporting use cases:
+for await (const current of watchAsyncResource(options)) {
+  console.log(current.status);
+}
+```
+
+The helper should accept a `Retry-After` value from both the initial `202` and each polling response and use it as the next interval when present, so that service-driven retry delays are respected automatically if the service starts sending them.
+
+#### Future option B.3: Polling integrated into the OpenApi request builder
```
marikaner: I didn't see options B.1 nor B.2.

adr/010-context-registry.md
```diff
@@ -0,0 +1,236 @@
...
+  intervalMs: 2_000,
+  signal
+});
+
+// An async generator covers progress-reporting use cases:
+for await (const current of watchAsyncResource(options)) {
+  console.log(current.status);
+}
+```
+
```
marikaner, line 138: I like the idea of a async pull utility and I think there might be potential to expose some of its internal funcitonality to allow user-driven polling (B.4).

I think putting it into the openapi requestbuilder (B.3) is not worth it.

marikaner (commented):
Thanks for shortening it, much better readable for humans now. Apparently, let's wait for the context-registry to be merged...

davidkna-sap:
> Thanks for shortening it, much better readable for humans now. Apparently, let's wait for the context-registry to be merged...

They have a spec now.

adr/010-context-registry.md
davidkna-sap: I've re-numbered it now.

PR #2185, opened 2026-08-21
feat: RPT 1.5 poc

marikaner:
## Context

Closes SAP/ai-sdk-js-backlog#632.

### DIFF
--- a/packages/rpt/src/client.ts
+++ b/packages/rpt/src/client.ts
@@ -1,8 +1,7 @@
 import {
   getFoundationModelDeploymentId,
   getResourceGroup,
-  isDeploymentIdConfig,
-  translateToFoundationModel
+  isDeploymentIdConfig
 } from '@sap-ai-sdk/ai-api/internal.js';
 import { compress as compressMiddleware } from '@sap-cloud-sdk/http-client';
 
--- a/packages/rpt/src/client.ts
+++ b/packages/rpt/src/client.ts
@@ -21,6 +21,14 @@ import type {
   ParquetPayload
 } from './types.ts';
 
+const rpt1ModelNames = new Set<string>(['sap-rpt-1-small', 'sap-rpt-1-large']);
+
+function isRpt1Model(modelDeployment: ModelDeployment<SapRptModel>): boolean {
+  return (
+    typeof modelDeployment === 'string' && rpt1ModelNames.has(modelDeployment)
+  );
+}
+
 /**
  * Representation of an RPT client to make predictions.
  * @experimental This class is experimental and may change at any time without prior notice.
@@ -35,7 +43,7 @@ export class RptClient {
    * @param destination - The destination to use for the request.
    */
   constructor(
-    modelDeployment: ModelDeployment<SapRptModel> = 'sap-rpt-1-small',
+    modelDeployment: ModelDeployment<SapRptModel>,
     destination?: HttpDestinationOrFetchOptions
   ) {
     this.modelDeployment = modelDeployment;
@@ -136,7 +144,13 @@ export class RptClient {
             ])
           )
         : null,
-      ...predictionData
+      ...predictionData,
+      // backwards compatibility with RPT-1.0
+      // TODO: Remove once RPT-1.0 is no longer supported
+      ...(isRpt1Model(this.modelDeployment) &&
+        predictionData.parse_data_types === undefined && {
+          parse_data_types: true
+        })
     } satisfies PredictRequestPayload;
 
     const { compress, ...customRequestConfig } = requestConfig;
--- a/packages/rpt/src/client.ts
+++ b/packages/rpt/src/client.ts
@@ -135,23 +135,7 @@ export class RptClient {
     const { resourceGroup, deploymentId } =
       await this.getResourceGroupAndDeploymentId();
 
-    const body = {
-      data_schema: dataSchema
-        ? Object.fromEntries(
-            dataSchema.map(({ name, ...schemaFieldConfig }) => [
-              name,
-              schemaFieldConfig
-            ])
-          )
-        : null,
-      ...predictionData,
-      // backwards compatibility with RPT-1.0
-      // TODO: Remove once RPT-1.0 is no longer supported
-      ...(isRpt1Model(this.modelDeployment) &&
-        predictionData.parse_data_types === undefined && {
-          parse_data_types: true
-        })
-    } satisfies PredictRequestPayload;
+    const body = this.buildBody(predictionData, dataSchema);
 
     const { compress, ...customRequestConfig } = requestConfig;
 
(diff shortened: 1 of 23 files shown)


packages/rpt/src/client/rpt/schema/column-type.ts
```diff
@@ -5,6 +5,27 @@
...
- * Representation of the 'ColumnType' schema.
+ * Supported column data types for the data schema.
+ *
+ * Includes base types (string, numeric, date) and additional types
+ * derived from SAP CDS (https://cap.cloud.sap/docs/cds/types#core-built-in-types).
+ * Additional types are mapped to the corresponding base type internally.
+ * All values are lowercase for case-insensitive matching.
  */
-export type ColumnType = 'string' | 'numeric' | 'date';
+export type ColumnType =
```
davidkna-sap, line 15: Since all types are mapped to the corresponding base type internally, we could down-level these for RPT-1 users.

marikaner: I don't fully understand. Before, we only had `'string' | 'numeric' | 'date'`, which still map to the same TS types and are handled same as before. That would be "downlevel support" to my understanding. Do you mean something different?

davidkna-sap: The comment sounds like all types still map to `'string' | 'numeric' | 'date'` internally - but this should probably be a server-side fix if they haven't already done this.

marikaner: > The comment sounds like all types still map to 'string' | 'numeric' | 'date' internally

I will adjust the comment to not single out the three types that existed before.

> - but this should probably be a server-side fix if they haven't already done this.

Now I think I get it: You mean, if someone uses e.g. 'time' with the rpt 1 API?
Not sure what would happen then, but I assume you would get an error. I can try to find out.
Assuming this is the case though, what are our options?
1. If the error is speaking leave it at that?
2. To avoid unnecessary calls throw an error before the request.
3. Fix the types when sending the request.

I understand that you are proposing 3. right?
I honestly think we should not do that, especially given, that people will not use this for a long time, it is supposed to stop working "no earlier" then the end of the year (which could be forever, but I think there is a willingness for this to be less than forever).

We should definitely cover this through documentation, though and I would be willing to consider 2.

marikaner: Actually, this comment is generated, I won't touch it then.

davidkna-sap: @marikaner Have you checked what happens if you send the new types to RPT-1?

marikaner: Yes, it works as if it always existed ;)

packages/rpt/src/client.ts
```diff
@@ -18,6 +18,14 @@ import type { SapRptModel } from '@sap-ai-sdk/core/internal.js';
 import type { ModelDeployment } from '@sap-ai-sdk/ai-api';
 import type { HttpDestinationOrFetchOptions } from '@sap-cloud-sdk/connectivity';
 
+const rpt1ModelNames = new Set<string>(['sap-rpt-1-small', 'sap-rpt-1-large']);
+
+function isRpt1Model(modelDeployment: ModelDeployment<SapRptModel>): boolean {
```
KavithaSiva, line 27: [q] Shouldn't we also consider the case where `ModelConfig<SapRptModel> & ResourceGroupConfig` is applied.

packages/rpt/src/client/rpt/schema/column-type.ts
davidkna-sap: In that case, I think the original `date` may have been wider than what we initially assumed.

packages/rpt/src/types.ts
```diff
@@ -25,14 +26,32 @@ export type DateString =
...
+ */
+export type TimeString =
+  `${number}${number}:${number}${number}:${number}${number}`;
+
+/**
+ * Maps the type from the spec to a TypeScript type.
  * @template T - Type of the data schema.
  */
-type TsType<T extends ColType> = T extends 'numeric'
+type TsType<T extends ColType> = T extends
```
davidkna-sap: [req] Map all column types.

packages/rpt/src/spec/rpt.json
```diff
@@ -673,43 +894,53 @@
...
-                    "type": "number"
-                  },
+                    "type": "null"
+                  }
+                ],
+                "default": null,
+                "description": "The name of the index column. If provided, the service will return this column's value in each prediction object to facilitate aligning the output predictions with the input rows on the client side. If not provided, the column will not be included in the output.",
+                "title": "Index Column"
+              },
+              "parse_data_types": {
```
ZhongpinWang, line 918: [q] Do you know why this field (and line 988) defaults to `true` for normal request and to `false` now only for parquet? This feels like a mistake?

I now have the feeling (also based on what you mentioned before, new types like time / int16 ... are now also supported for rpt 1.0), they have the same controller layer for different models and some changes are applied for both models (if they are not handled differently in their web server).

So spec version 1.5.0 might also contain changes applicable to rpt 1.0.

It is also confirmed by what Clemens wrote in the chat message

> this is the same web service code on server-side and the API is equivalent

So I doubt if `parse_data_types` now defaults for parquet to false, then for both models it is set to false (unless this change is simply a mistake from their side)

davidkna-sap: I haven't asked, but I think that parquet is handled differently because the input for parquet would generally be expected to already be in the relevant format without additional parsing.
I also remember all their parquet samples specifically disabling `parse_data_types`.

packages/rpt/src/client/rpt/schema/column-type.ts
marikaner: not necessarily, this is how it is documented https://cap.cloud.sap/docs/cds/types#core-built-in-types

packages/rpt/src/client.ts
marikaner: Yes, this is the case in the last "else" of the ternary.

packages/rpt/src/spec/rpt.json
marikaner: @cbiehlsap

packages/rpt/src/types.ts
marikaner: I agree, that's better 😊

packages/rpt/src/client/rpt/schema/column-type.ts
marikaner: I tried it out though, it seems that you can send arbitrary strings as timestamps, e.g. Reaching out to the RPT colleagues.

packages/rpt/src/types.ts
```diff
@@ -26,14 +27,45 @@ export type DateString =
...
+  double: number;
+  date: DateString;
+  time: TimeString;
+  datetime: DateTimeString;
+  timestamp: TimestampString;
+  boolean: boolean;
+  string: string;
+  largestring: string;
+  uuid: string;
+}
```
marikaner, line 68: @cbiehlsap

packages/rpt/src/types.ts
```diff
@@ -32,27 +32,40 @@ export type DateString =
...
-  | 'double'
-  ? number
-  : T extends 'date'
-    ? DateString
-    : T extends 'time'
-      ? TimeString
-      : T extends 'boolean'
-        ? boolean
-        : string;
+type TsType<T extends ColType> = ColTypeMap[T];
```
davidkna-sap, line 74: [pp] create a type test that asserts that `keyof ColTypeMap` matches `ColType`.

marikaner: What would be the purpose of that test? If a type from `ColType` is omitted in the `ColTypeMap` the `TsType` here will show an error.

davidkna-sap: Ok, that would be fine, I assumed it would just derive to e.g. `never`

PR #2188, opened 2026-08-21
fix(langchain): normalize invalid tool message content to empty string

anjan-poonacha:

Solves: #2204

When an MCP tool returns no results, the MCP protocol allows returning an empty
content array `[]`. LangChain wraps this as a `ToolMessage` with `content: []`.

`mapToolMessageToOrchestrationToolMessage` validates content using:

```ts
message.content.some(content => content.type !== 'text')
```

`Array.some()` on an empty array always returns `false`, so `[]` passes the
guard silently and is forwarded as-is to SAP AI Core Orchestration API.

AI Core requires tool message `content` to be a string. It rejects `[]` with:

```
400 - Request Body: [] is not of type 'string'
```

This makes any tool that returns an empty result uncallable — the entire agent
request fails with a 400, even though the tool call itself succeeded.

This PR normalises `[]` (and other non-string, non-array edge cases like `{}`)
to `''` before building the `ToolChatMessage`, so empty tool results are handled
gracefully without breaking the request.

## Definition of Done

- [x] Code is tested (Unit) — 4 new unit tests covering `[]`, `{}`, `null`,
      `undefined`; all fail without the fix, all pass with it
- [ ] Documentation updated — no public API change
```

### CODE REVIEW

File: packages/langchain/src/orchestration/util.ts [hunk 1]
```diff
@@ -216,17 +216,22 @@ function mapSystemMessageToOrchestrationSystemMessage(
...
+
+  if (typeof content !== 'string') {
+    if (!Array.isArray(content) || content.length === 0) {
+      return { role: 'tool', content: '', tool_call_id: message.tool_call_id };
+    }
+    if (content.some(block => block.type !== 'text')) {
+      throw new Error(
+        'The content type of tool message can only be "text" in the Orchestration Client.'
+      );
+    }
```
davidkna-sap: I would prefer also throwing if any truthy non-string is provided, and to simplify the check:

```suggestion
    if (!content?.length) {
      return { role: 'tool', content: '', tool_call_id: message.tool_call_id };
    }
    if (!Array.isArray(content) || content.some(block => block.type !== 'text')) {
      throw new Error(
        'The content type of tool message can only be "text" in the Orchestration Client.'
      );
    }
```

anjan-poonacha: Done

davidkna-sap (approved):

anjan-poonacha:
Thank you 👍🏻 
Please merge it as I do not have the access.

davidkna-sap:
Thanks for the contribution @anjan-poonacha!

PR #684, opened 2025-04-25
chore: Improve release note merge workflow in publish script

KavithaSiva:
## Context

Enable the `publish` script to also consider if the Docs PR gets merged already.

## What this PR does and why it is needed

We faced some issues with the last release process, adapting accordingly.

PR #2216, opened 2026-08-27
chore: fix sample code response input types

InjunPark-sap:

Replace `any[]` with `toResponseInputItems()` from `openai/lib/responses/ResponseInputItems` in `responsesApiMultiTurn()`.

Follows review feedback from sap/ai-sdk#684.


sample-code/src/openai.ts
```diff
@@ -149,13 +150,13 @@ export async function responsesApiMultiTurn(): Promise<string | undefined> {
     deployment: 'gpt-5.4-nano'
   });
 
-  let context: any[] = [
+  let context = toResponseInputItems([
```
davidkna-sap: [pp] This already looks like an input:
```suggestion
  let context =[
```

InjunPark-sap: The literal `{ role: 'user' `} widens to `role: string`, which isn't assignable to `ResponseInputItem` (`role` expects `'user' | 'assistant' | 'system' | 'developer'`). Would need `role: 'user' as const.` 

If you agree with, I can apply that if preferred over toResponseInputItems().

davidkna-sap: Please switch to as const.

InjunPark-sap: `role: 'user' as const` alone causes [TS2322] on reassignment — `{ role: "user"; ... }[]` can't hold `ResponseInputItem[].` 

Wrapping the initial array with `toResponseInputItems()` infers `ResponseInputItem[]` from the start and avoids the error. Kept this approach; let me know if you prefer something different.

sample-code/src/openai.ts
```diff
@@ -149,15 +153,15 @@ export async function responsesApiMultiTurn(): Promise<string | undefined> {
...
     { role: 'user', content: 'What is the capital of France?' }
```
davidkna-sap:
```suggestion
  let context = [
{ role: 'user' as const, content: 'What is the capital of France?' }
```

packages/openai/src/realtime/realtime-client.ts
```diff
@@ -121,7 +120,17 @@ export class SapOpenAiRealtimeWs extends OpenAIRealtimeWS {
...
+          const url = new URL(context.azureOptions.baseURL!);
+          url.protocol = 'wss:';
+          url.pathname = url.pathname.replace(/\/?$/, '') + '/v1/realtime';
+          url.searchParams.set(
+            'api-version',
+            context.azureOptions.apiVersion ?? '2024-10-21'
+          );
+          return url;
+        }
       },
```
davidkna-sap: Please don't add unrelated changes.

sample-code/src/openai.ts
InjunPark-sap: Applied the suggestion, but it causes `TS[2322]` in CI:

Type `({ role: "user"; content: string; } | ResponseInputItem)[]` is not assignable to type `{ role: "user"; content: string; }[]`.

Same error confirmed in the docs PR (#684) after applying the same suggestion. Reverting — `toResponseInputItems([...])` is necessary to infer `ResponseInputItem[]` from the start so the reassignment on the next line is type-safe.

InjunPark-sap:
Added explicit `ResponseInputItem[]` type annotation with `role: 'user' as const` — avoids [2322] without wrapping the initial array.

davidkna-sap (approved):

PR #2218, opened 2026-08-27
feat(openai): migrate Realtime URL to buildRealtimeURL callback, raise minimum openai to ^7.5.0

InjunPark-sap:
## Context

Closes SAP/ai-sdk-js-backlog#646.

Migrates Realtime URL construction from a prototype accessor hack (`Object.defineProperty`) to the native `buildRealtimeURL` callback (openai 7.5.0, [openai-node#2308](https://github.com/openai/openai-node/pull/2308)).
Removes `rewriteRealtimeUrl` as dead code.

**Breaking:** `openai` peer dependency raised to `^7.5.0`.


.changeset/realtime-openai-7.5.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/openai': major
+---
+
+Raise minimum required openai package version to 7.5.0 and migrate Realtime API URL construction to the native buildRealtimeURL callback.
```
davidkna-sap:
```suggestion
'@sap-ai-sdk/openai': minor
---

[compat] The minimum required openai package version was raised to 7.5.0.
```

packages/openai/src/realtime/realtime-client.ts
```diff
@@ -121,7 +73,23 @@ export class SapOpenAiRealtimeWs extends OpenAIRealtimeWS {
...
+            url = new URL(base + '/v1/realtime');
+          } catch {
+            throw new Error(
+              `Invalid SAP AI Core deployment URL: "${openAiClient.baseURL}". Ensure the deployment is resolved correctly before opening a Realtime connection.`
+            );
+          }
+          url.protocol = 'wss';
+          if (openAiClient.apiVersion) {
+            url.searchParams.set('api-version', openAiClient.apiVersion);
+          }
```
davidkna-sap: Please look at the BLI: `api-version` was deprecated and check if we need `buildRealtimeURL()` at all with the most recent openai package.
Please also check if the handling of any other query parameters can be optimised.

sample-code/src/openai.ts
```diff
@@ -107,7 +106,7 @@ export async function responsesApi(): Promise<string | undefined> {
  * @returns The stream of response events.
  */
 export async function responsesApiStream(): Promise<
-  AsyncIterable<{ type: string; delta?: string }>
+  AsyncIterable<{ type: string; delta?: unknown }>
```
davidkna-sap: [q] Why is this necessary?

davidkna-sap (changes requested):

packages/openai/src/realtime/realtime-client.ts
InjunPark-sap: Thanks! Will check the BLI and test without api-version. 

If the default `buildRealtimeURL` covers the path correctly,  I'll remove the callback entirely.

sample-code/src/openai.ts
InjunPark-sap: @davidkna-sap 

oxlint enforces `explicit-module-boundary-types` on exported functions which is removing the annotation breaks CI. 

The original `delta?: string` failed to compile against openai 7.5.0's updated `ResponseStreamEvent` types, so `delta?: unknown` is the minimal compatible annotation.

InjunPark-sap:
@davidkna-sap 
[q] Does the Realtime ws-proxy still require`?api-version=2024-10-21`? Upstream removed it in [8b1b4b6](https://github.com/openai/openai-node/commit/8b1b4b6) — kept here based on existing tests but unverified against SAP AI Core.

packages/openai/src/realtime/realtime-client.ts
davidkna-sap: Please do test this.

davidkna-sap (changes requested):
Please do remove `api-version`, confirm whether `buildRealtimeURL()` is necessary with the new version and investigate any other query parameters OpenAI SDK handles.

InjunPark-sap:
> Please do remove `api-version`, confirm whether `buildRealtimeURL()` is necessary with the new version and investigate any other query parameters OpenAI SDK handles.

Investigated openai 7.5.0's default Azure URL construction:
- Path: `baseURL + /v1/realtime` (same as ours)
- Adds `?model=gpt-realtime` automatically
- No `api-version`

So `api-version` removal is possible.

Without `buildRealtimeURL`, the SDK appends `?model=gpt-realtime` to the WebSocket URL by default.

Does SAP AI Core's Realtime endpoint accept that parameter?

Can you confirm @davidkna-sap ? I can remove the callback if the param is safe.

davidkna-sap:
> Does SAP AI Core's Realtime endpoint accept that parameter?

It should send its own value, but please confirm via the E2E test or realtime sample app-cli.

InjunPark-sap:
> > Does SAP AI Core's Realtime endpoint accept that parameter?
> 
> It should send its own value, but please confirm via the E2E test or realtime sample app-cli.

Removed `api-version` and `buildRealtimeURL` which the default Azure URL construction in openai 7.5.0 produces the correct path (`wss://<deploymentUrl>/v1/realtime?model=gpt-realtime`). 

Confirmed via [E2E tests](https://github.com/SAP/ai-sdk-js/actions/runs/33404148466) pass.

.changeset/realtime-openai-7.5.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/openai': minor
+---
+
+[compat] The minimum required openai package version was raised to 7.5.0.
```
marikaner: [req] use active voice

```suggestion
[compat] Raise the minimum required openai package version to 7.5.0.
```

davidkna-sap (approved):

PR #2214, opened 2026-08-26
chore: add tabular orchestration clients [poc]

davidkna-sap:
## Context

Closes SAP/ai-sdk-js-backlog#ISSUENUMBER..

PR #2234, opened 2026-09-01
feat: add context-registry package

davidkna-sap:
## Context

Part of https://github.com/SAP/ai-sdk-js-backlog/issues/663
Split from #2214


pnpm-workspace.yaml
```diff
@@ -21,10 +22,10 @@ allowBuilds:
...
   unrs-resolver: false
 
-gitBranchLockfile: true
-mergeGitBranchLockfilesBranchPattern:
-  - main
-  - renovate/*
+# gitBranchLockfile: true
+# mergeGitBranchLockfilesBranchPattern:
+#   - main
+#   - renovate/*
```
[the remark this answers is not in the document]
davidkna-sap:
```suggestion
# Re-evaluate in the future, this lead to lockfile errors.
# gitBranchLockfile: true
# mergeGitBranchLockfilesBranchPattern:
#   - main
#   - renovate/*
```

sample-code/src/utils.ts
```diff
@@ -0,0 +1,153 @@
...
+ * @param options.read - Function to read the current state of the resource.
+ * @param options.getResource - Optional function to extract the resource from the read response.
+ * @param options.isComplete - Function to determine if the resource is in a completed state.
+ * @param options.getFailure - Optional function to determine if the resource is in a failed state and return an error message.
+ * @param options.maxAttempts - Maximum number of poll attempts before throwing a timeout error. Defaults to 10.
+ * @param options.intervalMs - Milliseconds to wait between poll attempts. Defaults to 1,000.
+ * @param options.initialRetryAfter - Retry-After header value from the initial asynchronous response.
+ * @param options.getRetryAfter - Function to return the Retry-After header value from a polling response, if available.
+ * @param options.signal - Signal that cancels polling and waiting.
+ * @param options.sleep - Optional override for the sleep function, e.g. for testing.
```
marikaner: [pp] I think those are not needed

sample-code/src/utils.ts
```diff
@@ -0,0 +1,153 @@
...
+ * @throws If the signal is aborted.
+ */
+export function pollAsyncResource<T>(
+  options: PollAsyncResourceOptions<T, T> & { getResource?: undefined }
+): Promise<T>;
+export function pollAsyncResource<T, ReadT>(
+  options: PollAsyncResourceOptions<T, ReadT> & {
+    getResource: (response: ReadT) => T;
+  }
+): Promise<T>;
```
marikaner: [q] Why those overloads?

sample-code/src/utils.ts
```diff
@@ -0,0 +1,153 @@
...
+  initialRetryAfter,
+  getRetryAfter,
+  signal,
+  sleep = sleepWithSignal
+}: PollAsyncResourceOptions<T, ReadT>): Promise<T> {
+  const extractResource =
+    getResource ?? ((response: ReadT): T => response as unknown as T);
+
+  signal?.throwIfAborted();
+  if (maxAttempts > 0 && initialRetryAfter !== undefined) {
```
marikaner: #styleguide ;)

```suggestion
  if (maxAttempts && initialRetryAfter !== undefined) {
```

I assume this would also be safe:
```suggestion
  if (maxAttempts && initialRetryAfter) {
```

sample-code/src/utils.ts
```diff
@@ -0,0 +1,153 @@
...
+    await sleep(parseRetryAfter(initialRetryAfter) ?? intervalMs, signal);
+  }
+
+  for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+    signal?.throwIfAborted();
+    const response = await read(signal);
+    signal?.throwIfAborted();
+    const resource = extractResource(response);
```
marikaner: [q] could this be an async function?

sample-code/src/utils.ts
```diff
@@ -0,0 +1,153 @@
...
+  throw new Error(
+    `Async resource did not complete after ${maxAttempts} attempts`
+  );
+}
+
+async function sleepWithSignal(
+  ms: number,
+  signal?: AbortSignal
+): Promise<void> {
+  for (let remaining = ms; remaining > 0;) {
```
marikaner:
```suggestion
  for (let remaining = ms; remaining;) {
```

sample-code/src/utils.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  const trimmedValue = value.trim();
+  if (/^\d+$/.test(trimmedValue)) {
+    const seconds = Number(trimmedValue);
+    const milliseconds = seconds * 1_000;
+    return Number.isFinite(milliseconds) ? milliseconds : undefined;
+  }
+
+  if (
+    !/^(?:Mon|Tue|Wed|Thu|Fri|Sat|Sun), \d{2} (?:Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) \d{4} \d{2}:\d{2}:\d{2} GMT$/.test(
```
marikaner, line 125: [q] This seems odd to me. There must be a better way to test for a Date. Also: Is only GMT allowed?

sample-code/src/tabular-orchestration.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+  if ('status' in error && typeof error.status === 'number') {
+    return error.status;
+  }
+  if (
+    'response' in error &&
+    error.response &&
+    typeof error.response === 'object' &&
+    'status' in error.response &&
+    typeof error.response.status === 'number'
```
marikaner: [q] Would something like this work?

```suggestion
  if (
    typeof error.response?.status === 'number'
```

marikaner (changes requested):
looks quite good, a few questions and nitpicks.

sample-code/src/utils.ts
davidkna-sap: Only GMT is allowed: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Date

sample-code/src/utils.ts
davidkna-sap: Removed, this function is mostly AI-generated, but according to Claude it improved type safety: Here it looks like it ensures that `ReadT` is handled if provided.

sample-code/src/tabular-orchestration.ts
davidkna-sap: Simplified a bit differently, TS does not like this.

sample-code/src/utils.ts
davidkna-sap: Done.

davidkna-sap (commented):
[q] Should we add `tabular` to the package name after all? Java is putting everything into one package for the record. But I feel like something like `tabular-context` could be more appropriate as a package name.

ZhongpinWang:
> [q] Should we add `tabular` to the package name after all? Java is putting everything into one package for the record. But I feel like something like `tabular-context` could be more appropriate as a package name.

Sorry for the late reply.

- https://api.sap.com/api/CONTEXT_REGISTRY_API/overview
- https://help.sap.com/docs/sap-ai-core/generative-ai/tabular-ai

> The Context Registry is a central component of this architecture. It manages data source connectivity, schema metadata, and scenario configuration. This abstraction ensures that application developers do not need to manage these details directly.

Officially **Context Registry** the name for that service / component. 

We also don't name orchestration as gen-ai orchestration or generative orchestration.
Now for tabular AI, same approach, we align the package name on the official service name. 

So I personally prefer `context-registry`.

ZhongpinWang:
> [q] Should we add `tabular` to the package name after all? Java is putting everything into one package for the record. But I feel like something like `tabular-context` could be more appropriate as a package name.

Also I was not aware that Java is mixing terms. Some of their current understandings are not precise.

packages/context-registry/src/client/context-registry/schema/create-ta-request.ts
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * List of selected columns from the source file to be included in the TA
+     * @example [
+     *   "customer_id",
+     *   "customer_name",
+     *   "purchase_amount"
+     * ]
+     * Min Items: 1.
+     */
+    selectedColumns?: Set<string>;
```
InjunPark-sap, line 49:
```suggestion
    selectedColumns?: string[];
```

`JSON.stringify(new Set(['a','b']))` returns `'{}'`, in which any request with `selectedColumns` silently sends nothing to the server. 

The spec's `uniqueItems: true` is a validation constraint, not a TypeScript type; `string[]` is correct, and the response type in `tabular-artifact-details.ts` already uses it.

sample-code/resources/generate-parquet.ts
```diff
@@ -1,25 +1,28 @@
...
+ * CDS/HANA format (explicit Parquet schema, DECIMAL and DATE encoded for HANA):
+ *   node generate-parquet.ts --cds              // CDS-typed output for tabular artifacts
  */
 /* oxlint-disable no-console */
 
 import { join } from 'node:path';
 
 import { parquetWriteFile } from 'hyparquet-writer';
 
-import type { RowType } from '@sap-ai-sdk/rpt';
```
InjunPark-sap, line 16: `RowType` is still used in the unchanged `rowsToColumnData` signature and in the new data as `RowType<typeof rptSchema>[]` cast added by this PR 

Removing the import causes a compile error.

sample-code/src/tabular-orchestration.ts
```diff
@@ -0,0 +1,180 @@
...
+      contextSelectionStrategy: 'random',
+      tabularArtifacts: [{ name: artifactName }]
+    },
+    headers
+  ).execute();
+
+  return ScenarioConfigurationManagerApi.getScenarioConfigurationByName(
+    scenarioConfigName,
+    headers
+  ).execute();
```
InjunPark-sap: `createScenarioConfiguration` returns 202 Accepted (async) — the immediate `getByName` call will likely return a `PROCESSING` object. `createTabularArtifact` in the same file already uses the correct `executeRaw() + pollAsyncResource` pattern; please apply it here too.

pnpm-workspace.yaml
InjunPark-sap, line 1: Disabling `gitBranchLockfile` is a workspace-wide change that affects all contributors and Renovate, but the comment gives no root cause or re-enablement plan. 

How about splitting this into a separate PR with context on the CI failure?

InjunPark-sap (changes requested):
Some questions with comment

packages/context-registry/src/client/context-registry/schema/create-ta-request.ts
davidkna-sap: This is generated code.

sample-code/src/tabular-orchestration.ts
davidkna-sap: fixed.

pnpm-workspace.yaml
davidkna-sap: Will open PR for this, but keep around to make sure it still works.

packages/context-registry/src/client/context-registry/schema/create-ta-request.ts
davidkna-sap: Handled now, this was unhandled despite being around a few years.

.github/workflows/spec-update-cron.yml
```diff
@@ -65,6 +65,15 @@
     secrets: &spec-update-secrets
       GH_TOOLS_TOKEN: ${{ secrets.GH_TOOLS_TOKEN }}
 
+  update-context-registry:
+    permissions: *spec-update-permissions
+    uses: ./.github/workflows/spec-update.yaml
```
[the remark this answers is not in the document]
davidkna-sap: This should only be handled here if #2247 is merged first.

InjunPark-sap (approved):
LGTM!

PR #2244, opened 2026-09-02
chore(prompt-registry): Use mini zod

davidkna-sap:
## Context

This should reduce bundled-application size for consumers and increase startup speed:
- https://orval.dev/docs/guides/zod/#zod-mini
- https://zod.dev/packages/mini


davidkna-sap (commented):
The main draw would be that error messages are worse (by default, unless the English langpack its loaded).

PR #2250, opened 2026-09-04
chore: Small refactoring for `applyCacheControlToLastMessage`

ZhongpinWang:
While looking at JS cache control code, a small improvement can be done.

### DIFF
--- a/packages/langchain/src/orchestration/util.ts
+++ b/packages/langchain/src/orchestration/util.ts
@@ -15,7 +15,6 @@ import type {
   ChatMessage,
   ChatMessageContent,
   CompletionPostResponse,
-  DeveloperChatMessage,
   FunctionObject,
   MessageToolCalls,
   SystemChatMessage,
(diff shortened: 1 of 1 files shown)


packages/langchain/src/orchestration/util.ts
```diff
@@ -274,50 +273,31 @@ export function applyCacheControlToLastMessage(
...
-  const block = message.content.findLast(
-    b =>
-      b &&
-      typeof b === 'object' &&
-      (b.type === 'text' ||
-        (isUserMessage && (b.type === 'image_url' || b.type === 'file')))
-  );
-  if (block) {
-    (block as { cache_control?: CacheControl }).cache_control = cacheControl;
+  } else if (typeof message.content === 'string') {
```
davidkna-sap: I think it worked for me in practice, but assistant messages are not supposed to have cache control.
```suggestion
  } else if (typeof message.content === 'string' && ['user', 'developer', 'system'].includes(message.role)) {
```

davidkna-sap (approved):

PR #2251, opened 2026-09-04
chore: Improve release trigger handling

davidkna-sap:
## Context

Replaces #2148.

We no longer need a single entry workflow for canary releases and can use different envs for canary and stable releases: https://github.blog/changelog/2026-09-03-multiple-trusted-publishing-configurations-for-npm/


.github/workflows/publish-release.yml
```diff
@@ -107,3 +103,14 @@ jobs:
...
+  refresh-context7:
+    name: Refresh Context7 Docs
+    runs-on: ubuntu-latest
+    needs: [publish-npm]
+    steps:
+      - name: Trigger Context7 Refresh
+        run: |
+          curl -fs -X POST https://context7.com/api/v1/refresh \
+            -H "Content-Type: application/json" \
+            -H "Authorization: ${{ secrets.CONTEXT7_API_KEY }}" \
```
BrigittaK307: Was it intentional to use a simple Authorization header or should we use `Bearer` auth instead?

davidkna-sap: Good catch!
```suggestion
            -H "Authorization: Bearer ${{ secrets.CONTEXT7_API_KEY }}" \
```

davidkna-sap (commented):
This still requires changes to repo + npm.

PR #2266, opened 2026-09-07
fix(core): set 1.201s default HTTP agent timeout, disable keep-alive for AI Core requests

InjunPark-sap:
## Context

Closes SAP/ai-sdk-js-backlog#623.

The SDK inherited the SAP Cloud SDK's 5s HTTP socket timeout, which is too short for long-running chat and streaming completions. This PR sets a 12-minute (720000 ms) default socket timeout on the service-binding-derived AI Core destination (covering both client-secret and mTLS auth).

Keep-alive is disabled on these requests: with the old 5s timeout, idle pooled sockets were evicted before any load balancer closed them; the longer timeout would otherwise keep stale sockets around long enough to be reused and trigger ECONNRESET. Disabling pooling avoids stale-socket reuse while the long timeout still covers slow responses. Rationale and the researched load-balancer idle timeouts are in the decision record.

Both defaults apply only to the SDK-managed destination and can be overridden via a custom destination's `agentOptions` or per-request `CustomRequestConfig.httpsAgent`.

**Open for reviewers:** confirm disabling keep-alive is acceptable vs. adding `agentkeepalive` to preserve pooling (would need SAP BTP's actual LB idle timeout to tune). Docs handled in a separate SAP/ai-sdk PR — link to follow.

## Definition of Done

- [x] Documentation updated

### DIFF
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -18,9 +18,9 @@ const logger = createLogger({
   messageContext: 'context'
 });
 
-// Default HTTP agent socket timeout for AI Core requests (12 min), overriding the
+// Default HTTP agent socket timeout for AI Core requests (1,200s + 1s leeway), overriding the
 // Cloud SDK default of 5s, which is too short for chat/streaming completions.
-const DEFAULT_AGENT_TIMEOUT = 720_000;
+const DEFAULT_AGENT_TIMEOUT = 1_200_000 + 1e3;
 
 // Disabled so the long timeout above does not keep stale pooled sockets alive past
 // a load balancer's idle timeout, which would cause ECONNRESET on reuse.
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -23,7 +23,9 @@ const logger = createLogger({
 const DEFAULT_AGENT_TIMEOUT = 1_200_000 + 1e3;
 
 // Disabled so the long timeout above does not keep stale pooled sockets alive past
-// a load balancer's idle timeout, which would cause ECONNRESET on reuse.
+// a load balancer's idle timeout, which would cause ECONNRESET on reuse. Revisit if
+// we switch HTTP library (e.g. undici / a fetch adapter), which can isolate the idle
+// free-socket timeout from the active-request timeout and make keep-alive safe here.
 const DEFAULT_AGENT_KEEP_ALIVE = false;
 
 let aiCoreServiceBinding: Service | undefined;
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -23,9 +23,10 @@ const logger = createLogger({
 const DEFAULT_AGENT_TIMEOUT = 1_200_000 + 1e3;
 
 // Disabled so the long timeout above does not keep stale pooled sockets alive past
-// a load balancer's idle timeout, which would cause ECONNRESET on reuse. Revisit if
-// we switch HTTP library (e.g. undici / a fetch adapter), which can isolate the idle
-// free-socket timeout from the active-request timeout and make keep-alive safe here.
+// a load balancer's idle timeout, which would cause ECONNRESET on reuse.
+// TODO: revisit if we switch HTTP library (e.g. undici / a fetch adapter), which can
+// isolate the idle free-socket timeout from the active-request timeout and make
+// keep-alive safe here.
 const DEFAULT_AGENT_KEEP_ALIVE = false;
 
 let aiCoreServiceBinding: Service | undefined;
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -18,6 +18,14 @@ const logger = createLogger({
   messageContext: 'context'
 });
 
+// Default HTTP agent socket timeout for AI Core requests (12 min), overriding the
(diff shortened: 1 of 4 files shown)


packages/core/src/context.ts
```diff
@@ -18,6 +18,14 @@ const logger = createLogger({
   messageContext: 'context'
 });
 
+// Default HTTP agent socket timeout for AI Core requests (12 min), overriding the
+// Cloud SDK default of 5s, which is too short for chat/streaming completions.
+const DEFAULT_AGENT_TIMEOUT = 720_000;
```
davidkna-sap: [req] Please also keep in mind that timeouts are in ms, not in seconds.

The orchestration maximum timeout is actually 1200s.
```suggestion
// Default HTTP agent socket timeout for AI Core requests (1,200s), overriding the
// Cloud SDK default of 5s, which is too short for chat/streaming completions.
const DEFAULT_AGENT_TIMEOUT = 1_200_000;
```

But it may be useful to have some leeway:
```suggestion
// Default HTTP agent socket timeout for AI Core requests (1,200s + 1s leeway), overriding the
// Cloud SDK default of 5s, which is too short for chat/streaming completions.
const DEFAULT_AGENT_TIMEOUT = 1_200_000 + 1e3;
```

packages/core/src/context.ts
```diff
@@ -18,6 +18,14 @@ const logger = createLogger({
...
+
+// Disabled so the long timeout above does not keep stale pooled sockets alive past
+// a load balancer's idle timeout, which would cause ECONNRESET on reuse.
+const DEFAULT_AGENT_KEEP_ALIVE = false;
```
davidkna-sap, line 30: I would prefer enabling this by default. Please investigate if increasing the timeout actually leads to issues and if the keep alive timeout can be configured in isolation.

As far as I know, this is also the default.

packages/core/src/context.ts
```diff
@@ -63,7 +71,14 @@ export async function getAiCoreDestination(
       useCache: true
     }
   )) as HttpDestination;
-  return aiCoreDestination;
+  return {
```
davidkna-sap, line 77: [req] Ensure the other axios options mentioned in the documentation are added; I think we might already set `timeout: 0` somewhere.

davidkna-sap (changes requested):

packages/core/src/context.ts
InjunPark-sap: 1. Node core can't isolate idle vs active timeout — one `timeout` governs both (`keepAliveMsecs` is only the TCP probe delay). Isolation needs `agentkeepalive`(new dep).

2. 12min + keepAlive:true has a documented risk (nodejs/node#47130): if the LB drops idle conns first, the reused half-open socket → ECONNRESET, no core fix. The old 5s timeout avoided this implicitly.

So it hinges on the AI Core LB idle timeout. If it is more than 12min, keepAlive:true + 1200s is safe and I'll switch; if shorter, we add agentkeepalive or keep keepAlive:false. WDYT?

packages/core/src/context.ts
InjunPark-sap: `timeout: 0` is the axios request-level timeout in [Cloud SDK default ](https://github.com/SAP/cloud-sdk-js/blob/052efd32b668687051e9f0f91883a241ecd6a5f4/packages/http-client/src/http-client.ts#L494) — no request deadline, deliberate for streaming. 

Separate layer from the agent socket `timeout`here, no conflict. CustomRequestConfig already exposes httpsAgent/httpAgent/proxy/signal/middleware; axios `timeout` isn't picked.

Do you want it to be added, or other options?

davidkna-sap: https://github.com/cline/cline/blob/cdcaa744223465a9bf5beb4d5b5a5316d4a4ea65/apps/vscode/src/shared/net.ts#L185-L186

packages/core/src/context.ts
davidkna-sap: Sounds good, sadly axios does not appear to handle this, please add a comment here to re-visit this if we do switch to a different HTTP library like unidici.
The fetch adapter would be an option for non-mTLS connections, but that could also lead to issues with any custom HTTP middleware users might have.

InjunPark-sap: Done. Added a comment to revisit keepAlive if we switch HTTP library (`undici / fetch adapter`).

packages/core/src/context.ts
InjunPark-sap: Added `maxContentLength/maxBodyLength`: Infinity to the default request config.

Note it's a no-op on the current http adapter (axios defaults both to -1 = unlimited), but makes the intent explicit and future-proofs a switch to undici/fetch.

`timeout: 0` is already set by the Cloud SDK (https://github.com/SAP/cloud-sdk-js/blob/052efd32b668687051e9f0f91883a241ecd6a5f4/packages/http-client/src/http-client.ts#L494).

packages/core/src/context.ts
davidkna-sap: Please add the TODO: string to make it easier to find.

InjunPark-sap: Done — prefixed the revisit comment with `TODO:`.

davidkna-sap (approved):

PR #2305, opened 2026-09-15
feat: Support reasoning content in Orchestration langchain client

KavithaSiva:
## Context

Closes SAP/ai-sdk-js-backlog#653

  - Reasoning blocks from model responses are now surfaced on `AIMessage.content` as typed `ContentBlock.Reasoning` blocks alongside text
  blocks
  - Raw `ReasoningBlock[]` (including `signature`) is stored in `AIMessage.additional_kwargs.reasoning_content` for multi-turn round-trips
  - On subsequent turns, reasoning blocks are forwarded back to the API in `reasoning_content` and stripped from `content` to match the API's expected format — preserving the `signature` required for Claude's encrypted thinking
  - Fixed `additional_kwargs` (including `tool_calls`, `intermediate_results`, and `reasoning_content`) being incorrectly placed on the generation object instead of the `AIMessage`

### DIFF
--- a/packages/langchain/src/orchestration/util.test.ts
+++ b/packages/langchain/src/orchestration/util.test.ts
@@ -156,6 +156,53 @@ describe('mapBaseMessageToChatMessage', () => {
     });
   });
 
+  it('should round-trip reasoning_content from additional_kwargs into AssistantChatMessage', () => {
+    const rawBlocks = [{ content: 'Let me think.', signature: 'sig123' }];
+    const aiMessage = new AIMessage({
+      content: [
+        { type: 'reasoning', reasoning: 'Let me think.' },
+        { type: 'text', text: 'The answer is 42.' }
+      ],
+      additional_kwargs: { reasoning_content: rawBlocks }
+    });
+
+    const result = mapLangChainMessagesToOrchestrationMessages([aiMessage]);
+
+    expect(result[0]).toEqual({
+      role: 'assistant',
+      content: 'The answer is 42.',
+      reasoning_content: rawBlocks
+    });
+  });
+
+  it('should extract reasoning blocks from content array when additional_kwargs has none', () => {
+    const aiMessage = new AIMessage({
+      content: [
+        { type: 'reasoning', reasoning: 'Thinking hard.' },
+        { type: 'text', text: 'Result.' }
+      ]
+    });
+
+    const result = mapLangChainMessagesToOrchestrationMessages([aiMessage]);
+
+    expect(result[0]).toEqual({
+      role: 'assistant',
+      content: 'Result.',
+      reasoning_content: [{ content: 'Thinking hard.' }]
+    });
+  });
+
+  it('should not include reasoning_content when AIMessage has no reasoning', () => {
+    const aiMessage = new AIMessage('Plain answer.');
+
+    const result = mapLangChainMessagesToOrchestrationMessages([aiMessage]);
+
+    expect(result[0]).toEqual({
+      role: 'assistant',
+      content: 'Plain answer.'
+    });
+  });
+
   it('should map ToolMessage to ChatMessage with tool role', () => {
     const toolMessage = new ToolMessage('Tool message content', 'tool_call_id');
     const result = mapLangChainMessagesToOrchestrationMessages([toolMessage]);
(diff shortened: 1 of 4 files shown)


.changeset/fix-additional-kwargs-on-ai-message.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-ai-sdk/langchain': patch
+---
+
+[fix] Fix `additional_kwargs` (including `tool_calls`, `intermediate_results`, and `reasoning_content`) being incorrectly placed on the generation object instead of the `AIMessage`.
```
davidkna-sap, line 5:
```suggestion
[fix] Additional `additional_kwargs` properties (including `tool_calls`, `intermediate_results`, and `reasoning_content`) are now correctly placed on the generation object instead of the `AIMessage`.
```

packages/langchain/src/orchestration/util.ts
```diff
@@ -444,14 +502,17 @@ export function mapOrchestrationChunkToLangChainMessageChunk(
...
+  const content = buildContentBlocks(deltaReasoning, deltaText || null);
   const toolCallChunks = choice?.delta.tool_calls;
   const usage = chunk.getTokenUsage();
   return new AIMessageChunk({
     content,
     additional_kwargs: {
       // TODO: Fix duplicated intermediate results when using concat() method for streaming chunks.
-      intermediate_results: chunk._data.intermediate_results
+      intermediate_results: chunk._data.intermediate_results,
+      ...(deltaReasoning?.length && { reasoning_content: deltaReasoning })
```
davidkna-sap, line 515: [req] This should also have the reasoning signature, not just the content.

packages/langchain/src/orchestration/util.ts
```diff
@@ -366,6 +401,21 @@ function buildUsageMetadata(usage: TokenUsage): {
   };
 }
 
+function buildContentBlocks(
```
davidkna-sap, line 404: [pp] Streaming needs a stateful version that allocates one message-global index for every block in first-seen order, that is re-used it for later deltas.

Once reasoning establishes block index 0, later text-only chunks revert to strings, which LangChain’s stream-event bridge also assigns to index 0; tool-call index 0 can collide similarly. This may attach text/tool deltas to the active reasoning block.

packages/langchain/src/orchestration/util.ts
```diff
@@ -160,13 +162,46 @@ function mapAiMessageToOrchestrationAssistantMessage(
...
   return {
     ...(tool_calls?.length ? { tool_calls } : {}),
-    content: cloneMessageContent(message.content),
+    ...(reasoning_content?.length ? { reasoning_content } : {}),
+    content: cloneMessageContent(stripReasoningFromContent(message.content)),
     role: 'assistant'
   } as AssistantChatMessage;
 }
 
+function extractReasoningBlocks(
```
davidkna-sap, line 174: [pp] Only supporting reasoning-style lc content blocks could lead to issues, as some other v1 content blocks (e.g. tool calls) are not handled which may be dropped if unhandled.

As we don't opt-into v1 this should be acceptable but please create follow-up ticket for proper v1-content-block support.

davidkna-sap (changes requested):

packages/langchain/src/orchestration/util.ts
```diff
@@ -387,19 +437,27 @@ export function mapOutputToChatResult(
     generations: choices.map(choice => ({
       text: choice.message.content ?? '',
       message: new AIMessage({
-        content: choice.message.content ?? '',
+        content: buildContentBlocks(
+          choice.message.reasoning_content?.map(
+            (b: { content?: string }) => b.content ?? ''
+          ),
+          choice.message.content
+        ),
```
InjunPark-sap, line 445: [req] Content-less reasoning blocks (encrypted/redacted thinking, where only `signature` is present) become empty `{type:'reasoning', reasoning:''}` blocks in the message content. Streaming already excludes these — non-streaming should too. Filter out blocks without `content`.

suggestion e.g.: 
```ts
choice.message.reasoning_content
  ?.map(b => b.content)
  .filter((c): c is string => !!c)
```

packages/langchain/src/orchestration/util.test.ts
```diff
@@ -762,6 +883,33 @@ describe('mapOrchestrationChunkToLangChainMessageChunk', () => {
...
+    const result = mapOrchestrationChunkToLangChainMessageChunk(mockChunk);
+
+    expect(result.content).toEqual([
+      { type: 'reasoning', reasoning: 'Let me think.', index: 0 },
+      { type: 'text', text: 'The answer.' }
+    ]);
+    expect((result.additional_kwargs as any).reasoning_content).toEqual([
+      'Let me think.'
+    ]);
+  });
```
InjunPark-sap, line 902: [req] This assertion locks `reasoning_content` to `string[]`. When the signature fix (above) lands, update this to expect `ReasoningBlock[]` (`[{content, signature}]`)

Otherwise, the test certifies the pre-fix shape. Also drop the `as any`.

PR #2310, opened 2026-09-16
fix: extract prompt merging into dedicated function

marikaner:
- Extract prompt merging logic from `buildCompletionModulesConfig` into a dedicated `mergePromptWithMessages` function
- Fix bug where the validation error for missing prompt template or messages was silently skipped when no prompt was configured
- The new function is pure (no mutation of input)


BrigittaK307:
[q] Shouldn't one test be added to `module-config.test.ts` when `promptTemplating.prompt` is undefined?

PR #2315, opened 2026-09-17
docs: ADR 009 — prompt template message routing

marikaner:
## Summary

- Adds `adr/009-prompt-template-message-routing.md`
- Documents why `request.messages` exists (API has no top-level `messages` field), how it is routed across the four config paths, and the template echo problem
- Documents two multi-turn patterns: single-client (`getAllMessages()→messagesHistory`) and two-client (currently recommended)
- Records four open questions for future improvement (semantics of `request.messages`, ideal future DX, constructor rationale, TemplateRef review)

## Test plan

- [ ] Review ADR content for accuracy
- [ ] Verify code examples compile against current types


adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,166 @@
...
+
+`ChatCompletionRequest.messages` was introduced so callers do not have to decide between two very different API fields (`prompt.template` vs `messages_history`) depending on whether they configured a template.
+The SDK routes it automatically:
+
+| Config path | Where `request.messages` ends up |
+|---|---|
+| Local template (`prompt.template`) | Appended to `prompt.template` in the outgoing request |
+| Remote template reference (`TemplateRef`) | Appended to `messages_history` |
+| Config reference (by ID or name) | Appended to `messages_history` |
+| No template | Appended to `messages_history` |
```
ZhongpinWang: [req] This part seems wrong to me? Maybe I am not getting your idea here.

https://github.com/SAP/ai-sdk-js/blob/f219976a871222a4c0e79dc59295f29144c6d2cd/packages/orchestration/src/util/module-config.ts#L437

In which case will there be no `promptTemplating` module? Or you mean no `promptTemplating.prompt`?

adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,166 @@
...
+### The `request.messages` convenience field
+
+`ChatCompletionRequest.messages` was introduced so callers do not have to decide between two very different API fields (`prompt.template` vs `messages_history`) depending on whether they configured a template.
+The SDK routes it automatically:
+
+| Config path | Where `request.messages` ends up |
+|---|---|
+| Local template (`prompt.template`) | Appended to `prompt.template` in the outgoing request |
+| Remote template reference (`TemplateRef`) | Appended to `messages_history` |
+| Config reference (by ID or name) | Appended to `messages_history` |
```
ZhongpinWang:
```suggestion
| Orchestration Config reference (by ID or name) | Appended to `messages_history` |
```
[pp] To make it more clear that this is orchestration config reference.

adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,166 @@
...
+2. **Template duplication on subsequent turns**: If `getAllMessages()` is naïvely fed back as `request.messages` on the next turn, the template is merged into `prompt.template` again — the model receives the static template twice.
+
+## Decision
+
+### Two supported patterns for multi-turn conversation
+
+#### Pattern A — Single client, `getAllMessages()` as `messagesHistory`
+
+Use one client throughout the conversation.
+On each subsequent turn, pass the previous response's `getAllMessages()` output as `messagesHistory` (not `messages`).
```
ZhongpinWang, line 71: In addition, routing previous messages to `messagesHistory` instead of putting them again in the promptTemplating module needs special attention. You need to think if you enable translation for messages history or not. And all other modules, are they filtered or templated etc.

adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+**Caveat**: From turn 2 onward, `messagesHistory` contains the echoed template messages.
+On each subsequent turn, `request.messages` is again appended to `prompt.template`, so the model sees the static template messages twice — once in `messages_history` and once in the template array.
+In practice the model handles this gracefully, but it is redundant and may affect token usage.
+
+#### Pattern B — Two clients (currently recommended)
+
+Use one client that carries the template configuration for the initial turn, and a second client (with no template) to manage the ongoing conversation.
+The second client routes all messages through `messages_history`.
```
ZhongpinWang, line 102: Ideally, for one-time templating use case, it is okay to put prompts in prompt templating module. For conversational usecase, make sure the prompt templating module has zero prompt in it.

But sometimes for example, people use prompt template ref for system messages (for their agents) so that they can at runtime still control the agent behaviour. But then for all the upcoming conversational messages, better to have no content in prompt templating module.

adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,166 @@
...
+const conversationClient = new OrchestrationClient({
+  promptTemplating: {
+    model: { name: 'anthropic--claude-4.5-haiku' }
+    // no prompt template — messages go to messages_history
+  }
+});
+
+// Turn 1 — use the template client
+const resp1 = await templateClient.chatCompletion({
+  messages: [{ role: 'user', content: 'What is the capital of France?' }]
```
ZhongpinWang, line 123: As discussed, you don't even need to put anything in `chatCompletion()`, mostly people use templating module for system messages for initial instructions. But certainly it is also fine to have the first user message together with the initial templating messages to send everything in one round.

adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,166 @@
...
+});
+
+// Turn 2+ — use the conversation client; no template duplication
+const resp2 = await conversationClient.chatCompletion({
+  messagesHistory: resp1.getAllMessages(),
```
ZhongpinWang, line 128: SDK is more leaning to put messages all in the `messages` property including history to allow orchestration enabling modules for them. But it is up to user to decide. In langchain, we put all messages in templating module(, unless not possible? needs to be verified)

adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,166 @@
...
+
+**Why this avoids the duplication**: The second client has no local template, so `request.messages` is routed to `messages_history` and `prompt.template` is never touched.
+The model sees the history (including the echoed template from turn 1) exactly once.
+
+**Tradeoff**: Requires managing two client instances and coordinating which one to use for which turn.
+This pattern is oral knowledge today — it is not reflected in any sample code or documentation.
+
+### What not to do
+
+Do not feed `getAllMessages()` back as `request.messages` when using a local template.
```
ZhongpinWang, line 141: [req] As discussed, putting messages incl. history in messages is completely valid. In fact it is what we do in langchain. The right suggestion is rather to not put prompts in templating module for conversational usecase. If, as I mentioned above, they definitely need to put some initial instructions in the prompt templating module (as template ref for example), then do not reuse that client.

adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,178 @@
...
+When detected, it strips `messages` from the per-module request object and appends them after `messagesHistory` in the top-level `messages_history` field of the outgoing API request.
+
+This routing is mechanically forced — unlike the local-template path, there is no design decision being made.
+However, **no log is emitted** when this rerouting occurs.
+The config-reference path (`OrchestrationConfigRefById` / `OrchestrationConfigRefByName`) emits a `logger.debug` saying messages will be sent as `messages_history`; the inline `TemplateRef` path does not.
+A developer who passes `messages` to `chatCompletion()` alongside a `TemplateRef` config receives no signal that the routing differs from the local-template case.
+
+## Open Questions
+
+1. **Semantic meaning of `request.messages`**: It is unclear whether `request.messages` was intended to represent (a) the current user turn, (b) additional template messages, or (c) a convenience alias for `messages_history`.
```
ZhongpinWang: So it is eventually up to the user if they want to put everything in `messages` or just the current turn in it and all history in `messages_history`. Users need to make sure they understand some modules may not be enabled for `messages_history` and for translation module, they need to enable/disable the module etc.

adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,178 @@
...
+   The routing behaviour differs by config path, which suggests it may have been designed as "whatever makes sense given the config" rather than with a single semantics.
+   This should be clarified and documented.
+
+2. **Ideal future developer experience**: The current patterns require callers to understand internal routing details.
+   Candidate improvements include: (a) the client automatically strips echoed template messages from `getAllMessages()` so it returns only non-template turns; (b) `request.messages` is always routed to `messages_history` and never merged into the template; (c) a stateful client that accumulates history internally.
```
ZhongpinWang: For a)

My feeling is that it is possible to get only the latest response from remote. But users might also want the templated, masked history messages in their next call. So in the end they would need `getAllMessages()` to give them also all previous messages.

For b)

tbh, I don't know if this is feasible, as prompts in prompt templating module is probably appended to messages_history?

For c)

There was many discussions regarding the question: Do we support message history management for user in vanilla orchestration client. And the decision was we don't. 
One reason is that messages history can and may still need to be processed and it is up to users how they want them to be handled. Also no other AI SDK clients like OpenAI etc provide history management in this level of the SDK. Message history is handled more on the framework level like langchain. And it is a complex topic, for example, do we support in-memory history or other ways, do we allow modifying history?

adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,178 @@
...
+   No decision has been made.
+
+3. **Why is the prompt template in the constructor?**: Prompt template is one property of `promptTemplating` alongside model parameters, filters, and other module configs — all of which live in the constructor because they map to a stored config artifact.
```
ZhongpinWang: One issue I see here is that whenever user needs template ref, the same issue occurs.. Or do we allow users to provide template ref per call? Feels also okay. Then we completely remove `promptTemplating.prompt` and only support everything through `chatCompletion()`. Templating is anyway also done on the server side.

adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,178 @@
...
+   Whether it would be ergonomic to also allow a per-call template override (without breaking the config-artifact mental model) is an open question.
+
+4. **Is "messages are silently ignored" a correct description of `TemplateRef` behavior?**: The claim sometimes made is that messages passed alongside a `TemplateRef` are silently ignored.
+   This is not accurate: the messages reach the model via `messages_history`.
+   The actual gap is observability — no log fires to indicate the rerouting, unlike the config-reference path.
```
ZhongpinWang: 👍 you are right

adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,190 @@
...
+When detected, it strips `messages` from the per-module request object and appends them after `messagesHistory` in the top-level `messages_history` field of the outgoing API request.
+
+This routing is mechanically forced — unlike the local-template path, there is no design decision being made.
+However, **no log is emitted** when this rerouting occurs.
+The config-reference path (`OrchestrationConfigRefById` / `OrchestrationConfigRefByName`) emits a `logger.debug` saying messages will be sent as `messages_history`; the inline `TemplateRef` path does not.
+A developer who passes `messages` to `chatCompletion()` alongside a `TemplateRef` config receives no signal that the routing differs from the local-template case.
+
+#### Security implications
+
+It is not fully confirmed whether content in `messages_history` is subject to the same pipeline modules — masking and content filtering — as content in `prompt.template`.
```
KavithaSiva: I just talked to @rnewbigging and looks like `messages_history` passes through all of the orchestration pipeline modules by default, they have flags introduced now, for translation and masking to exclude it, but the implementation also differs on how to exclude, i.e some or all messages. 

@rnewbigging Maybe you can comment more.

rnewbigging: Here's a brief overview:

- **Grounding**: Does not touch messages history, only grounding input
- **Translation**: Can be disabled by setting `translate_messages_history` to `false`. Note this defaults to `true` for backwards compatibility reasons. [spec ref](https://github.tools.sap/AI/llm-orchestration/blob/757beaad4a6b567641ecc0c1369c75084141cb23/src/spec/v2.yaml#L2193)
- **Masking**: Currently cannot be disabled or mitigated. The entire messages_history will be processed.
- **Content Filtering**: as of release `2609b` we now support a `target_selector` for filtering input. This allows you to set that only the e.g. last _n_ messages are processed. [spec ref](https://github.tools.sap/AI/llm-orchestration/blob/757beaad4a6b567641ecc0c1369c75084141cb23/src/spec/v2.yaml#L2530)
- **LLM**: entire history will be sent, no mitigation currently possible.

Hope that helps! Let me know if you have more questions.

adr/009-prompt-template-message-routing.md
KavithaSiva: In the case of No template, the messages are appended  to `prompt.template` in the current implementation.

adr/009-prompt-template-message-routing.md
```diff
@@ -0,0 +1,190 @@
...
+});
+
+// Turn 2+ — use the conversation client; no template duplication
+const resp2 = await conversationClient.chatCompletion({
+  messagesHistory: resp1.getAllMessages(),
+  messages: [{ role: 'user', content: 'What is the typical food there?' }]
+});
+```
+
+**Why this avoids the duplication**: The second client has no local template, so `request.messages` is routed to `messages_history` and `prompt.template` is never touched.
```
KavithaSiva, line 133: No, I think even in this case it is currently routed to `prompt.template` and not to `messages_history`

adr/009-prompt-template-message-routing.md
KavithaSiva: >The right suggestion is rather to not put prompts in templating module for conversational usecase

Yeah, this looks like the correct way to do it, from orchestration service side, it looks like the template property is used to provide the current or latest message/query that the user wants to send to the service/llm.
But, whatever is passed in the `message_history` to the service is also pre-pended to the template.
This is why `intermediate_results.templating` also contain the history.

So then instead of the two client pattern, users can use a single client without the template configured for conversational use cases.

```TS
const conversationClient = new OrchestrationClient({
  promptTemplating: {
    model: { name: 'anthropic--claude-4.5-haiku' }
    // no prompt template
  }
});

const resp1 = await conversationClient.chatCompletion({
  messages: [{ role: 'system', content: 'You are a helpful assistant.' },{ role: 'user', content: 'What is the typical food there?' }]
});

const resp2 = await conversationClient.chatCompletion({
  messagesHistory: resp1.getAllMessages(),
  messages: [{ role: 'user', content: 'What is the typical food there?' }]
});
```
And you can continue using the `getAllMessages` of the previous response to continue the conversation.
I tested the sample out and it works.

adr/009-prompt-template-message-routing.md
KavithaSiva: With the behaviour that I verified with the service, even if they need system messages, for conversational use cases just pass it on to the `messagesHistory` instead.

adr/009-prompt-template-message-routing.md
ZhongpinWang: @marikaner @KavithaSiva I am thinking if we should just completely disallow setting prompt or ref in the constructor. Instead moving everything to `chatCompletion()` even the ref. With this, we can make the orchestration client somewhat stateless (except for configs of other modules). For orchestration config as a JSON string use case... well...

adr/009-prompt-template-message-routing.md
ZhongpinWang: Thank you so much @rnewbigging. I added your reply into my golden bookmark list. 🫡

adr/009-prompt-template-message-routing.md
ZhongpinWang: for template ref usecase, I would also suggest moving it from the constructor into the per call method `chatCompletion` and the stream method. So completely remove prompt (or ref) from constructor and make the client stateless with regard to messages.

adr/009-prompt-template-message-routing.md
ZhongpinWang: Yes. The only issue now is that we still allow them put something (either messages or ref) in the prompt templating module as part of the constructor. We should discourage / disallow this.

adr/009-prompt-template-message-routing.md
ZhongpinWang: Yes, I think if undefined `promptTemplating.prompt`, then we make it an empty list and append `messages` to it instead of to `message_history` (If I understand what was meant correctly)

adr/009-prompt-template-message-routing.md
davidkna-sap, line 1: [pp] rename to 012 to avoid conflicts with https://github.com/SAP/ai-sdk-js/pull/2234

PR #2320, opened 2026-09-18
chore: Added additonal compilation step for sample code in e2e tests workflow

BrigittaK307:
Added additional compilation for sample code folder


.github/workflows/e2e-tests.yml
```diff
@@ -34,6 +34,9 @@ jobs:
           echo "AICORE_SERVICE_KEY='$AICORE_SERVICE_KEY'" > tests/e2e-tests/.env
           url=$(echo "$AICORE_SERVICE_KEY" | jq -r '.serviceurls.AI_API_URL' | sed 's|^https://||')
           echo "Using AI Core ${{ matrix.environment }} instance on $url"
+      - name: 'Compile Sample Code'
```
davidkna-sap, line 37: Wouldn't this also run as part of post-install?

BrigittaK307: Theoretically it should be done, but the tests didn't pass no matter what I did, without adding this.

PR #2336, opened 2026-09-21
chore(scripts): add warn-only cross-check for batch models in sync-model-types

InjunPark-sap:
## Context

Closes SAP/ai-sdk-js-backlog#680.

`LlmBatchModel` is filled directly from `batchModels`, bypassing chat's retirement/exclusion logic — so a chat-excluded model (e.g. `gpt-4.1-nano`) silently lingers in the batch list.

## Approach A (agreed with @davidkna-sap)

Warn-only. Batch stays a separate ground-truth (batch/chat timelines differ) — exclusion is NOT applied to batch. Approach B rejected: risk of over-removing batch-only models the SAP Notes omit.

## Change

- `sync-model-types.ts`: `batchModelWarnings()` warns (`console.error`) for each batch entry where `isRetired(row) || MODEL_EXCLUSION_LIST.has(m)`. No auto-removal; reuses existing logic. Main-module guard so test imports don't run the script.
- 7 vitest cases; SKILL.md notes the warning.


scripts/sync-model-types.ts
```diff
@@ -482,7 +516,11 @@ async function writeChangeset(releaseNote: string): Promise<void> {
   console.error(`\nChangeset written: .changeset/${filename}`);
 }
 
-syncModelTypes().catch(err => {
-  console.error('Fatal error:', err);
-  process.exit(1);
-});
+// Only auto-run when invoked directly (e.g. `node scripts/sync-model-types.ts`),
+// not when imported by tests for the exported helpers.
+if (process.argv[1] === fileURLToPath(import.meta.url)) {
```
davidkna-sap:
```suggestion
if (process.argv[1] === import.meta.filename) {
```

scripts/sync-model-types.ts
```diff
@@ -155,6 +156,35 @@ function isRetired(row: ModelRow): boolean {
...
+  const warnings: string[] = [];
+  for (const m of batchModels) {
+    const row = rowByModel.get(m);
+    const retired = row ? isRetired(row) : false;
+    if (retired || MODEL_EXCLUSION_LIST.has(m)) {
+      warnings.push(
+        `⚠ Batch model "${m}" is retired/excluded for chat but still in batchModels — review manually and remove from the batchModels array if it is truly retired for batch.`
+      );
+    }
+  }
```
davidkna-sap: [pp] Please use `.map` and `.filter` to build a list of retired or deprecated batch models, and print everything (`", ".join) in a single message.

scripts/sync-model-types.ts
```diff
@@ -8,6 +8,7 @@
 import { existsSync } from 'node:fs';
 import { readFile, writeFile } from 'node:fs/promises';
 import { resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
```
davidkna-sap:
```suggestion
```

scripts/sync-model-types.ts
InjunPark-sap: done.

davidkna-sap (approved):

